PR #2, opened 2019-03-22
Fix bug when deserializing V::Threshold

joemphilips:
While I was porting this library in to F#, I think I found a bug.

`V::Threshold` and `E::Threshold` has almost same representation in script, but `V::Threshold` did not deserialize properly. I suppose this PR will fix.

Tell me if I'm misunderstanding something.

### DIFF
--- a/src/miniscript/astelem.rs
+++ b/src/miniscript/astelem.rs
@@ -1928,7 +1928,7 @@ pub fn parse_subexpression(tokens: &mut TokenIter) -> Result<Box<AstElem>, Error
         // vexpr [tfvq]expr AND
         if ret.is_t() || ret.is_f() || ret.is_v() || ret.is_q() {
             match tokens.peek() {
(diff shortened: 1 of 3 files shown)


joemphilips:
Found two more bugs. See commit messages for each for the detail.
(Tell me if you want me to split the PR for each.)

apoelstra:
Sorry for not noticing this til now! I didn't realize I wasn't "watching" this library on Github; I assumed I'd be watching all my own repos by default :).

Recently Pieter and I overhauled the whole library to make script fragments first class, and the E/T/V/etc properties of the script fragments (vs the old design where E/T/V were first-class and the script fragments were instances of these). I believe this bug should be fixed as a consequence of this.

joemphilips:
Thanks for the reply. I thought you are intentionally ignoring this repository until you release the first version.

Some bug seems fixed in refactored version butI think I've found different one. I will create a new PR for it.

PR #3, opened 2019-05-05
Fix bug in deserialization 2

joemphilips:
I'm not 100% sure what I'm doing here. Correct me if I'm wrong.


src/miniscript/astelem.rs
```diff
@@ -993,7 +994,14 @@ impl<P: ToPublicKey> AstElem<P> {
...
                 script_num_size(k) +
-                subs.iter().map(|s| s.script_size()).sum::<usize>()
+                subs.iter().enumerate().map(|(n, s)|
+                    if n == 0 {
+                        s.script_size()
+                    }
+                    else {
+                        1 + s.script_size()
+                    }
+                    ).sum::<usize>()
```
apoelstra: I think we can simplify this to simply adding `subs.len() - 1` to the old code (and maybe an `assert!(!subs.is_empty())` which I think/hope we enforce at parse time).

joemphilips: Fixed.
I didn't add the assertion since I thought this is not the right place to do.

apoelstra: The code will panic even without the assertion, if `subs.len()` is 0 and you try to subtract one.

apoelstra:
Can you also drop `Tuck` from lex.rs while you're changing things? It has nothing to do with the PR but it's simple and you're adding tokens anyway.

Hilarious that I screwed up thresholds so badly that I couldn't even recognize the ADD opcode in my lexer..

src/miniscript/astelem.rs
apoelstra: I agree it's not really appropriate to assert here, but if you don't want to, you need to add an explicit `subs.is_empty()` check. (Your original code did the right thing actually :})

joemphilips: Well, then I guess just assert here.
empty `subs` are something that shouldn't happen. so failing fast seems better than returning  nonsense value.(IMHO)

apoelstra:
ACK

PR #6, opened 2019-05-30
Fixed max satisfaction size calculation bug for PK case

sanket1729:
I think the witness for pubkey should be 73 bytes signature instead of `pubkey_size(p) + 2`. Or am I missing something?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "miniscript"
-version = "0.5.4"
+version = "0.5.5"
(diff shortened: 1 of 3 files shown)


sanket1729:
Seems really odd that this should cause a build fail. Travis is running some `test src/lib.rs`, but that file does not contain any tests.

apoelstra:
Good catch! The broken test is in the comment on lib.rs:76. It looks like the error reporting for documentation tests is still wrong.

BTW can you add a commit which bumps the minor version number in Cargo.toml

apoelstra:
Can you rebase to fix the broken test?

PR #7, opened 2019-06-03
Fixed max satisfaction size calculation bug for PK case

sanket1729:
From #6 
I think the witness for pubkey should be 73 bytes signature instead of pubkey_size(p) + 2. Or am I missing something?

Could not get #6 to reopen because I deleted the branch locally and remotely.


apoelstra:
Looks good. ACK. Will wait for travis to finish.

PR #9, opened 2019-06-08
Added interpreter for miniscript

sanket1729:
Added an interpreter for Miniscript which given a satisfying witness to a descriptor, returns the set of public keys, hash preimages, and timelocks that were used. 

I have also added tests to check the "all good" tests. I can add more to check error cases too, but I thought I should get some feedback on the design and code. I will add comments on places where I made some assumptions about the behaviour of the interpreter.

### DIFF
--- a/src/descriptor/create_descriptor.rs
+++ b/src/descriptor/create_descriptor.rs
@@ -1,0 +1,410 @@
+//! Tools for creating Descriptor and witness stack from given scriptpubkey and corresponding
+//! scriptsig and witness.
+//!
+
+use bitcoin::{self, Script, PublicKey};
+
+use descriptor::Descriptor;
+use miniscript::Miniscript;
+use miniscript::evaluate::StackElement;
+use Error;
+use ToPublicKey;
+use bitcoin::blockdata::script::Instruction;
+use bitcoin::blockdata::opcodes;
+
+/// Helper function for creating StackElement from Push instructions. Special case required for
+/// handling OP_PUSHNUM_1.
+/// Dissatisfied is Pushbytes0 and witness are mapped to Pushbytes instruction in scriptsig and
+/// satisfied is mapped as OP_PUSHNUM_1.
+/// Other opcodes are considered Non-standard in bitcoin core.
+/// Miniscript should not use other pushes apart from PUSHNUM_1. This will err on reciving anything
+/// which is not PUSHBYTES, OR PUSHNUM_1 as other things are expected to happen in Miniscript
+///
+/// NOTE: Miniscript pushes should only be either boolean, 1 or 0, signatures, and hash preimages.
+/// As per the current implementation, PUSH_NUM2 results in an error
+fn instr_to_stackelem(ins: Instruction) -> Result< StackElement, Error> {
+    match ins {
+        //Also covers the dissatisfied case as PushBytes0
+        Instruction::PushBytes(v) => Ok(StackElement::from(v.to_vec())),
+        Instruction::Op(opcodes::all::OP_PUSHNUM_1) => Ok(StackElement::Satisfied),
+        _ => Err(Error::BadDescriptor),
+    }
+}
+
+/// Helper function which splits the scriptsig into 2 parts returns the corresponding elements.
+/// Usually used for scripts which have top element as Pk (p2pkh) or redeem script(p2sh).
+/// Converts the other script elements into Vec<StackElement>
+fn parse_scriptsig_top(script_sig: &bitcoin::Script) -> Result<(Vec<u8>, Vec<StackElement>), Error>
+{
+    let instructions : Vec<Instruction> = script_sig.iter(false).collect();
+    let (pk_bytes, witness) = instructions.split_last().expect("public key/Redeem script");
+
+    let stack: Result< Vec<StackElement>, Error> = witness
+        .iter()
+        .map(|instr| instr_to_stackelem(instr.clone()))
+        .collect();
+
+    if let &Instruction::PushBytes(ref top) = pk_bytes {
+        Ok((top.to_vec(), stack?))
+    }
+    else {
+        Err(Error::BadDescriptor)
+    }
+}
+
+/// Helper to create a wpkh descriptor based on script_pubkey and witness. Validates the pubkey hash
+/// and pushes rest of witness to Vec<StackElement> as is.
+/// This does not check the signature, only creates the corresponding descriptor
+fn verify_wpkh(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &bitcoin::Script,
+    witness: Vec<Vec<u8>>,
+) -> Result<(PublicKey, Vec<StackElement>), Error>
+{
+    //script_sig must be empty
+    if *script_sig != Script::new(){
+        return Err(Error::CouldnotEvaluate)
+    }
+    let (pk_bytes, witness) = witness.split_last().expect("Public key");
+    let pk = bitcoin::PublicKey::from_slice(pk_bytes).expect("Public key parse");
+    let addr = bitcoin::Address::p2wpkh(
+        &pk.to_public_key(),
+        bitcoin::Network::Bitcoin,
+    );
+    if addr.script_pubkey() != *script_pubkey{
+        return Err(Error::CouldnotEvaluate)
+    }
+    let stack : Vec<StackElement> = witness.iter()
+        .map(|elem| StackElement::from(elem.clone())).collect();
+    Ok((pk, stack))
+}
+
+/// Helper for creating a wsh descriptor based on script_pubkey and witness. Validates the wsh
+/// hash based on witness script, pops the witness script from the stack and returns
+/// witness Vec<StackElement>. Does not interpret/check the witness against the miniscript inside
+/// the descriptor
+fn verify_wsh(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &bitcoin::Script,
+    witness: Vec<Vec<u8>>,
+) -> Result<(Miniscript<PublicKey>, Vec<StackElement>), Error>
+{
+    if *script_sig != Script::new(){
+        return Err(Error::BadDescriptor)
+    }
+    let (witness_script, witness) = witness.split_last().expect("Public key");
+    let witness_script = Script::from(witness_script.clone());
+    if witness_script.to_v0_p2wsh() != *script_pubkey{
+        return Err(Error::BadDescriptor)
+    }
+    let ms = Miniscript::parse(&witness_script)?;
+    //only iter till len -1 to not include the witness script
+    let stack : Vec<StackElement> = witness.iter()
+        .map(|elem| StackElement::from(elem.clone())).collect();
+    Ok((ms, stack))
+}
+
+
+/// Creates a pkh descriptor based on scriptsig and script_pubkey. Validates the hash checks for
+/// p2pkh against top element(pk) and pushes all remaining witness elements into witness<StackElement>
+fn verify_p2pkh(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &bitcoin::Script,
+    witness: Vec<Vec<u8>>,
+) -> Result<(Descriptor<PublicKey>, Vec<StackElement>), Error>
+{
+    let (pk_bytes, stack) = parse_scriptsig_top(script_sig)?;
+    let pk = bitcoin::PublicKey::from_slice(&pk_bytes).expect("Public key parse");
+    let addr = bitcoin::Address::p2pkh(
+        &pk.to_public_key(),
+        bitcoin::Network::Bitcoin,
+    );
+    if *script_pubkey != addr.script_pubkey() || witness.len() != 0{
+        return Err(Error::BadDescriptor)
+    }
+    Ok((Descriptor::Pkh(pk), stack))
+}
+
+/// Helper for creating a p2sh descriptor based on script_pubkey and witness. Validates the p2sh
+/// hash based on redeem script, pops the redeem script from the script sig stack, translates other
+/// elements from scriptsig into Vec<StackElement>
+fn verify_p2sh(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &bitcoin::Script,
+) -> Result<(Script, Vec<StackElement>), Error>
+{
+    let (redeem_script, stack) = parse_scriptsig_top(script_sig)?;
+    let redeem_script = Script::from(redeem_script);
+    if redeem_script.to_p2sh() != *script_pubkey {
+        return Err(Error::BadDescriptor)
+    }
+    Ok((redeem_script, stack))
+}
+
+/// Figures out the the type of descriptor based on scriptpubkey, witness and scriptsig.
+/// Outputs a descriptor and witness_stack for all descriptors which can be directly fed into the
+/// interpreter. All script_sig and witness are translated into a single witness stack
+/// Vec<StackElement>
+/// 1) Pkh: Removes top element(pk) and validates pubkey hash, pushes rest of witness to
+/// witness_stack<StackElement> and outputs a Pkh descriptor
+/// 2) Wphk: translates witness to witness_stack<StackElement>, validates sig and pubkey hash
+/// and outputs a Wpkh descriptor
+/// 3) Wsh: pops witness script and checks wsh output hash, translates remaining witness elements
+/// to Vec<StackElement> and outputs a Wsh descriptor. Does not check miniscript inside the descriptor
+/// 4) Bare: translates script_sig to witness_stack<StackElement> and script_pubkey to miniscript
+/// 5) Sh: Checks redeem_script hash, translates remaining elements from script_sig to
+/// witness_stack<StackElement> and redeem script to miniscript. Does not check the miniscript
+/// 6) ShWpkh: Sh: Checks redeem_script hash, translates remaining elements from script_sig to
+/// witness_stack<StackElement> and validates Wpkh sig, pubkey.
+/// 7) ShWsh: Checks witness script hash, pops witness script and converts it to miniscript.
+/// translates the remaining witness to witness_stack<StackElement>
+pub fn witness_stack(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &bitcoin::Script,
+    witness: Vec<Vec<u8>>,
+) -> Result<(Descriptor<PublicKey>, Vec<StackElement>), Error>
+{
+    if script_pubkey.is_p2pkh(){
+        verify_p2pkh(script_pubkey, script_sig, witness)
+    }
+    else if script_pubkey.is_v0_p2wpkh(){
+        let (pk, stack) = verify_wpkh(script_pubkey, script_sig, witness)?;
+        Ok((Descriptor::Wpkh(pk), stack))
+    }
+    else if script_pubkey.is_v0_p2wsh(){
+        let (ms, stack) = verify_wsh(script_pubkey, script_sig, witness)?;
+        Ok((Descriptor::Wsh(ms), stack))
+    }
+    else if script_pubkey.is_p2sh(){
+        let (redeem_script, stack) = verify_p2sh(script_pubkey, script_sig)?;
+        if redeem_script.is_v0_p2wpkh()
+        {
+            //ensures that scriptsig after popping redeem script contains 0 elements.
+            //Therefore while calling verify_wpkh, an argument of Script::new() is passed instead
+            //of script_sig. The redeem_script becomes the script_pubkey
+            if stack.len() != 0{
+                return Err(Error::BadDescriptor)
+            }
+            let (pk, stack) = verify_wpkh(&redeem_script, &Script::new(), witness)?;
+            Ok((Descriptor::ShWpkh(pk), stack))
+        }else if redeem_script.is_v0_p2wsh() {
+            //ensures that scriptsig after popping redeem script contains 0 elements.
+            //Therefore while calling verify_wpkh, an argument of Script::new() is passed instead
+            //of script_sig. The redeem_script becomes the script_pubkey
+            if stack.len() != 0{
+                return Err(Error::BadDescriptor)
+            }
+            let (ms, stack) = verify_wsh(&redeem_script, &Script::new(), witness)?;
+            Ok((Descriptor::ShWsh(ms), stack))
+        }
+        else{
+            if witness.len() != 0{
+                return Err(Error::BadDescriptor)
+            }
+            let ms = Miniscript::parse(&redeem_script)?;
+            Ok((Descriptor::Sh(ms), stack))
+        }
+    }
+    else{
+        //bare
+        let stack : Result <Vec<StackElement>, Error> = script_sig
+            .iter(false)
+            .map(|instr| instr_to_stackelem(instr.clone()))
+            .collect();
+        if witness.len() != 0{
+            return Err(Error::BadDescriptor)
+        }
+        let ms = Miniscript::parse(script_pubkey)?;
+        Ok((Descriptor::Bare(ms), stack?))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use ::{Descriptor, Miniscript};
+    use descriptor::create_descriptor::witness_stack;
+    use bitcoin::blockdata::script;
+    use bitcoin::{self, PublicKey};
+    use secp256k1::{self, Secp256k1, VerifyOnly};
+    use miniscript::evaluate::StackElement;
+    use miniscript::astelem::AstElem;
+
+    fn setup_keys_sigs(n: usize)
+                       -> ( Vec<PublicKey>, Vec<Vec<u8> >, secp256k1::Message, Secp256k1<VerifyOnly>) {
+        let secp_sign = secp256k1::Secp256k1::signing_only();
+        let secp_verify = secp256k1::Secp256k1::verification_only();
+        let msg = secp256k1::Message::from_slice(
+            &b"Yoda: btc, I trust. HODL I must!"[..]
+        ).expect("32 bytes");
+        let mut pks = vec![];
+        let mut sigs = vec![];
+        let mut sk = [0; 32];
+        for i in 1..n+1 {
+            sk[0] = i as u8;
+            sk[1] = (i >> 8) as u8;
+            sk[2] = (i >> 16) as u8;
+
+            let sk = secp256k1::SecretKey::from_slice(&sk[..]).expect("secret key");
+            let pk = PublicKey {
+                key: secp256k1::PublicKey::from_secret_key(
+                    &secp_sign,
+                    &sk,
+                ),
+                compressed: true,
+            };
+            let sig = secp_sign.sign(&msg, &sk);
+            let mut sigser = sig.serialize_der();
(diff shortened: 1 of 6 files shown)


src/descriptor/mod.rs
```diff
@@ -416,6 +418,31 @@ impl<P: ToPublicKey> Descriptor<P> {
...
+            Descriptor::Bare(ref d) |
+            Descriptor::Sh(ref d) |
+            Descriptor::ShWsh(ref d) |
+            Descriptor::Wsh(ref d) => {
+                d.interpret(witness, age, msg, secp)
+            },
+            Descriptor::Pkh(ref pk) |
+            Descriptor::ShWpkh(ref pk) |
+            Descriptor::Wpkh(ref pk) => {
+                Ok(Primitives { pks: vec![pk.to_public_key()], hashlocks: vec![], timelocks: vec![] })
```
sanket1729: Should the interpreter only check the signatures/hashes inside Miniscript or should it also interpret the entire descriptor and do a check for Pkh(P)? Right now, I am assuming that the descriptor has valid witnesses and only checking the interpretation of Miniscript.

src/descriptor/mod.rs
```diff
@@ -911,5 +941,374 @@ mod tests {
...
     }
+
+    #[test]
+    fn interpret()
+    {
+        fn setup_keys_sigs(n: usize)
+                           -> ( Vec<PublicKey>, Vec<Vec<u8> >, secp256k1::Message, Secp256k1<All>) {
+            let secp = secp256k1::Secp256k1::new();
+            let msg = secp256k1::Message::from_slice(
+                &b"Yoda: btc, I trust. HODL I must!"[..]
```
sanket1729: Assumed a standard message for checking signatures

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,560 @@
...
+    res.join(&right.interpret(witness, age, msg, secp)?);
+
+    let leftres = witness.pop().expect("Witness stack has at least one element");
+    let rightres = witness.pop().expect("Witness stack has at least one element");
+    let andres = leftres == vec![1] && rightres == vec![1];
+    if andres{
+        witness.push(vec![1]);
+    }
+    else{
+        //Even if the script fragment evaluates to false, this method will return the pks, hash
```
sanket1729: If the expression evaluates to false, but still has have valid pukbeys and signatures in it, should those pubkeys be included?
For example, a threshold statement might have valid pubkeys for some subexpressions but does still returns 0.

sanket1729: The comment in the code does not respect the code below it. I will fix based on the discussion

sanket1729: This is a general problem for all subexpressions that return 0. So, I am not commenting everywhere.

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,560 @@
...
+            }
+            AstElem::Multi(k, ref keys) => evaluate_multi(k, keys, witness, msg, secp),
+            AstElem::MultiV(k, ref keys) => {
+                let res = evaluate_multi(k, keys, witness, msg, secp)?;
+                evaluate_verify(witness)?;
+                Ok(res)
+            }
+            AstElem::TimeT(t) => {
+                let res = evaluate_csv(t, age)?;
+                witness.push(build_scriptint(t as i64));
```
sanket1729: I had copy `build_scriptint` from rust bitcoin library as it now a public function and I needed to convert an integer to `Vec<u8>`. This is only for the case where `TimeT` where the return value is not `1`.  

On a similar note, I think all miniscript of `T` form return 1 on success and this is only which is an exception. Is this expression `TimeT` really useful to be added in miniscript?

apoelstra: Yeah, unfortunately there are sane usecases where it saves a byte (for example `and_cat(pk(...), time_t(100))` is the shortest way to put a timelock restriction on `pk(...)` in a T context).

We can do a bit simpler than copying `build_scriptint` though .. since timelocks are always nonnegative we can avoid some of the sign craziness.

src/descriptor/mod.rs
apoelstra: I think it should only check inside the miniscript.

I'd like to have another function which, given a scriptpubkey and witness script, figures out what kind of descriptor it is, but I need to think more about what that might take as input. But that function would be responsible for doing hashchecks.

apoelstra:
I'm not thrilled with the terms "interpret" and "interpreter", since they imply an imperative programming model. But not sure what would be a better way of describing this.

src/miniscript/evaluate.rs
apoelstra: I also wonder whether you can just push `true` onto the stack, or even some poison value, since this element should never be popped by any other Miniscript expression.

src/miniscript/evaluate.rs
apoelstra: My feeling is that if the expression evaluates to false, we should ignore any data used to satisfy it. We make an assumption in Miniscript that such data can be 3rd-party-malleated away, so we might as well act as though we did that.

cc @sipa

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,561 @@
...
+        return Ok(Primitives{pks: vec![], hashlocks: vec![], timelocks: vec![]})
+    }
+    evaluate_hash(hash, &preimage, witness)
+}
+
+/// A and B. Evaluates both A and B and then computes BOOLAND
+fn interpret_and_bool<P>(
+    left: &AstElem<P>,
+    right: &AstElem<P>,
+    witness: &mut Vec<Vec<u8>>,
```
apoelstra: I think that the `witness`, rather than being a `Vec<Vec<u8>>`, should be renamed to `stack`, and be a vector of objects of the type
```
enum StackElement {
    /// Result of a satisfied E/T/W/F
    Satisfied,
    /// Result of a dissatisfied E/W
    Dissatisfied,
    /// Input from the witness stack
    WitnessElement(Vec<u8>),
}
```
rather than using `vec![]` or `vec[1]` the way you would with Script. This way you also eliminate the complexity of using `build_scriptint`.

apoelstra: Given a witness stack, you'd go through a preprocessing stage where you mapped every empty vector to `Satisfied`, every `vec![1]` to `Dissatisfied`, and everything else to `Witness`.

This has the slightly weird effect that missing hash preimages or signatures, which are required to be `vec![]` by Miniscript, will be interpreted as "Dissatisfied", but I think that makes sense.

src/descriptor/mod.rs
apoelstra: Yeah, so, it would be cool to have a function
```
pub fn witness_stack(
    script_pubkey: &bitcoin::Script,
    script_sig: &bitcoin::Script,
    witness: Vec<Vec<u8>>,
) -> Result<(Descriptor, Vec<StackElement>), Error>
```
where `StackElement` is an abstract stack element (I suggest a definition for this type in a later comment). This function would (a) check that the witness script is consistent with the scriptpubkey and error out otherwise; (b) figure out what kind of descriptor we're dealing with; (c) translate the `scriptSig` into a set of witness elements for `Bare`/`Pk`/`Pkh`/`Sh` descriptors; (d) drop the witness script from the stack for `Wsh`/`ShWsh` descriptors.

src/descriptor/mod.rs
```diff
@@ -911,5 +941,374 @@ mod tests {
...
+
+        //check satisfied case Pk
+        let ms = Miniscript(astelem::AstElem::Pk(pks[0].clone()));
+        let bare = Descriptor::Bare(ms.clone());
+        let mut witness = vec![sigs[0].clone()];
+        let res = bare.interpret(
+            &mut witness,
+            0,
+            &msg,
+            &secp).expect("Interpretation to succeed");
```
apoelstra: Style nit: everywhere you have a `)` on the last line of a multiline thing like this, it would be better to have
```
    ...
    0,
    &msg,
    &secp,
).expect("Interpretation to succeed");
```
Note the trailing comma on `secp` and that the `)` gets its own line.

Let me know if you can cite some style guide saying that your way is correct ... I swear the Rust community changes stuff like this every week.

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,561 @@
...
+    Ok(res)
+}
+
+/// If !A then 0 else B. Evaluates B only if A is true
+fn interpret_and_casc<P>(
+    left: &AstElem<P>,
+    right: &AstElem<P>,
+    witness: &mut Vec<Vec<u8>>,
+    age: u32,
+    msg: &secp256k1::Message,
```
apoelstra: I'd prefer `msg` be renamed to `sighash` everywhere

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,561 @@
...
+    secp: &secp256k1::Secp256k1<All>)
```
apoelstra: Nit: I'd prefer `secp` be the first argument of every function it appears in, rather than the last. You should also use a `Secp256k1<C>` with `C: secp256k1::Verification` rather than `Secp256k1<All>`.

apoelstra: In general I think the order of arguments should be
1. `secp`
2. `stack` (renamed from `witness`)
3. `sighash` (renamed from `msg`)
4. `age`
5. everything else (in this case `left` and `right`)

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,561 @@
...
+    -> Result<Primitives, Error >
+    where P: ToPublicKey + Clone,
+{
+    let mut res = left.interpret(witness, age, msg, secp)?;
+    res.join(&right.interpret(witness, age, msg, secp)?);
+
+    let leftres = witness.pop().expect("Witness stack has at least one element");
+    let rightres = witness.pop().expect("Witness stack has at least one element");
+    let orres = leftres == vec![1] || rightres == vec![1];
+    if orres{
```
apoelstra: style nit: space before `{`

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,561 @@
...
+        witness.push(vec![1]);
+    }
+    else{
```
apoelstra: style nit: `} else {` should be one line

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,561 @@
...
+
+///A or B. Evaluates both A and B and then computes BOOLAND
+fn interpret_or_bool<P>(
+    left: &AstElem<P>,
+    right: &AstElem<P>,
+    witness: &mut Vec<Vec<u8>>,
+    age: u32,
+    msg: &secp256k1::Message,
+    secp: &secp256k1::Secp256k1<All>)
+    -> Result<Primitives, Error >
```
apoelstra: Style nit: `)` should be at the beginning of the line, like `) -> Result<Primitives, Error>`, and there should be a comma after `Secp256k1<All>` above

src/descriptor/mod.rs
sanket1729: I am slightly confused by the order in which things are written. I would need to determine (b) in order to solve (a), right?

apoelstra: Yeah, sorry, (a) and (b) should be swapped.

src/miniscript/evaluate.rs
sanket1729: I think it is safe to push satisfied instead of int `time` for the current semantics of Miniscript, but could not be the case in the future.  What do you think should be done?

src/descriptor/mod.rs
sanket1729: I have also added the signature validation logic alongside the hash checks for PublicKey based types, `Pkh`/`ShWpkh`/`ShWsh`.

src/miniscript/evaluate.rs
sanket1729: Ignored all primitives when the expression evaluated to false

src/miniscript/evaluate.rs
sanket1729: Changed

sanket1729:
Addressed all the above comments and coding style nits. The PR is again ready for review. 

@apoelstra  This will require a rebase after merging #8  and #11, maybe we get those in first.

src/miniscript/evaluate.rs
apoelstra: The `T` calling convention is guaranteed to push a truthy value to the stack. It makes no guarantees about what that value is. If this were to change we'd need a new calling convention for that.

sanket1729: So, you suggest that I push the int instead?

sanket1729:
Interpreter can also be changed to `Evaluator` if that sounds better

src/miniscript/evaluate.rs
apoelstra: No, I'm suggesting you push `Satisfied`

apoelstra:
"evaluate" is even worse than "interpret" :) the point is that we aren't executing code, we're determining which data was used to satisfy the Miniscript conditions

src/descriptor/mod.rs
```diff
@@ -25,18 +25,22 @@
...
 use miniscript::Miniscript;
 use policy::AbstractPolicy;
+use miniscript::evaluate::{Primitives, StackElement};
 use Error;
 use pubkey_size;
 use ToPublicKey;
+use secp256k1::VerifyOnly;
 
+
+pub mod create_descriptor;
```
apoelstra: Rather than `pub mod create_descriptor` can you do
```
mod create_descriptor;
pub use create_descriptor::witness_stack;
```
so that `create_descriptor` is not exposed in the API as a separate module?

src/descriptor/create_descriptor.rs
```diff
@@ -0,0 +1,446 @@
...
+//! and outputs a Wpkh descriptor
+//! 3) Wsh: pops witness script and checks wsh output hash, translates remaining witness elements
+//! to Vec<StackElement> and outputs a Wsh descriptor. Does not check miniscript inside the descriptor
+//! 4) Bare: translates script_sig to witness_stack<StackElement> and script_pubkey to miniscript
+//! 5) Sh: Checks redeem_script hash, translates remaining elements from script_sig to
+//! witness_stack<StackElement> and redeem script to miniscript. Does not check the miniscript
+//! 6) ShWpkh: Sh: Checks redeem_script hash, translates remaining elements from script_sig to
+//! witness_stack<StackElement> and validates Wpkh sig, pubkey.
+//! 7) ShWsh: Checks witness script hash, pops witness script and converts it to miniscript.
+//! translates the remaining witness to witness_stack<StackElement>
```
apoelstra: 1. I think this whole block comment should actually go on `pub fn witness_stack` rather than on the module.
2. I don't think we should be verifying signatures in this function; that seems like a responsibility of the interpreter (and it seems like you're already doing it there)

src/descriptor/create_descriptor.rs
```diff
@@ -0,0 +1,446 @@
...
+/// handling OP_PUSHNUM_1.
+/// Dissatisfied and witness are mapped to Pushbytes instruction in scriptsig and satisfied is mapped
+/// as OP_PUSHNUM_1. Other opcodes are considered Non-standard in bitcoin core.
+/// Miniscript should not use other pushes apart from PUSHNUM_1. This will panic on reciving anything
+/// which is not PUSHBYTES, OR PUSHNUM_1 as other things are expected to happen in Miniscript
+fn instr_to_stackelem(ins: Instruction) -> StackElement {
+    match ins {
+        Instruction::PushBytes(v) => StackElement::from(v.to_vec()),
+        Instruction::Op(opcodes::all::OP_PUSHNUM_1) => StackElement::Satisfied,
+        _ => unreachable!(),
```
apoelstra: Why don't we translate empty pushes to `Dissatisfied`?

Also we can't panic on non-push data because we're starting from a scriptsig/witness stack which is user-controlled data. We should return an error instead.

sanket1729: `Dissatisfied` is actually translated as PUSHBYTES_0 which converted from `StackElement::from(v.to_vec()` . Will change the unreachable to error.

src/descriptor/create_descriptor.rs
sanket1729: The signatures are being checked for non-miniscript `Pkh/WPkh/ShWpkh` descriptors. 

For `Pkh/WPkh/ShWpkh` types, I should only check the pubkey hash matches and not the signatures. And the interpreter should then check only the signatures?

src/miniscript/evaluate.rs
sanket1729: Yup, that is the current implementation

src/descriptor/create_descriptor.rs
apoelstra: Ah, nice. Can you add a comment above `PushBytes(v)` line?

src/descriptor/create_descriptor.rs
apoelstra: I think so, yeah. Reason is that I want to be able to abuse `witness_stack` by giving it only the witnessScript/redeemScript and no actual satisfying data, and have it give me the `Descriptor` and an empty `Vec<StackElement>` back.

This gives me the ability to get a descriptor from a scriptpubkey/scriptsig/witness script.

sanket1729: So, `Pkh/WPkh/ShWpkh` should return empty `Vec<StackElement>` or vec![sig pk] or vec![sig] since is in descriptor.

sanket1729: Sorry, understood what you meant finally with the above comment. 

For `Pkh` if only pk is supplied as `witness` without a sig. It will return empty. If it provided with a `sig`, that will return `[StackElement::Witness(sig)]`. A similar logic for `Wpkh/Shwpkh`

apoelstra: Yep exactly :)

sanket1729:
Addressed the newer comments.

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,606 @@
...
+/// Others are directly pushed as witness
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub enum StackElement {
+    /// Result of a satisfied E/T/W/F
+    Satisfied,
+    /// Result of a dissatisfied E/W
+    Dissatisfied,
+    /// Input from the witness stack
+    Witness(Vec<u8>),
+}
```
apoelstra: Can you try
```

#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
pub enum StackElement<'a> {
    /// Result of a satisfied E/T/W/F, or input to `IF`/`NOTIF`
    Satisfied,
    /// Result of a dissatisfied E/W, or input to `IF`/`NOTIF`
    Dissatisfied,
    /// Other data from the witness stack
    Push(&'a [u8]),
}
```
You'll run into a bunch of weird lifetime issues doing the `'a` thing - if you have too much trouble let me know and I'll add a commit which does this. But the upside is that you'll avoid a bunch of extra allocations.

But at the very least can you rename `Witness` to `Push` (my, bad!). I didn't realize when I made my original suggestion that `Satisfied` and `Dissatisfied` might also be input from the witness stack, so naming one variant `Witness` is a bit confusing.

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,606 @@
...
+            _ if v == vec![1] => StackElement::Satisfied,
+            _ if v == vec![] => StackElement::Dissatisfied,
+            _ => StackElement::Witness(v)
+        }
+    }
+}
+
+impl StackElement{
+
+    pub fn is_witness_or_err(self) -> Result<Vec<u8>, Error>
```
apoelstra: I think you should implement this on `Option<StackElement>`, have it take `&self` rather than `self`, and return a `&[u8]` rather than `Vec<u8>`.

apoelstra: and rename the method to just `witness`. The `_or_err` is implied by the type.

src/descriptor/mod.rs
```diff
@@ -416,6 +419,37 @@ impl<P: ToPublicKey> Descriptor<P> {
...
+            Descriptor::Bare(ref d) |
+            Descriptor::Sh(ref d) |
+            Descriptor::ShWsh(ref d) |
+            Descriptor::Wsh(ref d) => {
+                d.interpret(secp, stack, sighash, age)
+            },
+            Descriptor::Pkh(ref pk) |
+            Descriptor::ShWpkh(ref pk) |
+            Descriptor::Wpkh(ref pk) => {
+                let sig = stack.pop().expect("Signature").is_witness_or_err()?;
```
apoelstra: This `expect` is invalid because there might not be anything on the stack when the user calls this function. See my below comment about `is_witness_or_err` - I think you should give the `Option<_>` from `pop` directly to that method.

src/miniscript/evaluate.rs
apoelstra: Similarly you should have a method `preimage(hash: sha256d::Hash)` which returns a preimage if it satisfies the length/hash check

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,606 @@
...
+use miniscript::astelem::AstElem;
+use ToPublicKey;
+use secp256k1::{Secp256k1, Signature, VerifyOnly};
+use bitcoin_hashes::Hash;
+use bitcoin;
+
+///Primitives for Miniscript: a vector of public keys(sigs), hashlocks and timelocks
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct Primitives{
+    pub pks: Vec<bitcoin::PublicKey>,
```
apoelstra: I think we should parameterize this structure over the public key type. Also it'd be cool to just return references into the descriptor rather than copying everything, like

```
pub struct Primitives<'desc, 'stack, P> {
    pub pks: Vec<&'desc P>,
    pub hashlocks: Vec<(&'stack [u8], sha256d::Hash)>,
    pub timelocks: Vec<u32>,
}
```
and you should also have doccomments (even trivial/redundant ones) on all the fields

apoelstra: Also, I think `pks` should be a `HashSet` (which puts a `std::hash::Hash` requirement on `P`, which kinda sucks, but should be fine in practice), `hashlocks` should be a `HashMap<&'desc sha256d::Hash, &'a stack[u8]>`, and `timelocks` should also be a `HashSet`

apoelstra: There are a few advantages of using references like this

1. Less allocation/copying
2. It gives the user a way to determine which part of the descriptor a public key came from, in case there are multiple instances of the same key (by casting to `*const _` and comparing the pointers)
3. It improves documentation because it's clear where each piece of data comes from

apoelstra: Actually maybe we want to change the API entirely to

```
/// A satisfied Miniscript condition (e.g. signature requirement)
enum SatisfiedConstraint<'desc, 'stack, P> {
    PublicKey {
        key: &'desc P,
        sig: secp256k1::Signature,
    },
    HashLock {
        hash: &'desc sha256d::Hash,
        preimage: &'stack [u8],
    },
    RelativeTimeLock {
        time: &'desc u32,
    },
}
```
then
```
/// An iterator over all the constraints satisfied by a given
/// descriptor/scriptSig/witness stack tuple
pub struct SatisfiedConstraints {
    // ...
}

impl<'desc, 'stack, P> Iterator for SatisfiedConstraints {
    type Item = SatisfiedConstraint<'desc, 'stack, P>;

    fn next(&mut self) -> Option<SatisfiedConstraint<'desc, 'stack, P>> {
        // ...
    }
}
```
Does this make sense?

src/miniscript/evaluate.rs
```diff
@@ -0,0 +1,606 @@
...
+            }
+        }
+    }
+
+}
+
+/// This functions is equivalent to VERIFY. This pops the top element and checks if it is non-zero.
+/// Gives an error in all other cases
+fn evaluate_verify(stack: &mut Vec<StackElement>) -> Result<(), Error>{
+    let top = stack.pop().expect("Non-zero element for verify");
```
apoelstra: These `expect`s all make me nervous. Suggest changing the whole function body to
```
stack.pop() == StackElement::Satisfied
```
(This is also a logic change, from "anything but `Dissatisfied`" to "must be `Satisfied`", which I think is more correct ... in particular there are some non-`Dissatisfied` pushes that the Script interpreter will evaluate as false, and in general I'd prefer to be as strict as we can possibly get away with.)

apoelstra: I also think (though maybe you disagree) that it'd look a bit nicer to `impl` all these methods on `Vec<StackElement>`, like
```
impl Vec<StackElement> {
    fn verify(&mut self) -> Result<(), Error> {
        if stack.pop() == Some(StackElement::Satisfied) {
            Ok(())
        } else {
            Err(Error::CouldNotEvaluate)
        }
    }
}
```

src/descriptor/create_descriptor.rs
```diff
@@ -0,0 +1,410 @@
...
+        Ok((Descriptor::Wsh(ms), stack))
+    }
+    else if script_pubkey.is_p2sh(){
+        let (redeem_script, stack) = verify_p2sh(script_pubkey, script_sig)?;
+        if redeem_script.is_v0_p2wpkh()
+        {
+            //ensures that scriptsig after popping redeem script contains 0 elements.
+            //Therefore while calling verify_wpkh, an argument of Script::new() is passed instead
+            //of script_sig. The redeem_script becomes the script_pubkey
+            if stack.len() != 0{
```
apoelstra: nit: it's a bit more idiomatic to use `stack.is_empty()` rather than comparing `len` to 0. (And also for some non-`Vec` collections it's much more efficient).

This is something `clippy` will warn you about BTW ... if it's not too noisy (as I mentioned, it hasn't been compiling for me lately so I haven't used it) it's definitely worth running.

src/descriptor/create_descriptor.rs
```diff
@@ -0,0 +1,410 @@
...
+        _ => Err(Error::BadDescriptor),
+    }
+}
+
+/// Helper function which splits the scriptsig into 2 parts returns the corresponding elements.
+/// Usually used for scripts which have top element as Pk (p2pkh) or redeem script(p2sh).
+/// Converts the other script elements into Vec<StackElement>
+fn parse_scriptsig_top(script_sig: &bitcoin::Script) -> Result<(Vec<u8>, Vec<StackElement>), Error>
+{
+    let instructions : Vec<Instruction> = script_sig.iter(false).collect();
```
apoelstra: I think we should use `iter(true)` everywhere, as part of our "be as strict as possible" ethos.

src/miniscript/evaluate.rs
apoelstra: Then `Descriptor::interpret` could be replaced by a function `SatisfiedConstraints::from_descriptor<'desc, 'stack>(descriptor: &'desc Descriptor, stack: Vec<StackElement<'stack>>)` which creates this iterator, (and all of the code could exist in its own module, perhaps reexported as part of the public `descriptor` module).

sanket1729:
Opening a new PR for this based on new version of miniscript

PR #11, opened 2019-06-11
Changed node names time -> time_e, time_t -> time, hash_t -> hash

sanket1729:
Changed node names `time` -> `time_e`, `time_t` -> `time`, `hash_t` -> `hash`. 

There adds a convention that the T version is always keyword and everything else (if distinct) is keyword_X.


apoelstra:
utACK. But needs rebase.

sanket1729:
Rebased.

PR #12, opened 2019-06-15
Be consistent with 73 bytes signature

sanket1729:
Be consistent with 73 bytes signature everywhere. Checked other portions of the codebase as well.
Minor code cleanup: print removal and comment correction


apoelstra:
Going to close this since we will rewrite the compiler for the next version of Miniscript.

PR #14, opened 2019-07-11
Added interpreter for miniscript and descriptor creation logic

sanket1729:
Added interpreter for miniscript and support for creating descriptors from transaction objects. Addresses the comments in #9 . 

This would again be needed to be rebased after structure changes in AstElem.

### DIFF
--- a/src/descriptor/create_descriptor.rs
+++ b/src/descriptor/create_descriptor.rs
@@ -1,0 +1,465 @@
+//! Tools for creating Descriptor and witness stack from given scriptpubkey and corresponding
+//! scriptsig and witness.
+//!
+
+use bitcoin::{self, Script, PublicKey};
+
+use descriptor::Descriptor;
+use miniscript::Miniscript;
+use descriptor::satisfied_contraints::StackElement;
+use Error;
+use ToPublicKey;
+use bitcoin::blockdata::script::Instruction;
+use bitcoin::blockdata::opcodes;
+use hash160;
+use descriptor::satisfied_contraints::Error as IntError;
+
+/// Helper function for creating StackElement from Push instructions. Special case required for
+/// handling OP_PUSHNUM_1.
+/// Dissatisfied is Pushbytes0 and witness are mapped to Pushbytes instruction in scriptsig and
+/// satisfied is mapped as OP_PUSHNUM_1.
+/// Other opcodes are considered Non-standard in bitcoin core.
+/// Miniscript should not use other pushes apart from PUSHNUM_1. This will err on reciving anything
+/// which is not PUSHBYTES, OR PUSHNUM_1 as other things are expected to happen in Miniscript
+///
+/// NOTE: Miniscript pushes should only be either boolean, 1 or 0, signatures, and hash preimages.
+/// As per the current implementation, PUSH_NUM2 results in an error
+fn instr_to_stackelem<'txin>(ins: &Instruction<'txin>) -> Result< StackElement<'txin>, Error> {
+    match *ins {
+        //Also covers the dissatisfied case as PushBytes0
+        Instruction::PushBytes(v) => Ok(StackElement::from(v)),
+        Instruction::Op(opcodes::all::OP_PUSHNUM_1) => Ok(StackElement::Satisfied),
+        _ => Err(Error::BadScriptSig),
+    }
+}
+
+/// Helper function which splits the scriptsig into 2 parts returns the corresponding elements.
+/// Usually used for scripts which have top element as Pk (p2pkh) or redeem script(p2sh).
+/// Converts the other script elements into Vec<StackElement>
+fn parse_scriptsig_top(script_sig: &bitcoin::Script) -> Result<(Vec<u8>, Vec<StackElement>), Error>
+{
+    let stack: Result< Vec<StackElement>, Error> = script_sig.iter(true)
+        .map(|instr| instr_to_stackelem(&instr))
+        .collect();
+    let mut stack = stack?;
+    if let Some(StackElement::Push(pk_bytes)) = stack.pop() {
+        Ok((pk_bytes.to_vec(), stack))
+    } else {
+        Err(Error::InterpreterError(IntError::UnexpectedStackEnd))
+    }
+}
+
+
+/// Creates a pk descriptor based on scriptsig and script_pubkey.
+/// Pushes all remaining witness elements into witness<StackElement>
+fn verify_p2pk<'txin>(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &'txin bitcoin::Script,
+    witness: & [Vec<u8>],
+) -> Result<(Descriptor<PublicKey, hash160::Hash>, Vec<StackElement<'txin>>), Error>
+{
+    let script_pubkey_len = script_pubkey.len();
+    let pk_bytes = &script_pubkey.to_bytes();
+    let pk = bitcoin::PublicKey::from_slice(&pk_bytes[1..script_pubkey_len-1]).unwrap();
+
+    let stack: Result< Vec<StackElement>, Error> = script_sig.iter(true)
+        .map(|instr| instr_to_stackelem(&instr))
+        .collect();
+    if !witness.is_empty(){
+        Err(Error::NonEmptyWitness)
+    } else {
+        Ok((Descriptor::Pk(pk), stack?))
+    }
+}
+/// Helper to create a wpkh descriptor based on script_pubkey and witness. Validates the pubkey hash
+/// and pushes rest of witness to Vec<StackElement> as is.
+/// This does not check the signature, only creates the corresponding descriptor
+fn verify_p2wpkh<'txin>(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &bitcoin::Script,
+    witness: &'txin [Vec<u8>],
+) -> Result<(PublicKey, Vec<StackElement<'txin>>), Error>
+{
+    //script_sig must be empty
+    if !script_sig.is_empty(){
+        return Err(Error::NonEmptyScriptSig)
+    }
+    if let Some((pk_bytes, witness)) = witness.split_last() {
+        if let Ok(pk) = bitcoin::PublicKey::from_slice(pk_bytes) {
+            let addr = bitcoin::Address::p2wpkh(
+                &pk.to_public_key(),
+                bitcoin::Network::Bitcoin,
+            );
+            if addr.script_pubkey() != *script_pubkey {
+                return Err(Error::InterpreterError(IntError::PkEvaluationError(pk)))
+            }
+            let stack: Vec<StackElement> = witness.iter()
+                .map(|elem| StackElement::from(elem)).collect();
+            Ok((pk, stack))
+        }else {
+            Err(Error::InterpreterError(IntError::PubkeyParseError))
+        }
+    } else{
+        Err(Error::InterpreterError(IntError::UnexpectedStackEnd))
+    }
+}
+
+/// Helper for creating a wsh descriptor based on script_pubkey and witness. Validates the wsh
+/// hash based on witness script, pops the witness script from the stack and returns
+/// witness Vec<StackElement>. Does not interpret/check the witness against the miniscript inside
+/// the descriptor
+fn verify_wsh<'txin>(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &bitcoin::Script,
+    witness: &'txin [Vec<u8>],
+) -> Result<(Miniscript<PublicKey, hash160::Hash>, Vec<StackElement<'txin>>), Error>
+{
+    if !script_sig.is_empty(){
+        return Err(Error::NonEmptyScriptSig)
+    }
+    if let Some((witness_script, witness)) = witness.split_last() {
+        let witness_script = Script::from(witness_script.clone());
+        if witness_script.to_v0_p2wsh() != *script_pubkey {
+            return Err(Error::IncorrectScriptHash)
+        }
+        let ms = Miniscript::parse(&witness_script)?;
+        //only iter till len -1 to not include the witness script
+        let stack: Vec<StackElement> = witness.iter()
+            .map(|elem| StackElement::from(elem)).collect();
+        Ok((ms, stack))
+    } else{
+        Err(Error::InterpreterError(IntError::UnexpectedStackEnd))
+    }
+}
+
+
+/// Creates a pkh descriptor based on scriptsig and script_pubkey. Validates the hash checks for
+/// p2pkh against top element(pk) and pushes all remaining witness elements into witness<StackElement>
+fn verify_p2pkh<'txin>(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &'txin bitcoin::Script,
+    witness: & [Vec<u8>],
+) -> Result<(Descriptor<PublicKey, hash160::Hash>, Vec<StackElement<'txin>>), Error>
+{
+    let (pk_bytes, stack) = parse_scriptsig_top(script_sig)?;
+    if let Ok(pk) = bitcoin::PublicKey::from_slice(&pk_bytes) {
+        let addr = bitcoin::Address::p2pkh(
+            &pk.to_public_key(),
+            bitcoin::Network::Bitcoin,
+        );
+        if !witness.is_empty(){
+            return Err(Error::NonEmptyWitness)
+        }
+        if *script_pubkey != addr.script_pubkey() {
+            return Err(Error::IncorrectPubkeyHash)
+        }
+        Ok((Descriptor::Pkh(pk), stack))
+    } else{
+        Err(Error::InterpreterError(IntError::PubkeyParseError))
+    }
+}
+
+/// Helper for creating a p2sh descriptor based on script_pubkey and witness. Validates the p2sh
+/// hash based on redeem script, pops the redeem script from the script sig stack, translates other
+/// elements from scriptsig into Vec<StackElement>
+fn verify_p2sh<'txin>(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &'txin bitcoin::Script,
+) -> Result<(Script, Vec<StackElement<'txin>>), Error>
+{
+    let (redeem_script, stack) = parse_scriptsig_top(script_sig)?;
+    let redeem_script = Script::from(redeem_script);
+    if redeem_script.to_p2sh() != *script_pubkey {
+        return Err(Error::IncorrectScriptHash)
+    }
+    Ok((redeem_script, stack))
+}
+
+/// Figures out the the type of descriptor based on scriptpubkey, witness and scriptsig.
+/// Outputs a descriptor and witness_stack for all descriptors which can be directly fed into the
+/// interpreter. All script_sig and witness are translated into a single witness stack
+/// Vec<StackElement>
+/// 1. `PK`: creates a `Pk` descriptor and transfers the scriptsig to witness_stack<StackElement>
+/// 2. `Pkh`: Removes top element(pk) and validates pubkey hash, pushes rest of witness to
+/// witness_stack<StackElement> and outputs a `Pkh` descriptor
+/// 3. `Wphk`: translates witness to witness_stack<StackElement>, validates sig and pubkey hash
+/// and outputs a `Wpkh` descriptor
+/// 4. `Wsh`: pops witness script and checks wsh output hash, translates remaining witness elements
+/// to Vec<StackElement> and outputs a `Wsh` descriptor. Does not check miniscript inside the descriptor
+/// 5. `Bare`: translates script_sig to witness_stack<StackElement> and script_pubkey to miniscript
+/// 6. `Sh`: Checks redeem_script hash, translates remaining elements from script_sig to
+/// witness_stack<StackElement> and redeem script to miniscript. Does not check the miniscript
+/// 7. `ShWpkh`: Checks redeem_script hash, translates remaining elements from script_sig to
+/// witness_stack<StackElement> and validates `Wpkh` sig, pubkey.
+/// 8. `ShWsh`: Checks witness script hash, pops witness script and converts it to miniscript.
+/// translates the remaining witness to witness_stack<StackElement>
+pub fn witness_stack<'txin>(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &'txin bitcoin::Script,
+    witness: &'txin [Vec<u8>],
+) -> Result<(Descriptor<PublicKey, hash160::Hash>, Vec<StackElement<'txin>>), Error>
+{
+    if script_pubkey.is_p2pk(){
+        verify_p2pk(script_pubkey, script_sig, witness)
+    }else if script_pubkey.is_p2pkh(){
+        verify_p2pkh(script_pubkey, script_sig, witness)
+    }
+    else if script_pubkey.is_v0_p2wpkh(){
+        let (pk, stack) = verify_p2wpkh(script_pubkey, script_sig, witness)?;
+        Ok((Descriptor::Wpkh(pk), stack))
+    }
+    else if script_pubkey.is_v0_p2wsh(){
+        let (ms, stack) = verify_wsh(script_pubkey, script_sig, witness)?;
+        Ok((Descriptor::Wsh(ms), stack))
+    }
+    else if script_pubkey.is_p2sh(){
+        let (redeem_script, stack) = verify_p2sh(script_pubkey, script_sig)?;
+        if redeem_script.is_v0_p2wpkh()
+        {
+            //Therefore while calling verify_wpkh, an argument of Script::new() is passed instead
+            //of script_sig. The redeem_script becomes the script_pubkey
+            let (pk, stack) = verify_p2wpkh(&redeem_script, &Script::new(), witness)?;
+            Ok((Descriptor::ShWpkh(pk), stack))
+        }else if redeem_script.is_v0_p2wsh() {
+            //Therefore while calling verify_wpkh, an argument of Script::new() is passed instead
+            //of script_sig. The redeem_script becomes the script_pubkey
+            let (ms, stack) = verify_wsh(&redeem_script, &Script::new(), witness)?;
+            Ok((Descriptor::ShWsh(ms), stack))
+        }
+        else{
+            if !witness.is_empty(){
+                return Err(Error::NonEmptyWitness)
+            }
+            let ms = Miniscript::parse(&redeem_script)?;
+            Ok((Descriptor::Sh(ms), stack))
+        }
+    }
+    else{
+        //bare
+        let stack : Result <Vec<StackElement>, Error> = script_sig
+            .iter(true)
+            .map(|instr| instr_to_stackelem(&instr))
+            .collect();
+        if !witness.is_empty(){
+            return Err(Error::NonEmptyWitness)
+        }
+        let ms = Miniscript::parse(script_pubkey)?;
+        Ok((Descriptor::Bare(ms), stack?))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use ::{Descriptor, Miniscript};
+    use descriptor::create_descriptor::witness_stack;
+    use bitcoin::blockdata::script;
+    use bitcoin::{self, PublicKey};
+    use secp256k1::{self, Secp256k1, VerifyOnly};
+    use descriptor::satisfied_contraints::StackElement;
+    use miniscript::astelem::AstElem;
+    use bitcoin::blockdata::opcodes;
+    use ToPublicKey;
+
+    fn setup_keys_sigs(n: usize)
+                       -> ( Vec<PublicKey>, Vec<Vec<u8> >, secp256k1::Message, Secp256k1<VerifyOnly>) {
+        let secp_sign = secp256k1::Secp256k1::signing_only();
+        let secp_verify = secp256k1::Secp256k1::verification_only();
+        let msg = secp256k1::Message::from_slice(
+            &b"Yoda: btc, I trust. HODL I must!"[..]
+        ).expect("32 bytes");
+        let mut pks = vec![];
+        let mut sigs = vec![];
+        let mut sk = [0; 32];
+        for i in 1..n+1 {
+            sk[0] = i as u8;
+            sk[1] = (i >> 8) as u8;
+            sk[2] = (i >> 16) as u8;
+
+            let sk = secp256k1::SecretKey::from_slice(&sk[..]).expect("secret key");
+            let pk = PublicKey {
+                key: secp256k1::PublicKey::from_secret_key(
+                    &secp_sign,
+                    &sk,
+                ),
+                compressed: true,
+            };
+            let sig = secp_sign.sign(&msg, &sk);
+            let mut sigser = sig.serialize_der();
+            sigser.push(0x01); // sighash_all
+            pks.push(pk);
+            sigs.push(sigser);
+        }
+        (pks, sigs, msg, secp_verify)
+    }
+
+    #[test]
+    fn create_witness_stack(){
+        let (pks,sigs, _, _) = setup_keys_sigs(10);
+
+        //test pkh
+        let script_pubkey =  bitcoin::Address::p2pkh(
+            &pks[0],
+            bitcoin::Network::Bitcoin,
+        ).script_pubkey();
+        let script_sig = script::Builder::new()
+            .push_slice(&sigs[0])
+            .push_key(&pks[0])
+            .into_script();
+        let witness = vec![] as Vec<Vec<u8>>;
+
+        let (des, stack) = witness_stack(
+            &script_pubkey,
+            &script_sig,
+            &witness,
+        ).expect("Descriptor/Witness stack creation to succeed");
+        assert_eq!(Descriptor::Pkh(pks[0]), des);
+        assert_eq!(stack,
+                   vec![StackElement::Push(&sigs[0])]);
+
+        //test pk
+        let script_pubkey = script::Builder::new()
+            .push_key(&pks[0].to_public_key())
+            .push_opcode(opcodes::all::OP_CHECKSIG)
+            .into_script();
+        let script_sig = script::Builder::new()
+            .push_slice(&sigs[0])
+            .into_script();
+        let witness = vec![] as Vec<Vec<u8>>;
+
+        let (des, stack) = witness_stack(
+            &script_pubkey,
+            &script_sig,
+            &witness,
+        ).expect("Descriptor/Witness stack creation to succeed");
+        assert_eq!(Descriptor::Pk(pks[0]), des);
+        assert_eq!(stack,
+                   vec![StackElement::Push(&sigs[0])]);
+
+        //test wpkh
+        let script_pubkey =  bitcoin::Address::p2wpkh(
+            &pks[1],
+            bitcoin::Network::Bitcoin,
+        ).script_pubkey();
+        let script_sig = script::Builder::new().into_script();
+        let witness = vec![sigs[1].clone(), pks[1].clone().to_bytes()];
+        let (des, stack) = witness_stack(
+            &script_pubkey,
+            &script_sig,
+            &witness,
+        ).expect("Descriptor/Witness stack creation to succeed");
+        assert_eq!(Descriptor::Wpkh(pks[1]), des);
+        assert_eq!(stack, vec![StackElement::Push(&sigs[1])]);
+
+        //test Wsh: and(pkv, pk). Note this does not check miniscript.
+        let ms = Miniscript::from(AstElem::AndV(
+            Box::new(AstElem::Verify(Box::new(AstElem::Check(
+                Box::new(AstElem::Pk(pks[0].clone())))))),
+            Box::new(AstElem::Check(Box::new(AstElem::Pk(pks[1].clone()))))));
+        let script_pubkey =  bitcoin::Address::p2wsh(
+            &ms.encode(),
+            bitcoin::Network::Bitcoin,
+        ).script_pubkey();
+        let script_sig = script::Builder::new().into_script();
+        let witness = vec![sigs[1].clone(), sigs[0].clone(), ms.encode().to_bytes()];
+        let (des, stack) = witness_stack(
+            &script_pubkey,
+            &script_sig,
+            &witness,
+        ).expect("Descriptor/Witness stack creation to succeed");
+        assert_eq!(Descriptor::Wsh(ms.clone()), des);
+        assert_eq!(stack,
+                   vec![StackElement::Push(&sigs[1]),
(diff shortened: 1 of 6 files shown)


src/descriptor/mod.rs
```diff
@@ -535,8 +537,10 @@ impl<'de, Pk, Pkh> de::Deserialize<'de> for Descriptor<Pk, Pkh> where
         struct StrVisitor<Qk, Qkh>(PhantomData<(Qk, Qkh)>);
 
         impl<'de, Qk, Qkh> de::Visitor<'de> for StrVisitor<Qk, Qkh> where
-            Qk: fmt::Debug + str::FromStr,
-            Qkh: fmt::Debug + str::FromStr,
+            Qk: fmt::Debug + str::FromStr + serde::export::fmt::Display
+            + serde::export::Clone,
```
apoelstra: Please just use `fmt::Display` and `Clone`. Using the serde exports for stdlib things feels really strange.

src/descriptor/create_descriptor.rs
```diff
@@ -0,0 +1,468 @@
...
+        _ => Err(Error::BadDescriptor),
+    }
+}
+
+/// Helper function which splits the scriptsig into 2 parts returns the corresponding elements.
+/// Usually used for scripts which have top element as Pk (p2pkh) or redeem script(p2sh).
+/// Converts the other script elements into Vec<StackElement>
+fn parse_scriptsig_top(script_sig: &bitcoin::Script) -> Result<(Vec<u8>, Vec<StackElement>), Error>
+{
+    let instructions : Vec<Instruction> = script_sig.iter(true).collect();
```
apoelstra: You could do the stack element conversion here and avoid an allocation:
```
let stack: Vec<StackElement> = script_sig
    .iter(true)
    .map(instr_to_stackelem)
    .collect()?
```
Then you can pop the last element off of the stack like
```
if let Some(StackElement::Push(pk_bytes)) = stack.pop() {
    Ok(pk_bytes.to_vec(), stack)
} else {
    Err(Error::BadDescriptor)
}
```
I'm also not thrilled about the `.to_vec` in there, but it seems necessary to use `Script::to_p2sh` on the result, so I'll allow it. (The alternative is to do the `to_p2sh` conversion manually.)

src/descriptor/create_descriptor.rs
```diff
@@ -0,0 +1,468 @@
...
+        return Err(Error::BadDescriptor)
+    }
+    Ok((redeem_script, stack))
+}
+
+/// Figures out the the type of descriptor based on scriptpubkey, witness and scriptsig.
+/// Outputs a descriptor and witness_stack for all descriptors which can be directly fed into the
+/// interpreter. All script_sig and witness are translated into a single witness stack
+/// Vec<StackElement>
+/// 1) PK: creates a Pk descriptor and transfers the scriptsig to witness_stack<StackElement>
```
apoelstra: Can you put `Pk`, `Pkh`, etc in backticks? It makes the documentation a bit easier to read.

I'd also prefer you use `1.` `2.` etc rather than `1)` `2)` etc, but I don't feel strongly about it. cargo doc does the right thing regardless.

src/descriptor/create_descriptor.rs
```diff
@@ -0,0 +1,468 @@
...
+/// translates the remaining witness to witness_stack<StackElement>
+pub fn witness_stack<'txin>(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &'txin bitcoin::Script,
+    witness: &'txin [Vec<u8>],
+) -> Result<(Descriptor<PublicKey, hash160::Hash>, Vec<StackElement<'txin>>), Error>
+{
+    if script_pubkey.is_p2pk(){
+        verify_p2pk(script_pubkey, script_sig, witness)
+    }else if script_pubkey.is_p2pkh(){
```
apoelstra: nit: `else if` should always be on the same line as the `}` and have a space before it. Similarly the `{` should have a space before it. (Same comment applies to all of the below.

apoelstra: Actually never mind; I'll do a PR after this one which runs `cargo fmt` on everything and then CI-enforces that it passes in future.

src/descriptor/create_descriptor.rs
```diff
@@ -0,0 +1,468 @@
...
+/// and pushes rest of witness to Vec<StackElement> as is.
+/// This does not check the signature, only creates the corresponding descriptor
+fn verify_p2wpkh<'txin>(
+    script_pubkey: &bitcoin::Script,
+    script_sig: &bitcoin::Script,
+    witness: &'txin [Vec<u8>],
+) -> Result<(PublicKey, Vec<StackElement<'txin>>), Error>
+{
+    //script_sig must be empty
+    if *script_sig != Script::new(){
```
apoelstra: It's a bit more idiomatic to use `!script_sig.is_empty()` rather than comparing against `Script::new`.

src/miniscript/mod.rs
```diff
@@ -34,8 +34,8 @@ use bitcoin_hashes::hash160;
 pub mod astelem;
 pub mod decode;
 pub mod lex;
-pub mod satisfy;
 pub mod types;
+pub mod satisfy;
```
apoelstra: Was there a need for this reordering? `s` comes before `t`.

src/miniscript/mod.rs
```diff
@@ -83,6 +83,10 @@ impl<Pk, Pkh> Miniscript<Pk, Pkh> {
     pub fn into_inner(self) -> astelem::AstElem<Pk, Pkh> {
         self.0
     }
+    /// Extracts the `AstElem` representing the root of the miniscript
+    pub fn to_inner(&self) -> &astelem::AstElem<Pk, Pkh> {
```
apoelstra: Conventional naming for this is `as_inner`; `to_inner` should clone the inner.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+                height,
+            },
+            &Descriptor::Sh(ref miniscript) |
+            &Descriptor::Bare(ref miniscript) |
+            &Descriptor::ShWsh(ref miniscript) |
+            &Descriptor::Wsh(ref miniscript) => SatisfiedConstraints{
+                secp,
+                sighash,
+                public_key: None,
+                state : vec![(&miniscript.to_inner() , NodeEvaluationState::NoChildrenCompleted)],
```
apoelstra: nit: you are taking the address of `miniscript.to_inner` but this is already a reference

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+use ::{Error, ToPublicKeyHash};
+use secp256k1::{self, Signature, VerifyOnly};
+use ::{ToPublicKey, Descriptor};
+use bitcoin;
+use ::Error::CouldnotEvaluate;
+
+/// Definition of Stack Element of the Witness stack used for interpretation of Miniscript.
+/// All stack elements with vec![] go to Dissatisfied and vec![1] are marked to Satisfied.
+/// Others are directly pushed as witness
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
```
apoelstra: You can also make this `Copy`.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+    pub fn from(v: &'stack [u8]) -> StackElement<'stack> {
+        match v{
+            _ if *v == [1] => StackElement::Satisfied,
+            _ if *v == [] => StackElement::Dissatisfied,
+            _ => StackElement::Push(v)
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
```
apoelstra: Can also be `Copy`

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+}
+
+impl<'stack> StackElement<'stack>{
+
+    pub fn from(v: &'stack [u8]) -> StackElement<'stack> {
+        match v{
+            _ if *v == [1] => StackElement::Satisfied,
+            _ if *v == [] => StackElement::Dissatisfied,
+            _ => StackElement::Push(v)
+        }
```
apoelstra: It would be better to use `if` `else if` than to use a match here.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+    ///Hash 256 hashlock
+    Hash256HashLock(&'desc sha256d::Hash),
+    ///Hash160 hashlock
+    Hash160HashLock(&'desc hash160::Hash),
+    ///Ripemd160 hashlock
+    Ripemd160HashLock(&'desc ripemd160::Hash),
+}
+/// A satisfied Miniscript condition (Signature, Hashlock, Timelock)
+/// 'desc represents the lifetime of descriptor and `stack represents the lifetime of witness
+#[derive(Clone, Debug, Eq, PartialEq)]
```
apoelstra: Can also be `Copy`

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+pub enum SatisfiedConstraint<'desc, 'stack, Pk: 'desc, Pkh: 'desc> {
+    ///Public key and corresponding signature
+    PublicKey {
+        key: &'desc Pk,
+        sig: secp256k1::Signature,
+    },
+    ///Public keyhash, corresponding pubkey and signature
+    PublicKeyHasH {
```
apoelstra: capitalization

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+            _ if *v == [] => StackElement::Dissatisfied,
+            _ => StackElement::Push(v)
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum HashLockType<'desc>{
+    ///SHA 256 hashlock
+    Sha256HashLock(&'desc sha256::Hash),
```
apoelstra: You should drop `HashLock` from the names of all these variants

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+    state : Vec <(&'desc AstElem<Pk, Pkh>, NodeEvaluationState)>,
+    stack : Vec<StackElement<'stack>>,
+    altstack: Vec<StackElement<'stack>>,
+    age: &'desc u32,
+    height: &'desc u32,
+}
+
+impl<'desc, 'stack, Pk, Pkh> SatisfiedConstraints<'desc, 'stack, Pk, Pkh>{
+
+    pub fn from_desriptor(
```
apoelstra: spelling

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+/// descriptor/scriptSig/witness stack tuple
+pub struct SatisfiedConstraints<'desc, 'stack, Pk: 'desc, Pkh: 'desc> {
+    secp: &'desc secp256k1::Secp256k1<VerifyOnly>,
+    sighash: &'desc secp256k1::Message,
+    public_key: Option<&'desc Pk>,
+    state : Vec <(&'desc AstElem<Pk, Pkh>, NodeEvaluationState)>,
+    stack : Vec<StackElement<'stack>>,
+    altstack: Vec<StackElement<'stack>>,
+    age: &'desc u32,
+    height: &'desc u32,
```
apoelstra: Why are `age` and `height` references here? Aren't these associated with a transaction/txout and have nothing to do with the descriptor?

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+            &Descriptor::Pk(ref pk) |
+            &Descriptor::Pkh(ref pk) |
+            &Descriptor::ShWpkh(ref pk) |
+            &Descriptor::Wpkh(ref pk) => SatisfiedConstraints{
+                secp,
+                sighash,
+                public_key: Some(&pk),
+                state : vec![],
+                stack,
+                altstack: vec![] as Vec<StackElement<'stack>>,
```
apoelstra: I had no idea this sort of cast was legal, but it's unnecessary.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+                    }
+                } else {
+                //All the script has been executed.
+                //Check that the stack must contain exactly 1 satisfied element
+                if self.stack.pop() == Some(StackElement::Satisfied)
+                    && self.stack.is_empty() {
+                    return None
+                } else {
+                    return Some(Err(
+                        Error::CouldnotEvaluate("Script Is not satisfied".to_owned())))
```
apoelstra: You could add an explicit error variant for this rather than using string-typed errors

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+    Pk: ToPublicKey, Pkh: ToPublicKeyHash
+    {
+        match des {
+            &Descriptor::Pk(ref pk) |
+            &Descriptor::Pkh(ref pk) |
+            &Descriptor::ShWpkh(ref pk) |
+            &Descriptor::Wpkh(ref pk) => SatisfiedConstraints{
+                secp,
+                sighash,
+                public_key: Some(&pk),
```
apoelstra: Why not just put a `Miniscript(AstNode::Pk(pk))` onto the stack rather than special-casing this?

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+            self.push(StackElement::Satisfied);
+            Some(Ok(SatisfiedConstraint::RelativeTimeLock { time: n }))
+        } else {
+            Some(Err(Error::RelativeLocktimeNotMet(n.clone())))
+        }
+    }
+
+    /// Helper function to evaluate a Older and OlderF Node. Takes no arugment from stack
+    /// `n CHECKLOCKTIMEVERIFY 0NOTEQUAL` and `n CHECKLOCKTIMEVERIFY`
+    /// Ideally this should return int value as n: build_scriptint(t as i64)),
```
apoelstra: "Ideally" satoshi's insane integer encoding would never exist

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+    /// But since this is a T value and it will never be popped, it is safe to
```
apoelstra: It's not necessarily a T. The reason we don't need to copy the Script semantics is that Miniscript never evaluates integers.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+        n: &'desc u32,
+        height: &'desc u32,
+    ) -> Option<Result<SatisfiedConstraint<'desc, 'stack, Pk, Pkh>, Error>>
+        where Pk: ToPublicKey, Pkh: ToPublicKeyHash
+    {
+        if *height >= *n {
+            self.push(StackElement::Satisfied);
+            Some(Ok(SatisfiedConstraint::AbsoluteTimeLock { time: n }))
+        } else {
+            Some(Err(Error::AbsoluteLocktimeNotMet(n.clone())))
```
apoelstra: Unnecessary clone

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+                return Some(Err(Error::CouldnotEvaluate("Hash preimage must be 32".to_owned())))
+            }
+            if sha256::Hash::hash(preimage) == *hash {
+                self.push(StackElement::Satisfied);
+                Some(Ok(SatisfiedConstraint::HashLock {
+                    hash: HashLockType::Sha256HashLock(hash),
+                    preimage,
+                }))
+            } else {
+                Some(Err(Error::CouldnotEvaluate("SHA2 hash failed".to_owned())))
```
apoelstra: Please use multiple variants rather than strings. Strings undermine the type system and require extra allocations.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+    ///Only partial evaluation completed(used for Thresh nodes)
+    PartialThreshCompleted,
+    ///Multi-sig verification which represents currently k out n are satisfied.
+    ///This is required because the children of multi are not AstElem and
+    ///cannot be interpreted directly. Thus, we need to keep track of them
+    ///separately
+    PartialMultiCompleted(usize,usize),
+    ///All chilren have been evaluated
+    AllChildrenCompleted
+}
```
apoelstra: Why not use
```
struct NodeEvaluationState {
    n_evaluated: usize,
    n_satisfied: usize,
}
```
which would replace all of these variants and also replace the altstack, whose use is confusing.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+                    return Some(Err(
+                        Error::CouldnotEvaluate("Script Is not satisfied".to_owned())))
+                }
+            }
+        }
+
+        let (root, node_state) = self.state.pop().unwrap();//non-empty stack
+
+        let node_res = match (root, node_state) {
+            (&AstElem::True, NodeEvaluationState::NoChildrenCompleted) => {
```
apoelstra: You can match on `*root` and avoid having these `&`s everywhere

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+                    && self.stack.is_empty() {
+                    return None
+                } else {
+                    return Some(Err(
+                        Error::CouldnotEvaluate("Script Is not satisfied".to_owned())))
+                }
+            }
+        }
+
+        let (root, node_state) = self.state.pop().unwrap();//non-empty stack
```
apoelstra: I think it would be much simpler to do

```
while let Some((root, node_state)) = self.state.pop() {
    // ...
}
```
rather than tail-call recursing here. Then you could move the above "empty stack" check to after the loop, replace all your `Some(x)` with `return Some(x)`, and all your `None` with nothing.

src/descriptor/satisfied_contraints.rs
apoelstra: It would also be simpler (I think) to match only on the AST element and then consider the node state when deciding what to do.

apoelstra: For example, if you took my above advice and had a `eval_state: NodeEvaluationState`, you could match on `node.ast` and then if it were `AstElem::True`, you'd
```
debug_assert_eq!(node.n_evaluated, 0);
debug_assert_eq!(node.n_satisfied, 0);
self.stack.push(StackElement::Satisfied);
```
rather than the existing logic, which returns an untyped error in case you somehow get an `AstElem::True` with evaluated children.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+            }
+            Err(e) => Err(Error::Secp(e))
+        }
+    }else{
+        return Err(Error::CouldnotEvaluate("Expected sighash type".to_owned()));
+    }
+}
+
+trait StackOperations<'stack>{
+//    fn evaluate_verify(&mut self) -> Result<(), Error>;
```
apoelstra: please delete unused code

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
...
+                }
+            }
+            Err(e) => Err(Error::Secp(e))
+        }
+    }else{
+        return Err(Error::CouldnotEvaluate("Expected sighash type".to_owned()));
+    }
+}
+
+trait StackOperations<'stack>{
```
apoelstra: It would probably be better to do
```
pub struct Stack(pub Vec<StackElement>)
```
and implement these methods directly on `Stack`, than to use this trait here.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1482 @@
+
+use bitcoin_hashes::{Hash, hash160, sha256, sha256d, ripemd160};
+use miniscript::astelem::AstElem;
+use ::{Error, ToPublicKeyHash};
```
apoelstra: You should define your own `Error` type in this module.

src/lib.rs
```diff
@@ -248,12 +248,24 @@ pub enum Error {
...
+    /// Could not satisfy, absolute locktime not met
+    AbsoluteLocktimeNotMet(u32),
     /// General failure to satisfy
     CouldNotSatisfy,
     /// Typechecking failed
     TypeCheck(String),
+    /// Could not evaluate a script (fragment) because of a invalid signature
+    VerifySigFail(bitcoin::PublicKey, secp256k1::Signature, secp256k1::Message),
+    /// Could not evaluate a script (fragment) because of a invalid hashlock
+    VerifyHashFail(sha256::Hash),
```
apoelstra: All of the above errors (`BadDescriptor` and `Secp` are okay, though `BadDescriptor` should be split into multiple errors) should be wrapped up in a single `interpreter::Error` type)

apoelstra:
On further reflection I would like to replace "interpret" with "evaluate" everywhere.

src/descriptor/mod.rs
sanket1729: Don't really think when I did it this way. The compiler told me to use these imports.

src/miniscript/mod.rs
sanket1729: No reason. I don't know why I did this 🤦‍♂

src/descriptor/satisfied_contraints.rs
sanket1729: What should it be called otherwise?

src/descriptor/satisfied_contraints.rs
sanket1729: Yes, you are right.

src/descriptor/satisfied_contraints.rs
sanket1729: I cannot move pk into `AstElem::Pk(pk)` without cloning it explicitly. Our `AstElems::Pk` hold pk not references to them.

src/descriptor/satisfied_contraints.rs
apoelstra: `PublicKeyHash`

src/descriptor/satisfied_contraints.rs
apoelstra: Ah. Ok, then I think you should just clone the `pk`.

src/descriptor/satisfied_contraints.rs
sanket1729: Yeah, forgot to fix this comment. This comment was based on the previous version on miniscript for timeT

src/descriptor/satisfied_contraints.rs
sanket1729: I cannot avoid it, n is reference from the descriptor. I must clone it for the Error enum.

src/descriptor/satisfied_contraints.rs
sanket1729: Do you want to propagate this all the way to the API inputs? Take input as `Stack` instead of `Vec<StackElement>`

sanket1729: It's a bit ugly as I used functions of vector pop, push, len etc. Maybe I will redefine those again for this type

src/descriptor/satisfied_contraints.rs
sanket1729: @apoelstra , I don't think I can avoid the special case here. The function expects the state vector object to have references to AstElems with lifetimes outside of the function scope. I cannnot use anything created in the function along with it

sanket1729:
Resolved all the comments but the following because of the mentioned reasons
a) https://github.com/apoelstra/rust-miniscript/pull/14#discussion_r303193224 
b) https://github.com/apoelstra/rust-miniscript/pull/14#discussion_r303146304

src/descriptor/satisfied_contraints.rs
apoelstra: It's a u32. You never need to clone it.

src/descriptor/satisfied_contraints.rs
apoelstra: Just use `.0` to get the inner vector.

src/descriptor/satisfied_contraints.rs
apoelstra: still unresolved

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+    ///number of children evaluated
+    n_evaluated: usize,
+    ///number of children satisfied
+    n_satisfied: usize,
+}
+
+/// An iterator over all the constraints satisfied by a given
+/// descriptor/scriptSig/witness stack tuple
+pub struct SatisfiedConstraints<'desc, 'stack, Pk: 'desc, Pkh: 'desc> {
+    secp: &'desc secp256k1::Secp256k1<VerifyOnly>,
```
apoelstra: I know we have a lot of lifetimes already but we should probably give `secp` its own

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+    sighash: &'desc secp256k1::Message,
```
apoelstra: `sighash` doesn't need to be a reference, it's `Copy`

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+    public_key: Option<&'desc Pk>,
+    state : Vec <NodeEvaluationState<'desc, Pk, Pkh>>,
+    stack : Stack<'stack>,
+    age: u32,
+    height: u32,
+}
+
+pub struct Stack<'stack>(pub Vec<StackElement<'stack>>);
```
apoelstra: Needs doccomment

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+                    if res.is_some(){
+                        return res;
+                    }
+                }
+                AstElem::Alt(ref sub) |
+                AstElem::Swap(ref sub) |
+                AstElem::Check(ref sub) => {
+                    debug_assert_eq!(node_state.n_evaluated, 0);
+                    debug_assert_eq!(node_state.n_satisfied, 0);
+                    self.state.push(NodeEvaluationState { node: sub, n_evaluated: 0, n_satisfied: 0 });
```
apoelstra: I'd implement a helper function `self.push_evaluation_state(sub, 0, 0)` to save yourself a bunch of typing. But I don't have a strong preference

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+                    }
+                }
+                AstElem::AndV(ref left, ref right) => {
+                    debug_assert_eq!(node_state.n_evaluated, 0);
+                    debug_assert_eq!(node_state.n_satisfied, 0);
+                    self.state.push(NodeEvaluationState { node: right, n_evaluated: 0, n_satisfied: 0 });
+                    self.state.push(NodeEvaluationState { node: left, n_evaluated: 0, n_satisfied: 0 });
+                }
+                AstElem::OrB(ref left, ref _right) |
+                AstElem::AndB(ref left, ref _right) if node_state.n_evaluated == 0 => {
```
apoelstra: Why not match on all the boolean combinators at once here?

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+                    self.state.push(NodeEvaluationState { node: node_state.node, n_evaluated: 1, n_satisfied: 0 });
+                    self.state.push(NodeEvaluationState { node: left, n_evaluated: 0, n_satisfied: 0 });
+                }
+                AstElem::OrB(ref _left, ref right) |
+                AstElem::AndB(ref _left, ref right) if node_state.n_evaluated == 1 => {
```
apoelstra: Same here - and you can change this to just
```
                    self.state.push(NodeEvaluationState { node: node_state.node, n_evaluated: 2, n_satisfied: 0 });
                    self.state.push(NodeEvaluationState { node: right, n_evaluated: 0, n_satisfied: 0 });
```
rather than checking on the result of the `left` branch here, which feels confusing to me.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+                        self.state.push(NodeEvaluationState {
+                            node: right,
+                            n_evaluated: 0,
+                            n_satisfied: 0
+                        });
+                    } else {
+                        return Some(Err(Error::UnexpectedStackEnd))
+                    }
+                }
+                AstElem::AndB(ref _left, ref _right) if node_state.n_evaluated == 2 => {
```
apoelstra: Then this stuff can be

```
match (self.stack.pop(), self.stack.pop()) {
    (Some(StackElement::Satisfied), Some(StackElement::Satisfied))
        => self.stack.push(StackElement::Satisfied),
    (Some(_), Some(_)) => self.stack.push(StackElement::Dissatisfied),
    _ => return Some(Err(Error::UnexpectedStackEnd)),
}
```
and similar for the other combinators.

src/descriptor/satisfied_contraints.rs
apoelstra: Oh, no, I see - the reason is that we need to get the result of `right` off of the stack before evaluating `left`.

apoelstra: Regardless I would change this tree of `if`s below to
```
match self.stack.pop() {
    Some(StackElement::Dissatisfied) => ...
    Some(StackElement::Satisfied) => ...
    Some(StackElement::Push(_)) => unreachable!("encountered non-boolean result of subexpression"),
    None => return Some(Err(Error::UnexpectedStackElementPush)),
}
```
which is shorter than the if-tree and easier to follow.

I'm ambivalent about having an `unreachable!` (and requiring a typeck before evaluating anything, to make sure this is actually unreachable) or returning an error like you did, and allowing evaluation of all sorts of weird stuff. I guess your strategy is probably the better one.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+                                    filter(|empty| *empty).count();
+                                if nonsat == *k {
+                                    self.stack.push(StackElement::Dissatisfied);
+                                } else {
+                                    return Some(Err(Error::MissingExtraZeroMultiSig))
+                                }
+                            }
+                            None => return Some(Err(Error::UnexpectedStackEnd)),
+                            _ => {
+                                match self.stack.evaluate_thres_m(self.secp, self.sighash, &subs[0]) {
```
apoelstra: Please change `thres` to `thresh`

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+                        return Some(Err(Error::UnexpectedStackEnd))
+                    }
+                }
+                AstElem::ThreshM(ref k, ref subs) if node_state.n_evaluated == 0 => {
+                    let len = self.stack.len();
+                    if len < k + 1 {
+                        return Some(Err(Error::InsufficientSignaturesMultiSig))
+                    } else {
+                        //Non-sat case. If the first sig is empty, others k elements must
+                        //be empty.
```
apoelstra: I don't see why this should be true.

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+            }
+        } else {
+            Some(Err(Error::UnexpectedStackEnd))
+        }
+    }
+    /// Helper function to evaluate a checkmultisig which takes the top of the stack as input
+    /// signature and validates it in order of pubkeys. If the first signature
+    /// is satisfied by second public keys, other signatures are not checked
+    /// against the first pubkey.
+    ///
```
apoelstra: I don't understand this comment

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1610 @@
...
+    /// Checks top sig against pks starting from pk vec starting index start_index
```
apoelstra: There are no more `index` or `start_index`

src/descriptor/satisfied_contraints.rs
sanket1729: MultiSig takes k-sigs all of which must be valid or all must be empty. Anything else aborts the script as per standardness.

src/descriptor/satisfied_contraints.rs
apoelstra: Yep, I think you're right.

src/descriptor/satisfied_contraints.rs
apoelstra: Oh, ok, fair enough

sanket1729: For reference: 
https://github.com/bitcoin/bitcoin/blob/0822b44d8a68a3f29b12719523f6cb5253eb679c/src/script/interpreter.cpp#L1034

src/descriptor/satisfied_contraints.rs
sanket1729: Looks like sighash is not `Copy`. I cannot move it.

src/descriptor/satisfied_contraints.rs
sanket1729: I am really confused, this is `PublicKeyHash`.

src/descriptor/satisfied_contraints.rs
sanket1729: In the reference specification for satisfied constraints, you wanted these to be reference. 
See: https://github.com/apoelstra/rust-miniscript/pull/9#discussion_r294046117

sanket1729: I think your motivation was to compare two different satisfied constraints using these references.

src/descriptor/satisfied_contraints.rs
sanket1729: Reclarified it. Hopefully, it's more clear now.

src/descriptor/satisfied_contraints.rs
sanket1729: Another way to avoid this is to change the descriptors `Pk`, `Pkh`, `Wpkh` and `ShWpkh` from Pk to accept Miniscript. That is Pk(Pk) --> Pk(Miniscript<Pk, Pkh>).

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1587 @@
...
+    /// supports standard scripts
+    PkEvaluationError(bitcoin::PublicKey),
+    /// Miniscript requires the entire top level script to be satisfied.
+    ScriptSatisfactionError,
+    /// The Public Key hash check for the given pubkey. This occurs in `PkH`
+    /// node when the given key does not match to Hash in script.
+    PkHashVerifyFail(hash160::Hash),
+    /// Parse Error while parsing a `StackElement::Push` as a Pubkey. Both
+    /// 33 byte and 65 bytes are supported.
+    PubkeyParseEror,
```
apoelstra: typo

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1587 @@
...
+    /// The preimage to the hash function must be exactly 32 bytes.
+    HashPreimageLenghtMismatch,
```
apoelstra: typo

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1587 @@
...
+    /// Forward-secp related errors
+    Secp(secp256k1::Error),
+}
+
+impl error::Error for Error {
+    fn description(&self) -> &str {
+        ""
+    }
+    fn cause(&self) -> Option<&error::Error> {
+        None
```
apoelstra: Should return the secp error

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1587 @@
...
+    ScriptSatisfactionError,
+    /// The Public Key hash check for the given pubkey. This occurs in `PkH`
+    /// node when the given key does not match to Hash in script.
+    PkHashVerifyFail(hash160::Hash),
+    /// Parse Error while parsing a `StackElement::Push` as a Pubkey. Both
+    /// 33 byte and 65 bytes are supported.
+    PubkeyParseEror,
+    /// The preimage to the hash function must be exactly 32 bytes.
+    HashPreimageLenghtMismatch,
+    /// Got `StackElement::Satisfied` or `StackElement::DisSatisfied` when the
```
apoelstra: capitalization in `DisSatisfied`

src/descriptor/satisfied_contraints.rs
apoelstra: The last character in `PublicKeyHasH` should not be a capital

src/descriptor/satisfied_contraints.rs
apoelstra: https://docs.rs/secp256k1/0.14.1/secp256k1/struct.Message.html says it is

src/descriptor/satisfied_contraints.rs
```diff
@@ -0,0 +1,1587 @@
...
+    fn push(&mut self, elem: StackElement<'stack>) -> () {
+        self.0.push(elem);
+    }
+
+    ///wrapper for self.0.push()
+    fn split_off(&mut self, k: usize) -> Vec<StackElement<'stack>> {
+        self.0.split_off(k)
+    }
+
+    ///wrapper for self.0.push()
```
apoelstra: these `wrapper` comments should be changed

src/descriptor/satisfied_contraints.rs
apoelstra: Ok, but `&'desc u32` is also `Copy`.

src/descriptor/satisfied_contraints.rs
apoelstra: No, I think the existing solution is fine

src/descriptor/satisfied_contraints.rs
sanket1729: oops, I thought really hard over this, still could not figure out until you told about this explicitly

src/descriptor/satisfied_contraints.rs
sanket1729: Works!.  I should never trust IDE showing move errors.

src/descriptor/satisfied_contraints.rs
sanket1729: On other cases, None? I am not really sure how to do this.

apoelstra:
ack

PR #15, opened 2019-07-20
Refactor for miniscript to save type information

sanket1729:
-Change miniscript to save type information
-Change tests to use string inputs instead of `decode::Terminal`(previsouly `asterm::AstElem`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ default = []
 
 [dependencies]
 bitcoin = "0.18"
-bitcoin_hashes = "0.4"
+bitcoin_hashes = "0.7"
 
 [dependencies.secp256k1]
 version = "0.12"
--- a/src/descriptor/create_descriptor.rs
+++ b/src/descriptor/create_descriptor.rs
@@ -49,7 +49,6 @@ fn parse_scriptsig_top(script_sig: &bitcoin::Script) -> Result<(Vec<u8>, Vec<Sta
     }
 }
 
-
 /// Creates a pk descriptor based on scriptsig and script_pubkey.
 /// Pushes all remaining witness elements into witness<StackElement>
 fn verify_p2pk<'txin>(
@@ -256,9 +255,10 @@ mod tests {
     use bitcoin::{self, PublicKey};
     use secp256k1::{self, Secp256k1, VerifyOnly};
     use descriptor::satisfied_contraints::StackElement;
-    use miniscript::astelem::AstElem;
     use bitcoin::blockdata::opcodes;
     use ToPublicKey;
+    use std::str::FromStr;
+    use bitcoin_hashes::hash160;
 
     fn setup_keys_sigs(n: usize)
                        -> ( Vec<PublicKey>, Vec<Vec<u8> >, secp256k1::Message, Secp256k1<VerifyOnly>) {
@@ -312,7 +312,7 @@ mod tests {
             &script_sig,
             &witness,
         ).expect("Descriptor/Witness stack creation to succeed");
-        assert_eq!(Descriptor::Pkh(pks[0]), des);
+        assert_eq!(des_str!("pkh({})",pks[0]), des);
         assert_eq!(stack,
                    vec![StackElement::Push(&sigs[0])]);
 
@@ -331,7 +331,7 @@ mod tests {
             &script_sig,
             &witness,
         ).expect("Descriptor/Witness stack creation to succeed");
-        assert_eq!(Descriptor::Pk(pks[0]), des);
+        assert_eq!(des_str!("pk({})",pks[0]), des);
         assert_eq!(stack,
                    vec![StackElement::Push(&sigs[0])]);
 
@@ -347,14 +347,12 @@ mod tests {
             &script_sig,
             &witness,
         ).expect("Descriptor/Witness stack creation to succeed");
-        assert_eq!(Descriptor::Wpkh(pks[1]), des);
+        assert_eq!(des_str!("wpkh({})",pks[1]), des);
         assert_eq!(stack, vec![StackElement::Push(&sigs[1])]);
 
         //test Wsh: and(pkv, pk). Note this does not check miniscript.
-        let ms = Miniscript::from(AstElem::AndV(
-            Box::new(AstElem::Verify(Box::new(AstElem::Check(
-                Box::new(AstElem::Pk(pks[0].clone())))))),
-            Box::new(AstElem::Check(Box::new(AstElem::Pk(pks[1].clone()))))));
+        let ms = ms_str!("and_v(vc:pk({}),c:pk({}))",
+                     pks[0], pks[1]);
         let script_pubkey =  bitcoin::Address::p2wsh(
             &ms.encode(),
             bitcoin::Network::Bitcoin,
@@ -372,10 +370,8 @@ mod tests {
                         StackElement::Push(&sigs[0])]);
 
         //test Bare: and(pkv, pk). Note this does not check miniscript.
-        let ms = Miniscript::from(AstElem::OrB(
-            Box::new(AstElem::Check(Box::new(AstElem::Pk(pks[1].clone())))),
-            Box::new(AstElem::Swap(
-                Box::new(AstElem::Check(Box::new(AstElem::Pk(pks[1].clone()))))))));
+        let ms = ms_str!("or_b(c:pk({}),sc:pk({}))",
+                     pks[0], pks[1]);
         let script_pubkey =  ms.encode();
         let script_sig = script::Builder::new()
             .push_int(0)
@@ -393,10 +389,8 @@ mod tests {
                         StackElement::Push(&sigs[0])]);
 
         //test Sh: and(pkv, pk). Note this does not check miniscript.
-        let ms = Miniscript::from(AstElem::Check
-            (Box::new(AstElem::OrI(
-            Box::new(AstElem::Pk(pks[0].clone())),
-            Box::new(AstElem::Pk(pks[1].clone()))))));
+        let ms = ms_str!("c:or_i(pk({}),pk({}))",
+                     pks[0], pks[1]);
         let script_pubkey =  bitcoin::Address::p2sh(
             &ms.encode(),
             bitcoin::Network::Bitcoin,
@@ -418,11 +412,9 @@ mod tests {
                         StackElement::Satisfied]);
 
         //test Shwsh: and(pkv, pk). Note this does not check miniscript.
-        //Check with incorrect witness
-        let ms = Miniscript::from(AstElem::AndV(
-            Box::new(AstElem::Verify(Box::new(AstElem::Check(
-                Box::new(AstElem::Pk(pks[0].clone())))))),
-            Box::new(AstElem::Check(Box::new(AstElem::Pk(pks[1].clone()))))));
+        //This test passes incorrect witness argument.
+        let ms = ms_str!("and_v(vc:pk({}),c:pk({}))",
+                     pks[0], pks[1]);
         let script_pubkey =  bitcoin::Address::p2shwsh(
             &ms.encode(),
             bitcoin::Network::Bitcoin,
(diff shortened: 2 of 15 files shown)


sanket1729:
@apoelstra  Will squash and add commit messages later. Still trying to figure out what is causing the fuzzer to crash. The build in 1.22 should be easily fixed, but I am not able to understand the fuzzer crash.

src/miniscript/astelem.rs
```diff
@@ -472,56 +410,48 @@ impl<Pk, Pkh> expression::FromTree for AstElem<Pk, Pkh> where
...
             ),
             ("hash160", 1) => expression::terminal(
                 &top.args[0],
-                |x| hash160::Hash::from_hex(x).map(AstElem::Hash160)
+                |x| hash160::Hash::from_hex(x).map(Terminal::Hash160)
             ),
-            ("true", 0) => Ok(AstElem::True),
-            ("and_v", 2) => {
-                let expr = expression::binary(top, AstElem::AndV)?;
-                if let AstElem::AndV(_, ref right) = expr {
```
sanket1729: @apoelstra I don't see why this was added in the code

apoelstra: Oh, oops, this should be dropped. It's because we used to have a `t:` prefix that you're supposed to use instead. But we don't have this now.

apoelstra: Now that we have `t:` back we shouldn't drop this :)

Cargo.toml
```diff
@@ -15,7 +15,7 @@ default = []
 
 [dependencies]
 bitcoin = "0.18"
-bitcoin_hashes = "0.4"
+bitcoin_hashes = "0.5"
```
apoelstra: Can you change to 0.7 instead

src/descriptor/create_descriptor.rs
```diff
@@ -292,6 +292,17 @@ mod tests {
...
 
+    fn des_str(s: &str) -> Descriptor<PublicKey, hash160::Hash>
+    {
+        Descriptor::<PublicKey, hash160::Hash>::from_str(s).unwrap()
+    }
+
+    fn ms_str(s: &str) -> Miniscript<PublicKey, hash160::Hash>
+    {
+        Miniscript::<PublicKey, hash160::Hash>::from_str(s).unwrap()
+    }
```
apoelstra: Can you make these macros so that users don't have to do `&format!` themselves?

apoelstra:
Can you split out the commits which

1. change the tests to use strings
2. rename `astelem::AstElem` to `decode::Terminal`
3. adds the type info

src/descriptor/create_descriptor.rs
sanket1729: @apoelstra How about another file, util.rs for the utility functions and separating them from lib.rs.
Maybe the macros go there

src/miniscript/astelem.rs
sanket1729: Leaving this out for this PR now.

src/descriptor/create_descriptor.rs
apoelstra: Sure, or we could just throw them in `lib.rs` with a `#[cfg(test)]` on them. Whatever you think is cleaner

src/miniscript/types/mod.rs
```diff
@@ -14,12 +14,15 @@
...
 pub mod malleability;
+pub mod extra_props;
 
 use std::{error, fmt};
 
-use miniscript::astelem::AstElem;
+use miniscript::decode::Terminal;
 pub use self::correctness::{Correctness, Base, Input};
 pub use self::malleability::{Dissat, Malleability};
+pub use self::extra_props::ExtData;
```
apoelstra: `extra_props` is not added til a later commit, so this commit doesn't compile

sanket1729: @apoelstra , 
I don't think it's easy to create a perfect bijection between taking strings as test inputs and changing the existing tests from Miniscript types.
 
The only reason we required test string types inputs was because typing in tests for Mini
script nodes using Miniscript{node: astelem, ty:type, ext: ext} was really hard.

sanket1729: I can however make a commit which removes the tests which relied on this. And then readd with the string inputs

src/miniscript/astelem.rs
apoelstra: Sounds good

src/macros.rs
```diff
@@ -9,20 +9,14 @@ macro_rules! ms_str {
...
 /// Allows tests to create a descriptor directly from string as
 /// `des_str!("wsh(c:or_i(pk({}),pk({})))", pk1, pk2)`
 macro_rules! des_str {
     ($($arg:tt)*) => (Descriptor::from_str(&format!($($arg)*)).unwrap())
 }
 
-/// Allows tests to create a descriptor directly from string as
-/// `des_str!("wsh(c:or_i(pk({}),pk({})))", pk1, pk2)`
+/// Allows tests to create a concrete policy directly from string as
+/// `polic_str!("wsh(c:or_i(pk({}),pk({})))", pk1, pk2)`
```
apoelstra: typo `polic_str`

src/miniscript/astelem.rs
```diff
@@ -32,9 +32,28 @@ use script_num_size;
...
+            AstElem::NonZero(ref sub) => Some(('j', sub)),
+            AstElem::ZeroNotEqual(ref sub) => Some(('u', sub)),
+            _ => None,
+        }
+    }
+}
 
 /// All AST elements
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
```
apoelstra: Why drop `Hash`?

src/miniscript/astelem.rs
```diff
@@ -61,68 +80,50 @@ pub enum AstElem<Pk, Pkh> {
...
-
 impl<Pk, Pkh: Clone> AstElem<Pk, Pkh> {
     /// Convert an AST element with one public key type to one of another
     /// public key type
     pub fn translate_pk<Func, Q, Error>(
         &self,
         mut translatefn: Func,
     ) -> Result<AstElem<Q, Pkh>, Error>
-        where Func: FnMut(&Pk) -> Result<Q, Error>,
+        where Func: FnMut(&Pk) -> Result<Q, Error>
```
apoelstra: Why drop this `,`? `cargo fmt` is just going to put it back.

src/miniscript/mod.rs
```diff
@@ -44,48 +44,56 @@ use ToPublicKeyHash;
...
-where
-    Pk: Clone + fmt::Debug,
-    Pkh: Clone + fmt::Debug,
+impl<Pk, Pkh> Miniscript<Pk, Pkh>
+    where Pk : fmt::Debug + fmt::Display + Clone, Pkh: fmt::Debug + fmt::Display + Clone
 {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{:?}", self.0)
+    /// Add type information(Type and Extdata) to Miniscript based on
+    /// `AstElem` fragment. Dependant on display and clone because it is used
```
apoelstra: typo `Dependant`

src/miniscript/mod.rs
```diff
@@ -44,48 +44,56 @@ use ToPublicKeyHash;
...
+    /// in creating Miniscript from String
+    pub fn from_ast(t: astelem::AstElem<Pk, Pkh>) -> Result< Miniscript<Pk, Pkh>, Error> {
```
apoelstra: I wonder if this should be `pub` or not, and whether it ought to return a `Box<Miniscript>` rather than a `Miniscript`.

src/miniscript/decode.rs
```diff
@@ -66,28 +72,70 @@ macro_rules! match_token {
...
 }
 
-fn rewrap1<Pk, Pkh, F>(term: &mut Vec<AstElem<Pk, Pkh>>, wrap: F)
-where
-    F: FnOnce(Box<AstElem<Pk, Pkh>>) -> AstElem<Pk, Pkh>
-{
-    let top = term.pop().unwrap();
-    term.push(wrap(Box::new(top)));
+trait Decodable <Pk, Pkh>{
+    fn push_ast0(&mut self, ms : AstElem<Pk, Pkh>) -> Result<(), Error>
```
apoelstra: I think it'd be a bit cleaner to write a new `TerminalStack<Pk,Pkh>(Vec<Miniscript<Pk,Pkh>>)` type and implement these methods on them, rather than using a trait.

I also think they should be called `reduce0`, `reduce1`, `reduce2` rather than `push_` because they do more than just push.

src/miniscript/mod.rs
```diff
@@ -158,17 +166,24 @@ impl<Pk: ToPublicKey, Pkh: ToPublicKeyHash> Miniscript<Pk, Pkh> {
...
 impl<Pk, Pkh: Clone> Miniscript<Pk, Pkh> {
     pub fn translate_pk<F, Q, E>(&self, translatefn: F)
         -> Result<Miniscript<Q, Pkh>, E>
         where F: FnMut(&Pk) -> Result<Q, E>
     {
-        let inner = self.0.translate_pk(translatefn)?;
-        Ok(Miniscript(inner))
+        let inner = self.node.translate_pk(translatefn)?;
+        Ok(Miniscript{
+            //directly copying the type nad ext is safe because translating public
```
apoelstra: typo `nad`

src/miniscript/mod.rs
```diff
@@ -177,8 +192,14 @@ impl<Pk: Clone, Pkh> Miniscript<Pk, Pkh> {
         -> Result<Miniscript<Pk, Q>, E>
         where F: FnMut(&Pkh) -> Result<Q, E>
     {
-        let inner = self.0.translate_pkh(translatefn)?;
-        Ok(Miniscript(inner))
+        let inner = self.node.translate_pkh(translatefn)?;
+        Ok(Miniscript{
+            //directly copying the type nad ext is safe because translating public
```
apoelstra: typo `nad`

src/miniscript/types/mod.rs
```diff
@@ -80,7 +83,7 @@ pub enum ErrorKind {
     },
 }
 
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
```
apoelstra: Why drop `Hash`?

src/miniscript/types/mod.rs
```diff
@@ -377,10 +367,10 @@ pub trait Property: Sized {
...
         mut child: C,
     ) -> Result<Self, Error<Pk, Pkh>>
-    where
-        C: FnMut(usize) -> Option<Self>,
-        Pk: Clone,
-        Pkh: Clone,
+        where
+            C: FnMut(usize) -> Option<Self>,
+            Pk: Clone,
+            Pkh: Clone,
```
apoelstra: More formatting changes that `cargo fmt` is going to undo.

sanket1729: Somehow my editor does this weird stuff automatically. If I run `cargo fmt` it would change many other things which are not supposed to be a part of this PR. 
Or if you want, I can add a `cargo fmt` commit to the this PR? Once we have `cargo fmt` ed version of the codebase, I can run `cargo fmt` before raising a PR. 
From the next PR onwards I can run `cargo fmt` locally without

src/miniscript/astelem.rs
sanket1729: I guess I forgot to implement Hash to Miniscirpt type which showed a compile error here. 
And maybe I did this as a quickfix 🤦‍♂ . Will fix it

src/miniscript/mod.rs
sanket1729: This is useful in compiler code. The comment to this code was incorrect. The `Debug` and `Display` bounds are because of the `Error` type `type_check` code.

sanket1729:
Adding commits directly on top this, I will squash and rebase once the review is finished.

src/miniscript/types/mod.rs
apoelstra: No, we'll do another `cargo fmt` PR because there are probably 100+ style issues in this PR. I was just commenting here because this introduces new style issues without even changing any code.

My suggestion is to switch to emacs ;)

apoelstra:
Looks good now. No need to squash/rebase.

PR #16, opened 2019-07-22
Adds safety logic for satisfier

sanket1729:
The satisfier now creates witness as the following safety rules.

1) If you can satisfy a weak branch, you must satisfy it. 2
2) Satisfy using only the minimum required branches. 
3) If there are more weak branches than required to satisfy, satisfy the cheapest ones.

This PR builds on the refactor PR

### DIFF
--- a/src/miniscript/satisfy.rs
+++ b/src/miniscript/satisfy.rs
@@ -25,6 +25,7 @@ use bitcoin_hashes::{hash160, ripemd160, sha256, sha256d};
 use secp256k1;
 
 use Terminal;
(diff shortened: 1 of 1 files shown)


apoelstra:
gross. but ACK

sanket1729:
@apoelstra wrt to "gross", What could have been better in this?

apoelstra:
Not having the satisfaction logic making complex assumptions about peers' trustworthiness, and producing inefficient satisfactions based on those assumptions

PR #17, opened 2019-07-25
 Added PubkeyTrait and changed Miniscript/Terminals to a single generic

sanket1729:
Ther are two commits in total
- Add PubkeyTrait:
- Changed Semantic Policy to only accept Pkh for the below reasons. 

This allows to following invariant.
`Lift(Concrete) == Concrete -> Miniscript -> Script -> Miniscript -> Semantic`
Lifting concrete policy is equivalent to the following:
Compiling Concrete to Miniscript followed by
Encoding the miniscript to Script
Decoding it back to Miniscript
And finally lifting it back to Semantic

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -33,8 +33,9 @@ use expression;
 use miniscript::Miniscript;
 use Error;
 use Satisfier;
+use MiniscriptKey;
 use ToPublicKey;
-use ToPublicKeyHash;
+use ToHash160;
 
 mod create_descriptor;
 pub mod satisfied_contraints;
@@ -42,9 +43,9 @@ pub use self::create_descriptor::witness_stack;
 
 /// Script descriptor
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
-pub enum Descriptor<Pk, Pkh> {
+pub enum Descriptor<Pk : MiniscriptKey> {
     /// A raw scriptpubkey (including pay-to-pubkey)
-    Bare(Miniscript<Pk, Pkh>),
+    Bare(Miniscript<Pk>),
     /// Pay-to-Pubkey
     Pk(Pk),
     /// Pay-to-PubKey-Hash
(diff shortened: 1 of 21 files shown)


sanket1729:
@apoelstra, I think we need a better name for PublicKey trait as it confusing with bitcoin::PublicKey

apoelstra:
How about `MiniscriptKey`?

sanket1729:
Tests using `to_public_key()` of `DummyKey`(which we don't have any as of now) would also fail because of the pubkey point not being on curve.

src/policy/semantic.rs
```diff
@@ -25,15 +25,22 @@ use std::str::FromStr;
...
+/// normalization.
+/// Semantic Policies don't have a `Pk` variant because they should have been
+/// converted to Pkh by `into_lift()` function.
+/// This is allow to following invariant.
+/// `Lift(Concrete) == Concrete -> Miniscript -> Script -> Miniscript -> Semantic`
+/// Lifting concrete policy is equiavalent to the following:
+/// Compiling Concrete to Miniscript followed by
+/// Encoding the miniscript to Script
+/// Decoding it back to Miniscript
+/// And finally lifting it back to Semantic
```
apoelstra: I find this comment very confusing. It could be replaced by

```
Semantic policies store only hashes of keys to ensure that objects representing the same policy are lifted to the same `Semantic`, regardless of their choice of `pk` or `pk_h` nodes.
```

src/lib.rs
```diff
@@ -109,6 +109,44 @@ pub use descriptor::Descriptor;
...
+
+impl ToHash160 for hash160::Hash{
+
+    fn to_hash160(&self) -> hash160::Hash{
+        *self
+    }
+}
+///Public key trait which can be converted to Hash type
+pub trait MiniscriptKey: Clone + Eq + Ord + str::FromStr + fmt::Debug +fmt::Display {
+    type Hash: Clone + Eq + Ord + str::FromStr + fmt::Display + fmt::Debug + ToHash160 + hash::Hash;
```
apoelstra: `ToHash160` shouldn't be one of these bounds; it prevents use of stringly-typed hashes.

src/descriptor/mod.rs
```diff
@@ -457,15 +461,14 @@ impl<Pk, Pkh> expression::FromTree for Descriptor<Pk, Pkh> where
...
 }
 
-impl<Pk, Pkh> FromStr for Descriptor<Pk, Pkh> where
-    Pk: Clone + fmt::Debug + fmt::Display + str::FromStr,
-    Pkh: Clone + fmt::Debug + fmt::Display + str::FromStr,
+impl<Pk> FromStr for Descriptor<Pk> where
+    Pk: MiniscriptKey,
     <Pk as FromStr>::Err: ToString,
-    <Pkh as FromStr>::Err: ToString,
+    <<Pk as MiniscriptKey>::Hash as str::FromStr>::Err: ToString,
```
apoelstra: I think this bound is unnecessary because `Pk: MiniscriptKey` already implies `Pk::Hash: str::FromStr`.

src/descriptor/mod.rs
```diff
@@ -496,9 +498,8 @@ where
     }
 }
 
-impl <Pk, Pkh> fmt::Display for Descriptor<Pk, Pkh> where
-    Pk: fmt::Display,
-    Pkh: fmt::Display,
+impl <Pk> fmt::Display for Descriptor<Pk> where
+    Pk: MiniscriptKey,
```
apoelstra: nit: I'd just say `impl<Pk: MiniscriptKey> fmt::Display` rather than using a `where` clause here

src/descriptor/mod.rs
apoelstra: Oh, no, I see, it's a bound on `Pk::Hash::Err`. Ok.

src/lib.rs
sanket1729: How should I encode the to `Script` without that bound?

src/descriptor/mod.rs
sanket1729: the compiler complains if I remove that

src/policy/mod.rs
```diff
@@ -35,22 +35,27 @@ pub use self::concrete::Policy as Concrete;
...
 
 /// Trait describing script representations which can be lifted into
-/// an abstract policy, by discarding information
-pub trait Liftable<Pk, Pkh> {
+/// an abstract policy, by discarding information.
+/// After Lifting all policies are converted into `KeyHash(Pk::HasH)` to
+/// maintain the following invariant:
+/// `Lift(Concrete) == Concrete -> Miniscript -> Script -> Miniscript -> Semantic`
+pub trait Liftable<Pk: MiniscriptKey> {
+>>>>>>> 027a708... Refactor miniscript Key
```
apoelstra: Can you remove this `>>>>>`? I guess it's removed in a later commit?

Please use `git rebase -x 'cargo test'` to check that all of your commits compile

sanket1729: Nice, I was manually checking out all the commits till now. that's pretty handy

src/lib.rs
```diff
@@ -200,10 +233,17 @@ impl fmt::Display for DummyKeyHash {
...
 }
 
-impl ToPublicKeyHash for DummyKeyHash {
-    fn to_public_key_hash(&self) -> hash160::Hash {
-        use bitcoin_hashes::hex::FromHex;
-        hash160::Hash::from_hex("0000000000000000000000000000000000000000").unwrap()
+impl ToHash160 for DummyKeyHash{
+
+    fn to_hash160(&self) -> hash160::Hash{
+        hash160::Hash::from_str("").unwrap()
```
apoelstra: I wonder about the consistency requirement that hashing the output of `ToPublicKey::to_publickey` should be the same as the output of `MiniscriptKey::to_pubkeyhash` then `ToHash160::to_hash160`.

But what you've got now is fine, if we want to change it later we can. But I think we'll probably just have to be sure to document this.

apoelstra:
ACK mod nits - and can you check that all the commits compile?

sanket1729:
git rebase -x 'cargo test' is awesome :) . Changed as per the suggestions

src/lib.rs
apoelstra: encoding to `Script` already requires the `ToPublicKey` bound. Add a `Pk::Hash: ToHash160` bound as well I guess

apoelstra:
Looks good

PR #18, opened 2019-07-31
Add cargo fmt

sanket1729:
2 commits:
- output of `cargo fmt`
- Adds cargo fmt check to travis. 

There were different commands for formatting check for 1.22.0 and other versions. I just added it for stable version.


apoelstra:
ACK

PR #19, opened 2019-07-31
Added Ref counting for Terminals 

sanket1729:
- Added Ref-counting for decode::Terminal
- Changed compiler logic based on refcounting instead of cloning

### DIFF
--- a/src/policy/mod.rs
+++ b/src/policy/mod.rs
@@ -44,20 +44,20 @@ use MiniscriptKey;
 /// `Lift(Concrete) == Concrete -> Miniscript -> Script -> Miniscript -> Semantic`
 pub trait Liftable<Pk: MiniscriptKey> {
     /// Convert the object into an abstract policy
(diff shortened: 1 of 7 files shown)


src/policy/mod.rs
```diff
@@ -44,20 +44,20 @@ use MiniscriptKey;
 /// `Lift(Concrete) == Concrete -> Miniscript -> Script -> Miniscript -> Semantic`
 pub trait Liftable<Pk: MiniscriptKey> {
     /// Convert the object into an abstract policy
-    fn into_lift(self) -> Semantic<Pk>;
+    fn into_lift(&self) -> Semantic<Pk>;
```
sanket1729: Should probably be renamed to `lift` or `to_lift`

sanket1729:
Pushed the wrong branch, so the PR got closed automatically.

src/policy/mod.rs
sanket1729: Changed to lift

PR #23, opened 2019-08-08
Type-checking rule bugs

sanket1729:


src/miniscript/types/malleability.rs
```diff
@@ -187,7 +187,11 @@ impl Property for Malleability {
     }
 
     fn cast_true(self) -> Result<Self, ErrorKind> {
-        Ok(self)
+        Ok(Malleability {
+            dissat: Dissat::None,
+            safe: self.safe,
+            non_malleable: self.non_malleable,
+        })
```
apoelstra: FWIW I think the old code was technically correct because this cast can only be applied to V types

sanket1729: Yeah, I just matched this exactly to represent the website.

PR #24, opened 2019-08-08
Included wrappers for `l`,`u`,`t`.

sanket1729:
Changed the naming to be consistent with bitcoin.sipa.be/miniscript

### DIFF
--- a/src/miniscript/astelem.rs
+++ b/src/miniscript/astelem.rs
@@ -49,7 +49,10 @@ impl<Pk: MiniscriptKey> Terminal<Pk> {
             Terminal::DupIf(ref sub) => Some(('d', sub)),
             Terminal::Verify(ref sub) => Some(('v', sub)),
             Terminal::NonZero(ref sub) => Some(('j', sub)),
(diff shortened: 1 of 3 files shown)


src/miniscript/astelem.rs
```diff
@@ -345,17 +364,34 @@ where
...
+                }
+                Ok(expr)
+            }
             ("and_b", 2) => expression::binary(top, Terminal::AndB),
-            ("and_or", 3) => Ok(Terminal::AndOr(
+            ("andor", 3) => Ok(Terminal::AndOr(
                 expression::FromTree::from_tree(&top.args[0])?,
-                expression::FromTree::from_tree(&top.args[2])?,
                 expression::FromTree::from_tree(&top.args[1])?,
+                expression::FromTree::from_tree(&top.args[2])?,
```
apoelstra: I think we need to add a `and_n` entry here

sanket1729: Nice catch! I forgot to add this

PR #25, opened 2019-08-08
Added pkh dissatisfaction logic

sanket1729:


src/miniscript/satisfy.rs
```diff
@@ -52,6 +52,9 @@ pub trait Satisfier<Pk: MiniscriptKey> {
     }
 
     /// Given a keyhash, look up the signature and the associated key
+    /// Even if signatures for public key Hashes are not available, the users
+    /// can use this map to provide pkh -> pk mapping which can use useful
+    /// for dissatisfying pkh.
```
sanket1729: We could also create a separate function for this mapping.

apoelstra: Yeah, we should. I don't understand how users can use the current function to look up public keys?

src/miniscript/satisfy.rs
```diff
@@ -51,11 +51,23 @@ pub trait Satisfier<Pk: MiniscriptKey> {
...
+    }
+
+    /// Given a keyhash, look up the signature and the associated key
+    fn lookup_pkh(&self, p: &Pk::Hash) -> Option<(bitcoin::PublicKey, BitcoinSig)> {
+        self.lookup_pkh_pk(p)
+            .and_then(|ref pk| Some((pk.to_public_key(), self.lookup_pk(pk)?)))
+    }
+
     /// Wrapper around `lookup_pkh` that witness-serializes a signature
     fn lookup_pkh_wit(&self, p: &Pk::Hash) -> Option<Vec<Vec<u8>>> {
```
apoelstra: Can you delete this method now that it's unused?

PR #31, opened 2019-08-12
Op Code Count for Miniscript 

sanket1729:
Added non-push op-code count for miniscript fragments

### DIFF
--- a/src/miniscript/types/extra_props.rs
+++ b/src/miniscript/types/extra_props.rs
@@ -3,6 +3,7 @@
 
 use super::{Error, ErrorKind, Property};
 use script_num_size;
+use std::cmp;
 use MiniscriptKey;
 use Terminal;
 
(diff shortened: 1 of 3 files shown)


apoelstra:
Very weird that github sees 7 commits. 3 of them are below `master`.

apoelstra:
ACK the non-opcode-related commits.

As discussed in person, because e.g. CHECKMULTISIG is counted as `n+1` opcodes when executed but only 1 opcode when not executed, we need more than one opcode count.

apoelstra:
> As discussed in person, because e.g. CHECKMULTISIG is counted as `n+1` opcodes when executed but only 1 opcode when not executed, we need more than one opcode count.

Aren't all the opcodes push opcodes though? What were we thinking here?

apoelstra:
Never mind, in interpreter.cpp there is explicit code that increases the opcount by the number of keys. I will improve the comment that we have there..

PR #32, opened 2019-08-20
cleanup `satisfy` function; bump rust-bitcoin version; remove unsafe code

apoelstra:


src/miniscript/satisfy.rs
```diff
@@ -212,339 +238,521 @@ impl_tuple_satisfier!(&'a A, &'b B, &'c C, &'d D, &'e E, &'f F,);
...
+
+    /// Produce a satisfaction
+    pub fn satisfy<Pk: MiniscriptKey + ToPublicKey, Sat: Satisfier<Pk>>(
+        term: &Terminal<Pk>,
+        stfr: &Sat,
+    ) -> Self {
+        match *term {
+            Terminal::Pk(ref pk) => Satisfaction {
+                stack: Witness::signature(stfr, pk),
+                has_sig: true,
```
sanket1729: Shouldn't has_sig return true only if signature actually exists? This may return has_sig true even if the signature is not present. 
Same comment for pkh, multi and thresh.

src/miniscript/satisfy.rs
```diff
@@ -212,339 +238,521 @@ impl_tuple_satisfier!(&'a A, &'b B, &'c C, &'d D, &'e E, &'f F,);
...
+                    .collect::<Vec<_>>();
+
+                // Sort everything by (sat cost - dissat cost), except that
+                // satisfactions without signatures beat satisfactions with
+                // signatures
+                let mut sat_indices = (0..subs.len()).collect::<Vec<_>>();
+                sat_indices.sort_by_key(|&i| {
+                    let stack_weight = match (&sats[i].stack, &ret_stack[i].stack) {
+                        (&Witness::Unavailable, _) => i64::MAX,
+                        (_, &Witness::Unavailable) => i64::MIN,
```
sanket1729: Can this be unreachable.

src/miniscript/satisfy.rs
```diff
@@ -212,339 +238,521 @@ impl_tuple_satisfier!(&'a A, &'b B, &'c C, &'d D, &'e E, &'f F,);
...
-                    if let Some(sat) = sub.node.satisfy(satisfier, age, height) {
-                        ret.push(sat);
-                        satisfied += 1;
-                    } else {
-                        ret.push(dissat.clone());
+                // The above loop should have taken everything without a sig
+                // (since those were sorted higher than non-sigs). If there
+                // are remaining non-sig satisfactions this indicates a
+                // malleability vector
+                if k < sats.len() && !sats[k].has_sig {
```
sanket1729: k should be sat_indices[k]?

src/miniscript/satisfy.rs
```diff
@@ -212,339 +238,521 @@ impl_tuple_satisfier!(&'a A, &'b B, &'c C, &'d D, &'e E, &'f F,);
...
+                    Satisfaction {
+                        stack: Witness::Unavailable,
+                        has_sig: false,
```
sanket1729: Should this technically be `ret_stack.iter().any(|sat| sat.has_sig),`? maybe false also works

src/miniscript/satisfy.rs
```diff
@@ -212,339 +238,521 @@ impl_tuple_satisfier!(&'a A, &'b B, &'c C, &'d D, &'e E, &'f F,);
...
-                                .unwrap()
-                                .0;
-                            ret.remove(max_idx);
+                    match Witness::signature(stfr, pk) {
+                        Witness::Stack(sig) => {
+                            sigs.push(sig);
+                            sig_count += 1;
+                        },
+                        Witness::Unavailable => {
+                            sigs.push(vec![]);
```
sanket1729: Unavailable should not be push anything onto the stack.

sanket1729: There seems to something wrong in the satisfaction for multi. I think you may have forgotten to remove the vec![] from the sigs vector

src/miniscript/satisfy.rs
```diff
@@ -212,339 +238,521 @@ impl_tuple_satisfier!(&'a A, &'b B, &'c C, &'d D, &'e E, &'f F,);
...
+                let lnsat = Self::dissatisfy(&l.node, stfr);
+                let dissat_1 = Satisfaction {
+                    stack: Witness::combine(lnsat.stack, Witness::push_1()),
+                    has_sig: lnsat.has_sig,
+                };
+
+                let rnsat = Self::dissatisfy(&r.node, stfr);
+                let dissat_2 = Satisfaction {
+                    stack: Witness::combine(rnsat.stack, Witness::push_1()),
+                    has_sig: rnsat.has_sig,
```
sanket1729: should be push_0

src/miniscript/satisfy.rs
```diff
@@ -212,339 +238,521 @@ impl_tuple_satisfier!(&'a A, &'b B, &'c C, &'d D, &'e E, &'f F,);
...
-                for sub in subs.iter().rev() {
-                    ret.extend(sub.node.dissatisfy(satisfier)?);
+                Satisfaction {
+                    stack: subs
+                        .iter()
+                        .fold(Witness::empty(), |acc, sub| Witness::combine(
+                            Self::dissatisfy(&sub.node, stfr).stack,
+                            acc,
+                        )),
+                    has_sig: false,
```
sanket1729: Should this is be any sig instead of false?

src/miniscript/satisfy.rs
```diff
@@ -212,339 +238,521 @@ impl_tuple_satisfier!(&'a A, &'b B, &'c C, &'d D, &'e E, &'f F,);
...
-                    })
-                }
+                let mut sats = subs
+                    .iter()
+                    .map(|s| Self::satisfy(&s.node, stfr))
+                    .collect::<Vec<_>>();
+                // Start with the to-return stack set to all dissatisfactions
+                let mut ret_stack = subs
+                    .iter()
+                    .map(|s| Self::satisfy(&s.node, stfr))
```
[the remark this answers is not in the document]
apoelstra: Yep

src/miniscript/satisfy.rs
```diff
@@ -212,339 +238,521 @@ impl_tuple_satisfier!(&'a A, &'b B, &'c C, &'d D, &'e E, &'f F,);
...
+                                - d.iter().map(Vec::len).sum::<usize>() as i64
+                        }
+                    };
+                    (sats[i].has_sig, stack_weight)
+                });
 
-                if k == 0 {
-                    return Some(vec![]);
+                for i in 0..k {
+                    mem::swap(&mut ret_stack[i], &mut sats[i]);
```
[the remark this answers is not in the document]
apoelstra: Yep

src/miniscript/satisfy.rs
apoelstra: These are all OK; if something is `Unavailable` and `has_sig` it means "no satisfaction, and no satisfaction that a 3rd party could produce".

compare `Unavailable` and `!has_sig` which means "no [nonmalleable] satisfaction, but there might be a satisfaction that a 3rd party could produce".

So `has_sig` isn't really about whether a signature exists, but whether a signature *would* exist if the call had succeeded.

src/miniscript/satisfy.rs
apoelstra: No, because `pk_h` is `d` but if the `Satisfier` doesn't know the preimage to the pubkey hash, dissatisfaction will fail nonetheless.

src/miniscript/satisfy.rs
apoelstra: Yep (in the second instance)

src/miniscript/satisfy.rs
apoelstra: I could add an assertion that `ret_stack.iter().any(|sat| sat.has_sig)` is false; none of the dissatisfactions should have signatures (because dissatisfactions of `d` fragments never do) while none of the satisfactions should have signatures either (because if this branch is being triggered, they're all weak).

src/miniscript/satisfy.rs
apoelstra: Oh, you're right, I had a broken model of CHECKMULTISIG in my head.

src/miniscript/satisfy.rs
apoelstra: Yep

src/miniscript/satisfy.rs
apoelstra: Again, any sig should always be false (because every argument to a `Thresh` in a well-typed script is `d`). But in this case it's hard to even assert this because of the way I've structured the code -- I don't ever store the complete dissatisfactions, I just fold all the stacks into one and assume that `has_sig` is false for every one of them.

I guess I could stick an assertion into the iterator.

sanket1729: @apoelstra, my logic at viewing at `thresh` is that it should share code similarity with `or_b`. Which is why I got confused by this. 
I guess it makes sense to club the `or_b` with `and_b` since has_sig for `or_b` dissat should be false but is written explicitly. 
Anyway, it does not matter for correctness of the code. So, everything should be fine.

apoelstra: I added an assert on every individual (dis)satisfaction to check that it has no signatures.

src/miniscript/satisfy.rs
apoelstra: Added an assertion.

src/miniscript/satisfy.rs
```diff
@@ -191,364 +282,573 @@ macro_rules! impl_tuple_satisfier {
...
+                stack: subs.iter().fold(Witness::empty(), |acc, sub| {
+                    let nsat = Self::dissatisfy(&sub.node, stfr);
+                    assert!(!nsat.has_sig);
+                    Witness::combine(nsat.stack, acc)
+                }),
+                has_sig: false,
+            },
+            Terminal::ThreshM(_, ref keys) => Satisfaction {
+                stack: Witness::Stack(vec![vec![]; keys.len() + 1]),
+                has_sig: false,
```
sanket1729: should be `k +1` instead of `keys.len() + 1`

src/policy/compiler.rs
```diff
@@ -1396,6 +1395,8 @@ mod tests {
                 vec![],
                 vec![],
                 vec![],
+                vec![],
```
sanket1729: . This would bring back to where it was :)

apoelstra: Lol, yes.

(FWIW I was not just blindly fixing the test to match the output. I once again had a faulty model of CHECKMULTISIG)

PR #33, opened 2019-08-21
Add failing test case for HTLC script

thomaseizinger:
According to http://bitcoin.sipa.be/miniscript/, the following is a valid miniscript:
`c:or_i(and_v(v:sha256(0000000000000000000000000000000000000000000000000000000000000000),pk_h(A)),and_v(after(3333333),pk_h(B)))`

Now if I plug that into the `Descriptor` type and replace `A` and `B` with data of the correct length, I get the following error:

```
TypeCheck("fragment «and_v(after(3333333),pk_h(2222222222222222222222222222222222222222))» cannot accept children of types B and K"
```

Is this a bug or am I creating the HTLC wrong with miniscript?

Thanks in advance!

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -874,4 +874,9 @@ mod tests {
                 .into_script()
         );
     }
+
+    #[test]
+    fn htlc() {
+        Descriptor::<bitcoin::PublicKey>::from_str("c:or_i(and_v(v:sha256(0000000000000000000000000000000000000000000000000000000000000000),pk_h(1111111111111111111111111111111111111111)),and_v(after(3333333),pk_h(2222222222222222222222222222222222222222)))").expect("htlc should be valid descriptor");
+    }
 }


sanket1729:
Hi @thomaseizinger, I am not able to type-check your Miniscript into bitcoin.sipa.be/miniscript . 
Are you missing a `v:` in front of the second `and_v`?

The typecheck error is correct. `and_v` does not accept first child as `B`, it must be `v`. 

`c:or_i(and_v(v:sha256(0000000000000000000000000000000000000000000000000000000000000000),pk_h(A)),and_v(v:after(3333333),pk_h(B)))` this miniscript should typecheck. 

Instead of directly creating a miniscript, you can use the policy `or(and(pk(A),sha256(0000000000000000000000000000000000000000000000000000000000000001),and(pk(B),after(333333))))`and compile it later. 

We are planning on adding a compiler example today which explain this. 

For now, this is the fixed example. 
https://gist.github.com/sanket1729/e600a774a2634ffebe047654caa83dd2

thomaseizinger:
Hi @sanket1729,
thanks! Through trial and error, we actually also came to the conclusion of adding `v:` however, that creates an inefficient script no?
There is no need for an `OP_VERIFY` after `OP_CHECKLOCKTIMEVERIFY`. Is this just a limitation of miniscript?

sanket1729:
Interesting( and for deliberate reasons which I won't get into here), `OP_CLTV` does not drop the argument from the stack, which is why additional `OP_VERIFY` is needed to drop the timelock value from the stack. 
`n OP_CLTV` leaves `n` on the stack. 
`n OP_CTLV OP_VERIFY` leaves nothing on stack. 

This is not any limitation of Miniscript.

thomaseizinger:
Ah yeah! I am actually familiar with that behaviour of OP_CLTV.
We are currently in the process of changing our "handcrafted" HTLC to miniscript and overlooked that we actually have a OP_DROP after OP_CLTV. That makes the contracts (almost) equally efficient, thanks for explaining again! :)

apoelstra:
It's more idiomatic to use `DROP` rather than `VERIFY` after CLTV (and actually, our use of `VERIFY` prevents using CLTV/CSV with a 0 argument). So it's a bit unfortunate that Miniscript does not support `DROP` here ... but on the other hand it would be an extra special case to the language with very little benefit, which is why we don't have it.

PR #34, opened 2019-08-22
Add example for creating an HTLC with miniscript

thomaseizinger:
Based on the findings in #33, here is an example that creates an HTLC with miniscript.

### DIFF
--- a/examples/htlc.rs
+++ b/examples/htlc.rs
@@ -1,0 +1,49 @@
+// Miniscript
+// Written in 2019 by
+//    Thomas Eizinger <thomas@coblox.tech>
+//
+// To the extent possible under law, the author(s) have dedicated all
+// copyright and related and neighboring rights to this software to
+// the public domain worldwide. This software is distributed without
+// any warranty.
+//
+// You should have received a copy of the CC0 Public Domain Dedication
+// along with this software.
+// If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
+//
+
+//! Example: Create an HTLC with miniscript
+
+extern crate bitcoin;
+extern crate miniscript;
+
+use bitcoin::Network;
+use std::str::FromStr;
+
+fn main() {
+    let descriptor_template = format!(
+        "wsh(c:or_i(and_v(v:sha256({secret_hash}),pk_h({redeem_identity})),and_v(v:after({expiry}),pk_h({refund_identity}))))",
+        secret_hash = "1111111111111111111111111111111111111111111111111111111111111111",
+        redeem_identity = "2222222222222222222222222222222222222222",
+        refund_identity = "3333333333333333333333333333333333333333",
+        expiry = "4444"
+    );
+
+    let htlc_descriptor =
+        miniscript::Descriptor::<bitcoin::PublicKey>::from_str(&descriptor_template).unwrap();
+
+    assert_eq!(
+        format!("{:x}", htlc_descriptor.script_pubkey()),
+        "0020b822548461760c6a7c3c51c7fdaa0ccbdc69ae39a66b752ec8a4772bfdd41e64"
+    );
+
+    assert_eq!(
+        format!("{:x}", htlc_descriptor.witness_script()),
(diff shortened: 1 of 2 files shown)


examples/htlc.rs
```diff
@@ -0,0 +1,49 @@
+// Miniscript
+// Written in 2019 by
+//     Andrew Poelstra <apoelstra@wpsoftware.net>
```
apoelstra: You're welcome to put your own name on this btw :)

apoelstra:
Is it possible to start from a HTLC policy and compile it? Or does Rust make it difficult to enable the "compiler" flag?

If not, could you `.to_lift()` the descriptor and assert that it equals the correct semantic policy?

examples/htlc.rs
thomaseizinger: Ups, sorry :sweat_smile: 
I just copy-pasted another example file. Fixed!

thomaseizinger:
> Is it possible to start from a HTLC policy and compile it? Or does Rust make it difficult to enable the "compiler" flag?

It is not particularly difficult but slightly cumbersome. One can add a section to `Cargo.toml` to specify required features:

```toml
[[example]]
name = "htlc"
required-features = ["compiler"]
```

However, in Rust 2015, this makes the Rust compiler turn off "autodiscovery" of other examples and we would have to list all the others then as-well.

I briefly tried to change the edition to 2018 but that caused many compile errors.

> If not, could you .to_lift() the descriptor and assert that it equals the correct semantic policy?

Did that instead!

sanket1729:
Hi @thomaseizinger, thanks for converting this into an example. 

I am curious, how did you obtain the HTLC miniscript?  
`or(10@and(sha256(H),pk(A)),1@and(after(10),pk(B)))` (note the odds) compiles to this miniscript:
`andor(c:pk(A),sha256(H),and_v(vc:pk_h(B),after(10)))`. Even without the odds, this compiles to a different minsicript. 

The example you gave is a valid htlc miniscript, just not the optimal compiled one. 

Generally speaking, it is always better to let the compiler construct Miniscripts instead of manually creating them.

thomaseizinger:
@sanket1729

> I am curious, how did you obtain the HTLC miniscript?

we just wrote it ourselves :sweat_smile: 

> Generally speaking, it is always better to let the compiler construct Miniscripts instead of manually creating them.

Honestly, I couldn't figure out how to use the compiler in this crate, so we opted for using the `Descriptor` directly.
Maybe this example should make use of the compiler to showcase that?
Could you point me into the right direction of how to achieve this?

If we were to do that, we would still need to answer the question of how we deal with the examples. I see two ways forward:

- We stay on Rust 2015 for this crate and list all examples manually in the `Cargo.toml`. Manually listing the example allows us to define the required features.
- We port the crate to Rust 2018 and only list the example where we have to specify the required feature

Rust 2018 would require bumping the minimum supported Rust version to 1.31.

sanket1729:
Based on my previous discussions with @apoelstra, I think we have to stick with rust 2015 and list all examples. But I will wait for his comment on this issue before moving forward.

thomaseizinger:
> Based on my previous discussions with @apoelstra, I think we have to stick with rust 2015 and list all examples. But I will wait for his comment on this issue before moving forward.

Fine with me :)
In the meantime, can you point me to how to use the compiler feature of the crate?

Is `miniscript::Descriptor` still the entry point if all I have is a policy?

sanket1729:
@thomaseizinger, I wrote a quick example here.  
https://github.com/sanket1729/rust-miniscript/blob/pr-34/examples/htlc.rs constructs a Descriptor from Policy. 

Please let me know if you still struggling.

apoelstra:
We are not bumping the min compiler version to 1.31.

apoelstra:
But I definitely do think we need a compiler example - so I think having a long/messy Cargo.toml is the best way forward.

thomaseizinger:
Revised the example to use the policy compiler! Thanks for the input @sanket1729!

sanket1729:
tAck d5cd66eb9d5c1281c4db79f0a28c080d6f9f94d8

sanket1729:
Hi @thomaseizinger , Please note that we had a bug in `andor` encoding logic which is now fixed and htlc test was changed. Please check if that affects you.

PR #36, opened 2019-08-22
Compilation of `after` and `older` is flipped according to the documentation

thomaseizinger:
According to http://bitcoin.sipa.be/miniscript/, `after` should be compiled to  `OP_CLTV` and `older` should be compiled to `OP_CSV`.

The current code does the opposite.

Is this a bug or is the documentation wrong?

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -528,6 +528,8 @@ where
 
 #[cfg(test)]
 mod tests {
+    use bitcoin::blockdata::opcodes::all::{OP_CLTV, OP_CSV};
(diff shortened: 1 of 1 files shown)


sanket1729:
This is indeed a bug, thanks for looking into this and finding out the differences in implementations. I think we should be consistent with Pieter's C++ code and Ivy's code for `older/after` even though I think the way we did it is more to me. 

@apoelstra, I have made a PR to fix this #37 .

thomaseizinger:
Closed in favor of https://github.com/apoelstra/rust-miniscript/pull/37.

PR #37, opened 2019-08-22
Older-after interchange bug

sanket1729:
Looks like we are inconsistent with Pieter's and Ivy's implementation with Older and After. Changed


apoelstra:
Can you cherry-pick @thomaseizinger's unit test commit?

sanket1729:
@apoelstra, done

PR #42, opened 2019-09-06
fix some API bugs; new version

apoelstra:


sanket1729:
tACK b44a05fcf2fae0f0716b7a9f189de5b34a1cd4f7

practicalswift:
@apoelstra @sanket1729 What bug was the `thresh_m` reversal bug? :-) Do you have an example to share?

apoelstra:
@practicalswift when you try to satisfy a `thresh_m`, the signatures are in the wrong order.

I don't have a good example; I encountered it in Liquid, but our Miniscript code there does a bunch of complicated munging to make Miniscript compatible with the deployed production network, so it's not so easy to extract a simple test case.

PR #49, opened 2019-09-08
Changed typing rules

sanket1729:


apoelstra:
This all looks great. Thanks for being so careful!

apoelstra:
Note for future readers - this PR (re-)reverses the SHA256d hashes because after discussion on IRC, we agreed that Miniscript encoding shouldn't use reversed hashes.

This gives us consistency with script encoding, is the more natural thing to do, and also makes the hashes consistent (previously we'd reverse HASH256 but nothing else).

PR #63, opened 2019-11-26
add tests for script deserialization

joemphilips:
While I was trying to [port this library into C#](https://github.com/MetacoSA/NBitcoin/pull/764) , I found a bug in script deserialization.

I was trying to fix it by myself, but I'm afraid that the spec will change again. So I will not do it for now (unless we are sure that the spec won't change drastically anymore.)

Here is the failing case, tell me if I'm doing something wrong.

### DIFF
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -438,6 +438,7 @@ mod tests {
         match (ms, valid) {
             (Ok(ms), true) => {
                 assert_eq!(format!("{:x}", ms.encode()), expected_hex);
+                assert_eq!(format!("{:x}", Miniscript::parse(&ms.encode()).unwrap().encode()), expected_hex);
                 assert_eq!(ms.ty.mall.non_malleable, non_mal);
                 assert_eq!(ms.ty.mall.safe, need_sig);
                 assert_eq!(ms.ext.ops_count_sat.unwrap(), ops);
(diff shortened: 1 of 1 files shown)


apoelstra:
Thanks for this report - sorry for letting it drop. I will try to investigate today.

apoelstra:
offending script:
`OP_IF OP_SIZE OP_PUSHBYTES_1 20 OP_EQUALVERIFY OP_HASH256 OP_PUSHBYTES_32 131772552c01444cd81360818376a040b7c3b2b7b0a53550ee3edde216cec61b OP_EQUAL OP_ELSE OP_0 OP_ENDIF OP_VERIFY OP_SIZE OP_PUSHBYTES_1 20 OP_EQUALVERIFY OP_SHA256 OP_PUSHBYTES_32 ec4916dd28fc4c10d78e287ca5d9cc51ee1ae73cbfde08c6b37324cbfaac8bc5 OP_EQUALVERIFY OP_PUSHNUM_1`

or `6382012088aa20131772552c01444cd81360818376a040b7c3b2b7b0a53550ee3edde216cec61b876700686982012088a820ec4916dd28fc4c10d78e287ca5d9cc51ee1ae73cbfde08c6b37324cbfaac8bc58851`

apoelstra:
Yep, I see ... the problem is that when we are parsing backward after seeing `VERIFY`, we parse `OP_EQUAL` and think we're in a pubkeyhash construction, but we're actually in a hashlock construction. We need to split up `Terminal::PkH` into a `NonTerminal::PkHOrHashLock` so that we can differentiate these cases.

Great find!

apoelstra:
This PR needs an update to rename `c:pk` to `pk`, `thresh_m` to `multi`, and to use `Segwitv0Script` rather than `Miniscript::<bitcoin::PublicKey>`.

It still does not pass by the way, even with these fixes.

sanket1729:
@apoelstra This test has been included in the #78  and this can be closed.

PR #70, opened 2020-01-23
Miniscript iterators

dr-orlovsky:
I'll start with this sample one for concept (n)ACK and than will work on other types (like public keys)

### DIFF
--- a/src/miniscript/iter.rs
+++ b/src/miniscript/iter.rs
@@ -14,7 +14,6 @@
 
 use super::decode::Terminal;
 use super::{Miniscript, MiniscriptKey, ScriptContext};
-use std::collections::VecDeque;
 use std::ops::Deref;
 use std::sync::Arc;
 
@@ -51,30 +50,27 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
     /// Enumerates all child nodes of the current AST node (`self`) and returns a `Vec` referencing
     /// them.
     pub fn branches(&self) -> Vec<&Miniscript<Pk, Ctx>> {
-        use Terminal::*;
-        match &self.node {
-            &PkK(_) | &PkH(_) | &Multi(_, _) => vec![],
+        match self.node {
+            Terminal::PkK(_) | Terminal::PkH(_) | Terminal::Multi(_, _) => vec![],
 
-            &Alt(ref node)
-            | &Swap(ref node)
-            | &Check(ref node)
-            | &DupIf(ref node)
-            | &Verify(ref node)
-            | &NonZero(ref node)
-            | &ZeroNotEqual(ref node) => vec![node.deref()],
+            Terminal::Alt(ref node)
+            | Terminal::Swap(ref node)
+            | Terminal::Check(ref node)
+            | Terminal::DupIf(ref node)
+            | Terminal::Verify(ref node)
+            | Terminal::NonZero(ref node)
+            | Terminal::ZeroNotEqual(ref node) => vec![node],
 
-            &AndV(ref node1, ref node2)
-            | &AndB(ref node1, ref node2)
-            | &OrB(ref node1, ref node2)
-            | &OrD(ref node1, ref node2)
-            | &OrC(ref node1, ref node2)
-            | &OrI(ref node1, ref node2) => vec![node1.deref(), node2.deref()],
+            Terminal::AndV(ref node1, ref node2)
+            | Terminal::AndB(ref node1, ref node2)
+            | Terminal::OrB(ref node1, ref node2)
+            | Terminal::OrD(ref node1, ref node2)
+            | Terminal::OrC(ref node1, ref node2)
+            | Terminal::OrI(ref node1, ref node2) => vec![node1, node2],
 
-            &AndOr(ref node1, ref node2, ref node3) => {
-                vec![node1.deref(), node2.deref(), node3.deref()]
-            }
+            Terminal::AndOr(ref node1, ref node2, ref node3) => vec![node1, node2, node3],
 
-            &Thresh(_, ref node_vec) => node_vec.iter().map(Arc::deref).collect(),
+            Terminal::Thresh(_, ref node_vec) => node_vec.iter().map(Arc::deref).collect(),
 
             _ => vec![],
         }
@@ -87,9 +83,9 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
     /// To obtain a list of all public keys within AST use [`iter_pk()`] function, for example
     /// `miniscript.iter_pubkeys().collect()`.
     pub fn get_leaf_pk(&self) -> Vec<Pk> {
-        match self.node.clone() {
-            Terminal::PkK(key) => vec![key],
-            Terminal::Multi(_, keys) => keys,
+        match self.node {
+            Terminal::PkK(ref key) => vec![key.clone()],
+            Terminal::Multi(_, ref keys) => keys.clone(),
             _ => vec![],
         }
     }
@@ -98,21 +94,21 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
     /// Otherwise returns an empty `Vec`.
     ///
     /// For each public key the function computes hash; for each hash of the public key the function
-    /// returns it's cloned copy.
+    /// returns it cloned copy.
     ///
     /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
     /// To obtain a list of all public key hashes within AST use [`iter_pkh()`] function,
     /// for example `miniscript.iter_pubkey_hashes().collect()`.
     pub fn get_leaf_pkh(&self) -> Vec<Pk::Hash> {
-        match self.node.clone() {
-            Terminal::PkH(hash) => vec![hash],
-            Terminal::PkK(key) => vec![key.to_pubkeyhash()],
-            Terminal::Multi(_, keys) => keys.iter().map(Pk::to_pubkeyhash).collect(),
+        match self.node {
+            Terminal::PkH(ref hash) => vec![hash.clone()],
+            Terminal::PkK(ref key) => vec![key.to_pubkeyhash()],
+            Terminal::Multi(_, ref keys) => keys.iter().map(Pk::to_pubkeyhash).collect(),
             _ => vec![],
         }
     }
 
-    /// Returns `Vec` of [PubkeyOrHash] entries, representing either public keys or public key
+    /// Returns `Vec` of [PkPkh] entries, representing either public keys or public key
     /// hashes, depending on the data from the current miniscript item. If there is no public
     /// keys or hashes, the function returns an empty `Vec`.
     ///
@@ -120,12 +116,57 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
     /// To obtain a list of all public keys or hashes within AST use [`iter_pk_pkh()`]
     /// function, for example `miniscript.iter_pubkeys_and_hashes().collect()`.
     pub fn get_leaf_pk_pkh(&self) -> Vec<PkPkh<Pk>> {
-        use self::PkPkh::*;
-        match self.node.clone() {
-            Terminal::PkH(hash) => vec![HashedPubkey(hash)],
-            Terminal::PkK(key) => vec![PlainPubkey(key)],
-            Terminal::Multi(_, keys) => keys.into_iter().map(PlainPubkey).collect(),
+        match self.node {
+            Terminal::PkH(ref hash) => vec![PkPkh::HashedPubkey(hash.clone())],
+            Terminal::PkK(ref key) => vec![PkPkh::PlainPubkey(key.clone())],
+            Terminal::Multi(_, ref keys) => keys
+                .into_iter()
+                .map(|key| PkPkh::PlainPubkey(key.clone()))
+                .collect(),
             _ => vec![],
+        }
+    }
+
+    /// Returns `Option::Some` with cloned n'th public key from the current miniscript item,
+    /// if any. Otherwise returns `Option::None`.
+    ///
+    /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
+    pub fn get_nth_pk(&self, n: usize) -> Option<Pk> {
+        match (&self.node, n) {
+            (&Terminal::PkK(ref key), 0) => Some(key.clone()),
+            (&Terminal::Multi(_, ref keys), _) => keys.get(n).cloned(),
+            _ => None,
+        }
+    }
+
+    /// Returns `Option::Some` with hash of n'th public key from the current miniscript item,
+    /// if any. Otherwise returns `Option::None`.
+    ///
+    /// For each public key the function computes hash; for each hash of the public key the function
+    /// returns it cloned copy.
+    ///
+    /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
+    pub fn get_nth_pkh(&self, n: usize) -> Option<Pk::Hash> {
+        match (&self.node, n) {
+            (&Terminal::PkH(ref hash), 0) => Some(hash.clone()),
+            (&Terminal::PkK(ref key), 0) => Some(key.to_pubkeyhash()),
+            (&Terminal::Multi(_, ref keys), _) => keys.get(n).map(Pk::to_pubkeyhash),
+            _ => None,
+        }
+    }
+
+    /// Returns `Option::Some` with hash of n'th public key or hash from the current miniscript item,
+    /// if any. Otherwise returns `Option::None`.
+    ///
+    /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
+    pub fn get_nth_pk_pkh(&self, n: usize) -> Option<PkPkh<Pk>> {
+        match (&self.node, n) {
+            (&Terminal::PkH(ref hash), 0) => Some(PkPkh::HashedPubkey(hash.clone())),
+            (&Terminal::PkK(ref key), 0) => Some(PkPkh::PlainPubkey(key.clone())),
+            (&Terminal::Multi(_, ref keys), _) => {
+                keys.get(n).map(|key| PkPkh::PlainPubkey(key.clone()))
+            }
+            _ => None,
         }
     }
 }
@@ -134,7 +175,12 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
 /// node which constructs the iterator via [Miniscript::iter] method.
 pub struct Iter<'a, Pk: 'a + MiniscriptKey, Ctx: 'a + ScriptContext> {
     next: Option<&'a Miniscript<Pk, Ctx>>,
-    path: Vec<(&'a Miniscript<Pk, Ctx>, usize)>,
+    // Here we store vec of path elements, where each element is a tuple, consisting of:
+    // 1. Miniscript node on the path
+    // 2. It's branches stored as a vec (used for avoiding multiple vec allocations
+    //    during path traversal)
+    // 3. Index of the current branch
+    path: Vec<(&'a Miniscript<Pk, Ctx>, Vec<&'a Miniscript<Pk, Ctx>>, usize)>,
 }
 
 impl<'a, Pk: MiniscriptKey, Ctx: ScriptContext> Iter<'a, Pk, Ctx> {
@@ -167,23 +213,24 @@ impl<'a, Pk: MiniscriptKey, Ctx: ScriptContext> Iterator for Iter<'a, Pk, Ctx> {
     ///     +--> K
     /// ```
     /// `Iter::next()` will iterate over the nodes in the following order:
-    /// `A > B > C > D > E > F > G > I > J > K`
+    /// `A > B > C > D > E > F > G > H > I > J > K`
     ///
     /// To enumerate the branches iterator uses [Miniscript::branches] function.
     fn next(&mut self) -> Option<Self::Item> {
         let mut curr = self.next;
         if let None = curr {
-            while let Some((node, child)) = self.path.pop() {
-                curr = node.branches().get(child).map(|x| *x);
+            while let Some((node, branches, child)) = self.path.pop() {
+                curr = branches.get(child).map(|x| *x);
                 if curr.is_some() {
-                    self.path.push((node, child + 1));
+                    self.path.push((node, branches, child + 1));
                     break;
                 }
             }
         }
         if let Some(node) = curr {
-            self.next = node.branches().first().map(|x| *x);
-            self.path.push((node, 1));
+            let branches = node.branches();
+            self.next = branches.first().map(|x| *x);
+            self.path.push((node, branches, 1));
         }
         curr
     }
@@ -193,14 +240,17 @@ impl<'a, Pk: MiniscriptKey, Ctx: ScriptContext> Iterator for Iter<'a, Pk, Ctx> {
 /// constructs the iterator via [Miniscript::iter_pk] method.
 pub struct PkIter<'a, Pk: 'a + MiniscriptKey, Ctx: 'a + ScriptContext> {
     node_iter: Iter<'a, Pk, Ctx>,
-    keys_buff: VecDeque<Pk>,
+    curr_node: Option<&'a Miniscript<Pk, Ctx>>,
+    key_index: usize,
 }
 
 impl<'a, Pk: MiniscriptKey, Ctx: ScriptContext> PkIter<'a, Pk, Ctx> {
     fn new(miniscript: &'a Miniscript<Pk, Ctx>) -> Self {
+        let mut iter = Iter::new(miniscript);
         PkIter {
-            node_iter: Iter::new(miniscript),
-            keys_buff: VecDeque::new(),
+            curr_node: iter.next(),
+            node_iter: iter,
+            key_index: 0,
         }
     }
 }
@@ -209,15 +259,22 @@ impl<'a, Pk: MiniscriptKey, Ctx: ScriptContext> Iterator for PkIter<'a, Pk, Ctx>
     type Item = Pk;
 
     fn next(&mut self) -> Option<Self::Item> {
-        if self.keys_buff.is_empty() {
-            self.keys_buff = VecDeque::from(loop {
-                let data = self.node_iter.next()?.get_leaf_pk();
-                if !data.is_empty() {
-                    break data;
-                }
-            });
+        loop {
+            match self.curr_node {
+                None => break None,
+                Some(node) => match node.get_nth_pk(self.key_index) {
+                    None => {
+                        self.curr_node = self.node_iter.next();
+                        self.key_index = 0;
+                        continue;
+                    }
+                    Some(pk) => {
+                        self.key_index += 1;
+                        break Some(pk);
+                    }
+                },
+            }
         }
-        self.keys_buff.pop_front()
     }
 }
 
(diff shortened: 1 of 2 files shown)


apoelstra:
I think this is useful but dislike the approach - you should be able to do this with space proportional to the depth of the script, rather than the number of keys.

dr-orlovsky:
Hm, not sure I get why it uses space proportional to the number of keys (and I assume by "keys" you mean "nodes" here, not "public keys", iterator for which is not yet implemented)...

As far as I see, the maximum size of the internal `stack` for `MiniscriptIter` is proportional to the depth of the script tree multiplied by a branching factor, which is `1...3` (statistically more likely to be 0.5-1.5).

apoelstra:
Ah, I think you are right - but I'd still prefer that you get rid of the branching factor.

dr-orlovsky:
Ok, I'll do that. BTW maybe you have at hand some complex Miniscript samples so I can use them in test cases?

dr-orlovsky:
I have refactored it: iterator structure "remembers" (i.e. keeps in stack) only upsteam nodes and their direct children on other (than the current) branches. Don't know how it is possible to optimize its further...

src/miniscript/iter.rs
```diff
@@ -0,0 +1,82 @@
...
+            ZeroNotEqual(node) =>
+                vec![node],
+
+            AndV(node1, node2) |
+            AndB(node1, node2) |
+            OrB(node1, node2) |
+            OrD(node1, node2) |
+            OrC(node1, node2) |
+            OrI(node1, node2) =>
+                vec![node1, node2],
```
apoelstra: This is where the linear space comes from.

apoelstra: There is no need to create a vector here; just have the stack remember the last-used child index.

dr-orlovsky:
Well, I have made it not to store the stack of expanded branches, but I am not happy with the code and doubt that reducing memory requirement from ~`depth*1.5*(pointer_size)` to `depth*(pointer_size+usize)` worth the increase in code complexity. And actually storing 8-byte pointer and 8-byte index value for each step of the tree takes more memory than before...

apoelstra:
It is not "1.5", it is the maximum fan-out of any node in the proof. This is not a constant. It is indeed worth reducing the asymptotic space usage.

dr-orlovsky:
Ok, so will this version of iterator work? If yes, I will proceed to the other parts of the code

dr-orlovsky:
I have finished implementing all functionality I was thinking of for AST iterators, public key/hash traversing and replacement and the code is ready for the review. It does not compile for 1.22, but I will work on that after we will be happy about the whole concept.

dr-orlovsky:
@apoelstra can you pls ack/nack concept? I am already relying on this branch in my own code, so it would be nice to know do you plan to adopt this — or it's better to do it other way

apoelstra:
oops - concept ACK the API. will try to review today.

apoelstra:
Could you rebase this PR so that it has a sensible history?

dr-orlovsky:
Will do it tomorrow, I have to remember it myself :)

dr-orlovsky:
I've summarized it up to six commits, showcasing the process of feature & docs development so it will be easier to review the design logic step by step. Please let me know if you need more brief version — but it will probably mean just a single commit squashing everything together

dr-orlovsky:
Added unit tests and fixed issues found by tests. Also have done rebase on master

apoelstra:
Looks like the use of `impl Trait` is preventing this from working on 1.22

dr-orlovsky:
Fixed. Build failed on stable, some strange fuzz crash, can't figure out what's that

dr-orlovsky:
> Build failed on stable, some strange fuzz crash, can't figure out what's that

Seems the same story as #89

kiminuo:
I think I have figured out the fuzz crashes:

https://github.com/apoelstra/rust-miniscript/blob/96105bb45c6efc60d0bac41866e4786b2276ed67/src/miniscript/astelem.rs#L366 

This `match` is easy to mess up later and that's what actually one can see in the following code listing:

https://github.com/apoelstra/rust-miniscript/blob/96105bb45c6efc60d0bac41866e4786b2276ed67/src/miniscript/astelem.rs#L431-L432

for `thresh` and similarly for `multi`. Meaning there is no check that there are actually any arguments but at least one is expected.

dr-orlovsky:
Now, with the fuzz problem being figured out, I did a rebase, so the thing can be reviewed once again

src/miniscript/iter.rs
```diff
@@ -0,0 +1,703 @@
...
+        }
+    }
+
+    /// Returns `Vec` with cloned version of all public keys from the current miniscript item,
+    /// if any. Otherwise returns an empty `Vec`.
+    ///
+    /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
+    /// To obtain a list of all public keys within AST use [`iter_pubkeys()`] function, for example
+    /// `miniscript.iter_pubkeys().collect()`.
+    pub fn get_pubkeys(&self) -> Vec<PK> {
```
sanket1729, line 90: Is it necessary for these functions to be `pub`? Seems like this is only required for internal iterator purposes. The name might even be confusing for users(who did not look at the doc comment) as they might expect a recursive implementation. 
I think we can get rid of the pub of this and other similar functions.

src/miniscript/iter.rs
```diff
@@ -0,0 +1,703 @@
...
+    /// Returns `Vec` with hashes of all public keys from the current miniscript item, if any.
+    /// Otherwise returns an empty `Vec`.
+    ///
+    /// For each public key the function computes hash; for each hash of the public key the function
+    /// returns it's cloned copy.
+    ///
+    /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
+    /// To obtain a list of all public key hashes within AST use [`iter_pubkey_hashes()`] function,
+    /// for example `miniscript.iter_pubkey_hashes().collect()`.
+    pub fn get_pubkey_hashes(&self) -> Vec<PK::Hash> {
```
sanket1729, line 107: Same comment about `pub`

src/miniscript/iter.rs
```diff
@@ -0,0 +1,703 @@
...
+    }
+
+    /// Returns `Vec` of [PubkeyOrHash] entries, representing either public keys or public key
+    /// hashes, depending on the data from the current miniscript item. If there is no public
+    /// keys or hashes, the function returns an empty `Vec`.
+    ///
+    /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
+    /// To obtain a list of all public keys or hashes within AST use [`iter_pubkeys_and_hashes()`]
+    /// function, for example `miniscript.iter_pubkeys_and_hashes().collect()`.
+    pub fn get_pubkeys_and_hashes(&self) -> Vec<PubkeyOrHash<PK>> {
```
sanket1729, line 123: Same comment about `pub`

src/miniscript/iter.rs
```diff
@@ -0,0 +1,703 @@
...
+    }
+
+    /// Iterates over all public keys and public key hashes by traversing a complete Miniscript
+    /// AST; calls `processor` callback providing each of the found public key or hash and
+    /// replaces miniscript item with the new one constructed out of results returned from the
+    /// callback. The function returns a new Miniscript AST.
+    ///
+    /// If you need to fail the process due to error, return `None` value from the processor
+    /// and replacement will fail with [Error::PubkeyProcessorFailure] error.
+    pub fn replace_pubkeys_and_hashes<T>(&self, processor: &T) -> Result<Self, Error>
```
sanket1729, line 140: We already have a function called `translate_pk` which does something similar but at a more generic level. You can transform a Miniscript<String> to Miniscript<publickey>. 
https://github.com/apoelstra/rust-miniscript/blob/3610c1d508781f25752eb8336e9f2e159973f3af/src/descriptor/mod.rs#L73

I don't think this adds any new functionality to what already possible and can be removed from this PR.

src/miniscript/iter.rs
dr-orlovsky: I am using these function in my downstream dependency on this library. And in general, it is quite useful to be able to get a list of all unique public keys and/or hashes from a given script of any complexity

src/miniscript/iter.rs
dr-orlovsky: Well, actually I was creating this because I needed this sort of functionality and `translate_pk` was not there yet... Seems like while this PR was waiting for review, somebody have added this part

sanket1729: That API is present from the start of the codebase, but maybe the name is not intuitive and difficult to look for the end-user. I think we should add an example of the `translate_pk` functionality.

src/miniscript/iter.rs
sanket1729: I think we should either 
1. change the name of the function to highlight that it is not considering the descendants. 
OR
2. Make the `is_leaf()` type function and downstream call it `miniscript.is_leaf() && miniscript.iter_pubkeys.collect()`

dr-orlovsky:
Can you guys re-review the request: all the issues were fixed and it compiles wrll

src/miniscript/iter.rs
```diff
@@ -0,0 +1,577 @@
...
+    #[test]
+    fn get_hashes() {
+        gen_testcases()
+            .into_iter()
+            .enumerate()
+            .for_each(|(idx, (ms, k, h, test_top_level))| {
+                if !test_top_level {
+                    return;
+                }
+                println!("Test case {}: {:?}", idx + 1, ms);
```
sanket1729, line 494: Can we remove the `println` from tests? Currently, no other tests do any printing, it would be good to stick to that.

src/lib.rs
```diff
@@ -445,6 +450,12 @@ impl fmt::Display for Error {
             Error::IncorrectPubkeyHash => {
                 f.write_str("Incorrect pubkey hash for given descriptor pkh/wpkh")
             }
+            Error::UnexpectedPubkeyHash => {
+                f.write_str("Was expecting public key, not a public key hash")
+            }
+            Error::PubkeyProcessorFailure => {
+                f.write_str("Internal fialure of the public key processor")
```
sanket1729, line 457: I don't think any of these error types are needed anymore.

dr-orlovsky: Correct, will remove them

src/miniscript/iter.rs
```diff
@@ -0,0 +1,577 @@
...
+            });
+        }
+        self.buff.pop_front()
+    }
+}
+
+// Module is public since it export testcase generation which may be used in
+// dependent libraries for their own tasts based on Miniscript AST
+#[cfg(test)]
+pub mod test {
```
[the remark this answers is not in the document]
dr-orlovsky: The tests are not public, public are fns that generate dumb data. I use them in the downstream dependencies for instance. They are quite common (generation of pubkeys)

src/miniscript/iter.rs
dr-orlovsky: Sorry, missed that. Will do

apoelstra:
Can you rebase/squash the commits into a more straightforward ordering?

src/miniscript/iter.rs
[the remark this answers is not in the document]
apoelstra: I think this is fine for a project like this, though we'll have to be mindful about changing these functions. I'm not sure if it's a "breaking change" to break an API gated behind `#[cfg(test)]` but we should at least make an effort not to do it.

apoelstra:
In particular commit `f2a979a` is a merge commit which isn't even labelled as a merge, and I can't make sense of the git history/diff. I need a linear history to review.

apoelstra:
I think all you need to do is

`git rebase master`

and then follow the instructions to resolve the conflicts that appear. Once that's working, if you're feeling generous, you can run

`git rebase -i master`

and rearrange/squash commits which logically belong together.

dr-orlovsky:
Plain rebase onto master unfortunately is quite complex to get working: most of the commits need to be re-written from scratch for that, since master had significantly diverged (new generic type parameter was introduced for instance). So I have to either merge master into this branch again - or just re-implement the iterators from scratch on top of the current master...

sanket1729:
Hi @dr-orlovsky , can you try the following:

Make sure your local master is updated to upstream master. Below, `ms_iter` is the name of local branch that has the code. 
```
sanket1729@mypc:~/rust-miniscript$ git merge-base master ms_iter
e98989c4c1d4140f755dfaf09f0297ba50fc0243
sanket1729@mypc:~/rust-miniscript$ git reset --mixed e98989c4c1d4140f755dfaf09f0297ba50fc0243
Unstaged changes after reset:
M	src/miniscript/mod.rs
sanket1729@mypc:~/rust-miniscript$ git status
On branch ms_iter
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   src/miniscript/mod.rs

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	src/miniscript/iter.rs

no changes added to commit (use "git add" and/or "git commit -a")
sanket1729@mypc:~/rust-miniscript$ git diff src/miniscript/mod.rs
diff --git a/src/miniscript/mod.rs b/src/miniscript/mod.rs
index e4a027e..ca275a8 100644
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -36,6 +36,7 @@ pub use self::context::Segwitv0;
 pub mod astelem;
 pub(crate) mod context;
 pub mod decode;
+pub mod iter;
 pub mod lex;
 pub mod satisfy;
 pub mod types;
sanket1729@mypc:~/rust-miniscript$ git add src/miniscript/iter.rs
```

dr-orlovsky:
Done with a single commit: the functionality is quite holistic and there is no reason to split it across multiple commits.

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+}
+
+impl<'a, Pk: MiniscriptKey, Ctx: ScriptContext> Iterator for Iter<'a, Pk, Ctx> {
+    type Item = &'a Miniscript<Pk, Ctx>;
+
+    /// First, the function returns `self`, then the first child of the self (if any),
+    /// then proceeds to the child of the child — down to a leaf of the tree in its first branch.
+    /// When the leaf is reached, it goes in the reverse direction on the same branch until it
+    /// founds a first branching node that had more than a single branch and returns it, traversing
+    /// it with the same algorithm again.
```
apoelstra, line 156: I'm having trouble understanding this comment. Do you just mean that it does an in-order traversal of the entire tree?

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+    ///     |                 |
+    ///     |                 +--> F
+    ///     |                 |
+    ///     |                 +--> G --+--> H
+    ///     |                          |
+    ///     |                          +--> I -----> J
+    ///     +--> K
+    /// ```
+    /// `Iter::next()` will iterate over the nodes in the following order:
+    /// `A > B > C > D > E > F > G > I > J > K`
```
apoelstra, line 170: `H` is missing from this list

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+        match &self.node {
+            &PkK(_) | &PkH(_) | &Multi(_, _) => vec![],
+
+            &Alt(ref node)
+            | &Swap(ref node)
+            | &Check(ref node)
+            | &DupIf(ref node)
+            | &Verify(ref node)
+            | &NonZero(ref node)
+            | &ZeroNotEqual(ref node) => vec![node.deref()],
```
apoelstra, line 64: What does `.deref()` do here? Why not just use `*`?

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// For the specific algorithm please see [PkPkhIter::next] function.
+    pub fn iter_pk_pkh(&self) -> PkPkhIter<Pk, Ctx> {
+        PkPkhIter::new(self)
+    }
+
+    /// Enumerates all child nodes of the current AST node (`self`) and returns a `Vec` referencing
+    /// them.
+    pub fn branches(&self) -> Vec<&Miniscript<Pk, Ctx>> {
+        use Terminal::*;
+        match &self.node {
```
apoelstra, line 55: Why the `&`? None of the match arms try to move anything.

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// `Iter::next()` will iterate over the nodes in the following order:
+    /// `A > B > C > D > E > F > G > I > J > K`
+    ///
+    /// To enumerate the branches iterator uses [Miniscript::branches] function.
+    fn next(&mut self) -> Option<Self::Item> {
+        let mut curr = self.next;
+        if let None = curr {
+            while let Some((node, child)) = self.path.pop() {
+                curr = node.branches().get(child).map(|x| *x);
+                if curr.is_some() {
```
apoelstra, line 178: It looks like this logic calls `node.branches()`, allocating a vector, purely to check whether the node has `child + 1`-many branches?

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// Creates a new [PkPkhIter] iterator that will iterate over all plain public keys and
+    /// key hash values present in Miniscript items within AST by traversing all its branches.
+    /// For the specific algorithm please see [PkPkhIter::next] function.
+    pub fn iter_pk_pkh(&self) -> PkPkhIter<Pk, Ctx> {
+        PkPkhIter::new(self)
+    }
+
+    /// Enumerates all child nodes of the current AST node (`self`) and returns a `Vec` referencing
+    /// them.
+    pub fn branches(&self) -> Vec<&Miniscript<Pk, Ctx>> {
```
apoelstra, line 53: It appears this function is only ever used to obtain the nth child of a node. It would save some allocations and simplify the logic if you replaced it with
```
pub fn nth_child(&self) -> Option<&Miniscript<Pk, Ctx>> { ... }
```

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+    }
+
+    /// Returns `Vec` with cloned version of all public keys from the current miniscript item,
+    /// if any. Otherwise returns an empty `Vec`.
+    ///
+    /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
+    /// To obtain a list of all public keys within AST use [`iter_pk()`] function, for example
+    /// `miniscript.iter_pubkeys().collect()`.
+    pub fn get_leaf_pk(&self) -> Vec<Pk> {
+        match self.node.clone() {
```
apoelstra, line 90: Why do we unconditionally clone here even though for most nodes we just throw away the result?

apoelstra: It's also unclear to me what the value is in making this method be public.

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+            Terminal::Multi(_, keys) => keys,
+            _ => vec![],
+        }
+    }
+
+    /// Returns `Vec` with hashes of all public keys from the current miniscript item, if any.
+    /// Otherwise returns an empty `Vec`.
+    ///
+    /// For each public key the function computes hash; for each hash of the public key the function
+    /// returns it's cloned copy.
```
apoelstra, line 101: Wrong `it's`

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+    ///
+    /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
+    /// To obtain a list of all public key hashes within AST use [`iter_pkh()`] function,
+    /// for example `miniscript.iter_pubkey_hashes().collect()`.
+    pub fn get_leaf_pkh(&self) -> Vec<Pk::Hash> {
+        match self.node.clone() {
```
apoelstra, line 107: In this one we never use the clone under any circumstances.

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+
+    /// Returns `Vec` of [PubkeyOrHash] entries, representing either public keys or public key
+    /// hashes, depending on the data from the current miniscript item. If there is no public
+    /// keys or hashes, the function returns an empty `Vec`.
+    ///
+    /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
+    /// To obtain a list of all public keys or hashes within AST use [`iter_pk_pkh()`]
+    /// function, for example `miniscript.iter_pubkeys_and_hashes().collect()`.
+    pub fn get_leaf_pk_pkh(&self) -> Vec<PkPkh<Pk>> {
+        use self::PkPkh::*;
```
apoelstra, line 123: I dislike `*` imports as well as importing enum variants

src/miniscript/iter.rs
```diff
@@ -0,0 +1,552 @@
...
+        }
+    }
+}
+
+impl<'a, Pk: MiniscriptKey, Ctx: ScriptContext> Iterator for PkIter<'a, Pk, Ctx> {
+    type Item = Pk;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.keys_buff.is_empty() {
+            self.keys_buff = VecDeque::from(loop {
```
apoelstra, line 213: It seems like all this "compute a `Vec`, convert to `VecDeque`, repeatedly pop" logic could also be replaced with a counter and a `get_nth_pkh` method, eliminating every allocation.

src/miniscript/iter.rs
dr-orlovsky: Yes, this is what was meant

src/miniscript/iter.rs
dr-orlovsky: I'm using `deref` below in line 77 to do dereferencing inside an interator (`node_vec.iter().map(Arc::deref).collect()`), so for consistency I've used it as a function here. Not a problem to correct.

dr-orlovsky: With the commend below, either `deref` nor `*` is not required anymore, will push new version

src/miniscript/iter.rs
dr-orlovsky: Function is public since it may be useful. It is different from iterator methods: iterators traverse the tree, while this function returns keys only from the current node.

dr-orlovsky:
@apoelstra I made all requested changes and re-factored iterators without allocations

src/miniscript/iter.rs
dr-orlovsky: Fixed (don't know why it's not shown as "Outdated"

src/miniscript/iter.rs
dr-orlovsky: I've kept this fn for future possible external use but refactored the iterator code in such way that it call branches only once

src/miniscript/iter.rs
dr-orlovsky: Refactored as you've suggested

src/miniscript/iter.rs
```diff
@@ -98,43 +94,93 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
     /// Otherwise returns an empty `Vec`.
     ///
     /// For each public key the function computes hash; for each hash of the public key the function
-    /// returns it's cloned copy.
+    /// returns it cloned copy.
```
apoelstra, line 97: `its`

src/miniscript/iter.rs
```diff
@@ -98,43 +94,93 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
...
     /// NB: The function analyzes only single miniscript item and not any of its descendants in AST.
     /// To obtain a list of all public key hashes within AST use [`iter_pkh()`] function,
     /// for example `miniscript.iter_pubkey_hashes().collect()`.
     pub fn get_leaf_pkh(&self) -> Vec<Pk::Hash> {
-        match self.node.clone() {
-            Terminal::PkH(hash) => vec![hash],
-            Terminal::PkK(key) => vec![key.to_pubkeyhash()],
-            Terminal::Multi(_, keys) => keys.iter().map(Pk::to_pubkeyhash).collect(),
+        match self.node {
+            Terminal::PkH(ref hash) => vec![hash.clone()],
```
apoelstra, line 104: Hashes are `Copy`, no need for `ref` or `clone`

src/miniscript/iter.rs
```diff
@@ -98,43 +94,93 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
...
 }
 
 /// Iterator for traversing all [Miniscript] miniscript AST references starting from some specific
 /// node which constructs the iterator via [Miniscript::iter] method.
 pub struct Iter<'a, Pk: 'a + MiniscriptKey, Ctx: 'a + ScriptContext> {
     next: Option<&'a Miniscript<Pk, Ctx>>,
-    path: Vec<(&'a Miniscript<Pk, Ctx>, usize)>,
+    // Here we store vec of path elements, where each element is a tuple, consisting of:
+    // 1. Miniscript node on the path
+    // 2. It's branches stored as a vec (used for avoiding multiple vec allocations
```
apoelstra, line 180: wrong "It's"

apoelstra: Also there is no need to do **any** `Vec` allocations here

src/miniscript/iter.rs
dr-orlovsky: Hases are, but not `MiniscriptKey::Hash`. This does not compile, I either have to leave it as is, or add `Hash: Copy` type in `MiniscriptKey` trait

src/miniscript/iter.rs
dr-orlovsky: No allocations now. Also, removed `branches` fn as you asked.

dr-orlovsky:
Travis fails only for 1.22, not 1.31, due to ancient syntax requirements. So I think this can be ignored

src/miniscript/iter.rs
```diff
@@ -76,6 +76,38 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
...
+            | (1, Terminal::AndB(_, ref node))
+            | (1, Terminal::OrB(_, ref node))
+            | (1, Terminal::OrD(_, ref node))
+            | (1, Terminal::OrC(_, ref node))
+            | (1, Terminal::OrI(_, ref node))
+            | (0, Terminal::AndOr(ref node, _, _))
+            | (1, Terminal::AndOr(_, ref node, _))
+            | (2, Terminal::AndOr(_, _, ref node)) => Some(node),
+
+            (n, Terminal::Thresh(_, ref node_vec)) => node_vec.get(n).map(Deref::deref),
```
apoelstra, line 105: A minor nit - I think it's more idiomatic (and a bit shorter) to write `|x| **x` rather than `Deref::deref`

dr-orlovsky: Addressed this with latest force-push. But it seems to me `|x| &**x` looks really stranger than `Deref` :)

apoelstra: why the `&`?

dr-orlovsky: B/c it does not compile without: we return Option<&>, not just option

apoelstra: I think you meant `|x| *x` rather than `|x| &**x`

dr-orlovsky: It really takes `&**` to get from `&Arc<Miniscript>` (which we need to double-deref first with **) to `&Miniscript`

dr-orlovsky: > I think you meant |x| *x rather than |x| &**x

Nope, it does not compile that way.

`*x = *&Arc = Arc`

dr-orlovsky: Proof:

```console
error[E0308]: match arms have incompatible types
   --> src/miniscript/iter.rs:81:9
    |
81  | /         match (n, &self.node) {
82  | |             (0, &Terminal::Alt(ref node))
83  | |             | (0, &Terminal::Swap(ref node))
84  | |             | (0, &Terminal::Check(ref node))
...   |
107 | |             _ => None,
108 | |         }
    | |_________^ expected reference, found struct `std::sync::Arc`
    |
    = note: expected type `std::option::Option<&miniscript::Miniscript<Pk, Ctx>>`
               found type `std::option::Option<std::sync::Arc<miniscript::Miniscript<Pk, Ctx>>>`
note: match arm with an incompatible type
   --> src/miniscript/iter.rs:105:56
    |
105 |             (n, &Terminal::Thresh(_, ref node_vec)) => node_vec.get(n).map(|x| *x),
    |                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

apoelstra: Sorry, I forgot we had `Arc`s in there.

apoelstra (approved):
ack b8360326316c7695fee7beee931a25e8349674de

Thanks for the many iterations!

PR #72, opened 2020-02-07
Cure RAS Syndrome

OttoAllmendinger:
From https://en.wikipedia.org/wiki/RAS_syndrome

> RAS syndrome (where "RAS" stands for "redundant acronym syndrome",
making the phrase "RAS syndrome" self-referential) is the use of one
or more of the words that make up an acronym (or other initialism) in
conjunction with the abbreviated form.


apoelstra:
Looks like a breaking change in rustfmt is preventing compilation

Issue #67, opened 2019-12-13
Bugfix: roundtrip_miniscript_script fails on `006400675168`


apoelstra:
This is `OP_0 OP_NOTIF OP_0 OP_ELSE OP_PUSHNUM_1 OP_ENDIF` which winds up swapping the 0 and the 1.

kiminuo:
@apoelstra My attempt to tackle it: https://github.com/apoelstra/rust-miniscript/pull/75

PR #75, opened 2020-02-14
Fix roundtrip_miniscript_script that fails on `006400675168`

kiminuo:
Fixes issue #67

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -542,9 +542,7 @@ mod tests {
 
     type StdDescriptor = Descriptor<PublicKey>;
     const TEST_PK: &'static str =
-        "pk(\
-         020000000000000000000000000000000000000000000000000000000000000002\
-         )";
+        "pk(020000000000000000000000000000000000000000000000000000000000000002)";
 
     #[test]
     fn parse_descriptor() {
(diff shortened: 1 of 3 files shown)


fuzz/fuzz_targets/roundtrip_miniscript_script.rs
```diff
@@ -60,4 +60,12 @@ mod tests {
...
         super::do_test(&a);
     }
+
+    #[test]
+    fn test_issue_67() {
+        let mut a = Vec::new();
+        // Script corresponds to: OP_0 OP_NOTIF OP_0 OP_ELSE OP_PUSHNUM_1 OP_ENDIF
+        extend_vec_from_hex("006400675168", &mut a);
+        super::do_test(&a);
+    }
```
apoelstra: Can you instead add a unit test to `src/miniscript/mod.rs`. See the `true_false` test for a template for how to round-trip test a specific script. I'd prefer not to have real unit tests in the fuzz repo.

apoelstra:
ACK except I'd like you to move/replace the unit test as suggested

fuzz/fuzz_targets/roundtrip_miniscript_script.rs
kiminuo: You are right about the test line. It just forces me to use the equivalent form `andor(0,1,0)` instead of `and_n(0,1)`.

kiminuo:
@apoelstra Done

fuzz/fuzz_targets/roundtrip_miniscript_script.rs
apoelstra: Weird. I tested locally and it did the opposite :)

kiminuo: Travis is on my side. Haha.

PR #80, opened 2020-03-04
Compiler code deduplication

kiminuo:
`best_e` and `best_w` were almost identical.


apoelstra:
Nice, looks good to me

Issue #81, opened 2020-03-10
Update for latest (and hopefully final) changes to the language


apoelstra:
See https://github.com/sipa/miniscript/pull/33

These changes make descriptor/miniscript parsing unambiguous (or at least, equivalent in terms of script whenever there is an ambiguity).

sanket1729:
Should `s:pk(K)` be parsed as `sc:pk_k(K)`, or is it just applicable to `pk(K) -> c:pk_k(K)`

sanket1729:
- [ ] Rename pk to pk_k
- [ ] Rename thresh_m to multi
- [ ] Add alias pk(K) = c:pk_k(K)
- [ ] Add alias pkh(K) = c:pk_h(K)

PR #82, opened 2020-04-07
Rename `thresh_m` to `multi`

kiminuo:
Helps with #81. It does not solve whole issue.


apoelstra:
This is great! Thanks so much

PR #85, opened 2020-04-27
Miniscript produces witness stacks with wrong order of signatures for `and` instructions

thomaseizinger:
Given an `and` instruction, miniscript produces a witness stack that
is wrongly ordered:

The left branch of an `and` instruction is evaluated first, which means
whatever witness stack is produced by this branch needs to come _after_
the witness stack of the right branch.

Unfortunately, Witness::combine simply concatenates the two vectors,
resulting in the witness stack for the left branch to end up _before_ the
one for the right branch.

This is what we currently have:

```
Witness in memory:

|----A-----|----B-----|--and(A,B)--|

Witness stack evaluation:

Round 0:
        Stack:          Instructions:

                        check(A) check(B)
        |----B-----|
        |----A-----|

Round 1:
        Stack:          Instructions:

                        |----B-----| check(A) check(B)
        |----A-----|

Round 2:
        Failure due to wrong signature.
```

With this patch, this changes to what we actually want:

```
Witness in memory:

|----B-----|----A-----|--and(A,B)--|

Witness stack evaluation:

Round 0:
        Stack:          Instructions:

                        check(A) check(B)
        |----A-----|
        |----B-----|

Round 1:
        Stack:          Instructions:

                        |----A-----| check(A) check(B)
        |----B-----|

Round 2:
        Stack:          Instructions:

                        check(B)
        |----B-----|

Round 3:
        Stack:          Instructions:

                        |----B-----| check(B)

Round 4:
        Script successfully executed!
```

This PR is branched of the 0.12.0 tag because we couldn't make the descriptor parsing work with latest master. It seems that there are some bigger changes incoming.

I also don't know, for which other instructions this behaviour also applies but I guess it might actually apply to all combinators.

We tested this against bitcoind 0.19.1.

Co-authored-by: Lucas Soriano del Pino <l.soriano.del.pino@gmail.com>

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -555,6 +555,7 @@ mod tests {
     use bitcoin::hashes::{hash160, sha256};
     use bitcoin::{self, secp256k1, PublicKey};
     use miniscript::satisfy::BitcoinSig;
+    use std::collections::HashMap;
     use std::str::FromStr;
     use {Descriptor, DummyKey, Miniscript, Satisfier};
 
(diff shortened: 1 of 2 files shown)


eupn:
@apoelstra

thomaseizinger:
The build is failing because it is a pull-request build and hence tested against latest master which includes already the refactorings mentioned in the PR description that make the descriptor unparsable.

apoelstra:
Ok, but when we merge we will merge onto latest master, right? :)

Can you rebase to fix the conflict?

thomaseizinger:
> Ok, but when we merge we will merge onto latest master, right? :)

Yes of course :)

The problem we had was that master seems to undergo some changes at the moment and we actually didn't manage to construct a valid descriptor but we wanted to provide a passing test to showcase the bug.

Once I rebase, the test is not gonna pass because the descriptor is invalid and if I remember correctly, the compiler also didn't give us a valid miniscript.

I'll try again :)

thomaseizinger:
Updated the test with a descriptor using the new language (obtained through the compiler) but it doesn't seem to like it. Fails with: `Unexpected("vpk(1 args) while parsing Miniscript")`

sanket1729:
@thomaseizinger, good find and sorry for the delayed response. This looks correct and we have fixed our CI issues.

I also reviewed rest of the file and there is another error in AndOr satisfaction code where the code Witness::Combine needs to be swapped for both of the satisfactions. 

If you can rebase to the latest master and also the AndOr case, we can merge this. 

AndOr code lines: https://github.com/apoelstra/rust-miniscript/blob/3610c1d508781f25752eb8336e9f2e159973f3af/src/miniscript/satisfy.rs#L594-L599

thomaseizinger:
@sanket1729 Done!

Let me know if this is correct. I didn't any test for the latter two changes. I am generally not really happy with the approach we took for testing this.
The test setup is super verbose and having to take apart the witness stack manually is kind of tedious and error prone to the same bug that was in the actual satisfaction code.

Would be nice to have a set of (signed) transactions that we know are correct and we could run the satisfaction code against that, asserting that it generates a witness stack that is correct.

sanket1729:
@thomaseizinger that was quick! Looks like the Travis is complaining about some rust 1.22 issue, but it's passing everywhere else. Unfortunately, we have to stick 1.22 for being compatible with the rest of the rust bitcoin ecosystem.

We will be working on a more robust testing suite for Satisfaction logic with real transactions which remains an undertested part of this library.

thomaseizinger:
@sanket1729 Should be fixed now :)

sanket1729:
Fixes #87

PR #94, opened 2020-06-05
Fixed pk_k bug, added test

sanket1729:
I think #81 is now completely resolved.

### DIFF
--- a/src/miniscript/astelem.rs
+++ b/src/miniscript/astelem.rs
@@ -305,9 +305,19 @@ impl<Pk: MiniscriptKey> fmt::Display for Terminal<Pk> {
                     }
 
                     fmt::Write::write_char(f, ch)?;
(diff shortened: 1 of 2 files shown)


src/miniscript/astelem.rs
```diff
@@ -356,8 +367,14 @@ where
                 if wrap.is_empty() {
                     return Err(Error::Unexpected(top.name.to_owned()));
                 }
-                frag_name = name;
-                frag_wrap = wrap;
+                if name == "pk" {
+                    frag_name = "pk_k";
+                    aliased_wrap = wrap.to_owned() + &"c";
```
apoelstra, line 372: As a matter of style, I don't like using `+` on strings, because I think it is an abuse of operator overloading to use `+` for a noncommutative operator.

PR #96, opened 2020-06-06
Fix for: Tests in fuzz directory [issue #91]

### LINKED ISSUE #91: Tests in fuzz directory
I guess the purpose of duplicate_crash() functions to replicate crashes on Travis for debugging, but I don't think they should belong in the fuzz directory. 
The current codebase seems to have the same fuzz test extend_from_vec across all files.

kiminuo:
* Move `duplicate_crash_2` test to `semantic.rs`
* Move `extend_vec_from_hex` to lib.rs (de-duplicate)

Fixes #91


kiminuo:
Unfortunately, the fuzz test fails to pass thanks to `or(pk())`

kiminuo:
https://github.com/apoelstra/rust-miniscript/pull/96/files#diff-4374b9b87c971ea1a2293a875f41efabR4 is not that nice. It would be nicer to put it somehow to https://github.com/apoelstra/rust-miniscript/blob/master/fuzz/Cargo.toml. Not sure how. Rust gurus are welcomed here :-)

sanket1729:
Hi @kiminuo , even I am not sure what would be the best way to address this. I also made an attempt for this in #95 . 

My idea was to simply put those in README. My understanding was that code should not be a part of fuzz repo at all. We had it because it was easy to reproduce crashes. So, whenever some contributor wants to reproduce a crash, they can paste the code from README, debug the crash, remove the crash code before we merge the PR. 
We can wait for @apoelstra's opinion on how to address this.

kiminuo:
> 
> 
> Hi @kiminuo , even I am not sure what would be the best way to address this. I also made an attempt for this in #95 .
> 
Your PR looks very good.

> My idea was to simply put those in README. My understanding was that code should not be a part of fuzz repo at all. We had it because it was easy to reproduce crashes. So, whenever some contributor wants to reproduce a crash, they can paste the code from README, debug the crash, remove the crash code before we merge the PR.

To be honest, I don't find that really user friendly. But it might be what @apoelstra had in mind, I don't know.

PR #97, opened 2020-06-15
Add ScriptContext to Miniscript

sanket1729:
Adds context to Miniscript. I have made a bunch of design decisions here with minor changes in API for satisfied constraints. 

I could not find a clean way to enforce constraints for the only bitcoin::PublicKey type because our code was generic, so had to hack in `IS_BITCOIN_PUBKEY` to the Miniscript Trait. 

This builds on top of #95 , would rebase and squash as needed.

### DIFF
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -43,6 +43,23 @@ then
     )
 fi
 
+# Miri Checks if told to
+# Only supported in nightly
+if [ "$DO_MIRI" = true ]
+then
+    (
+        MIRI_NIGHTLY=nightly-$(curl -s https://rust-lang.github.io/rustup-components-history/x86_64-unknown-linux-gnu/miri)
+        echo "Installing latest nightly with Miri: $MIRI_NIGHTLY"
+        rustup set profile minimal
+        rustup default "$MIRI_NIGHTLY"
+        rustup component add miri
+        cargo miri test -- -- miri_
+
+        # Change back to latest nightly possibly without Miri
+        rustup default nightly
+    )
+fi
+
 # Bench if told to
 if [ "$DO_BENCH" = true ]
 then
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -32,7 +32,8 @@ use std::str::{self, FromStr};
 
 use expression;
 use miniscript;
-use miniscript::Miniscript;
+use miniscript::context::ScriptContextError;
+use miniscript::{Legacy, Miniscript, Segwitv0};
 use Error;
 use MiniscriptKey;
 use Satisfier;
@@ -50,8 +51,8 @@ pub use self::satisfied_constraints::Stack;
 /// Script descriptor
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub enum Descriptor<Pk: MiniscriptKey> {
-    /// A raw scriptpubkey (including pay-to-pubkey)
-    Bare(Miniscript<Pk>),
+    /// A raw scriptpubkey (including pay-to-pubkey) under Legacy context
+    Bare(Miniscript<Pk, Legacy>),
     /// Pay-to-Pubkey
     Pk(Pk),
     /// Pay-to-PubKey-Hash
@@ -60,16 +61,19 @@ pub enum Descriptor<Pk: MiniscriptKey> {
     Wpkh(Pk),
     /// Pay-to-Witness-PubKey-Hash inside P2SH
     ShWpkh(Pk),
-    /// Pay-to-ScriptHash
-    Sh(Miniscript<Pk>),
-    /// Pay-to-Witness-ScriptHash
-    Wsh(Miniscript<Pk>),
-    /// P2SH-P2WSH
-    ShWsh(Miniscript<Pk>),
+    /// Pay-to-ScriptHash with Legacy context
+    Sh(Miniscript<Pk, Legacy>),
+    /// Pay-to-Witness-ScriptHash with Segwitv0 context
+    Wsh(Miniscript<Pk, Segwitv0>),
+    /// P2SH-P2WSH with Segwitv0 context
+    ShWsh(Miniscript<Pk, Segwitv0>),
 }
 
 impl<Pk: MiniscriptKey> Descriptor<Pk> {
     /// Convert a descriptor using abstract keys to one using specific keys
+    /// This will panic if translatefpk returns an uncompressed key when
+    /// converting to a Segwit descriptor. To prevent this panic, ensure
+    /// translatefpk returns an error in this case instead.
     pub fn translate_pk<Fpk, Fpkh, Q, E>(
         &self,
         mut translatefpk: Fpk,
@@ -86,8 +90,18 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
             )),
             Descriptor::Pk(ref pk) => translatefpk(pk).map(Descriptor::Pk),
             Descriptor::Pkh(ref pk) => translatefpk(pk).map(Descriptor::Pkh),
-            Descriptor::Wpkh(ref pk) => translatefpk(pk).map(Descriptor::Wpkh),
-            Descriptor::ShWpkh(ref pk) => translatefpk(pk).map(Descriptor::ShWpkh),
+            Descriptor::Wpkh(ref pk) => {
+                if pk.is_uncompressed() {
+                    panic!("Uncompressed pubkeys are not allowed in segwit v0 scripts");
+                }
+                translatefpk(pk).map(Descriptor::Wpkh)
+            }
+            Descriptor::ShWpkh(ref pk) => {
+                if pk.is_uncompressed() {
+                    panic!("Uncompressed pubkeys are not allowed in segwit v0 scripts");
+                }
+                translatefpk(pk).map(Descriptor::ShWpkh)
+            }
             Descriptor::Sh(ref ms) => Ok(Descriptor::Sh(
                 ms.translate_pk(&mut translatefpk, &mut translatefpkh)?,
             )),
@@ -201,7 +215,8 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
                 let addr = bitcoin::Address::p2wpkh(&pk.to_public_key(), bitcoin::Network::Bitcoin);
                 addr.script_pubkey()
             }
-            Descriptor::Sh(ref d) | Descriptor::Wsh(ref d) | Descriptor::ShWsh(ref d) => d.encode(),
+            Descriptor::Sh(ref d) => d.encode(),
+            Descriptor::Wsh(ref d) | Descriptor::ShWsh(ref d) => d.encode(),
         }
     }
 
@@ -396,7 +411,12 @@ where
                 expression::terminal(&top.args[0], |pk| Pk::from_str(pk).map(Descriptor::Pkh))
             }
             ("wpkh", 1) => {
-                expression::terminal(&top.args[0], |pk| Pk::from_str(pk).map(Descriptor::Wpkh))
+                let wpkh = expression::terminal(&top.args[0], |pk| Pk::from_str(pk))?;
+                if wpkh.is_uncompressed() {
+                    Err(Error::ContextError(ScriptContextError::CompressedOnly))
+                } else {
+                    Ok(Descriptor::Wpkh(wpkh))
+                }
             }
             ("sh", 1) => {
                 let newtop = &top.args[0];
@@ -409,9 +429,14 @@ where
                             Ok(Descriptor::ShWsh(sub))
                         }
                     }
-                    ("wpkh", 1) => expression::terminal(&newtop.args[0], |pk| {
-                        Pk::from_str(pk).map(Descriptor::ShWpkh)
-                    }),
+                    ("wpkh", 1) => {
+                        let wpkh = expression::terminal(&newtop.args[0], |pk| Pk::from_str(pk))?;
+                        if wpkh.is_uncompressed() {
+                            Err(Error::ContextError(ScriptContextError::CompressedOnly))
+                        } else {
+                            Ok(Descriptor::ShWpkh(wpkh))
+                        }
+                    }
                     _ => {
                         let sub = Miniscript::from_tree(&top.args[0])?;
                         if sub.ty.corr.base != miniscript::types::Base::B {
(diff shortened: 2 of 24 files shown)


src/lib.rs
```diff
@@ -113,13 +113,15 @@ pub use miniscript::Miniscript;
 pub trait MiniscriptKey:
     Clone + Eq + Ord + str::FromStr + fmt::Debug + fmt::Display + hash::Hash
 {
+    const IS_BITCOIN_PUBKEY: bool = false;
```
apoelstra, line 116: I think it would be better to have method `is_uncompressed()` which has a default implementation which returns false. Then **only** for `bitcoin::PublicKey` do we implement this and have it actually return the uncompressedness.

src/miniscript/astelem.rs
```diff
@@ -146,11 +147,13 @@ impl<Pk: MiniscriptKey> Terminal<Pk> {
                 let keys: Result<Vec<Q>, _> = keys.iter().map(&mut *translatefpk).collect();
                 Terminal::Multi(k, keys?)
             }
-        })
+        };
+        Ctx::check_frag_validity(&frag).expect("Translate failed!");
```
sanket1729, line 151: I am not completely sure how to handle this error. There is already one error type because of the Function argument. Secondly, we would also another because the translation may not be valid(for ex: translating string to uncompressed pubkeys in segwit descriptor).

sanket1729: We can 
1. Create a enum sum type for both errors
2. Return a Result<Result<ms, err1>,err2>
3. Panic!

sanket1729:
Fuzz failures seem unrelated, I can make another PR to address those.

contrib/test.sh
```diff
@@ -43,6 +43,23 @@ then
...
+# Only supported in nightly
+if [ "$DO_MIRI" = true ]
+then
+    (
+        MIRI_NIGHTLY=nightly-$(curl -s https://rust-lang.github.io/rustup-components-history/x86_64-unknown-linux-gnu/miri)
+        echo "Installing latest nightly with Miri: $MIRI_NIGHTLY"
+        rustup set profile minimal
+        rustup default "$MIRI_NIGHTLY"
+        rustup component add miri
+        cargo miri test -- -- any_unsafe_transmute_miri_test
```
apoelstra, line 56: This line filters to only run the `any_unsafe_transmute_miri_test` test? Maybe we should shorten it to `miri_`, so that in future, we can add MIRI-executable tests just by prefixing them with `miri_`

src/descriptor/mod.rs
```diff
@@ -74,29 +75,43 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
         &self,
         mut translatefpk: Fpk,
         mut translatefpkh: Fpkh,
-    ) -> Result<Descriptor<Q>, E>
+    ) -> Result<Result<Descriptor<Q>, Error>, E>
```
apoelstra, line 78: I think we should do `Result<Result<Descriptor<Q>, E>, Error>` instead

apoelstra: Actually, after some discussion on IRC, I think we should panic on uncompressed keys, and put a clear warning in the doccomment for this function.

If the user does not want to panic, this is totally in her control -- she can change `translatefpk` and `translatefpkh` to return an error rather than an uncompressed key.

src/descriptor/satisfied_constraints.rs
```diff
@@ -200,9 +201,9 @@ pub enum SatisfiedConstraint<'desc, 'stack> {
 ///the top of the stack, we need to decide whether to execute right child or not.
 ///This is also useful for wrappers and thresholds which push a value on the stack
 ///depending on evaluation of the children.
-struct NodeEvaluationState<'desc> {
+struct NodeEvaluationState<'desc, Ctx: 'desc + ScriptContext> {
     ///The node which is being evaluated
-    node: &'desc Miniscript<bitcoin::PublicKey>,
+    node: &'desc Miniscript<bitcoin::PublicKey, Ctx>,
```
apoelstra, line 206: I think we should put `Any` here rather than genericizing over `Ctx` and then using `Any` down below.

src/descriptor/satisfied_constraints.rs
```diff
@@ -1074,8 +1112,8 @@ mod tests {
         fn from_stack<'stack, 'elem, F>(
             verify_fn: F,
             stack: Stack<'stack>,
-            ms: &'elem Miniscript<bitcoin::PublicKey>,
-        ) -> SatisfiedConstraints<'elem, 'stack, F>
+            ms: &'elem Miniscript<bitcoin::PublicKey, Legacy>,
+        ) -> SatisfiedConstraints<'elem, 'stack, Legacy, F>
```
apoelstra, line 1116: Curious why we use `Legacy` here rather than `Any`

src/miniscript/context.rs
```diff
@@ -0,0 +1,200 @@
...
+            }
+            ScriptContextError::CompressedOnly => {
+                write!(f, "Uncompressed pubkeys not allowed in segwit context")
+            }
+        }
+    }
+}
+
+pub trait ScriptContext:
+    fmt::Debug + Clone + Ord + PartialOrd + Eq + PartialEq + private::Sealed
```
apoelstra, line 52: Lol, do we actually need any of these extra traits beyond `private::Sealed`?

src/miniscript/context.rs
```diff
@@ -0,0 +1,200 @@
...
+/// Any ScriptContext. The implementation has the most conservative rules
+/// for validity and malleability
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub enum Any {}
+impl ScriptContext for Any {
+    fn check_frag_non_malleable<Pk: MiniscriptKey, Ctx: ScriptContext>(
+        frag: &Terminal<Pk, Ctx>,
+    ) -> Result<(), ScriptContextError> {
+        Segwitv0::check_frag_non_malleable(frag)?;
+        Legacy::check_frag_non_malleable(frag)
```
apoelstra, line 131: I think these should be `unimplemented!()` rather than doing both checks. We shouldn't ever actually use `Any` except in places where the context doesn't matter.

src/miniscript/context.rs
```diff
@@ -0,0 +1,200 @@
...
+    pub(crate) fn from_legacy<Pk: MiniscriptKey>(
+        ms: &Miniscript<Pk, Legacy>,
+    ) -> &Miniscript<Pk, Any> {
+        // The fields Miniscript<Pk, Legacy> and Miniscript<Any, Legacy> only
+        // differ in PhantomData. This unsafe assumes that the unlerlying byte
+        // representation of the both should be the same. There is a Miri test
+        // checking the same.
+        unsafe {
+            use std::mem::transmute;
+            transmute::<&Miniscript<Pk, Legacy>, &Miniscript<Pk, Any>>(ms)
```
apoelstra, line 152: FWIW you don't need to provide types here, Rust can infer them

src/lib.rs
```diff
@@ -113,13 +115,25 @@ pub use miniscript::Miniscript;
 pub trait MiniscriptKey:
     Clone + Eq + Ord + str::FromStr + fmt::Debug + fmt::Display + hash::Hash
 {
+    /// Check if the publicKey is uncompressed. The default
+    /// implementation returns false for all Generic types of
+    /// MiniscriptKey, and returns the true only for
+    /// bitcoin::Publickey if the underlying key is uncompressed.
```
apoelstra, line 121: This comment could be shortened to just "The default implementation returns false"

src/lib.rs
```diff
@@ -113,13 +115,25 @@ pub use miniscript::Miniscript;
...
     type Hash: Clone + Eq + Ord + str::FromStr + fmt::Display + fmt::Debug + hash::Hash;
 
     ///Converts an object to PublicHash
     fn to_pubkeyhash(&self) -> Self::Hash;
 }
 
 impl MiniscriptKey for bitcoin::PublicKey {
+    fn is_uncompressed(&self) -> bool {
+        !self.compressed
+    }
```
apoelstra, line 135: ...and we can add a doccomment here explaining that we are returning the compressedness flag of the bitcoin pubkey

src/miniscript/context.rs
sanket1729: Turns out we do. Rust is complaining that it can't compare, equate without those

src/miniscript/context.rs
sanket1729: Changed to `!unreachable!()`

src/miniscript/astelem.rs
```diff
@@ -58,99 +59,158 @@ impl<Pk: MiniscriptKey> Terminal<Pk> {
...
+impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
     /// Convert an AST element with one public key type to one of another
     /// public key type
-    pub fn translate_pk<FPk, FPkh, Q, Error>(
+    pub fn translate_pk<FPk, FPkh, Q, FuncError>(
         &self,
         translatefpk: &mut FPk,
         translatefpkh: &mut FPkh,
-    ) -> Result<Terminal<Q>, Error>
+    ) -> Result<Result<Terminal<Q, Ctx>, Error>, FuncError>
```
apoelstra, line 98: Here too we can drop the double-result (and probably leave the old logic in rather than calling out to `translate_bin`?)

src/miniscript/mod.rs
```diff
@@ -26,114 +26,124 @@
...
 use self::types::Property;
+use miniscript::decode::Terminal;
 use miniscript::types::extra_props::ExtData;
 use miniscript::types::Type;
 use std::cmp;
 use std::sync::Arc;
 use MiniscriptKey;
 use {expression, Error, ToPublicKey};
 
+pub use self::context::{Any, Legacy, ScriptContext, Segwitv0};
```
apoelstra, line 52: I think we shouldn't export `Any`, it's a bit too dangerous to expose outside the library.

src/miniscript/mod.rs
```diff
@@ -26,114 +26,124 @@
...
 #[cfg(feature = "serde")]
 use serde::{de, ser};
+use std::marker::PhantomData;
 use std::{fmt, str};
 
 use bitcoin;
 use bitcoin::blockdata::script;
 
 pub mod astelem;
+pub mod context;
```
apoelstra, line 36: I think we should keep this module private (we publicly rexport some things below)

src/miniscript/mod.rs
```diff
@@ -196,29 +206,33 @@ impl<Pk: MiniscriptKey + ToPublicKey> Miniscript<Pk> {
...
 
-impl<Pk: MiniscriptKey> Miniscript<Pk> {
-    pub fn translate_pk<FPk, FPkh, Q, Error>(
+impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
+    pub fn translate_pk<FPk, FPkh, Q, FuncError>(
         &self,
         translatefpk: &mut FPk,
         translatefpkh: &mut FPkh,
-    ) -> Result<Miniscript<Q>, Error>
+    ) -> Result<Result<Miniscript<Q, Ctx>, Error>, FuncError>
```
apoelstra, line 214: another double-result we shuold change back to a panic

src/policy/compiler.rs
```diff
@@ -633,17 +642,19 @@ fn all_casts<Pk: MiniscriptKey>() -> [Cast<Pk>; 10] {
...
 ) -> bool {
-    // return malleable types directly. If a elem is malleable, all the casts
-    // to it are also going to be malleable
-    if !elem.ms.ty.mall.non_malleable {
+    // return malleable types directly. If a elem is malleable under current context,
+    // all the casts to it are also going to be malleable
+    if !elem.ms.ty.mall.non_malleable && Ctx::check_frag_non_malleable(&elem.ms.node).is_ok() {
         return false;
     }
+    // Check if the element is malleable under the current script Context
```
apoelstra, line 656: I think this comment line is misplaced

src/descriptor/satisfied_constraints.rs
sanket1729: As you mentioned in a later comment, we should never actually use `Any` other than whenever required. 
As such, there is no way to create an `Any` type Miniscript from string or decoding, we must specify a `Segwitv0` or `Legacy` context along with it. If we try to create an `Any` type Miniscript, the context checks would map to `unreachable!()`.

src/descriptor/mod.rs
```diff
@@ -86,8 +89,18 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
             )),
             Descriptor::Pk(ref pk) => translatefpk(pk).map(Descriptor::Pk),
             Descriptor::Pkh(ref pk) => translatefpk(pk).map(Descriptor::Pkh),
-            Descriptor::Wpkh(ref pk) => translatefpk(pk).map(Descriptor::Wpkh),
-            Descriptor::ShWpkh(ref pk) => translatefpk(pk).map(Descriptor::ShWpkh),
+            Descriptor::Wpkh(ref pk) => {
+                if pk.is_uncompressed() {
+                    panic!("Uncompressed Pubkeys are non-standard in Segwit Context");
```
apoelstra, line 94: Not "non-standard", actually not allowed. I suggest "Uncompressed pubkeys are not allowed in segwit v0 scripts".

src/descriptor/mod.rs
```diff
@@ -86,8 +89,18 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
+                }
+                translatefpk(pk).map(Descriptor::Wpkh)
+            }
+            Descriptor::ShWpkh(ref pk) => {
+                if pk.is_uncompressed() {
+                    panic!("Uncompressed Pubkeys are non-standard in Segwit Context");
```
apoelstra, line 100: same here

src/descriptor/satisfied_constraints.rs
```diff
@@ -257,24 +260,14 @@ impl<'desc, 'stack, F> SatisfiedConstraints<'desc, 'stack, F>
...
-
-    /// Creates a new iterator over all constraints satisfied for a given
+    // Creates a new iterator over all constraints satisfied for a given
     /// descriptor by a given witness stack. Because this iterator is lazy,
     /// it may return satisfied constraints even if these turn out to be
     /// irrelevant to the final (dis)satisfaction of the descriptor.
+    /// Because the iterator is dependent evaluating the miniscript, it is
+    /// dependent on scriptContext. Since, the caller of the function should
+    /// not care about Miniscript Context, resulting contextis transformed into
+    /// ScriptContext::Any
```
apoelstra, line 270: We can remove this extra comment now that it's an invisible implementation detail.

src/miniscript/mod.rs
```diff
@@ -26,114 +26,127 @@
...
+pub struct Miniscript<Pk: MiniscriptKey, Ctx: ScriptContext> {
     ///A node in the Abstract Syntax Tree(
-    pub node: decode::Terminal<Pk>,
+    pub node: Terminal<Pk, Ctx>,
     ///The correctness and malleability type information for the AST node
     pub ty: types::Type,
     ///Additional information helpful for extra analysis.
     pub ext: types::extra_props::ExtData,
+    /// Context PhantomData
+    pub phantom: PhantomData<Ctx>,
```
apoelstra, line 67: I think we shouldn't make this public.

This has the weird effect of making it impossible to construct a `Miniscript` without going through the constructor, but I'm not sure we ever wanted that. The other fields are public to make introspection possible, but ideally we don't want the user to be modifying these things.

src/miniscript/mod.rs
```diff
@@ -196,15 +209,17 @@ impl<Pk: MiniscriptKey + ToPublicKey> Miniscript<Pk> {
     }
 }
 
-impl<Pk: MiniscriptKey> Miniscript<Pk> {
-    pub fn translate_pk<FPk, FPkh, Q, Error>(
+impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
+    /// This will panic while converting to Segwit descriptor using
+    /// using uncompressed pubkeys.
```
apoelstra, line 214: Can you expand on this? I suggest "This will panic if `translatefpk` returns an uncompressed key when converting to a Segwit descriptor. To prevent this panic, ensure `translatefpk` returns an error in this case instead."

PR #78, opened 2020-02-24
Issue from pull/63

kiminuo:
This is a tentative attempt to do some good with regard to: https://github.com/apoelstra/rust-miniscript/pull/63

It's far from done.

PR #106, opened 2020-07-06
Issue from pull/63

apoelstra:
Copy of #78 with the unit test fixed


src/miniscript/decode.rs
```diff
@@ -472,3 +482,10 @@ pub fn parse(tokens: &mut TokenIter) -> Result<Miniscript<bitcoin::PublicKey>, E
     assert_eq!(term.0.len(), 1);
     Ok(term.pop().unwrap())
 }
+
+fn isAndV(tokens: &mut TokenIter) -> bool {
```
sanket1729, line 486: rustc warning: "convert the identifier to snake case: `is_and_v`"

apoelstra: Oh nice, thanks

sanket1729:
Branch build is expected to fail as it does not have the latest CI fixes

PR #109, opened 2020-07-08
Switch Policy to use BTreeMap from HashMap

JeremyRubin:
Switch the Concrete Policy Compilation to use BTreeMap instead of HashMap so that compilation is deterministic for a given compiler release.

In some contexts it can be useful to be able to generate identical miniscripts for a given policy (e.g., testing or generating scripts in multiparty/non interactive contexts).

This shouldn't be relied on across versions of the Policy compiler, as the output is otherwise unstable.

If non determinism is desired, it can be re-introduced with an explicit randomization of values before iterating.

Things to check in this PR:

1) This does eliminate present non-determinism in Policy
2) No tests are currently relying on non-determinism to compare Policy runs
3) BTreeMap is an acceptable substitute for HashMap

### DIFF
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -17,10 +17,10 @@
 //! Optimizing compiler from concrete policies to Miniscript
 //!
 
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 use std::convert::From;
 use std::marker::PhantomData;
-use std::{cmp, error, f64, fmt};
+use std::{cmp, error, f64, fmt, mem};
 
 use miniscript::types::extra_props::MAX_OPS_PER_SCRIPT;
 use miniscript::types::{self, ErrorKind, ExtData, Property, Type};
@@ -33,7 +33,7 @@ use {policy, Terminal};
 use {Miniscript, MiniscriptKey};
 
 type PolicyCache<Pk, Ctx> =
-    HashMap<(Concrete<Pk>, OrdF64, Option<OrdF64>), HashMap<CompilationKey, AstElemExt<Pk, Ctx>>>;
+    BTreeMap<(Concrete<Pk>, OrdF64, Option<OrdF64>), BTreeMap<CompilationKey, AstElemExt<Pk, Ctx>>>;
 
 ///Ordered f64 for comparison
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
@@ -107,7 +107,7 @@ impl hash::Hash for OrdF64 {
 
 /// Compilation key: This represents the state of the best possible compilation
 /// of a given policy(implicitly keyed).
-#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]
+#[derive(Copy, Clone, PartialEq, Eq, Debug, PartialOrd, Ord)]
 struct CompilationKey {
     /// The type of the compilation result
     ty: Type,
@@ -643,7 +643,7 @@ fn all_casts<Pk: MiniscriptKey, Ctx: ScriptContext>() -> [Cast<Pk, Ctx>; 10] {
 /// In general, we maintain the invariant that if anything is inserted into the
 /// map, it's cast closure must also be considered for best compilations.
 fn insert_elem<Pk: MiniscriptKey, Ctx: ScriptContext>(
-    map: &mut HashMap<CompilationKey, AstElemExt<Pk, Ctx>>,
+    map: &mut BTreeMap<CompilationKey, AstElemExt<Pk, Ctx>>,
     elem: AstElemExt<Pk, Ctx>,
     sat_prob: f64,
     dissat_prob: Option<f64>,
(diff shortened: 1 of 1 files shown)


JeremyRubin:
CI issues fixed up, build seems correct now :)

apoelstra:
I'm ok with this, but determinism is not a goal of the compiler and I don't know that this will achieve it, and I have no idea how I would verify that. What are you trying to do that requires determinism?

dr-orlovsky (approved):
I like the idea; since have seen how `HashMap` elements can change the order depending from the compilation platform. `BTreeMap` relies on `Ord` to follow the same order always. Regarding testing I think simple tests for ordering + tests for `Cmp` trait run using CI for multiple platforms would suffice.

apoelstra:
Simple testing definitely will not suffice to verify determinism.

JeremyRubin:
One thing that might help improve confidence in the platform independence would be to convert use of f64's to a rational type (e.g., from crate num_rational or fixed) which would guarantee better cross platform consistency.

I don't view this as a 'one and done' task, and I certainly don't think determinism across versions is required or feasible, but I do think it would be positive for changes that make compilation as reproducible as possible and introduce randomization in a deterministic way (e.g., a seed or hashing the input's string representation). 

Proving that a program is fully deterministic on all architectures may be difficult, especially since rust may call unsafe code somewhere in the stack (e.g., to read an atomic global random variable), but I think as a general direction to proceed it is good because it makes it much easier to detect compiler defects or changes.

apoelstra:
I think it might make sense to create a separate compiler crate which has more coherent design goals.

I generally view the compiler here as a QA utility; it's feature gated because it's both more complex and less reviewed than other parts of the cadebase, and because my uses of rust-miniscript never touch it. I wouldn't want to add extra dependencies for it.

JeremyRubin:
I think this is a more general discussion, on where the code should live. I mostly agree, especially if the purpose of the compiler is internal & is limited in scope of purpose it's good to not expose it as something external projects can use. Then, if a separate crate exists, it can be pulled in for testing dependencies / to do integration tests as needed (or tests can live in that crate). 

What's your preferred way to do this? another repo under rust-bitcoin? or should I just fork my own.

on topic: If you don't mind this patch (and the tests go through), it does at least help with QA/testing repeatability.

apoelstra:
I mean, I'm sure Sanket and Pieter would like the compiler to be exposed -- it is intended to be user-facing, it's just that my opinion is that it has fairly limited and ad-hoc usecases. So it's definitely not an "internal-only tool that should live in rust-miniscript because that's where it's used" ... more like the opposite, it's an independent tool that is not getting a lot of love here and which isn't used for the same sorts of things that the rest of the library does.

You're welcome to fork your own of course, but if you want an "official" compiler repo we should probably discuss on IRC.

Having said that, I don't mind merging this PR - it's pretty easy to read, has a small impact on performance (looks like on the order of 10% slowdown, which for an exp-time algorithm is irrelevant :)). @sanket1729 what do you think?

sanket1729:
@apoelstra @JeremyRubin On the topic of this PR, this may be helpful in testing and there is no loss in having it. 

> The question is more general whether this `rust-miniscript` should make effort towards making the current compiler deterministic. 
I don't think targetting optimality and determinism would be easy to review, so probably that should be a separate project. 

Based on the interest of @JeremyRubin and Shesek, I think having a simple deterministic compiler targetting miniscript with just one fragment (per `and/or`)  has value. 
If I understand correctly, the thing that you guys want is composing miniscripts directly. 
`compile(comp(p1,p2)) = comp(compile(p1),compile(p2)` ?  

Even if we have a deterministic compiler, guaranteeing the above property make require additional work.

sanket1729:
Cherry-picked and merged in #141 .

JeremyRubin:
The main motivation is actually a bit different. I don't care about compile(comp(p1, p2)) == comp(compile(p1), compile(p2)), I care about compile(P) == compile(P). Miniscript already has this property afaict!

So why not just directly use Miniscript if it already has the property I want?

The issue is the rules for Miniscript composition are non-obvious sometimes with which wrapper should be selected when, because the nested types need be "aware" of what type they're being plugged into. A deterministic (but non optimal!) compiler (don't even need the probability stuff, but I think it can be supported easily) is very useful for writing metaprogramming languages that target miniscript purely as a program which handles "miniscript type coercion". You can still do some optimizing of course, as long as it's not introducing any true randomness.

As such, it doesn't need the composition law that you state, it's actually 100% acceptable that this not be true (at least... to me). I just care that for a given input to compile, the same output comes out. This is useful for writing functions that might, e.g., given  a HD Key  generate all the appropriate scripts and addresses for a wallet vault. If you don't have the determinism, then you always have to cache the result of compilation which limits recovery options.


It very well might be the case that it's easy to implement the composition rule given the determinism constraint, but it's not implied as one could e.g. seed a rng with the entire input.

PR #110, opened 2020-07-20
policy: Corrected an inverted documentation for `Older` and `After`

darosior:
A small typo I noticed on docs.rs


sanket1729:
This was swapped in #37, but the comments are not fixed. Looks like comments here for miniscript opcodes are also incorrect https://github.com/rust-bitcoin/rust-miniscript/blob/e98989c4c1d4140f755dfaf09f0297ba50fc0243/src/miniscript/decode.rs#L75-L79

PR #113, opened 2020-07-22
Optimize policy compilation for N-of-N tresholds

darosior:
This adds a special case to the compilation of the `thresh(k, subs)` policy to
miniscript: when all sub-policies are keys and k == subs.len() this
compiles to a combination of and()s instead of a thresh().

The first commit adds some tests for the policy->miniscript compilation of different combinations of `k` and pubkeys under the `thresh()` policy.

The second commit special cases `k == pubkeys.len()` (when `> 20`) to eventually produce slightly more efficient Bitcoin Script:
```
<pubkey1> CHECKSIGVERIFY <pubkey2> CHECKSIGVERIFY  <pubkey3> CHECKSIGVERIFY ... <pubkeyN> CHECKSIG
```
Instead of
```
<pubkey1> CHECKSIG SWAP <pubkey2> CHECKSIG ADD SWAP  <pubkey3> CHECKSIG ADD SWAP ... <pubkeyN> CHECKSIG ADD N EQUAL
```

As mentioned by sipa on IRC we can also special-case `k==1` to a combination of `or()`s, but I left a `FIXME` as I intend to do this in a follow-up (can also push a commit here, as you prefer).

### DIFF
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -992,9 +992,22 @@ where
                     }
                 })
                 .collect();
+
             if key_vec.len() == subs.len() && subs.len() <= 20 {
                 insert_wrap!(AstElemExt::terminal(Terminal::Multi(k, key_vec)));
             }
+            // Not a threshold, it's always more optimal to translate it to and()s as we save the
+            // resulting threshold check (N EQUAL) in any case.
+            else if k == subs.len() {
+                let mut policy = subs.first().expect("No sub policy in thresh() ?").clone();
+                for sub in &subs[1..] {
+                    policy = Concrete::And(vec![sub.clone(), policy]);
+                }
+
+                ret = best_compilations(policy_cache, &policy, sat_prob, dissat_prob)?;
+            }
+
+            // FIXME: Should we also optimize thresh(1, subs) ?
         }
     }
     for k in ret.keys() {
@@ -1393,7 +1406,7 @@ mod tests {
                 keys[2]
             );
             let small_thresh_ms: SegwitMiniScript = small_thresh.compile().unwrap();
-            let small_thresh_ms_expected =
+            let small_thresh_ms_expected: SegwitMiniScript =
                 ms_str!("multi({},{},{},{})", k, keys[0], keys[1], keys[2]);
             assert_eq!(small_thresh_ms, small_thresh_ms_expected);
         }
@@ -1406,16 +1419,21 @@ mod tests {
                 keys.iter().map(|pubkey| Concrete::Key(*pubkey)).collect();
             let big_thresh = Concrete::Threshold(*k, pubkeys);
             let big_thresh_ms: SegwitMiniScript = big_thresh.compile().unwrap();
-            // N * (PUSH + pubkey + CHECKSIG + ADD + SWAP) + N EQUAL
-            assert_eq!(
-                big_thresh_ms.script_size(),
-                keys.len() * (1 + 33 + 3) + script_num_size(*k) + 1 - 2 // minus one SWAP and one ADD
-            );
-            let big_thresh_ms_expected = ms_str!(
+            if *k == 21 {
+                // N * (PUSH + pubkey + CHECKSIGVERIFY)
+                assert_eq!(big_thresh_ms.script_size(), keys.len() * (1 + 33 + 1));
+            } else {
+                // N * (PUSH + pubkey + CHECKSIG + ADD + SWAP) + N EQUAL
+                assert_eq!(
+                    big_thresh_ms.script_size(),
+                    keys.len() * (1 + 33 + 3) + script_num_size(*k) + 1 - 2 // minus one SWAP and one ADD
+                );
+                let big_thresh_ms_expected = ms_str!(
                 "thresh({},pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}),s:pk({}))",
                 k, keys[0], keys[1], keys[2], keys[3], keys[4], keys[5], keys[6], keys[7], keys[8], keys[9],keys[10], keys[11], keys[12], keys[13], keys[14], keys[15], keys[16], keys[17], keys[18], keys[19], keys[20]
             );
-            assert_eq!(big_thresh_ms, big_thresh_ms_expected);
+                assert_eq!(big_thresh_ms, big_thresh_ms_expected);
+            };
         }
     }
 }
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -1153,6 +1153,7 @@ mod tests {
 
     use miniscript::{satisfy, Segwitv0};
     use policy::Liftable;
+    use script_num_size;
     use BitcoinSig;
     use DummyKey;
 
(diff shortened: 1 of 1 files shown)


src/policy/compiler.rs
```diff
@@ -1397,18 +1412,27 @@ mod tests {
...
+        // FIXME: k = 1 (or 2, 3) doesn't work !
         for k in &[10, 21] {
             let pubkeys: Vec<Concrete<bitcoin::PublicKey>> =
                 keys.iter().map(|pubkey| Concrete::Key(*pubkey)).collect();
             let big_thresh = Concrete::Threshold(*k, pubkeys);
             let big_thresh_ms: SegwitMiniScript = big_thresh.compile().unwrap();
-            let big_thresh_ms_expected = ms_str!(
+            if *k == 21 {
+                assert!(
+                    matches!(big_thresh_ms.node, Terminal::AndV { .. } | Terminal::Check { .. }),
```
darosior, line 1425: This test is a bit hacky but I had no idea how to test it an other way as the compilation is not deterministic. Open to suggestion

src/policy/compiler.rs
```diff
@@ -992,8 +992,23 @@ where
...
                 .collect();
-            if key_vec.len() == subs.len() && subs.len() <= 20 {
-                insert_wrap!(AstElemExt::terminal(Terminal::Multi(k, key_vec)));
+            if key_vec.len() == subs.len() {
+                if subs.len() <= 20 {
+                    insert_wrap!(AstElemExt::terminal(Terminal::Multi(k, key_vec)));
+                } else if subs.len() == k {
+                    // If this is a N-of-N, then we can save some bytes in the witness
+                    // program by just CHECKSIGVERIFY on all keys instead of counting.
+                    let mut policy = Concrete::And(vec![
```
sanket1729, line 1001: I think this is not special to keys, The else if code can be moved to match thresh rule. Transforming the 

Concrete::thresh(k, ref subs) into chain of `Ands` if `k == n`. That way this will also work for things that are not keys and for n < 20 too.

src/policy/compiler.rs
sanket1729: We don't try to match the exact Miniscripts generated by the compilations. We only check whether the expected satisfaction costs are the same.

sanket1729:
The complexity of the compiler is proportional to the number of nested or's. I am a bit concerned in transforming `Thresh(1, subs)` into a chain of `or`s for higher value of `k`. cc @apoelstra

src/policy/compiler.rs
darosior: Hmm, I'm not sure it's efficient for `n < 20` ?

If using `CHECKMULTISIG` you have `4` bytes taken by OP codes (`n` pushed 2 times, `CHECKMULTISIG`, and the empty byte array), while you have `n` bytes taken by OP codes if using a set of `CHECKSIGVERIFY`s.

So, yeah, it's worth it for `n < 4 || n > 20`, will change that.

sanket1729: I mean the there can be other `subs` apart from keys that can benefit from chaining `And` compilation. It is most likely the case that `thresh(1,sha256(H),pk(C),pk(C),pk(C))` should be treated as `and(sha256(H),and(pk(C),and(pk(C),pk(C))` would produce a better compilation

src/policy/compiler.rs
darosior: But that's what other tests do ? (I mostly used `compile_misc()` as an example)

Ok so I kept the miniscript assertion for those that compiled straight (thresh => multi and thresh => thresh), and added script_size assertion on all.

darosior:
Amended to not use inclusive range syntax..

src/policy/compiler.rs
darosior: Ok, so we want this for policies where `k == subs.len()` *but not* when all `subs` are keys and `k in 4..=20` (in which case we want `multi`) ?

darosior: Finally went for:
- All keys and less than 20 keys ?
  - Yes => `multi`
  - No
    - => `k == subs.len()` ?
      - Yes => Nested `and()`s
      - No => `thresh()`

src/policy/compiler.rs
```diff
@@ -1377,6 +1378,46 @@ mod tests {
...
+            );
+            let small_thresh_ms: SegwitMiniScript = small_thresh.compile().unwrap();
+            let small_thresh_ms_expected =
+                ms_str!("multi({},{},{},{})", k, keys[0], keys[1], keys[2]);
+            assert_eq!(small_thresh_ms, small_thresh_ms_expected);
+        }
+
+        // Above 20 keys, thresh is compiled to a combination of and()s if it's a N of N,
+        // and to a ms thresh otherwise.
+        // FIXME: k = 1 (or 2, 3) doesn't work !
```
apoelstra, line 1415: What does "doesn't work" mean?

src/policy/compiler.rs
```diff
@@ -992,9 +992,21 @@ where
...
                 insert_wrap!(AstElemExt::terminal(Terminal::Multi(k, key_vec)));
             }
+            // Not a threshold, more optimal to convert it to and()s
+            else if k == subs.len() {
+                let mut policy = subs.first().expect("No sub policy in thresh() ?").clone();
+                for sub in &subs[1..] {
+                    policy = Concrete::And(vec![sub.clone(), policy]);
+                }
+
+                ret = best_compilations(policy_cache, &policy, sat_prob, dissat_prob)?;
```
apoelstra, line 1006: It is probably better to compare the `best_compilations` returned here to the ones we computed using the threshold logic, and actually make sure that we're taking the best ones. It seems intuitively true that nested `and`s will beat `thresh` every time but I'm not sure.

sanket1729: Agreed that `best_compilations` makes more sense here. But the current code always produces the most optimal compilation. Consider a particular compilation with all AND's replaced as BOOLAND. 
We would have 

`[E] [W] BOOLAND [W] BOOLAND .....`

The corresponding thresh would be:
`[E] [W] ADD [W] ADD [W] ADD.... k EQUAL`. 

Therefore, there exists at least one compilation that is better than the `thresh` one. In any case, this deserves a comment explaining this scenario.

src/policy/compiler.rs
sanket1729: I know you saw this issue. But for completeness of future viewers( refer #114 )

darosior: Will link the issue instead, and maybe remove the "FIXME" as it seems we're not going to fix it ?

sanket1729: Yes, that would be great.

src/policy/compiler.rs
darosior: I've expanded a bit more on the comment just above (was not sure how to phrase it though):
> // Not a threshold, it's always more optimal to translate it to and()s as we save the
> // resulting threshold check (N EQUAL) in any case.

sanket1729:
tAck 688b0a84673f52e463f77dc1338e786da2fb1742 . Will merge once #118 is merged so that the current invariant fuzz tests
`policy -> compile -> lift == policy -> lift` is preserved.

src/policy/compiler.rs
apoelstra: I think this phrasing is fine

apoelstra (approved):
Neat, lgtm. I'm not sure we want a "FIXME" comment for a future optimization we're unsure how to do, but we can address that later

darosior:
Will open an issue regarding the uncertainty of fixing the FIXME, this'll end up with either me implementing it or removing the FIXME.

PR #116, opened 2020-07-24
Add `DescriptorSecretKey` and `Descriptor::parse_secret()`

afilini:
I'm opening this right PR now mostly to gather feedback, since many things are still missing and need a lot of polishing.

This PR builds on top of #93 ([here](https://github.com/sgeisler/rust-miniscript/compare/2020-06-descriptor-key...afilini:sgeisler-descriptor-key)'s the diff with it), so it incorporates the `DescriptorKey` concept from @sgeisler and extends it to also support parsing of secrets (WIF or xprv) in Descriptors.

It implements pretty much what @apoelstra described in #93, which is having a way to parse a descriptor with secrets, but immediately separate them out and store them in a map, and only have public types in the descriptor. This is achieved with the `SplitSecret` trait that can be implemented on types that are `MiniscriptKey` that may contain secrets, and it adds a method `split_secret()` that when called returns a tuple of, roughly speaking, `(PublicType, Option<SecretType>)`.

This PR also adds the `Signer` trait which at the moment is very simple but again, I wanted to hear some feedback on how to move forward with it, and especially how to make it interact in a nice way with the `Satisfier` trait.

Also, I don't particularly like the name I chose for some structs, like `DescriptorKeyWithSecrets` or `DescriptorAndSigners` so if you have better ideas I'd love to hear them!

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -34,6 +34,7 @@ use bitcoin::blockdata::{opcodes, script};
 use bitcoin::hashes::hash160;
 use bitcoin::hashes::hex::FromHex;
 use bitcoin::secp256k1;
+use bitcoin::secp256k1::{Secp256k1, SignOnly};
 use bitcoin::util::bip32;
 use bitcoin::{self, Script};
 
@@ -155,7 +156,8 @@ impl fmt::Display for DescriptorSecretKey {
 /// handling of `bip32::ExtendedPubKey` and `bip32::ExtendedPrivKey`.
 pub trait InnerXKey: fmt::Display + str::FromStr {
     /// Returns the fingerprint of the key
-    fn xkey_fingerprint(&self) -> bip32::Fingerprint;
+    fn xkey_fingerprint(&self, secp: &Secp256k1<SignOnly>) -> bip32::Fingerprint;
+
     /// Returns whether hardened steps can be derived on the key
     ///
     /// `true` for `bip32::ExtendedPrivKey` and `false` for `bip32::ExtendedPubKey`.
@@ -163,7 +165,7 @@ pub trait InnerXKey: fmt::Display + str::FromStr {
 }
 
 impl InnerXKey for bip32::ExtendedPubKey {
-    fn xkey_fingerprint(&self) -> bip32::Fingerprint {
+    fn xkey_fingerprint(&self, _secp: &Secp256k1<SignOnly>) -> bip32::Fingerprint {
         self.fingerprint()
     }
 
@@ -173,8 +175,8 @@ impl InnerXKey for bip32::ExtendedPubKey {
 }
 
 impl InnerXKey for bip32::ExtendedPrivKey {
-    fn xkey_fingerprint(&self) -> bip32::Fingerprint {
-        self.fingerprint(&secp256k1::Secp256k1::signing_only())
+    fn xkey_fingerprint(&self, secp: &Secp256k1<SignOnly>) -> bip32::Fingerprint {
+        self.fingerprint(secp)
     }
 
     fn can_derive_hardened() -> bool {
@@ -197,10 +199,11 @@ pub struct DescriptorXKey<K: InnerXKey> {
 
 impl DescriptorSinglePriv {
     /// Returns the public key of this key
-    fn as_public(&self) -> Result<DescriptorSinglePub, DescriptorKeyParseError> {
-        let secp = secp256k1::Secp256k1::new();
-
-        let pub_key = self.key.public_key(&secp);
+    fn as_public(
+        &self,
+        secp: &Secp256k1<SignOnly>,
+    ) -> Result<DescriptorSinglePub, DescriptorKeyParseError> {
+        let pub_key = self.key.public_key(secp);
 
         Ok(DescriptorSinglePub {
             origin: self.origin.clone(),
@@ -216,9 +219,10 @@ impl DescriptorXKey<bip32::ExtendedPrivKey> {
     /// If the key already has an origin, the derivation steps applied will be appended to the path
     /// already present, otherwise this key will be treated as "root" key and an origin will be
     /// added with this key's fingerprint and the derivation steps applied.
-    fn as_public(&self) -> Result<DescriptorXKey<bip32::ExtendedPubKey>, DescriptorKeyParseError> {
-        let secp = secp256k1::Secp256k1::new();
-
+    fn as_public(
+        &self,
+        secp: &Secp256k1<SignOnly>,
+    ) -> Result<DescriptorXKey<bip32::ExtendedPubKey>, DescriptorKeyParseError> {
         let path_len = (&self.derivation_path).as_ref().len();
         let public_suffix_len = (&self.derivation_path)
             .into_iter()
@@ -309,12 +313,17 @@ impl DescriptorSecretKey {
     /// If the key is an "XPrv", the hardened derivation steps will be applied before converting it
     /// to a public key. See the documentation of [`DescriptorXKey<bip32::ExtendedPrivKey>::as_public`]
     /// for more details.
-    pub fn as_public(&self) -> Result<DescriptorPublicKey, DescriptorKeyParseError> {
+    pub fn as_public(
+        &self,
+        secp: &Secp256k1<SignOnly>,
+    ) -> Result<DescriptorPublicKey, DescriptorKeyParseError> {
         Ok(match self {
             &DescriptorSecretKey::SinglePriv(ref sk) => {
-                DescriptorPublicKey::SinglePub(sk.as_public()?)
+                DescriptorPublicKey::SinglePub(sk.as_public(secp)?)
             }
-            &DescriptorSecretKey::XPrv(ref xprv) => DescriptorPublicKey::XPub(xprv.as_public()?),
+            &DescriptorSecretKey::XPrv(ref xprv) => {
+                DescriptorPublicKey::XPub(xprv.as_public(secp)?)
+            }
         })
     }
 }
@@ -546,31 +555,39 @@ impl<K: InnerXKey> DescriptorXKey<K> {
     /// use miniscript::bitcoin::util::bip32;
     /// use miniscript::descriptor::DescriptorPublicKey;
     ///
+    /// let ctx = miniscript::bitcoin::secp256k1::Secp256k1::signing_only();
+    ///
     /// let key = DescriptorPublicKey::from_str("[d34db33f/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*")?;
     /// let xpub = match key {
     ///     DescriptorPublicKey::XPub(xpub) => xpub,
     ///     _ => panic!("Parsing Error"),
     /// };
     ///
-    /// assert_eq!(xpub.matches(&(bip32::Fingerprint::from_str("d34db33f")?, bip32::DerivationPath::from_str("m/44'/0'/0'/1/42")?)), Some(bip32::DerivationPath::from_str("m/44'/0'/0'/1")?));
-    /// assert_eq!(xpub.matches(&(bip32::Fingerprint::from_str("ffffffff")?, bip32::DerivationPath::from_str("m/44'/0'/0'/1/42")?)), None);
-    /// assert_eq!(xpub.matches(&(bip32::Fingerprint::from_str("d34db33f")?, bip32::DerivationPath::from_str("m/44'/0'/0'/100/0")?)), None);
+    /// assert_eq!(xpub.matches(&(bip32::Fingerprint::from_str("d34db33f")?, bip32::DerivationPath::from_str("m/44'/0'/0'/1/42")?), &ctx), Some(bip32::DerivationPath::from_str("m/44'/0'/0'/1")?));
+    /// assert_eq!(xpub.matches(&(bip32::Fingerprint::from_str("ffffffff")?, bip32::DerivationPath::from_str("m/44'/0'/0'/1/42")?), &ctx), None);
+    /// assert_eq!(xpub.matches(&(bip32::Fingerprint::from_str("d34db33f")?, bip32::DerivationPath::from_str("m/44'/0'/0'/100/0")?), &ctx), None);
     /// # Ok(())
     /// # }
     /// # body().unwrap()
     /// ```
-    pub fn matches(&self, keysource: &bip32::KeySource) -> Option<bip32::DerivationPath> {
+    pub fn matches(
+        &self,
+        keysource: &bip32::KeySource,
+        secp: &Secp256k1<SignOnly>,
+    ) -> Option<bip32::DerivationPath> {
         let (fingerprint, path) = keysource;
 
         let (compare_fingerprint, compare_path) = match &self.origin {
             &Some((fingerprint, ref path)) => (
-                fingerprint.clone(),
+                fingerprint,
                 path.into_iter()
                     .chain(self.derivation_path.into_iter())
-                    .cloned()
                     .collect(),
             ),
-            &None => (self.xkey.xkey_fingerprint(), self.derivation_path.clone()),
+            &None => (
+                self.xkey.xkey_fingerprint(secp),
+                self.derivation_path.into_iter().collect::<Vec<_>>(),
+            ),
         };
 
         let path_excluding_wildcard = if self.is_wildcard && path.as_ref().len() > 0 {
@@ -582,7 +599,11 @@ impl<K: InnerXKey> DescriptorXKey<K> {
             path.clone()
         };
 
-        if &compare_fingerprint == fingerprint && compare_path == path_excluding_wildcard {
+        if &compare_fingerprint == fingerprint
+            && compare_path
+                .into_iter()
+                .eq(path_excluding_wildcard.into_iter())
+        {
             Some(path_excluding_wildcard)
         } else {
             None
@@ -1083,21 +1104,22 @@ impl Descriptor<DescriptorPublicKey> {
     /// Internally turns every secret key found into the corresponding public key and then returns a
     /// a descriptor that only contains public keys and a map to lookup the secret key given a public key.
     pub fn parse_secret(s: &str) -> Result<(Descriptor<DescriptorPublicKey>, KeyMap), Error> {
-        fn parse_key(
-            s: &String,
-            keymap: &mut KeyMap,
-        ) -> Result<DescriptorPublicKey, DescriptorKeyParseError> {
+        let secp = secp256k1::Secp256k1::signing_only();
+
+        let parse_key = |s: &String,
+                         key_map: &mut KeyMap|
+         -> Result<DescriptorPublicKey, DescriptorKeyParseError> {
             let (public_key, secret_key) = match DescriptorSecretKey::from_str(s) {
-                Ok(sk) => (sk.as_public()?, Some(sk)),
+                Ok(sk) => (sk.as_public(&secp)?, Some(sk)),
                 Err(_) => (DescriptorPublicKey::from_str(s)?, None),
             };
 
             if let Some(secret_key) = secret_key {
-                keymap.insert(public_key.clone(), secret_key);
+                key_map.insert(public_key.clone(), secret_key);
             }
 
             Ok(public_key)
-        }
+        };
 
         let mut keymap_pk = KeyMap::new();
         let mut keymap_pkh = KeyMap::new();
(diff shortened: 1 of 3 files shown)


afilini:
I pushed some experiments I've been doing with the `Signer` trait. I still don't feel 100% confident about the interface, specifically if it's better to take a `&mut psbt::PartiallySignedTransaction` and apply the signatures inside the signer (what's currently implemented) as opposed to taking a normal reference, returning the signature and letting the caller handle everything else.

I initially liked the first approach better because it hides the logic of pushing the signature in the `partial_sigs` map from the caller, but that actually forces me to clone the psbt somewhere, which is not optimal.

I guess the `&mut` approach is better if we expect people to directly use the signers outside of the code provided by this crate, while the other one is less "user-friendly" but it allows for some minor optimizations inside this code.

The other big question is whether I should add an explicit flag to ask for a BIP143 signature or try to automatically determine it by looking at the `witness_utxo`/`non_witness_utxo` fields like I'm doing now. Considering the recent issue with Trezors and the fact that they decided to require `non_witness_utxo`s for segwit psbts too, I think it might be necessary to have an explicit flag to avoid issues down the road.

sanket1729:
> The other big question is whether I should add an explicit flag to ask for a BIP143 signature or try to automatically determine it by looking at the witness_utxo/non_witness_utxo fields like I'm doing now. Considering the recent issue with Trezors and the fact that they decided to require non_witness_utxos for segwit psbts too, I think it might be necessary to have an explicit flag to avoid issues down the road.

I don't completely follow this, what is the issue to determine by looking at `witness_utxo` field? Can you elaborate how this might lead to more issues ahead?

As for the role of miniscript in PSBT stack, we originally thought that miniscript should only be required for finalizing `psbt`. Other roles must fill in the partial sigs as required. I read BIP32, BIP 174 a couple days ago, so maybe I am wrong with these. We only considered the following workflow

1) The creator creates a `psbt` with unsigned tx and passes it on for signing. 
2) The signers can analyze the Miniscript, infer which keys are in the miniscript; sign the tx with appropriate sighash flag and add the partial sigs to the pbst. 
3) The finalizer finally looks at the partial sigs and each inputs and creates the transaction. The role of satisfier triat is to construct the final pbst as implemented here. https://github.com/rust-bitcoin/rust-miniscript/blob/87a47e242dc19c517ca06c558cf79b43eec22abf/src/psbt/mod.rs#L125 

Please note that the `pbst` implementation is incomplete and not tested. 


My understanding that the overall flow should be as follows: (cc @apoelstra )
```
fn main() {
    // get descriptor and signers
    let descriptor, signers = .....

    /// Create unsigned initial psbt
    let psbt: bitcoin::util::psbt::PartiallySignedTransaction = .......

    // Note that this could be a distributed process
    // Sign each input and
    for input in psbt.inputs{
    // Add sig for each input. Again could be distributed process.
        for signer in signers{
        //Add partial sig here.
        signer.sign(&mut psbt, input_index)    
        }
    }
    // Pass on to the finalizer. The Satisfier trait is already implemented for psbt input.
    pbst::finalize(& mut psbt);
}
```
I did not look at all of the code yet, but most of it looks good. 
A couple of questions that I am still unclear about: 
Is there any special need for `DescriptorWithSigners` struct to implement `FromTree`. I think that this only requires `FromStr` and then the struct `MiniscriptWithSigners` should no longer be required.

afilini:
> I don't completely follow this, what is the issue to determine by looking at witness_utxo field? Can you elaborate how this might lead to more issues ahead?

I don't know how realistic this is, I was just thinking that maybe some wallets could decide in the future to always include `non_witness_utxo`s and potentially remove `witness_utxo`s entirely to save some space, because after all you can find everyting you need from the `non_witness_utxo`. If that was the case, PSBTs built by these wallets would be mistaken for non-segwit with the logic I have implemented currently.

> As for the role of miniscript in PSBT stack, we originally thought that miniscript should only be required for finalizing psbt. Other roles must fill in the partial sigs as required.

If you feel like the signers are out of scope I can totally remove them. There was a comment on #93 (https://github.com/rust-bitcoin/rust-miniscript/pull/93#discussion_r450488856) that said "and have our [signers] map use the Satisfier trait to allow users to produce signatures etc with it." so I interpreted that as "implement Satisfier on SignersContaner so that it can produce signatures on-demand", but I might have just misunderstood something.

> I think that this only requires FromStr and then the struct MiniscriptWithSigners should no longer be required.

Yeah, I think you are right. I was trying to keep the structs "parallel" (Miniscript -> MiniscriptWithSigners, Descriptor -> DescriptorWithSigners) but now that I think about it, the signers pretty much only make sense with descriptors, so I could remove it `MiniscriptWithSigners`. I'll try working on that today.

apoelstra:
I agree that PSBT signing is out of scope ... if only because it is so simple. As Sanket says, you just compute the sighash (which rust-bitcoin has functionality for), produce a signature (i.e. call `sign` in rust-secp) and attach it to the PSBT (and the PSBT struct exposes a `BTreeMap` for you to insert the signature in). Given this, I'm not sure I see the purpose of the `Signer` trait.

As for implementing `Satisfier` on a map from public keys to private keys (and from `Pk::Hash` to private keys), I think this seems totally reasonable. And I agree that the way to do this is to create something like the `PSBTSigningContext` structure which contains a reference to a PSBT, a reference to a map, and an input index. But the implementation in this PR seems overly complicated. I will do a detailed review with more comments.

apoelstra:
Some more thoughts:

* I think `DescriptorWithSigners` and and `MiniscriptWithSigners` and `DescriptorKeyWithSecrets` is too clever by half. I think it'd be better to have two concrete types, `DescriptorPublicKey` and `DescriptorSecretKey`, which are basically the same except that one has secret-key variants of the other.
* Then I think rather than having a separate type `DescriptorWithSigners` where we impl `FromStr` on it, we should just add a method to `Descriptor` called `parse_secret` or something, which returns a `Descriptor<DescriptorPublicKey>` as well as a `HashMap<DescriptorPublicKey, DescriptorPrivateKey>`. I think it's fine to wrap the `HashMap` in a new type, maybe call it `KeyMap`
* Then you can implement a method on `KeyMap` which takes an input index and a mutable reference to a PSBT, which computes a signature and inserts it in the right place. (Notice that this does not actually involve Miniscript or descriptors at all, it's just a convenience function...I am ambivalent about whether we ought to have it in this library at all.)
* Then you can implement `Satisfier` directly on `Psbt`, which takes signatures from the input map, scripts from the input map, and outputs a valid witness.

apoelstra:
Regarding my comment in 93 about "using `Satisfier` to produce signatures", I was just wrong. The `Satisfier` trait's job is to produce a witness, given signatures, not to produce signatures (which it cannot do because the API provides nowhere to input a sighash nor anywhere to output a signature). If we add such a function, and I don't necessarily think we should, it should be a standalone function.

src/miniscript/context.rs
```diff
@@ -71,6 +80,11 @@ pub trait ScriptContext:
     fn check_frag_validity<Pk: MiniscriptKey, Ctx: ScriptContext>(
         _frag: &Terminal<Pk, Ctx>,
     ) -> Result<(), ScriptContextError>;
+
+    fn sighash(
+        psbt: &psbt::PartiallySignedTransaction,
+        input_index: usize,
+    ) -> Result<(SigHash, SigHashType), SignerError>;
```
apoelstra, line 87: This is a cool idea, putting this into `ScriptContext`. I think it should be changed though as

```
fn sighash(tx: &bitcoin::Transaction, input_index: usize, sighash_type: SigHashType) -> SigHash;
```

Then the function is more generic than just PSBT, and also it can't fail.

apoelstra: The existing functionality I think is still worth including in the library, but as a standalone function which takes a PSBT and input index, does the appropriate lookups (which may fail) then calls `ScriptContext::sighash` to do the actual sighash computation.

afilini:
The reason why I kinda liked the idea of connecting the signing flow with the `Satisfier` trait is that it seemed nice to be able to only generate the signatures you require and nothing more, on-demand, but that really makes a difference only when somebody has the ability to provide multiple signatures in the same script, which is kind of a weird situation and probably doesn't justify all the complexity that signing carries: it's much easier to just sign with every key you have (probably just one in most cases) and then the satisfier will only pick the ones it needs.

So here's what I'll try to do for this PR:

* Remove the `Signer` trait, `PSBTSigningContext`, etc
* Rename `DescriptorKey` to `DescriptorPublicKey`
* Rename `DescriptorKeyWithSecrets` to `DescriptorSecretKey`, but don't make it `MiniscriptKey` since we won't care about parsing directly in `Descriptor::from_str` (which also means removing all the custom implementation for `Eq`, `Ord`, etc since we won't need them anymore)
* Add `Descriptor::parse_secret` that parses the string as `Descriptor<String>` internally first, then parses and maps every individual key using `translate_pk()` and at the same time populates the `KeyMap` with secrets, if they keys contain any
* Modify `ScriptContext::sighash` to use transactions instead of PSBTs and move the PSBT stuff in a separate, standalone function

Hope I understood everything right, thanks for the feedback!

sgeisler:
Just so we are on the same page: `Descriptor::parse_secret` would return `Result<(Descriptor<DescriptorPublicKey>, HashMap<DescriptorPublicKey, DescriptorSecretKey>), _>`?

Another thing I noticed is that we might want to add a `Hash` variant to the `DescriptorPublicKey` enum. It's not pretty as some methods will need to panic or we'll need to sacrifice the `ToPublicKey` impl. But not being able to build a `pkh(<hash>)` descriptor is very annoying because one doesn't even know the pubkey.

apoelstra:
That's a neat idea to generate only the actually-used signatures. Unfortunately it doesn't really fit into the workflows we imagined or the API of the `Satisfier` trait, and I don't think it's worth the added complexity to try to support it.

If you wanted to do this, it is possible to do outside the library, by writing your own type that knows how to sign (I guess it would include a sighash or means to compute this) and implementing `Satisfier` on it. But my feeling is that this is too niche to be worth including in the library.

Everything else you say @afilini sounds good to me! @sgeisler that does sound like the right method signature.

apoelstra:
@sgeisler regarding a `Hash` variant, can you bring this up on IRC and ping me and sipa? Even with PSBT, I don't think there should be any situations where you don't know a public key but still want to create a descriptor.

sanket1729:
Even I don't know why but there is an `addr` descriptor in bitcoin-core that is roughly equivalent to `Hash`. But, even I am curious about use-cases where one might need `Hash` descriptor. 

Also, the current implementation of PSBT finalizer is incomplete. I hope to have PR for it tomorrow.

dr-orlovsky:
@afilini what do you think about extending xpub/xpriv derivation paths with arbitrary tweaks, which can be useful for situations like in RGB?

afilini:
It's probably out of scope in my opinion, considering that you can very easily define your own `MiniscriptKey` types outside of this library. It's trivial to wrap `DescriptorPublicKey` and store whatever you need in your custom struct.

darosior:
Hi @afilini, i'd be pretty interested to see this moving forward (at least the xpub part).
I'd happily help you with it, thought about making a PR for the `DescriptorPublicKey` of your above list mainly based of #93. What do you think ?

afilini:
I think I will have some time this weekend to come back to this and finish it up. I think most of the code is already there, it needs cleanup and docs. I was also planning to add commit https://github.com/rust-bitcoin/rust-miniscript/pull/93/commits/d9fa9455b1e8e815a7babb65bc27e2abb86d9dcb from @sgeisler in this branch since I think it could be useful.

The only thing that's missing with the "refactoring" I've done is the sighash computation part, which I was trying to rewrite as suggested to take a transaction instead of a PSBT, but for whatever reason I got kinda stuck (IIRC the interface was awful and I couldn't figure out a way to improve it) and never ended up finishing it. But since this PR doesn't do signing anymore maybe we could also go ahead without sighash and open a separate PR if we really want it.

The only thing (other than me rebasing and finishing the code) that could block merging this PR is the fact that it needs a git version of rust-bitcoin, since I don't think there has been a release yet since the required changes have been merged. I don't know the release schedule of rust-bitcoin, but I think it's been a while since the last release, so hopefully that's coming soon.

As a side note, I've been using this branch in some of my projects for the past few weeks and I don't think I've noticed any particular issue/regression with it, which is nice I guess.

darosior:
> The only thing that's missing with the "refactoring" I've done is the sighash computation part, which I was trying to rewrite as suggested to take a transaction instead of a PSBT, but for whatever reason I got kinda stuck (IIRC the interface was awful and I couldn't figure out a way to improve it) and never ended up finishing it. But since this PR doesn't do signing anymore maybe we could also go ahead without sighash and open a separate PR if we really want it.

Yeah, @sanket1729 has been doing some PSBT stuff related to sighash too lately if i'm not mistaking.

> The only thing (other than me rebasing and finishing the code) that could block merging this PR is the fact that it needs a git version of rust-bitcoin, since I don't think there has been a release yet since the required changes have been merged. I don't know the release schedule of rust-bitcoin, but I think it's been a while since the last release, so hopefully that's coming soon.

It has just been bumped, and my cleaned up / improved version of #93 is rebased on top of #130 can be found here https://github.com/darosior/rust-miniscript/commits/descriptor_publickey (currently WIP, need some more tests squashed and raw pubkey with origin implementation: working on it locally).

afilini:
This is now rebased on #131

sanket1729 (commented):
Left some minor nits. It would be great if we can use `Keysource` from bitcoin 0.25 wherever possible.

afilini:
Just rebased this on master, it's ready to be reviewed :)

src/descriptor/mod.rs
```diff
@@ -87,10 +87,71 @@ pub struct DescriptorSinglePub {
...
+    }
+
+    fn can_derive_hardened() -> bool {
+        false
+    }
+}
+
+impl InnerXKey for bip32::ExtendedPrivKey {
+    fn xkey_fingerprint(&self) -> bip32::Fingerprint {
+        self.fingerprint(&secp256k1::Secp256k1::signing_only())
```
apoelstra, line 145: :skull_and_crossbones: we definitely should not be creating a one-off secp context for this

src/descriptor/mod.rs
```diff
@@ -280,35 +398,71 @@ impl DescriptorPublicKey {
...
+    /// };
+    ///
+    /// assert_eq!(xpub.matches(&(bip32::Fingerprint::from_str("d34db33f")?, bip32::DerivationPath::from_str("m/44'/0'/0'/1/42")?)), Some(bip32::DerivationPath::from_str("m/44'/0'/0'/1")?));
+    /// assert_eq!(xpub.matches(&(bip32::Fingerprint::from_str("ffffffff")?, bip32::DerivationPath::from_str("m/44'/0'/0'/1/42")?)), None);
+    /// assert_eq!(xpub.matches(&(bip32::Fingerprint::from_str("d34db33f")?, bip32::DerivationPath::from_str("m/44'/0'/0'/100/0")?)), None);
+    /// # Ok(())
+    /// # }
+    /// # body().unwrap()
+    /// ```
+    pub fn matches(&self, keysource: &bip32::KeySource) -> Option<bip32::DerivationPath> {
```
apoelstra, line 510: I think most all of the clones in this function can be avoided

src/descriptor/mod.rs
```diff
@@ -156,7 +159,64 @@ pub struct DescriptorXKey<K: InnerXKey> {
     pub is_wildcard: bool,
 }
 
-#[derive(Debug, PartialEq, Clone, Copy)]
+impl DescriptorSinglePriv {
+    fn as_public(&self) -> Result<DescriptorSinglePub, DescriptorKeyParseError> {
+        let secp = secp256k1::Secp256k1::new();
```
apoelstra, line 164: Please do not create ephemeral secp contexts

src/descriptor/mod.rs
apoelstra: Ah, not the ones in the assignment to `path_excluding_wildcard`, since that result gets returned and needs to be owned.

src/descriptor/mod.rs
```diff
@@ -156,7 +159,64 @@ pub struct DescriptorXKey<K: InnerXKey> {
     pub is_wildcard: bool,
 }
 
-#[derive(Debug, PartialEq, Clone, Copy)]
+impl DescriptorSinglePriv {
+    fn as_public(&self) -> Result<DescriptorSinglePub, DescriptorKeyParseError> {
```
apoelstra, line 163: I think this should be renamed to emphasize that it's dropping information

apoelstra: Oh, no, I misunderstood it

src/descriptor/mod.rs
```diff
@@ -196,6 +256,17 @@ impl fmt::Display for DescriptorPublicKey {
     }
 }
 
+impl DescriptorSecretKey {
+    pub fn as_public(&self) -> Result<DescriptorPublicKey, DescriptorKeyParseError> {
```
apoelstra, line 260: Needs a doccomment, and should mention that it will drop some of the derivation path if this involves hardened derivations. (Maybe it should be renamed to reflect this, but I'm not sure what a better name would be.)

src/descriptor/mod.rs
```diff
@@ -827,6 +898,57 @@ impl Descriptor<DescriptorPublicKey> {
         )
         .expect("Translation fn can't fail.")
     }
+
+    pub fn parse_secret(s: &str) -> Result<(Descriptor<DescriptorPublicKey>, KeyMap), Error> {
```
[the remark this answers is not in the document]
apoelstra: I think this should be renamed to `parse_descriptor`

src/descriptor/mod.rs
```diff
@@ -871,6 +942,61 @@ impl Descriptor<DescriptorPublicKey> {
...
+                |pkh| parse_key(pkh, &mut keymap_pkh),
+            )
+            .map_err(|e| Error::Unexpected(e.to_string()))?;
+
+        keymap_pk.extend(keymap_pkh.into_iter());
+
+        Ok((descriptor, keymap_pk))
+    }
+
+    pub fn to_string_with_secret(&self, key_map: &KeyMap) -> String {
```
apoelstra, line 983: This could also be done with far fewer allocations, but we can leave it to a followup PR

afilini:
Thanks, sorry it took me a while to fix this!

I think most of the comments should have been addressed: everything public is now documented and I removed pretty much all the ephemeral secp contexts except for one in `Descriptor::parse_secret`, mostly because I think it's not a critical method in terms of performance, and in general I think it keeps the public API a bit more "clean". But if you think this should also be removed and taken as argument I'm happy to change it, I don't have any strong opinion on that and it was just my personal preference.

Other than that, I don't think I see what you mean with your comment on `Descriptor::to_string_with_secret`: I mean, I can see that it does many allocations and it clones the public keys needlessly, but I don't think `&String` is a `MiniscriptKey`, so I don't think you can avoid cloning them, unless you completely change approach. If it's good for you to do that in a separate PR I could try to work on it, I guess I just need a bit of help!

apoelstra:
What I was thinking with `to_string_with_secret` was that you'd make a type like
```
struct MapKey<'a, 'b> {
    map: &'a KeyMap,
    key: &'b DescriptorPublicKey,
}
```
then impl `fmt::Display` on this in a way that did the lookup and then either wrote the secret key or public key to the `fmt::Formatter` without allocating.

Unfortunately you still have to translate the input descriptor to a descriptor which uses this weird type as keys, so you can't avoid the allocations inherent in that.

src/descriptor/mod.rs
apoelstra: Can you rebase so that the doccomment was there originally? Otherwise the individual commits do not compile.

You can check whether the commits compile by running `git rebase -x 'cargo test && cargo +1.29.0 test' master` on your branch.

apoelstra:
I think we can fix the dissat thing after the fact.

@afilini can you rebase so that acb1164d4779e36062ce6d25c333954a9f9474d8 compiles?

src/descriptor/mod.rs
afilini: It should be good now..

src/descriptor/mod.rs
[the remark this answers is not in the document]
apoelstra: I'm still not thrilled but I can't think of anything better :). `parse_descriptor_with_keymap` is probably too long for such an important function.

In any case, I'm not going to hold up the PR for this.

apoelstra (approved):
ack 4a1e94c0a440c67702dc8cbb8b407f9b1f807913

There are still a few clones that I think we could eliminate, but we can deal with them in a later PR. We should probably allocate everything because we've added a lot of new APIs recently.

apoelstra:
Lol, I meant "audit" not "allocate"

src/descriptor/mod.rs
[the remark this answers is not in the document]
afilini: Yeah it sounds a bit weird to me as well, but I also don't have any better ideas unfortunately..

PR #118, opened 2020-07-31
Policy entailment

sanket1729:
Implemented policy entailment algorithm. It is also possible to optimize this even further by memorizing results in hashmap, but this works fast for desired use cases. 

The first commit fixes some bugs in normalization code and also changes the logic so that we convert everything to `thresh`. As discussed, we do not require `and` and `or` in semantic analysis. Note that the `FromStr` and `Debug`, `Display` are updated to the to print `and` and `or` while are internally represented as `thresh`.

 This will be helpful in #113 where it is possible to switch between `And` and `thresh` in order to maintain the current invariant that  
`policy -> compile -> lift == policy -> lift`.

### DIFF
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -72,6 +72,10 @@ pub enum PolicyError {
     ZeroTime,
     /// `after` fragment can only have ` n < 2^31`
     TimeTooFar,
+    /// Semantic Policy Error: `And` `Or` fragments must take args: k > 1
+    InsufficientArgsforAnd,
+    /// Semantic Policy Error: `And` `Or` fragments must take args: k > 1
+    InsufficientArgsforOr,
 }
 
 impl error::Error for PolicyError {
@@ -97,6 +101,12 @@ impl fmt::Display for PolicyError {
                 f.write_str("Relative/Absolute time must be less than 2^31; n < 2^31")
             }
             PolicyError::ZeroTime => f.write_str("Time must be greater than 0; n > 0"),
+            PolicyError::InsufficientArgsforAnd => {
+                f.write_str("Semantic Policy 'And' fragment must have atleast 2 args ")
+            }
+            PolicyError::InsufficientArgsforOr => {
+                f.write_str("Semantic Policy 'Or' fragment must have atleast 2 args ")
+            }
         }
     }
 }
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -19,6 +19,7 @@ use bitcoin::hashes::hex::FromHex;
 use bitcoin::hashes::{hash160, ripemd160, sha256, sha256d};
 use std::{error, fmt, str};
 
+use super::ENTAILMENT_MAX_TERMINALS;
 use errstr;
 use expression::{self, FromTree};
 #[cfg(feature = "compiler")]
@@ -76,6 +77,8 @@ pub enum PolicyError {
     InsufficientArgsforAnd,
     /// Semantic Policy Error: `And` `Or` fragments must take args: k > 1
     InsufficientArgsforOr,
+    /// Entailment max terminals exceeded
+    EntailmentMaxTerminals,
 }
 
 impl error::Error for PolicyError {
@@ -107,6 +110,11 @@ impl fmt::Display for PolicyError {
             PolicyError::InsufficientArgsforOr => {
                 f.write_str("Semantic Policy 'Or' fragment must have atleast 2 args ")
             }
+            PolicyError::EntailmentMaxTerminals => write!(
+                f,
+                "Policy entailment only supports {} terminals",
+                ENTAILMENT_MAX_TERMINALS
+            ),
         }
     }
 }
--- a/src/policy/mod.rs
+++ b/src/policy/mod.rs
@@ -37,6 +37,8 @@ pub use self::concrete::Policy as Concrete;
 pub use self::semantic::Policy as Semantic;
 use MiniscriptKey;
 
+/// Policy entailment algorithm maximum number of terminals allowed
+const ENTAILMENT_MAX_TERMINALS: usize = 20;
 /// Trait describing script representations which can be lifted into
 /// an abstract policy, by discarding information.
 /// After Lifting all policies are converted into `KeyHash(Pk::HasH)` to
--- a/src/policy/mod.rs
+++ b/src/policy/mod.rs
@@ -74,17 +74,20 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Liftable<Pk> for Terminal<Pk, Ctx> {
             | Terminal::NonZero(ref sub)
             | Terminal::ZeroNotEqual(ref sub) => sub.node.lift(),
             Terminal::AndV(ref left, ref right) | Terminal::AndB(ref left, ref right) => {
-                Semantic::And(vec![left.node.lift(), right.node.lift()])
+                Semantic::Threshold(2, vec![left.node.lift(), right.node.lift()])
             }
-            Terminal::AndOr(ref a, ref b, ref c) => Semantic::Or(vec![
-                Semantic::And(vec![a.node.lift(), c.node.lift()]),
-                b.node.lift(),
-            ]),
+            Terminal::AndOr(ref a, ref b, ref c) => Semantic::Threshold(
+                1,
+                vec![
+                    Semantic::Threshold(2, vec![a.node.lift(), c.node.lift()]),
+                    b.node.lift(),
+                ],
+            ),
             Terminal::OrB(ref left, ref right)
             | Terminal::OrD(ref left, ref right)
             | Terminal::OrC(ref left, ref right)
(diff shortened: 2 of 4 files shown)


src/policy/semantic.rs
```diff
@@ -286,36 +286,93 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+                        k => Policy::Threshold(k, ret_subs),
+                    }
+                } else if k == 1 {
+                    let mut ret_subs = Vec::with_capacity(subs.len());
+                    for sub in subs {
+                        match sub.normalized() {
+                            Policy::Trivial => return Policy::Trivial,
+                            Policy::Unsatisfiable => {}
+                            // sanity check,
+                            // this should reach as we should have substituted all or's
```
darosior, line 395: *never* reach ?

sanket1729: good catch

src/policy/semantic.rs
```diff
@@ -237,36 +199,44 @@ where
...
                     return Err(errstr(top.args[0].args[0].name));
                 }
 
                 let thresh = expression::parse_num(top.args[0].name)?;
-                if thresh >= nsubs {
+
+                // thresh(1) and thresh(n) are disallowed in semantic policies
+                if thresh <= 1 || thresh >= (nsubs as u32 - 1) {
+                    return Err(errstr(
+                        "Semantic Policy thresh cannot have k = 1 or k =n, use 'and'/'or' instead",
```
apoelstra, line 308: Should use backticks around `and` and `or`; also missing a space before `n`

src/policy/semantic.rs
```diff
@@ -286,36 +256,86 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
-                for sub in subs {
-                    match sub.normalized() {
-                        Policy::Trivial => {}
-                        Policy::Unsatisfiable => return Policy::Unsatisfiable,
-                        Policy::And(and_subs) => ret_subs.extend(and_subs),
-                        x => ret_subs.push(x),
+            Policy::Threshold(k, subs) => {
+                if k == subs.len() {
+                    let mut ret_subs = Vec::with_capacity(subs.len());
+                    for sub in subs.clone() {
```
apoelstra, line 334: :grimacing: why this clone?

src/policy/semantic.rs
```diff
@@ -286,36 +256,86 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+                            }
+                            x => ret_subs.push(x),
+                        }
+                    }
+                    match ret_subs.len() {
+                        0 => Policy::Unsatisfiable,
+                        1 => ret_subs.pop().unwrap(),
+                        _ => Policy::Threshold(1, ret_subs),
+                    }
+                } else {
```
apoelstra, line 370: Do we actually need the `and` and `or` branches or can we just use the general one always?

src/policy/semantic.rs
```diff
@@ -82,6 +82,78 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
     }
+
+    /// This function computes whether the current policy entails the second one.
+    /// A |- B means every sasifaction of A is also a satisfaction of B.
+    /// This implementation will run slow for larger policies but should be sufficient for
+    /// most practical policies.
+
+    // This algorithm has a naive implementation. It is possible to optimize this
+    // by memoizing and maintaining a hashmap.
+    pub fn entails(self, other: Policy<Pk>) -> bool {
```
apoelstra, line 93: I think we should require the caller do the normalization step, and we should be able to take `other` by reference

src/policy/semantic.rs
```diff
@@ -82,6 +82,78 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+    // Assumes that the current policy is normalized.
+    fn first_constraint(&self) -> Policy<Pk> {
+        debug_assert!(self.clone().normalized() == self.clone());
+        match self {
+            &Policy::Threshold(_k, ref subs) => subs[0].first_constraint(),
+            first => first.clone(),
+        }
+    }
+
+    // Helper function that takes in witness and it's availability,
```
apoelstra, line 126: Wrong `it's`

src/policy/semantic.rs
apoelstra: Ah I see, the algorithm itself does no normalization, so after setting things to trivial/unsatisfiable, we need to renormalize anyway.

apoelstra: I'd be okay with this if you put a recursion depth counter in and refused to go further than 16 deep say. As written this function may run effectively forever and there is no good way for the user to detect or prevent this.

src/policy/semantic.rs
sanket1729: We do a recursive combination of branches. `thresh(1,pk(A),thresh(1,pk(B),pk(C)))` would be reread as `thresh(1,pk(A),pk(B),pk(C))`

src/policy/semantic.rs
sanket1729: oops. I don't remember why I have it.

src/policy/semantic.rs
sanket1729: Actually, you may be right. Thinking a bit more about this

sanket1729: You are right, all of them can be covered in one general condition. I blindly translated the `And` and `or` thinking that those were necessary for merging

src/policy/semantic.rs
sanket1729: A simpler way might be limit to the number of distinct witnesses as the depth is exactly equal to that. I think 24 should be good because we can support liquid as that is 15 + 3 = 18 keys

apoelstra: Agreed, we should add a method to `Policy` which returns this. Call it `n_terminals` or something.

src/policy/semantic.rs
```diff
@@ -401,16 +401,16 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
     }
 
     /// Helper function to detect a true/trivial policy
-    pub fn is_trivial(&self) -> bool {
-        match *self {
+    pub fn is_trivial(self) -> bool {
```
apoelstra, line 404: These should definitely not consume `self`

src/policy/semantic.rs
```diff
@@ -401,16 +401,16 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+        match self.normalized() {
```
apoelstra, line 405: This function is now very expensive and may fail for large policies

src/policy/concrete.rs
```diff
@@ -19,6 +19,7 @@ use bitcoin::hashes::hex::FromHex;
 use bitcoin::hashes::{hash160, ripemd160, sha256, sha256d};
 use std::{error, fmt, str};
 
+use super::ENTAILMENT_RECURSION_DEPTH;
```
apoelstra, line 22: This should have a different name, it is a limit on the number of constraints rather than the recursion depth

src/policy/concrete.rs
```diff
@@ -76,6 +77,8 @@ pub enum PolicyError {
     InsufficientArgsforAnd,
     /// Semantic Policy Error: `And` `Or` fragments must take args: k > 1
     InsufficientArgsforOr,
+    /// Entailment recrusion depth exceeded
+    EntailmentRecursionDepthExceeded,
```
apoelstra, line 81: Ditto here, I think we shouldn't call this a recursion depth limit

src/policy/semantic.rs
sanket1729: My rationale was trivial should return `true` for the normalized policy. Or(true, largepolicy) should also be trivial

src/policy/semantic.rs
sanket1729: The PR already has such a function `num_contraints`.

src/policy/concrete.rs
```diff
@@ -76,6 +77,8 @@ pub enum PolicyError {
     InsufficientArgsforAnd,
     /// Semantic Policy Error: `And` `Or` fragments must take args: k > 1
     InsufficientArgsforOr,
+    /// Entailment max terminals allowed
```
apoelstra, line 80: `allowed` should be `exceeded`

src/policy/mod.rs
```diff
@@ -37,6 +37,8 @@ pub use self::concrete::Policy as Concrete;
 pub use self::semantic::Policy as Semantic;
 use MiniscriptKey;
 
+/// Policy entailment algorithm maximum number of terminals allowed
+const ENTAILMENT_MAX_TERMINALS: usize = 24;
```
apoelstra, line 41: Some quick benchmarks on map laptop show that this allows an entailment to take nearly 10 seconds in debug mode, 0.3s in release mode. I'd prefer this value be reduced to 20 so that everything stays under a second.

apoelstra (approved):

PR #119, opened 2020-08-05
Psbt: Add finalizer and extractor

### LINKED ISSUE #117: Fix Psbt finalizing for Legacy inputs
I think the witness for Legacy should go into ScriptSig component for PSBT. Would also require translation from witness into script. 

See: 
https://github.com/rust-bitcoin/rust-miniscript/blob/87a47e242dc19c517ca06c558cf79b43eec22abf/src/psbt/mod.rs#L171

sanket1729:
Implemented psbt finalizer and extractor. Fixes #117 . 

This PR also has some small refactors.

- Decouples satisfaction API for reusing most code in psbt input and Tx input
- Completed psbt finalizer. Miniscript primarily relies on three types of constraints: timelocks, hashlocks and sigs. 
1. `Timelocks`: There is no special witness required from `psbt`. The transaction input contains `sequence` and transaction has the corresponding `locktime`.
2. `Hashlocks`: Awaiting on https://github.com/bitcoin/bips/pull/955 and to be merged, psbt support in rust-bitcoin.
3. `Sigs`: Partial sigs contain the data required for the finalizer.

- Once the finalizer fills in the required fields, we double-check everything by running the interpreter and checking that there are no errors.
- Similarly, before extracting we double-check everything with the interpreter. 

The interpreter signature checks rely on sighash calculation from #116. Once the sighash functionality is added this PR can be updated to use it. Currently, this returns true. We can even merge this first and add the check in a followup PR once #116 is in.


src/miniscript/satisfy.rs
```diff
@@ -25,12 +25,25 @@ use bitcoin::hashes::{hash160, ripemd160, sha256, sha256d};
...
 /// Type alias for a signature/hashtype pair
 pub type BitcoinSig = (secp256k1::Signature, bitcoin::SigHashType);
 
+/// Helper function to create BitcoinSig from Rawsig
+/// Useful for downstream when implementing Satisfier.
+/// Returns underlying secp if the Signature is not of correct formart
+pub fn bitcoinsig_from_rawsig(rawsig: &[u8]) -> Result<BitcoinSig, Error> {
+    let (flag, sig) = rawsig.split_last().unwrap();
+    let flag = bitcoin::SigHashType::from_u32(*flag as u32);
+    let sig = match secp256k1::Signature::from_der(sig) {
```
apoelstra, line 41: Will
```
let sig = secp256k1::Signature::from_der(sig)?;
```
work?

src/psbt/mod.rs
```diff
@@ -15,41 +15,113 @@
...
-    WrongInputCount {
-        in_tx: usize,
-        in_map: usize,
-    },
+use bitcoin;
+
+#[derive(Debug)]
+pub enum InputError {
+    /// Public key parsing error
+    InvalidPubkey { pubkey: Vec<u8> },
```
apoelstra, line 39: We should retain the `secp256k1::Error` as well here

src/psbt/mod.rs
```diff
@@ -15,41 +15,113 @@
...
+    /// Redeem script does not match the p2sh hash
+    InvalidRedeemScript,
```
apoelstra, line 41: We should retain the expected and actual hashes here

src/psbt/mod.rs
```diff
@@ -15,41 +15,113 @@
...
+    /// Witness script does not match the p2wsh hash
+    InvalidWitnessScript,
```
apoelstra, line 43: We should retain the expected and actual hashes here

src/psbt/mod.rs
```diff
@@ -15,41 +15,113 @@
...
+                write!(f, "Redeem script does not match the p2sh script hash")
+            }
+            InputError::InvalidWitnessScript => {
+                write!(f, "Witness script does not match the p2wsh script hash")
+            }
+            InputError::MiniscriptError(ref e) => write!(f, "Miniscript Error: {}", e),
+            InputError::MissingWitness => write!(f, "PSBT is missing witness"),
+            InputError::MissingRedeemScript => write!(f, "PSBT is Redeem script"),
+            InputError::MissingUtxo => {
+                write!(f, "PSBT is missing both witness and non-witnes UTXO")
```
apoelstra, line 95: "witness" spelled wrong

src/miniscript/satisfy.rs
sanket1729: Yup, that should be cleaner. We already have a `From` implementation from the `Secp` error

src/miniscript/satisfy.rs
```diff
@@ -25,12 +25,25 @@ use bitcoin::hashes::{hash160, ripemd160, sha256, sha256d};
 use bitcoin::{self, secp256k1};
 use {MiniscriptKey, ToPublicKey};
 
+use Error;
 use ScriptContext;
 use Terminal;
 
 /// Type alias for a signature/hashtype pair
 pub type BitcoinSig = (secp256k1::Signature, bitcoin::SigHashType);
```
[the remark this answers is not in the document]
sanket1729: Conceptually yes. But I think along with that we would also want to add new APIs to use this type instead.

src/psbt/finalizer.rs
```diff
@@ -0,0 +1,369 @@
...
+        let vfyfn = |pk: &bitcoin::PublicKey, (sig, _)| secp.verify(&msg, &sig, &pk.key).is_ok();
+        let constraints_iter = SatisfiedConstraints::from_descriptor(&des, stack, vfyfn, cltv, csv);
+        let constraints: Result<Vec<_>, _> = constraints_iter.collect();
+        constraints
+            .map_err(|e| Error::InputError(InputError::MiniscriptError(MsError::from(e)), index))?;
+    }
+    Ok(())
+}
+
+pub fn finalize(psbt: &mut Psbt) -> Result<(), super::Error> {
```
sgeisler, line 269: Mutably borrowing the PSBT will leave it in a potentially invalid state in case of an error. I'd rather consume and return the PSBT.

sanket1729: We do check the validity of the psbt before returning. 

Somehow, it seems to be that allocating a new psbt when we are only adding two fields per input(final sciptsig and final witness) seems wrong.

dr-orlovsky: While working on PSBTs in rust-bitcoin, I am not getting why finalize shouldn't be part of `PartiallySignedTransaction` structure with access to `self`
https://github.com/rust-bitcoin/rust-bitcoin/pull/472/files#diff-77e0d8cfeb8d4f84f6f901f6fc14139fR337

sanket1729: I think the method finalize could belong there, but rust-miniscript offers a way to construct the final witness and final scriptsig in a generic way by the satisfier trait. 

We don't want to introduce a rust-miniscript dependancy to rust-bitcoin. 

Maybe one way might be to have the method at both places, the generic finalizer in rust-miniscript and one method in rust-bitcoin that only fills in the required fields.

sanket1729: The more I think about this, I think there definitely should be a finalize method in rust-bitcoin that works for all scripts. Miniscript is more of helper to the finalizer. The job of a finalizer is to fill in the final values for witness and script_sig even if they are not miniscripts. 

I think the API should be as follows:

rust-miniscript:
```
fn finalize_miniscript(&mut psbt) -> (Script, Vec<u8>) 
// Final script_sig and witness
// Only works when all scripts used are miniscripts
```

rust-bitcoin:
```
fn finalize(&mut self, witness: Vec<u8>, script_sig: Script)
// This fills in the values in psbt.
```

sgeisler: > `fn finalize(&mut self, witness: Vec<u8>, script_sig: Script)`

Do I understand it correctly that this finalize method would work on per-input basis? Otherwise the witness/script sig had to be matched to an input which would make the finalization at least `O(n^2)`. This also means that miniscript `finalize` had to return multiple `(Script, Vec<u8>)` tuples (one for each input).

sanket1729: Yup, my bad. Edited the original post. We can discuss the API even further. Maybe it should only work on a per input basis, but I wanted to share my opinion on how to the split between rust-miniscript's role and rust-bitcoin's role on finalizing.

src/psbt/finalizer.rs
```diff
@@ -0,0 +1,371 @@
...
+use bitcoin::{self, PublicKey, Script, SigHashType};
+use descriptor::{from_txin_with_witness_stack, SatisfiedConstraints};
+use Descriptor;
+use Legacy;
+use Miniscript;
+use Segwitv0;
+// Get the scriptpubkey for the psbt input
+fn get_scriptpubkey(psbt: &Psbt, index: usize) -> Result<&Script, InputError> {
+    let script_pubkey;
+    let inp = psbt.inputs.get(index).expect("Invalid Input index");
```
apoelstra, line 36: nit: here (and elsewhere) you can just write `let inp = &psbt.inputs[index];`

src/psbt/finalizer.rs
```diff
@@ -0,0 +1,371 @@
...
+    // 1. `PK`: creates a `Pk` descriptor(does not check if partial sig is given)
+    if script_pubkey.is_p2pk() {
+        let script_pubkey_len = script_pubkey.len();
+        let pk_bytes = &script_pubkey.to_bytes();
+        match bitcoin::PublicKey::from_slice(&pk_bytes[1..script_pubkey_len - 1]) {
+            Ok(pk) => return Ok(Descriptor::Pk(pk)),
+            Err(e) => return Err(InputError::from(e)),
+        }
+    } else if script_pubkey.is_p2pkh() {
+        // 1. `Pkh`: creates a `PkH` descriptor if partial_sigs has the correponding pk
```
apoelstra, line 89: `1.` sholud be `2.` here

src/psbt/finalizer.rs
```diff
@@ -0,0 +1,371 @@
...
+                *script_pubkey == addr.script_pubkey()
+            })
+            .next();
+        match partial_sig_contains_pk {
+            Some((pk, _sig)) => {
+                return Ok(Descriptor::Pkh(pk.to_owned()));
+            }
+            None => {
+                return Err(InputError::MissingPubkey);
+            }
```
apoelstra, line 106: more idiomatic (and readable) to move the `return` out of the match so it is clear there are no non-returning branches
```
return match partial_sig_contains_pk {
    Some((pk, _sig)) => Ok(Descriptor::Pkh(pk.to_owned())),
    None => Err(InputError::MissingPubkey),
};
```

apoelstra: also every single `return` in this function could be removed since they appear at the end of the block (you need to change a couple `if`s to `else if`s). This would also be clearer since it emphasizes that this giant if statement is the only thing going on.

src/psbt/finalizer.rs
```diff
@@ -0,0 +1,371 @@
...
+            let ms = Miniscript::<bitcoin::PublicKey, Segwitv0>::parse(witness_script)?;
+            return Ok(Descriptor::Wsh(ms));
+        } else {
+            return Err(InputError::MissingWitnessScript);
+        }
+    } else if script_pubkey.is_p2sh() {
+        if inp.redeem_script.is_none() {
+            return Err(InputError::MissingRedeemScript);
+        }
+        let redeem_script = inp.redeem_script.as_ref().unwrap();
```
apoelstra, line 150: This `is_none` followed by `unwrap` should be replaced with a `match` (which will add an indentation level but will eliminate a panic case)

src/psbt/finalizer.rs
```diff
@@ -0,0 +1,371 @@
...
+/// script is correctly interpreted according to the context
+/// The psbt must have included final script sig and final witness.
+/// In other words, this checks whether the finalized psbt interprets
+/// correctly
+pub fn interpreter_check(psbt: &Psbt) -> Result<(), Error> {
+    for (index, input) in psbt.inputs.iter().enumerate() {
+        let script_pubkey =
+            get_scriptpubkey(psbt, index).map_err(|e| Error::InputError(e, index))?;
+        let empty_script_sig = Script::new();
+        let empty_witness = Vec::new();
```
apoelstra, line 223: nit: i would drop both of these variables, they are only used once and both have longer names than the expressions they replace :P

src/psbt/finalizer.rs
sanket1729: Removing all returns as it is the last statement

sanket1729: Rather, should I keep the short circuit return paths? Maybe those are more clear

apoelstra: I only saw one that I'd consider a "short-circuit return path", which is the one I complained about below because it was immediately followed by an `.unwrap()`.

src/psbt/finalizer.rs
sanket1729: I think I need a let binding for the following?

`fn_name(&Script::new())`

apoelstra (approved):
c2b1fa7c52b355469e1d5ec56043ba07e76d7689

PR #120, opened 2020-08-05
More csv ctlv fixes

sanket1729:
Still discovering remains of #37.

### DIFF
--- a/src/descriptor/satisfied_constraints.rs
+++ b/src/descriptor/satisfied_constraints.rs
@@ -871,7 +871,7 @@ impl<'stack> Stack<'stack> {
     }
 
     /// Helper function to evaluate a After Node. Takes no argument from stack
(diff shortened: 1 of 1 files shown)


src/descriptor/satisfied_constraints.rs
```diff
@@ -870,8 +870,8 @@ impl<'stack> Stack<'stack> {
         }
     }
 
-    /// Helper function to evaluate a After Node. Takes no argument from stack
-    /// `n CHECKSEQUENCEVERIFY 0NOTEQUAL` and `n CHECKSEQUENCEVERIFY`
+    /// Helper function to evaluate a Older Node. Takes no argument from stack
```
apoelstra, line 873: I don't think you mean to change `After` to `Older` here; ditto in the other function's comment.

apoelstra (approved):

PR #121, opened 2020-08-06
Detect miniscripts that have a satisfaction with a combination of time locks and height locks

sanket1729:


src/miniscript/types/extra_props.rs
```diff
@@ -25,6 +27,13 @@ pub struct ExtData {
     pub ops_count_sat: Option<usize>,
     /// The worst case ops-count for dissatisfying this Miniscript fragment.
     pub ops_count_nsat: Option<usize>,
+    /// Whether any satisfaction of this fragment contains any timelocks
+    /// contains_timelock[0] = csv with heights
+    /// contains_timelock[1] = csv with times
+    /// contains_timelock[2] = cltv with heights
+    /// contains_timelock[3] = cltv with times
+    /// contains_timelock[4] = combination of any heightlocks and timelocks
```
apoelstra, line 35: Please create a structure for this instead of using a `[bool; 5]` :)

apoelstra: You can impl `Default` on it so you can construct it quickly

src/policy/mod.rs
```diff
@@ -162,6 +183,21 @@ mod tests {
...
+    #[test]
+    fn test_timelock_validity() {
+        // only height
+        assert!(ConcretePol::from_str("after(100)").is_ok());
+        // only time
+        assert!(ConcretePol::from_str("after(1000000000)").is_ok());
+        // disjunction
+        assert!(ConcretePol::from_str("or(after(1000000000),after(100))").is_ok());
+        // conjunction
+        assert!(ConcretePol::from_str("and(after(1000000000),after(100))").is_err());
```
apoelstra, line 195: Why should this fail?

sanket1729: A conjunction of height lock and timelock?

apoelstra: Oh, right, I forgot that timelocks are defined implicitly by having a value >500000.

I really don't like this, we shuold modify Miniscript to use a different syntax for them.

sanket1729: Any suggestions? For starters, I like the way shesek did it for minsc. https://min.sc/#time-durations

src/miniscript/satisfy.rs
```diff
@@ -94,7 +95,13 @@ pub struct Older(pub u32);
 
 impl<Pk: MiniscriptKey> Satisfier<Pk> for Older {
     fn check_older(&self, n: u32) -> bool {
-        n <= self.0
+        if !((n < HEIGHT_TIME_THRESHOLD && self.0 < HEIGHT_TIME_THRESHOLD)
```
apoelstra, line 98: It may be simpler to check either
```
(n < HEIGHT_TIME_THRESHOLD) == (self.0 < HEIGHT_TIME_THRESHOLD)
```
or just
```
n < HEIGHT_TIME_THRESHOLD && self.0 >= HEIGHT_TIME_THRESHOLD
```

src/miniscript/satisfy.rs
```diff
@@ -104,7 +111,13 @@ pub struct After(pub u32);
 
 impl<Pk: MiniscriptKey> Satisfier<Pk> for After {
     fn check_after(&self, n: u32) -> bool {
-        n <= self.0
+        if !((n < HEIGHT_TIME_THRESHOLD && self.0 < HEIGHT_TIME_THRESHOLD)
+            || (n >= HEIGHT_TIME_THRESHOLD && self.0 >= HEIGHT_TIME_THRESHOLD))
```
apoelstra, line 115: ditto here

src/policy/concrete.rs
```diff
@@ -115,6 +118,9 @@ impl fmt::Display for PolicyError {
                 "Policy entailment only supports {} terminals",
                 ENTAILMENT_MAX_TERMINALS
             ),
+            PolicyError::HeightTimeLockCombination => {
+                f.write_str("Cannot lift polcies that have a heightlock and timelock combination")
```
apoelstra, line 122: typo `policies`

src/miniscript/satisfy.rs
sanket1729: My logic for these is trying to implement _as it is_ written in the BIP.

apoelstra: Which BIP?

My suggestion is logically equivalent to what's written here, for what it's worth, it's just shorter.

sanket1729: https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki . I can change this too, just was a bit pedantic in trying to copy the logic conditions directly.

src/miniscript/types/extra_props.rs
apoelstra: To be clear, I can't review this PR any further until this is fixed :)

sanket1729: @apoelstra, This meaning the logic condition? I can change that quickly

src/policy/concrete.rs
```diff
@@ -168,9 +174,72 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+    }
+
+    // Checks whether the given concrete policy contains a combination of
+    // timelocks and heightlocks
+    // Returns a tuple of five elements:
+    // timelocks[0]: satisfaction contains csv heightlock
+    // timelocks[1]: satisfaction contains csv timelock
+    // timelocks[2]: satisfaction contains cltv heightlock
+    // timelocks[3]: satisfaction contains cltv timelock
+    // timelocks[4]: satisfaction contains an invalid combination
```
apoelstra, line 197: delete this comment

src/miniscript/types/extra_props.rs
```diff
@@ -8,6 +8,90 @@ use MiniscriptKey;
...
+    /// Whether the current contains any possible unspendable
+    /// path
+    pub fn contains_unspendable_path(self) -> bool {
+        self.contains_combination
+    }
+
+    // handy function for combining `and` timelocks
+    // This can be operator overloaded in future
+    pub(crate) fn comb_and_timelocks(a: Self, b: Self) -> Self {
+        Self::combine_thresh_timelocks(2, vec![a, b].into_iter())
```
apoelstra, line 51: Can you do `once(a).chain(once(b))` here rather than `Vec::into_iter`. (You need to `use std::iter::once`.) Saves the allocation.

src/miniscript/types/extra_props.rs
```diff
@@ -8,6 +8,90 @@ use MiniscriptKey;
...
+    // csv with heights
+    pub csv_with_height: bool,
+    /// csv with times
+    pub csv_with_time: bool,
+    /// cltv with heights
+    pub cltv_with_height: bool,
+    /// cltv with times
+    pub cltv_with_time: bool,
+    /// combination of any heightlocks and timelocks
+    pub contains_combination: bool,
```
apoelstra, line 26: Do we actually need this field? Or can we just check `(csv_with_height && csv_with_time) || (cltv_with_height && cltv_with_time)`?

apoelstra: Ah - it is there to cover the disjunction case, where you may have a heighlock branch and a timelock branch, but they are independent so it is ok.

src/miniscript/types/extra_props.rs
```diff
@@ -8,6 +8,90 @@ use MiniscriptKey;
...
+        // timelocks.csv_with_height and timelocks.csv_with_time can be directly
+        // combined together instead of iterating through all the pairs:
+        // consider the two cases:
+        // 1) Only one subfragment contains timelock and hieghtlock: In this case,
+        // timelocks.contains_combination would already by `true`. and other combinations would
+        // not add any effect.
+        // 2) hieghtlocks and timelocks are distributed across two branches:
+        // and we check if any of the existing branches has correponding lock
+        // that cannot be combined
+        // If k == 1, then this is no combination possible.
```
apoelstra, line 77: I find this whole comment block confusing. I think you should replace it with
```
// Propagate all fields of `TimelockInfo` from each of the node's children to the node
// itself (by taking the logical-or of all of them). In case `k == 1` (this is a disjunction)
// this is all we need to do: the node may behave like any of its children, for purposes
// of timelock accounting.
//
// If `k > 1` we have the additional consideration that if any two children have conflicting
// timelock requirements, this represents an inaccessible spending branch. If there are
// two such children we must set the `contains_combination` field.
```

src/miniscript/types/extra_props.rs
```diff
@@ -8,6 +8,90 @@ use MiniscriptKey;
...
+        for sub_timelock in sub_timelocks {
+            if k >= 2 {
+                timelock_info.contains_combination |= (timelock_info.csv_with_height
+                    && sub_timelock.csv_with_time)
+                    || (timelock_info.csv_with_time && sub_timelock.csv_with_height)
+                    || (timelock_info.cltv_with_time && sub_timelock.cltv_with_height)
+                    || (timelock_info.cltv_with_height && sub_timelock.cltv_with_time);
+            }
```
apoelstra, line 85: You can move this block out of the loop and simplify it:
```
if k >= 2 {
    timelock_info.contains_combination |= timelock_info.csv_with_time && sub_timelock.csv_with_height;
    timelock_info.contains_combination |= timelock_info.cltv_with_time && sub_timelock.cltv_with_height;
}
```
(or however you want to format it)

apoelstra: Once it is out of the loop you can replace the loop with `fold`, as
```
let mut ret = sub_timelocks.fold(TimelockInfo::Default, |acc, x| TimelockInfo {
    csv_with_time: acc.csv_with_time | x.csv_with_time,
    csv_with_height: acc.csv_with_height | x.csv_with_height,
    cltv_with_time: acc.cltv_with_time | x.cltv_with_time;
    cltv_with_height: acc.cltv_with_height | x.cltv_with_height;
    contains_combination: acc.contains_combination | acc.contains_combination;
});
if k > 1 {
    ret.contains_combination |= ret.csv_with_time && ret.csv_with_height;
    ret.contains_combination |= ret.cltv_with_time && ret.cltv_with_height;
}
```

sanket1729: This is incorrect. As it will set `contains_combination` for the following policy too. 
`thresh(2,pk(),pk(),or(time,height))`

sanket1729: Unforetunately, I think it is necessary to check k>2 inside of loop.

apoelstra: I think your existing code has the same issue.

apoelstra: Ah, no, I see -- the distinction is that only the one child has any timelocks at all. Interesting.

apoelstra: Ok, so I leave it up to you whether you want to bother using `fold` instead of a loop. But can you add a comment above the `k >= 2` line saying
```
// If more than one branch may be taken, and some other branch has a requirement
// that conflicts with this one, set `contains_combination`
```
or something

sanket1729: Changing to fold

apoelstra (approved):
ack a6621050a860f89175116ce2a819df9d5c42c9f7

PR #123, opened 2020-08-19
descriptor: add a method to get the `scriptCode` as per bip-0143

darosior:
To generate valid Segwit sighashes, the `witness_script()` could be used
for P2WSH outputs's `scriptCode`. However a P2PKH Script would have to be
reconstructed by-hand for P2WPKH.

This unifies the way to get the `scriptCode` for any Segwit output
defined using a rust-miniscript descriptor.

Note that we don't bother checking for CODESEPARATOR as we don't produce
Scripts containing such opcode.

I did not add a test yet, but wanted your input on how to test this: could some tests be modified to use real signatures and can i introduce libbitcoinconsensus as a dev-dependency in order to check the scripts against it ? This would indirectly check that the `scriptCode` is correct.

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -392,6 +392,32 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             }
         }
     }
+
+    /// Get the `scriptCode` as [defined by
+    /// bip-0143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification),
+    /// used to generate sighashes for spending Segwit outputs.
+    ///
+    /// # Errors
+    /// - If the script descriptor type is non-segwit, as the script code is only defined for
+    /// segregated witness transaction outputs.
+    pub fn script_code(&self) -> Result<Script, Error> {
+        match *self {
+            // Spending non-segwit outputs requires "legacy" sighash and don't use bip143's `scriptCode`.
+            Descriptor::Bare(..)
+            | Descriptor::Pk(..)
+            | Descriptor::Pkh(..)
+            | Descriptor::Sh(..) => Err(Error::BadDescriptor),
+            // The item 5:
+            //     - For P2WPKH witness program, the scriptCode is `0x1976a914{20-byte-pubkey-hash}88ac`.
+            Descriptor::Wpkh(ref pk) | Descriptor::ShWpkh(ref pk) => {
+                let addr = bitcoin::Address::p2pkh(&pk.to_public_key(), bitcoin::Network::Bitcoin);
+                Ok(addr.script_pubkey())
+            }
+            //     - For P2WSH witness program, if the witnessScript does not contain any `OP_CODESEPARATOR`,
+            //       the `scriptCode` is the `witnessScript` serialized as scripts inside CTxOut.
+            Descriptor::Wsh(ref d) | Descriptor::ShWsh(ref d) => Ok(d.encode()),
+        }
+    }
 }
 
 impl<Pk> expression::FromTree for Descriptor<Pk>
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -1015,4 +1015,57 @@ mod tests {
         assert_eq!(sig1, sig_a);
         assert_eq!(sig0, sig_b);
     }
+
+    #[test]
+    fn test_scriptcode() {
+        // P2WPKH (from bip143 test vectors)
+        let descriptor = Descriptor::<PublicKey>::from_str(
+            "wpkh(025476c2e83188368da1ff3e292e7acafcdb3566bb0ad253f62fc70f07aeee6357)",
+        )
+        .unwrap();
+        assert_eq!(
+            *descriptor
+                .script_code()
+                .expect("script_code() on P2WPKH descriptor")
+                .as_bytes(),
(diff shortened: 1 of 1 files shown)


sanket1729:
I don't see any downsides of having this. The sighash method that will probably be implemented as a part #116 can use this directly. 
As for testing this, we can add tests similar to ones for `script_pubkey()` .

> could some tests be modified to use real signatures

Modifications to current tests with real signatures are welcome. 

> Can I introduce libbitcoinconsensus as a dev-dependency in order to check the scripts against it ?

Can you elaborate more on what do you mean by script checking? Do you mean creating a complete transaction and using libbitcoinconsensus to check it's validity? 
I think we should probably stick to rust-ecosystem(rust-bitcoin/rust-secp). But adding cross-checking against libbitcoinconsensus might not be a bad idea. cc @apoelstra.

darosior:
> I don't see any downsides of having this. The sighash method that will probably be implemented as a part #116 can use this directly.

Didn't read #116 in detail yet, will check how it was implemented.

> As for testing this, we can add tests similar to ones for script_pubkey() .

I finally went for the bip143 sighash_all vectors and some handmade tests with bitcoind rpc. Will push a commit.

> Can you elaborate more on what do you mean by script checking?

Checking that the input witness satisfies the witness program of the spent txo's scriptPubKey.

> Do you mean creating a complete transaction and using libbitcoinconsensus to check it's validity?

Yeah, that's what i meant. But finally it appeared to be more on-scope to directly check the scriptCode.

However i still wonder if it could make sense to have this check more generally. The drawback would be a slightly higher execution time for the tests.

> I think we should probably stick to rust-ecosystem(rust-bitcoin/rust-secp).

There are rust binding to libbitcoinconsensus in the rust-bitcoin organization.

—

You are receiving this because you authored the thread.

Reply to this email directly, view it on GitHub, or unsubscribe.￼

[ { "@context": "http://schema.org", "@type": "EmailMessage", "potentialAction": { "@type": "ViewAction", "target": "[https://github.com](<a href=)/rust-bitcoin/rust-miniscript/pull/123#issuecomment-676802642">https://github.com/rust-bitcoin/rust-miniscript/pull/123#issuecomment-676802642", "url": "[https://github.com](<a href=)/rust-bitcoin/rust-miniscript/pull/123#issuecomment-676802642">https://github.com/rust-bitcoin/rust-miniscript/pull/123#issuecomment-676802642", "name": "View Pull Request" }, "description": "View this Pull Request on GitHub", "publisher": { "@type": "Organization", "name": "GitHub", "url": "https://github.com" } } ]

sanket1729:
Sorry for the delay in responding. Somehow I thought that I commented on this post, but turns out I didn't. I will create a separate issue for consensus testing to library created miniscripts. I think we should have this for sure. 

tAck 283676ee431079a1334e83202edb91d1e885557b

src/descriptor/mod.rs
```diff
@@ -392,6 +392,32 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
...
+    /// Get the `scriptCode` as [defined by
+    /// bip-0143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification),
+    /// used to generate sighashes for spending Segwit outputs.
+    ///
+    /// # Errors
+    /// - If the script descriptor type is non-segwit, as the script code is only defined for
+    /// segregated witness transaction outputs.
+    pub fn script_code(&self) -> Result<Script, Error> {
+        match *self {
+            // Spending non-segwit outputs requires "legacy" sighash and don't use bip143's `scriptCode`.
```
apoelstra, line 405: Legacy outputs still have a `scriptCode`

src/descriptor/mod.rs
```diff
@@ -392,6 +392,32 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
...
+            Descriptor::Bare(..)
+            | Descriptor::Pk(..)
+            | Descriptor::Pkh(..)
+            | Descriptor::Sh(..) => Err(Error::BadDescriptor),
+            // The item 5:
+            //     - For P2WPKH witness program, the scriptCode is `0x1976a914{20-byte-pubkey-hash}88ac`.
+            Descriptor::Wpkh(ref pk) | Descriptor::ShWpkh(ref pk) => {
+                let addr = bitcoin::Address::p2pkh(&pk.to_public_key(), bitcoin::Network::Bitcoin);
+                Ok(addr.script_pubkey())
```
apoelstra, line 414: why can't we just call `self.script_pubkey()`?

apoelstra:
I'm really confused about the motivation of this PR. It appears to be identical to `.witness_script()` except that it sometimes returns errors?

src/descriptor/mod.rs
darosior: Because it returns `OP_0 <key hash>` while bip143 specifies it as `DUP HASH160 <key hash> EQUALVERIFY CHECKSIG` https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#specification

apoelstra: Ah, you're right.

apoelstra:
Talked to @darosior on IRC, who points out that BIP143 defines the scriptCode for Wpkh and ShWpkh to resemble a pay-to-pkh scriptPubKey, and currently there is no function on `Descriptor` which returns this object.

PR #130, opened 2020-09-11
update rust-bitcoin to 0.24.0

apoelstra:

PR #131, opened 2020-09-11
DescriptorPublicKey, second pass

darosior:
This is a rebased, cleaned up and amended version of #93.

It's based on #130.

<details>

<summary>Here is the diff with a [rebased-on-master #93 ](https://github.com/darosior/rust-miniscript/commits/sgeisler_2020-06-descriptor-key_rebased) at commit cdfab02acfe13aeb3c2743b95208f255da4856a0</summary>

<p>

```patch
diff --git a/.github/workflows/rust.yml b/.github/workflows/rust.yml
index dac83c0..3d34f23 100644
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -58,8 +58,8 @@ jobs:
           - 1.22.0
           - beta
           - stable
-    steps:      
-      - name: Checkout Crate 
+    steps:
+      - name: Checkout Crate
         uses: actions/checkout@v2
       - name: Checkout Toolchain
         uses: actions-rs/toolchain@v1
@@ -67,6 +67,8 @@ jobs:
           profile: minimal
           toolchain: ${{ matrix.rust }}
           override: true
+      - name: Pin cc if rust 1.22
+        if: matrix.rust == '1.22.0'
+        run: cargo generate-lockfile --verbose && cargo update -p cc --precise "1.0.41" --verbose
       - name: Running cargo
         run: ./contrib/test.sh
-          
diff --git a/Cargo.toml b/Cargo.toml
index 429d336..d4397e6 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ unstable = []
 default = []
 
 [dependencies]
-bitcoin = {git = "https://github.com/sgeisler/rust-bitcoin/", branch = "2020-06-bip32-derive-more"}
+bitcoin = "0.24"
 
 [dependencies.serde]
 version = "1.0"
diff --git a/src/descriptor/create_descriptor.rs b/src/descriptor/create_descriptor.rs
index b698dbc..d36a7cb 100644
--- a/src/descriptor/create_descriptor.rs
+++ b/src/descriptor/create_descriptor.rs
@@ -24,8 +24,7 @@ use ToPublicKey;
 /// NOTE: Miniscript pushes should only be either boolean, 1 or 0, signatures, and hash preimages.
 /// As per the current implementation, PUSH_NUM2 results in an error
 fn instr_to_stackelem<'txin>(
-    ins: &Result<Instruction<'txin>,
-    bitcoin::blockdata::script::Error>,
+    ins: &Result<Instruction<'txin>, bitcoin::blockdata::script::Error>,
 ) -> Result<StackElement<'txin>, Error> {
     match *ins {
         //Also covers the dissatisfied case as PushBytes0
@@ -322,8 +321,9 @@ mod tests {
         assert_eq!(stack, stack![Push(&sigs[0])]);
 
         //test wpkh
-        let script_pubkey =
-            bitcoin::Address::p2wpkh(&pks[1], bitcoin::Network::Bitcoin).unwrap().script_pubkey();
+        let script_pubkey = bitcoin::Address::p2wpkh(&pks[1], bitcoin::Network::Bitcoin)
+            .unwrap()
+            .script_pubkey();
         let script_sig = script::Builder::new().into_script();
         let witness = vec![sigs[1].clone(), pks[1].clone().to_bytes()];
         let (des, stack) = from_txin_with_witness_stack(&script_pubkey, &script_sig, &witness)
@@ -385,10 +385,12 @@ mod tests {
         assert_eq!(stack, stack![Push(&sigs[1]), Push(&sigs[3])]);
 
         //test shwpkh
-        let script_pubkey =
-            bitcoin::Address::p2shwpkh(&pks[2], bitcoin::Network::Bitcoin).unwrap().script_pubkey();
-        let redeem_script =
-            bitcoin::Address::p2wpkh(&pks[2], bitcoin::Network::Bitcoin).unwrap().script_pubkey();
+        let script_pubkey = bitcoin::Address::p2shwpkh(&pks[2], bitcoin::Network::Bitcoin)
+            .unwrap()
+            .script_pubkey();
+        let redeem_script = bitcoin::Address::p2wpkh(&pks[2], bitcoin::Network::Bitcoin)
+            .unwrap()
+            .script_pubkey();
         let script_sig = script::Builder::new()
             .push_slice(&redeem_script.to_bytes())
             .into_script();
diff --git a/src/descriptor/mod.rs b/src/descriptor/mod.rs
index 6e857a0..6b251bf 100644
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -23,16 +23,19 @@
 //! these with BIP32 paths, pay-to-contract instructions, etc.
 //!
 
-use bitcoin::blockdata::{opcodes, script};
-use bitcoin::{self, PublicKey, Script};
-#[cfg(feature = "serde")]
-use serde::{de, ser};
 use std::fmt;
 use std::str::{self, FromStr};
 
 use bitcoin::blockdata::{opcodes, script};
+use bitcoin::hashes::hash160;
+use bitcoin::hashes::hex::FromHex;
+use bitcoin::secp256k1;
+use bitcoin::util::bip32;
 use bitcoin::{self, Script};
 
+#[cfg(feature = "serde")]
+use serde::{de, ser};
+
 use expression;
 use miniscript;
 use miniscript::context::ScriptContextError;
@@ -50,12 +53,6 @@ pub use self::satisfied_constraints::Error as InterpreterError;
 pub use self::satisfied_constraints::SatisfiedConstraint;
 pub use self::satisfied_constraints::SatisfiedConstraints;
 pub use self::satisfied_constraints::Stack;
-use bitcoin::hashes::core::fmt::Formatter;
-use bitcoin::hashes::hash160;
-use bitcoin::hashes::hex::FromHex;
-use bitcoin::secp256k1::Secp256k1;
-use bitcoin::util::bip32::{ChildNumber, DerivationPath, Error as Bip32Error, ExtendedPubKey};
-use std::fmt::{Display, Write};
 
 /// Script descriptor
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
@@ -79,35 +76,44 @@ pub enum Descriptor<Pk: MiniscriptKey> {
 }
 
 #[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
-pub enum DescriptorKey {
-    PukKey(bitcoin::PublicKey),
+pub enum DescriptorPublicKey {
+    ShortPub(DescriptorShortPub),
     XPub(DescriptorXPub),
 }
 
+pub struct DescriptorShortPub {
+    origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+    key: bitcoin::PublicKey,
+}
+
 #[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
 pub struct DescriptorXPub {
-    source: Option<([u8; 4], DerivationPath)>,
-    xpub: bitcoin::util::bip32::ExtendedPubKey,
-    derivation_path: DerivationPath,
+    origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+    xpub: bip32::ExtendedPubKey,
+    derivation_path: bip32::DerivationPath,
     is_wildcard: bool,
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct DescriptorKeyParseError(&'static str);
 
-impl Display for DescriptorKey {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        match self {
-            DescriptorKey::PukKey(pk) => pk.fmt(f),
-            DescriptorKey::XPub(xpub) => {
-                if let Some((master_id, ref master_deriv)) = &xpub.source {
-                    f.write_char('[')?;
-                    for byte in master_id {
-                        write!(f, "{:02x}", byte)?;
-                    }
-                    fmt_derivation_path(f, master_deriv)?;
-                    f.write_char(']')?;
-                }
+impl fmt::Display for DescriptorKeyParseError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str(self.0)
+    }
+}
+
+impl fmt::Display for DescriptorPublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            DescriptorPublicKey::ShortPub(ref pk) => {
+                maybe_fmt_master_id(f, &pk.origin)?;
+                pk.key.fmt(f)?;
+                Ok(())
+            }
+            DescriptorPublicKey::XPub(ref xpub) => {
+                maybe_fmt_master_id(f, &xpub.origin)?;
                 xpub.xpub.fmt(f)?;
                 fmt_derivation_path(f, &xpub.derivation_path)?;
                 if xpub.is_wildcard {
@@ -119,29 +125,52 @@ impl Display for DescriptorKey {
     }
 }
 
-fn fmt_derivation_path(f: &mut Formatter<'_>, path: &DerivationPath) -> std::fmt::Result {
+/// Writes the fingerprint of the origin, if there is one.
+fn maybe_fmt_master_id(
+    f: &mut fmt::Formatter,
+    origin: &Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+) -> fmt::Result {
+    if let Some((ref master_id, ref master_deriv)) = *origin {
+        fmt::Formatter::write_str(f, "[")?;
+        for byte in master_id.into_bytes().iter() {
+            write!(f, "{:02x}", byte)?;
+        }
+        fmt_derivation_path(f, master_deriv)?;
+        fmt::Formatter::write_str(f, "]")?;
+    }
+
+    Ok(())
+}
+
+/// Writes a derivation path to the formatter, no leading 'm'
+fn fmt_derivation_path(f: &mut fmt::Formatter, path: &bip32::DerivationPath) -> fmt::Result {
     for child in path {
         write!(f, "/{}", child)?;
     }
     Ok(())
 }
 
-impl FromStr for DescriptorKey {
+impl FromStr for DescriptorPublicKey {
     type Err = DescriptorKeyParseError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // A "raw" public key without any origin is the least we accept.
         if s.len() < 66 {
-            Err(DescriptorKeyParseError(
+            return Err(DescriptorKeyParseError(
                 "Key too short (<66 char), doesn't match any format",
-            ))
-        } else if s.chars().next().unwrap() == '[' {
+            ));
+        }
+
+        // They may specify an origin
+        let mut origin = None;
+        if s.chars().next().unwrap() == '[' {
             let mut parts = s[1..].split(']');
-            let mut origin = parts
+            let mut raw_origin = parts
                 .next()
                 .ok_or(DescriptorKeyParseError("Unclosed '['"))?
                 .split('/');
 
-            let origin_id_hex = origin.next().ok_or(DescriptorKeyParseError(
+            let origin_id_hex = raw_origin.next().ok_or(DescriptorKeyParseError(
                 "No master fingerprint found after '['",
             ))?;
 
@@ -150,55 +179,59 @@ impl FromStr for DescriptorKey {
                     "Master fingerprint should be 8 characters long",
                 ));
             }
-
-            let origin_id: [u8; 4] = FromHex::from_hex(origin_id_hex).map_err(|_| {
+            let origin_raw_id: [u8; 4] = FromHex::from_hex(origin_id_hex).map_err(|_| {
                 DescriptorKeyParseError("Malformed master fingerprint, expected 8 hex chars")
             })?;
+            let parent_fingerprint = bip32::Fingerprint::from(origin_raw_id.as_ref());
 
-            let origin_path = origin
-                .map(|p| ChildNumber::from_str(p))
-                .collect::<Result<DerivationPath, Bip32Error>>()
+            let origin_path = raw_origin
+                .map(|p| bip32::ChildNumber::from_str(p))
+                .collect::<Result<bip32::DerivationPath, bip32::Error>>()
                 .map_err(|_| {
                     DescriptorKeyParseError("Error while parsing master derivation path")
                 })?;
+            origin = Some((parent_fingerprint, origin_path));
+        }
 
-            let key_deriv = parts.next().ok_or(DescriptorKeyParseError(
-                "No key found after origin description",
-            ))?;
+        let key_part = if origin == None {
+            Ok(s)
+        } else {
+            s.split(']')
+                .collect::<Vec<&str>>()
+                .pop()
+                .ok_or(DescriptorKeyParseError("No key after origin."))
+        }?;
 
-            let (xpub, derivation_path, is_wildcard) = Self::parse_xpub_deriv(key_deriv)?;
+        if key_part.starts_with("xpub") {
+            let (xpub, derivation_path, is_wildcard) = Self::parse_xpub_deriv(key_part)?;
 
-            Ok(DescriptorKey::XPub(DescriptorXPub {
-                source: Some((origin_id, origin_path)),
+            Ok(DescriptorPublicKey::XPub(DescriptorXPub {
+                origin,
                 xpub,
                 derivation_path,
                 is_wildcard,
             }))
-        } else if s.starts_with("02") || s.starts_with("03") || s.starts_with("04") {
-            let pk = PublicKey::from_str(s)
-                .map_err(|_| DescriptorKeyParseError("Error while parsing simple public key"))?;
-            Ok(DescriptorKey::PukKey(pk))
         } else {
-            let (xpub, derivation_path, is_wildcard) = Self::parse_xpub_deriv(s)?;
-            Ok(DescriptorKey::XPub(DescriptorXPub {
-                source: None,
-                xpub,
-                derivation_path,
-                is_wildcard,
+            let key = bitcoin::PublicKey::from_str(key_part)
+                .map_err(|_| DescriptorKeyParseError("Error while parsing simple public key"))?;
+            Ok(DescriptorPublicKey::ShortPub(DescriptorShortPub {
+                key,
+                origin,
             }))
         }
     }
 }
 
-impl DescriptorKey {
+impl DescriptorPublicKey {
+    /// Parse an extended public key concatenated to a derivation path.
     fn parse_xpub_deriv(
         key_deriv: &str,
-    ) -> Result<(ExtendedPubKey, DerivationPath, bool), DescriptorKeyParseError> {
+    ) -> Result<(bip32::ExtendedPubKey, bip32::DerivationPath, bool), DescriptorKeyParseError> {
         let mut key_deriv = key_deriv.split('/');
         let xpub_str = key_deriv.next().ok_or(DescriptorKeyParseError(
             "No key found after origin description",
         ))?;
-        let xpub = ExtendedPubKey::from_str(xpub_str)
+        let xpub = bip32::ExtendedPubKey::from_str(xpub_str)
             .map_err(|_| DescriptorKeyParseError("Error while parsing xpub."))?;
 
         let mut is_wildcard = false;
@@ -212,12 +245,12 @@ impl DescriptorKey {
                         "'*' may only appear as last element in a derivation path.",
                     )))
                 } else {
-                    Some(ChildNumber::from_str(p).map_err(|_| {
+                    Some(bip32::ChildNumber::from_str(p).map_err(|_| {
                         DescriptorKeyParseError("Error while parsing key derivation path")
                     }))
                 }
             })
-            .collect::<Result<DerivationPath, _>>()?;
+            .collect::<Result<bip32::DerivationPath, _>>()?;
 
         if (&derivation_path).into_iter().all(|c| c.is_normal()) {
             Ok((xpub, derivation_path, is_wildcard))
@@ -232,15 +265,15 @@ impl DescriptorKey {
     /// self.
     ///
     /// Panics if derivation path contains a hardened child number
-    pub fn derive(&self, path: &[ChildNumber]) -> DescriptorKey {
-        assert!(path.into_iter().all(|c| c.is_normal()));
+    pub fn derive(&self, path: &[bip32::ChildNumber]) -> DescriptorPublicKey {
+        debug_assert!(path.into_iter().all(|c| c.is_normal()));
 
-        match self {
-            DescriptorKey::PukKey(pk) => DescriptorKey::PukKey(*pk),
-            DescriptorKey::XPub(xpub) => {
+        match *self {
+            DescriptorPublicKey::ShortPub(ref pk) => DescriptorPublicKey::ShortPub(pk.clone()),
+            DescriptorPublicKey::XPub(ref xpub) => {
                 if xpub.is_wildcard {
-                    DescriptorKey::XPub(DescriptorXPub {
-                        source: xpub.source.clone(),
+                    DescriptorPublicKey::XPub(DescriptorXPub {
+                        origin: xpub.origin.clone(),
                         xpub: xpub.xpub.clone(),
                         derivation_path: (&xpub.derivation_path)
                             .into_iter()
@@ -257,17 +290,17 @@ impl DescriptorKey {
     }
 }
 
-impl MiniscriptKey for DescriptorKey {
+impl MiniscriptKey for DescriptorPublicKey {
     type Hash = hash160::Hash;
 
     fn to_pubkeyhash(&self) -> Self::Hash {
-        match self {
-            DescriptorKey::PukKey(pk) => pk.to_pubkeyhash(),
-            DescriptorKey::XPub(xpub) => {
-                let ctx = Secp256k1::verification_only();
+        match *self {
+            DescriptorPublicKey::ShortPub(ref spub) => spub.key.to_pubkeyhash(),
+            DescriptorPublicKey::XPub(ref xpub) => {
+                let ctx = secp256k1::Secp256k1::verification_only();
                 xpub.xpub
                     .derive_pub(&ctx, &xpub.derivation_path)
-                    .expect("Shouldn't fail, only normal derivations")
+                    .expect("Won't fail, only normal derivations")
                     .public_key
                     .to_pubkeyhash()
             }
@@ -275,12 +308,12 @@ impl MiniscriptKey for DescriptorKey {
     }
 }
 
-impl ToPublicKey for DescriptorKey {
-    fn to_public_key(&self) -> PublicKey {
-        match self {
-            DescriptorKey::PukKey(pk) => *pk,
-            DescriptorKey::XPub(xpub) => {
-                let ctx = Secp256k1::verification_only();
+impl ToPublicKey for DescriptorPublicKey {
+        match *self {
+            DescriptorPublicKey::ShortPub(ref spub) => spub.key.to_public_key(),
+            DescriptorPublicKey::XPub(ref xpub) => {
+                let ctx = secp256k1::Secp256k1::verification_only();
                 xpub.xpub
                     .derive_pub(&ctx, &xpub.derivation_path)
                     .expect("Shouldn't fail, only normal derivations")
@@ -294,12 +327,6 @@ impl ToPublicKey for DescriptorKey {
     }
 }
 
-impl Display for DescriptorKeyParseError {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        f.write_str(self.0)
-    }
-}
-
 impl<Pk: MiniscriptKey> Descriptor<Pk> {
     /// Convert a descriptor using abstract keys to one using specific keys
     /// This will panic if translatefpk returns an uncompressed key when
@@ -659,11 +686,14 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
     }
 }
 
-impl Descriptor<DescriptorKey> {
+impl Descriptor<DescriptorPublicKey> {
     /// Derives all wildcard keys in the descriptor using the supplied `path`
-    pub fn derive(&self, path: &[ChildNumber]) -> Descriptor<DescriptorKey> {
-        self.translate_pk(|pk| Result::<_, ()>::Ok(pk.derive(path)), |pkh| Ok(*pkh))
-            .expect("Translation fn can't fail.")
+    pub fn derive(&self, path: &[bip32::ChildNumber]) -> Descriptor<DescriptorPublicKey> {
+        self.translate_pk(
+            |pk| Result::Ok::<DescriptorPublicKey, ()>(pk.derive(path)),
+            |pkh| Result::Ok::<hash160::Hash, ()>(*pkh),
+        )
+        .expect("Translation fn can't fail.")
     }
 }
 
@@ -793,15 +823,18 @@ serde_string_impl_pk!(Descriptor, "a script descriptor");
 
 #[cfg(test)]
 mod tests {
+    use super::DescriptorKeyParseError;
+
     use bitcoin::blockdata::opcodes::all::{OP_CLTV, OP_CSV};
     use bitcoin::blockdata::script::Instruction;
     use bitcoin::blockdata::{opcodes, script};
     use bitcoin::hashes::{hash160, sha256};
-    use bitcoin::util::bip32::{ChildNumber, DerivationPath, ExtendedPubKey};
+    use bitcoin::util::bip32;
     use bitcoin::{self, secp256k1, PublicKey};
-    use descriptor::{DescriptorKey, DescriptorXPub};
+    use descriptor::{DescriptorPublicKey, DescriptorShortPub, DescriptorXPub};
     use miniscript::satisfy::BitcoinSig;
+    use policy;
     use std::collections::HashMap;
     use std::str::FromStr;
     use {Descriptor, DummyKey, Miniscript, Satisfier};
@@ -1320,53 +1353,137 @@ mod tests {
 
     #[test]
     fn parse_descriptor_key() {
-        let key = "[d34db33f/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*";
-        let expected = DescriptorKey::XPub(DescriptorXPub {
-            source: Some((
-                [0xd3, 0x4d, 0xb3, 0x3f],
+        // With a wildcard
+        let key = "[78412e3a/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*";
+        let expected = DescriptorPublicKey::XPub(DescriptorXPub {
+            origin: Some((
+                bip32::Fingerprint::from(&[0x78, 0x41, 0x2e, 0x3a][..]),
                 (&[
-                    ChildNumber::from_hardened_idx(44).unwrap(),
-                    ChildNumber::from_hardened_idx(0).unwrap(),
-                    ChildNumber::from_hardened_idx(0).unwrap(),
+                    bip32::ChildNumber::from_hardened_idx(44).unwrap(),
+                    bip32::ChildNumber::from_hardened_idx(0).unwrap(),
+                    bip32::ChildNumber::from_hardened_idx(0).unwrap(),
                 ][..])
                 .into(),
             )),
-            xpub: ExtendedPubKey::from_str("xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL").unwrap(),
-            derivation_path: (&[ChildNumber::from_normal_idx(1).unwrap()][..]).into(),
+            xpub: bip32::ExtendedPubKey::from_str("xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL").unwrap(),
+            derivation_path: (&[bip32::ChildNumber::from_normal_idx(1).unwrap()][..]).into(),
             is_wildcard: true,
         });
         assert_eq!(expected, key.parse().unwrap());
         assert_eq!(format!("{}", expected), key);
 
+        // Without origin
         let key = "xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1";
-        let expected = DescriptorKey::XPub(DescriptorXPub {
-            source: None,
-            xpub: ExtendedPubKey::from_str("xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL").unwrap(),
-            derivation_path: (&[ChildNumber::from_normal_idx(1).unwrap()][..]).into(),
+        let expected = DescriptorPublicKey::XPub(DescriptorXPub {
+            origin: None,
+            xpub: bip32::ExtendedPubKey::from_str("xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL").unwrap(),
+            derivation_path: (&[bip32::ChildNumber::from_normal_idx(1).unwrap()][..]).into(),
             is_wildcard: false,
         });
         assert_eq!(expected, key.parse().unwrap());
         assert_eq!(format!("{}", expected), key);
 
+        // Without derivation path
         let key = "xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL";
-        let expected = DescriptorKey::XPub(DescriptorXPub {
-            source: None,
-            xpub: ExtendedPubKey::from_str("xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL").unwrap(),
-            derivation_path: DerivationPath::from(&[][..]),
+        let expected = DescriptorPublicKey::XPub(DescriptorXPub {
+            origin: None,
+            xpub: bip32::ExtendedPubKey::from_str("xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL").unwrap(),
+            derivation_path: bip32::DerivationPath::from(&[][..]),
             is_wildcard: false,
         });
         assert_eq!(expected, key.parse().unwrap());
         assert_eq!(format!("{}", expected), key);
 
+        // Raw (compressed) pubkey
         let key = "03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8";
-        let expected = DescriptorKey::PukKey(
-            bitcoin::PublicKey::from_str(
+        let expected = DescriptorPublicKey::ShortPub(DescriptorShortPub {
+            key: bitcoin::PublicKey::from_str(
                 "03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8",
             )
             .unwrap(),
-        );
+            origin: None,
+        });
         assert_eq!(expected, key.parse().unwrap());
         assert_eq!(format!("{}", expected), key);
+
+        // Raw (uncompressed) pubkey
+        let key = "04f5eeb2b10c944c6b9fbcfff94c35bdeecd93df977882babc7f3a2cf7f5c81d3b09a68db7f0e04f21de5d4230e75e6dbe7ad16eefe0d4325a62067dc6f369446a";
+        let expected = DescriptorPublicKey::ShortPub(DescriptorShortPub {
+            key: bitcoin::PublicKey::from_str(
+                "04f5eeb2b10c944c6b9fbcfff94c35bdeecd93df977882babc7f3a2cf7f5c81d3b09a68db7f0e04f21de5d4230e75e6dbe7ad16eefe0d4325a62067dc6f369446a",
+            )
+            .unwrap(),
+            origin: None,
+        });
+        assert_eq!(expected, key.parse().unwrap());
+        assert_eq!(format!("{}", expected), key);
+
+        // Raw pubkey with origin
+        let desc =
+            "[78412e3a/0'/42/0']0231c7d3fc85c148717848033ce276ae2b464a4e2c367ed33886cc428b8af48ff8";
+        let expected = DescriptorPublicKey::ShortPub(DescriptorShortPub {
+            key: bitcoin::PublicKey::from_str(
+                "0231c7d3fc85c148717848033ce276ae2b464a4e2c367ed33886cc428b8af48ff8",
+            )
+            .unwrap(),
+            origin: Some((
+                bip32::Fingerprint::from(&[0x78, 0x41, 0x2e, 0x3a][..]),
+                (&[
+                    bip32::ChildNumber::from_hardened_idx(0).unwrap(),
+                    bip32::ChildNumber::from_normal_idx(42).unwrap(),
+                    bip32::ChildNumber::from_hardened_idx(0).unwrap(),
+                ][..])
+                    .into(),
+            )),
+        });
+        assert_eq!(expected, desc.parse().expect("Parsing desc"));
+        assert_eq!(format!("{}", expected), desc);
+    }
+
+    #[test]
+    fn parse_descriptor_key_errors() {
+        // We refuse creating descriptors which claim to be able to derive hardened childs
+        let desc = "[78412e3a/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/42'/*";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "Hardened derivation is currently not supported."
+            ))
+        );
+
+        // And ones with misplaced wildcard
+        let desc = "[78412e3a/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*/44";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "\'*\' may only appear as last element in a derivation path."
+            ))
+        );
+
+        // And ones with invalid fingerprints
+        let desc = "[NonHexor]xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "Malformed master fingerprint, expected 8 hex chars"
+            ))
+        );
+
+        // And ones with invalid xpubs..
+        let desc = "[78412e3a]xpub1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaLcgJvLJuZZvRcEL/1/*";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError("Error while parsing xpub."))
+        );
+
+        // ..or invalid raw keys
+        let desc = "[78412e3a]0208a117f3897c3a13c9384b8695eed98dc31bc2500feb19a1af424cd47a5d83/1/*";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "Error while parsing simple public key"
+            ))
+        );
     }
 
     #[test]
@@ -1376,16 +1493,16 @@ mod tests {
 pk([d34db33f/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*),\
 pk(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1),\
 pk(03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8))";
-        let policy: crate::policy::concrete::Policy<DescriptorKey> =
-            descriptor_str.parse().unwrap();
+        let policy: policy::concrete::Policy<DescriptorPublicKey> = descriptor_str.parse().unwrap();
         let descriptor = Descriptor::Sh(policy.compile().unwrap());
-        let derived_descriptor = descriptor.derive(&[ChildNumber::from_normal_idx(42).unwrap()]);
+        let derived_descriptor =
+            descriptor.derive(&[bip32::ChildNumber::from_normal_idx(42).unwrap()]);
 
         let res_descriptor_str = "thresh(2,\
 pk([d34db33f/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/42),\
 pk(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1),\
 pk(03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8))";
-        let res_policy: crate::policy::concrete::Policy<DescriptorKey> =
+        let res_policy: policy::concrete::Policy<DescriptorPublicKey> =
             res_descriptor_str.parse().unwrap();
         let res_descriptor = Descriptor::Sh(res_policy.compile().unwrap());
```

</p>

</details>

@sgeisler i took the liberty to put myself as co-author of d84350d, which is the commit i mainly reworked - the rest were mostly rebase conflicts and fixes for 1.22.

Closes #93.

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -245,17 +245,26 @@ impl MiniscriptKey for DescriptorPublicKey {
     type Hash = hash160::Hash;
 
     fn to_pubkeyhash(&self) -> Self::Hash {
+        self.to_public_key().to_pubkeyhash()
+    }
+}
+
+impl ToPublicKey for DescriptorPublicKey {
+    fn to_public_key(&self) -> bitcoin::PublicKey {
         match *self {
-            DescriptorPublicKey::PubKey(ref pk) => pk.to_pubkeyhash(),
+            DescriptorPublicKey::PubKey(ref pk) => *pk,
             DescriptorPublicKey::XPub(ref xpub) => {
                 let ctx = secp256k1::Secp256k1::verification_only();
                 xpub.xpub
                     .derive_pub(&ctx, &xpub.derivation_path)
-                    .expect("Won't fail, only normal derivations")
+                    .expect("Shouldn't fail, only normal derivations")
                     .public_key
-                    .to_pubkeyhash()
             }
         }
+    }
+
+    fn hash_to_hash160(hash: &Self::Hash) -> hash160::Hash {
+        *hash
     }
 }
 
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -27,7 +27,12 @@ use std::fmt;
 use std::str::{self, FromStr};
 
 use bitcoin::blockdata::{opcodes, script};
+use bitcoin::hashes::hex::FromHex;
+use bitcoin::util::bip32;
 use bitcoin::{self, Script};
+
+#[cfg(feature = "serde")]
+use serde::{de, ser};
 
 use expression;
 use miniscript;
@@ -66,6 +71,140 @@ pub enum Descriptor<Pk: MiniscriptKey> {
     Wsh(Miniscript<Pk, Segwitv0>),
     /// P2SH-P2WSH with Segwitv0 context
     ShWsh(Miniscript<Pk, Segwitv0>),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum DescriptorPublicKey {
+    PubKey(bitcoin::PublicKey),
+    XPub(DescriptorXPub),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct DescriptorXPub {
+    pub origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+    pub xpub: bip32::ExtendedPubKey,
+    pub derivation_path: bip32::DerivationPath,
+    pub is_wildcard: bool,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct DescriptorKeyParseError(&'static str);
+
+impl FromStr for DescriptorPublicKey {
+    type Err = DescriptorKeyParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.len() < 66 {
+            return Err(DescriptorKeyParseError(
+                "Key too short (<66 char), doesn't match any format",
+            ));
+        }
+
+        for ch in s.as_bytes() {
+            if *ch < 20 || *ch > 127 {
+                return Err(DescriptorKeyParseError(
+                    "Encountered an unprintable character",
+                ));
+            }
+        }
+
+        if s.chars().next().unwrap() == '[' {
+            let mut parts = s[1..].split(']');
+            let mut raw_origin = parts
+                .next()
+                .ok_or(DescriptorKeyParseError("Unclosed '['"))?
+                .split('/');
+
+            let origin_id_hex = raw_origin.next().ok_or(DescriptorKeyParseError(
+                "No master fingerprint found after '['",
+            ))?;
+
+            if origin_id_hex.len() != 8 {
+                return Err(DescriptorKeyParseError(
+                    "Master fingerprint should be 8 characters long",
+                ));
+            }
+            let parent_fingerprint = bip32::Fingerprint::from_hex(origin_id_hex).map_err(|_| {
+                DescriptorKeyParseError("Malformed master fingerprint, expected 8 hex chars")
+            })?;
+
+            let origin_path = raw_origin
+                .map(|p| bip32::ChildNumber::from_str(p))
+                .collect::<Result<bip32::DerivationPath, bip32::Error>>()
+                .map_err(|_| {
+                    DescriptorKeyParseError("Error while parsing master derivation path")
+                })?;
+
+            let key_deriv = parts
+                .next()
+                .ok_or(DescriptorKeyParseError("No key after origin."))?;
+
+            let (xpub, derivation_path, is_wildcard) = Self::parse_xpub_deriv(key_deriv)?;
+
+            Ok(DescriptorPublicKey::XPub(DescriptorXPub {
+                origin: Some((parent_fingerprint, origin_path)),
+                xpub,
+                derivation_path,
+                is_wildcard,
+            }))
+        } else if s.starts_with("02") || s.starts_with("03") || s.starts_with("04") {
+            let pk = bitcoin::PublicKey::from_str(s)
+                .map_err(|_| DescriptorKeyParseError("Error while parsing simple public key"))?;
+            Ok(DescriptorPublicKey::PubKey(pk))
+        } else {
+            let (xpub, derivation_path, is_wildcard) = Self::parse_xpub_deriv(s)?;
+            Ok(DescriptorPublicKey::XPub(DescriptorXPub {
+                origin: None,
+                xpub,
+                derivation_path,
+                is_wildcard,
+            }))
+        }
+    }
+}
+
+impl DescriptorPublicKey {
+    /// Parse an extended public key concatenated to a derivation path.
+    fn parse_xpub_deriv(
+        key_deriv: &str,
+    ) -> Result<(bip32::ExtendedPubKey, bip32::DerivationPath, bool), DescriptorKeyParseError> {
+        let mut key_deriv = key_deriv.split('/');
+        let xpub_str = key_deriv.next().ok_or(DescriptorKeyParseError(
+            "No key found after origin description",
+        ))?;
+        let xpub = bip32::ExtendedPubKey::from_str(xpub_str)
+            .map_err(|_| DescriptorKeyParseError("Error while parsing xpub."))?;
+
+        let mut is_wildcard = false;
+        let derivation_path = key_deriv
+            .filter_map(|p| {
+                if !is_wildcard && p == "*" {
+                    is_wildcard = true;
+                    None
+                } else if !is_wildcard && p == "*'" {
+                    Some(Err(DescriptorKeyParseError(
+                        "Hardened derivation is currently not supported.",
+                    )))
+                } else if is_wildcard {
+                    Some(Err(DescriptorKeyParseError(
+                        "'*' may only appear as last element in a derivation path.",
+                    )))
+                } else {
+                    Some(bip32::ChildNumber::from_str(p).map_err(|_| {
+                        DescriptorKeyParseError("Error while parsing key derivation path")
+                    }))
+                }
+            })
+            .collect::<Result<bip32::DerivationPath, _>>()?;
+
+        if (&derivation_path).into_iter().all(|c| c.is_normal()) {
+            Ok((xpub, derivation_path, is_wildcard))
+        } else {
+            Err(DescriptorKeyParseError(
+                "Hardened derivation is currently not supported.",
+            ))
+        }
+    }
 }
 
 impl<Pk: MiniscriptKey> Descriptor<Pk> {
@@ -551,16 +690,23 @@ serde_string_impl_pk!(Descriptor, "a script descriptor");
 
 #[cfg(test)]
 mod tests {
+    use super::DescriptorKeyParseError;
+
     use bitcoin::blockdata::opcodes::all::{OP_CLTV, OP_CSV};
     use bitcoin::blockdata::script::Instruction;
     use bitcoin::blockdata::{opcodes, script};
     use bitcoin::hashes::hex::FromHex;
     use bitcoin::hashes::{hash160, sha256};
+    use bitcoin::util::bip32;
     use bitcoin::{self, secp256k1, PublicKey};
+    use descriptor::{DescriptorPublicKey, DescriptorXPub};
     use miniscript::satisfy::BitcoinSig;
     use std::collections::HashMap;
     use std::str::FromStr;
     use {Descriptor, DummyKey, Miniscript, Satisfier};
+
+    #[cfg(feature = "compiler")]
+    use policy;
 
     type StdDescriptor = Descriptor<PublicKey>;
     const TEST_PK: &'static str =
@@ -1063,6 +1209,121 @@ mod tests {
                 .as_bytes(),
             Vec::<u8>::from_hex("522103789ed0bb717d88f7d321a368d905e7430207ebbd82bd342cf11ae157a7ace5fd2103dbc6764b8884a92e871274b87583e6d5c2a58819473e17e107ef3f6aa5a6162652ae")
                 .unwrap()[..]
+        );
+    }
+
+    #[test]
+    fn parse_descriptor_key() {
+        // With a wildcard
+        let key = "[78412e3a/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*";
+        let expected = DescriptorPublicKey::XPub(DescriptorXPub {
+            origin: Some((
+                bip32::Fingerprint::from(&[0x78, 0x41, 0x2e, 0x3a][..]),
+                (&[
+                    bip32::ChildNumber::from_hardened_idx(44).unwrap(),
+                    bip32::ChildNumber::from_hardened_idx(0).unwrap(),
+                    bip32::ChildNumber::from_hardened_idx(0).unwrap(),
+                ][..])
+                .into(),
+            )),
+            xpub: bip32::ExtendedPubKey::from_str("xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL").unwrap(),
+            derivation_path: (&[bip32::ChildNumber::from_normal_idx(1).unwrap()][..]).into(),
+            is_wildcard: true,
+        });
+        assert_eq!(expected, key.parse().unwrap());
+
+        // Without origin
+        let key = "xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1";
+        let expected = DescriptorPublicKey::XPub(DescriptorXPub {
+            origin: None,
+            xpub: bip32::ExtendedPubKey::from_str("xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL").unwrap(),
+            derivation_path: (&[bip32::ChildNumber::from_normal_idx(1).unwrap()][..]).into(),
+            is_wildcard: false,
+        });
+        assert_eq!(expected, key.parse().unwrap());
+
+        // Without derivation path
+        let key = "xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL";
+        let expected = DescriptorPublicKey::XPub(DescriptorXPub {
+            origin: None,
+            xpub: bip32::ExtendedPubKey::from_str("xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL").unwrap(),
+            derivation_path: bip32::DerivationPath::from(&[][..]),
+            is_wildcard: false,
+        });
+        assert_eq!(expected, key.parse().unwrap());
+
+        // Raw (compressed) pubkey
+        let key = "03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8";
+        let expected = DescriptorPublicKey::PubKey(
+            bitcoin::PublicKey::from_str(
+                "03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8",
+            )
+            .unwrap(),
+        );
+        assert_eq!(expected, key.parse().unwrap());
+
+        // Raw (uncompressed) pubkey
+        let key = "04f5eeb2b10c944c6b9fbcfff94c35bdeecd93df977882babc7f3a2cf7f5c81d3b09a68db7f0e04f21de5d4230e75e6dbe7ad16eefe0d4325a62067dc6f369446a";
+        let expected = DescriptorPublicKey::PubKey(
+            bitcoin::PublicKey::from_str(
+                "04f5eeb2b10c944c6b9fbcfff94c35bdeecd93df977882babc7f3a2cf7f5c81d3b09a68db7f0e04f21de5d4230e75e6dbe7ad16eefe0d4325a62067dc6f369446a",
+            )
+            .unwrap(),
+        );
+        assert_eq!(expected, key.parse().unwrap());
+    }
+
+    #[test]
+    fn parse_descriptor_key_errors() {
+        // origin is only supported for xpubs
+        let desc =
+            "[78412e3a/0'/0'/0']0231c7d3fc85c148717848033ce276ae2b464a4e2c367ed33886cc428b8af48ff8";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError("Error while parsing xpub."))
+        );
+
+        // We refuse creating descriptors which claim to be able to derive hardened childs
+        let desc = "[78412e3a/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/42'/*";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "Hardened derivation is currently not supported."
+            ))
+        );
+
+        // And even if they they claim it for the wildcard!
+        let desc = "[78412e3a/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/42/*'";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "Hardened derivation is currently not supported."
+            ))
+        );
+
+        // And ones with misplaced wildcard
+        let desc = "[78412e3a/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*/44";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "\'*\' may only appear as last element in a derivation path."
+            ))
+        );
+
+        // And ones with invalid fingerprints
+        let desc = "[NonHexor]xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "Malformed master fingerprint, expected 8 hex chars"
+            ))
+        );
+
+        // And ones with invalid xpubs
+        let desc = "[78412e3a]xpub1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaLcgJvLJuZZvRcEL/1/*";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError("Error while parsing xpub."))
         );
     }
 }
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -92,6 +92,12 @@ pub struct DescriptorXPub {
 #[derive(Debug, PartialEq, Clone, Copy)]
 pub struct DescriptorKeyParseError(&'static str);
 
+impl fmt::Display for DescriptorKeyParseError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str(self.0)
+    }
+}
+
 impl fmt::Display for DescriptorPublicKey {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
@@ -239,13 +245,37 @@ impl DescriptorPublicKey {
             ))
         }
     }
+
+    /// Derives the specified child key if self is a wildcard xpub. Otherwise returns self.
+    ///
+    /// Panics if given a hardened child number
+    pub fn derive(self, child_number: bip32::ChildNumber) -> DescriptorPublicKey {
+        debug_assert!(child_number.is_normal());
+
+        match self {
+            DescriptorPublicKey::PubKey(_) => self,
+            DescriptorPublicKey::XPub(xpub) => {
+                if xpub.is_wildcard {
+                    DescriptorPublicKey::XPub(DescriptorXPub {
+                        origin: xpub.origin,
+                        xpub: xpub.xpub,
+                        derivation_path: xpub.derivation_path.into_child(child_number),
+                        is_wildcard: false,
+                    })
+                } else {
+                    DescriptorPublicKey::XPub(xpub)
+                }
+            }
+        }
+    }
 }
 
 impl MiniscriptKey for DescriptorPublicKey {
-    type Hash = hash160::Hash;
+    // This allows us to be able to derive public keys even for PkH s
+    type Hash = Self;
 
-    fn to_pubkeyhash(&self) -> Self::Hash {
-        self.to_public_key().to_pubkeyhash()
+    fn to_pubkeyhash(&self) -> Self {
+        self.clone()
     }
 }
 
@@ -264,7 +294,7 @@ impl ToPublicKey for DescriptorPublicKey {
     }
 
     fn hash_to_hash160(hash: &Self::Hash) -> hash160::Hash {
-        *hash
+        hash.to_public_key().to_pubkeyhash()
     }
 }
 
@@ -622,6 +652,17 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             //       the `scriptCode` is the `witnessScript` serialized as scripts inside CTxOut.
             Descriptor::Wsh(ref d) | Descriptor::ShWsh(ref d) => d.encode(),
         }
+    }
+}
+
+impl Descriptor<DescriptorPublicKey> {
+    /// Derives all wildcard keys in the descriptor using the supplied `child_number`
+    pub fn derive(&self, child_number: bip32::ChildNumber) -> Descriptor<DescriptorPublicKey> {
+        self.translate_pk(
+            |pk| Result::Ok::<DescriptorPublicKey, ()>(pk.clone().derive(child_number)),
+            |pk| Result::Ok::<DescriptorPublicKey, ()>(pk.clone().derive(child_number)),
+        )
+        .expect("Translation fn can't fail.")
     }
 }
 
@@ -1391,5 +1432,28 @@ mod tests {
             DescriptorPublicKey::from_str(desc),
             Err(DescriptorKeyParseError("Error while parsing xpub."))
         );
+    }
+
+    #[test]
+    #[cfg(feature = "compiler")]
+    fn parse_and_derive() {
+        let descriptor_str = "thresh(2,\
+pk([d34db33f/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*),\
+pk(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1),\
+pk(03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8))";
+        let policy: policy::concrete::Policy<DescriptorPublicKey> = descriptor_str.parse().unwrap();
+        let descriptor = Descriptor::Sh(policy.compile().unwrap());
+        let derived_descriptor =
+            descriptor.derive(bip32::ChildNumber::from_normal_idx(42).unwrap());
+
+        let res_descriptor_str = "thresh(2,\
+pk([d34db33f/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/42),\
+pk(xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1),\
+pk(03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8))";
+        let res_policy: policy::concrete::Policy<DescriptorPublicKey> =
+            res_descriptor_str.parse().unwrap();
+        let res_descriptor = Descriptor::Sh(res_policy.compile().unwrap());
+
+        assert_eq!(res_descriptor, derived_descriptor);
     }
 }
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -27,7 +27,9 @@ use std::fmt;
 use std::str::{self, FromStr};
 
 use bitcoin::blockdata::{opcodes, script};
+use bitcoin::hashes::hash160;
 use bitcoin::hashes::hex::FromHex;
+use bitcoin::secp256k1;
 use bitcoin::util::bip32;
 use bitcoin::{self, Script};
 
@@ -73,13 +75,13 @@ pub enum Descriptor<Pk: MiniscriptKey> {
     ShWsh(Miniscript<Pk, Segwitv0>),
 }
 
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
 pub enum DescriptorPublicKey {
     PubKey(bitcoin::PublicKey),
     XPub(DescriptorXPub),
 }
 
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
 pub struct DescriptorXPub {
     pub origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
     pub xpub: bip32::ExtendedPubKey,
@@ -87,8 +89,40 @@ pub struct DescriptorXPub {
     pub is_wildcard: bool,
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone, Copy)]
 pub struct DescriptorKeyParseError(&'static str);
+
+impl fmt::Display for DescriptorPublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            DescriptorPublicKey::PubKey(ref pk) => pk.fmt(f),
+            DescriptorPublicKey::XPub(ref xpub) => {
+                if let Some((ref master_id, ref master_deriv)) = xpub.origin {
+                    fmt::Formatter::write_str(f, "[")?;
+                    for byte in master_id.into_bytes().iter() {
+                        write!(f, "{:02x}", byte)?;
+                    }
+                    fmt_derivation_path(f, master_deriv)?;
+                    fmt::Formatter::write_str(f, "]")?;
+                }
+                xpub.xpub.fmt(f)?;
+                fmt_derivation_path(f, &xpub.derivation_path)?;
+                if xpub.is_wildcard {
+                    write!(f, "/*")?;
+                }
+                Ok(())
+            }
+        }
+    }
+}
+
+/// Writes a derivation path to the formatter, no leading 'm'
+fn fmt_derivation_path(f: &mut fmt::Formatter, path: &bip32::DerivationPath) -> fmt::Result {
+    for child in path {
+        write!(f, "/{}", child)?;
+    }
+    Ok(())
+}
 
 impl FromStr for DescriptorPublicKey {
     type Err = DescriptorKeyParseError;
@@ -203,6 +237,24 @@ impl DescriptorPublicKey {
             Err(DescriptorKeyParseError(
                 "Hardened derivation is currently not supported.",
             ))
+        }
+    }
+}
+
+impl MiniscriptKey for DescriptorPublicKey {
+    type Hash = hash160::Hash;
+
+    fn to_pubkeyhash(&self) -> Self::Hash {
+        match *self {
+            DescriptorPublicKey::PubKey(ref pk) => pk.to_pubkeyhash(),
+            DescriptorPublicKey::XPub(ref xpub) => {
+                let ctx = secp256k1::Secp256k1::verification_only();
+                xpub.xpub
+                    .derive_pub(&ctx, &xpub.derivation_path)
+                    .expect("Won't fail, only normal derivations")
+                    .public_key
+                    .to_pubkeyhash()
+            }
         }
     }
 }
@@ -1231,6 +1283,7 @@ mod tests {
             is_wildcard: true,
         });
         assert_eq!(expected, key.parse().unwrap());
+        assert_eq!(format!("{}", expected), key);
 
         // Without origin
         let key = "xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1";
@@ -1241,6 +1294,7 @@ mod tests {
             is_wildcard: false,
         });
         assert_eq!(expected, key.parse().unwrap());
+        assert_eq!(format!("{}", expected), key);
 
         // Without derivation path
         let key = "xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL";
@@ -1251,6 +1305,7 @@ mod tests {
             is_wildcard: false,
         });
         assert_eq!(expected, key.parse().unwrap());
+        assert_eq!(format!("{}", expected), key);
 
         // Raw (compressed) pubkey
         let key = "03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8";
@@ -1261,6 +1316,7 @@ mod tests {
             .unwrap(),
         );
         assert_eq!(expected, key.parse().unwrap());
+        assert_eq!(format!("{}", expected), key);
 
         // Raw (uncompressed) pubkey
         let key = "04f5eeb2b10c944c6b9fbcfff94c35bdeecd93df977882babc7f3a2cf7f5c81d3b09a68db7f0e04f21de5d4230e75e6dbe7ad16eefe0d4325a62067dc6f369446a";
@@ -1271,6 +1327,7 @@ mod tests {
             .unwrap(),
         );
         assert_eq!(expected, key.parse().unwrap());
+        assert_eq!(format!("{}", expected), key);
     }
 
     #[test]
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -77,8 +77,14 @@ pub enum Descriptor<Pk: MiniscriptKey> {
 
 #[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
 pub enum DescriptorPublicKey {
-    PubKey(bitcoin::PublicKey),
+    SinglePub(DescriptorSinglePub),
     XPub(DescriptorXPub),
+}
+
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
+pub struct DescriptorSinglePub {
+    pub origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+    pub key: bitcoin::PublicKey,
 }
 
 #[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
@@ -101,16 +107,13 @@ impl fmt::Display for DescriptorKeyParseError {
 impl fmt::Display for DescriptorPublicKey {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
-            DescriptorPublicKey::PubKey(ref pk) => pk.fmt(f),
+            DescriptorPublicKey::SinglePub(ref pk) => {
+                maybe_fmt_master_id(f, &pk.origin)?;
+                pk.key.fmt(f)?;
+                Ok(())
+            }
             DescriptorPublicKey::XPub(ref xpub) => {
-                if let Some((ref master_id, ref master_deriv)) = xpub.origin {
-                    fmt::Formatter::write_str(f, "[")?;
-                    for byte in master_id.into_bytes().iter() {
-                        write!(f, "{:02x}", byte)?;
-                    }
-                    fmt_derivation_path(f, master_deriv)?;
-                    fmt::Formatter::write_str(f, "]")?;
-                }
+                maybe_fmt_master_id(f, &xpub.origin)?;
                 xpub.xpub.fmt(f)?;
                 fmt_derivation_path(f, &xpub.derivation_path)?;
                 if xpub.is_wildcard {
@@ -120,6 +123,23 @@ impl fmt::Display for DescriptorPublicKey {
             }
         }
     }
+}
+
+/// Writes the fingerprint of the origin, if there is one.
+fn maybe_fmt_master_id(
+    f: &mut fmt::Formatter,
+    origin: &Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+) -> fmt::Result {
+    if let Some((ref master_id, ref master_deriv)) = *origin {
+        fmt::Formatter::write_str(f, "[")?;
+        for byte in master_id.into_bytes().iter() {
+            write!(f, "{:02x}", byte)?;
+        }
+        fmt_derivation_path(f, master_deriv)?;
+        fmt::Formatter::write_str(f, "]")?;
+    }
+
+    Ok(())
 }
 
 /// Writes a derivation path to the formatter, no leading 'm'
@@ -134,6 +154,7 @@ impl FromStr for DescriptorPublicKey {
     type Err = DescriptorKeyParseError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // A "raw" public key without any origin is the least we accept.
         if s.len() < 66 {
             return Err(DescriptorKeyParseError(
                 "Key too short (<66 char), doesn't match any format",
@@ -148,8 +169,11 @@ impl FromStr for DescriptorPublicKey {
             }
         }
 
+        let mut parts = s[1..].split(']');
+
+        // They may specify an origin
+        let mut origin = None;
         if s.chars().next().unwrap() == '[' {
-            let mut parts = s[1..].split(']');
             let mut raw_origin = parts
                 .next()
                 .ok_or(DescriptorKeyParseError("Unclosed '['"))?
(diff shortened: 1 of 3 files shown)


darosior:
Added a commit to fix the 1.22.0 build (and fixed the actual build under 1.22)

src/descriptor/mod.rs
```diff
@@ -68,6 +75,258 @@ pub enum Descriptor<Pk: MiniscriptKey> {
     ShWsh(Miniscript<Pk, Segwitv0>),
 }
 
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
+pub enum DescriptorPublicKey {
+    ShortPub(DescriptorShortPub),
```
sgeisler, line 80: I'd rather call it `DescriptorSinglePub` to avoid confusion with the compressed/uncompressed dichotomy.

src/descriptor/mod.rs
```diff
@@ -68,6 +75,258 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+    XPub(DescriptorXPub),
+}
+
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
+pub struct DescriptorShortPub {
+    origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
```
sgeisler, line 86: As both `DescriptorShortPub` and `DescriptorXPub` have an `origin` wouldn't it fit better into a top level `DescriptorPublicKey` struct?

src/descriptor/mod.rs
```diff
@@ -68,6 +75,258 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
+pub struct DescriptorXPub {
+    origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+    xpub: bip32::ExtendedPubKey,
+    derivation_path: bip32::DerivationPath,
+    is_wildcard: bool,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct DescriptorKeyParseError(&'static str);
```
sgeisler, line 99: I used this as a placeholder in my WIP PR, but we should probably give more contextualized and structural errors.

src/descriptor/mod.rs
```diff
@@ -68,6 +75,258 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+    fn to_pubkeyhash(&self) -> Self::Hash {
+        match *self {
+            DescriptorPublicKey::ShortPub(ref spub) => spub.key.to_pubkeyhash(),
+            DescriptorPublicKey::XPub(ref xpub) => {
+                let ctx = secp256k1::Secp256k1::verification_only();
+                xpub.xpub
+                    .derive_pub(&ctx, &xpub.derivation_path)
+                    .expect("Won't fail, only normal derivations")
+                    .public_key
+                    .to_pubkeyhash()
```
sgeisler, line 305: This could be extracted into an implementation of `MiniscriptKey` for `DescriptorXPub`. Also the derivation could be used from `ToPublicKey`.

src/descriptor/mod.rs
```diff
@@ -68,6 +75,258 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+        }
+    }
+}
+
+impl ToPublicKey for DescriptorPublicKey {
+    fn to_public_key(&self) -> bitcoin::PublicKey {
+        match *self {
+            DescriptorPublicKey::ShortPub(ref spub) => spub.key.to_public_key(),
+            DescriptorPublicKey::XPub(ref xpub) => {
+                let ctx = secp256k1::Secp256k1::verification_only();
```
sgeisler, line 316: I think we can use the [global context](https://github.com/rust-bitcoin/rust-secp256k1/pull/224) by now.

sgeisler (changes requested):
I noticed you based your PR on #93 instead of #116. I guess this makes reviewing/merging quicker, I just hope it won't be too annoying for @afilini.

src/descriptor/mod.rs
```diff
@@ -68,6 +75,258 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+        let key_part = if origin == None {
+            Ok(s)
+        } else {
+            s.split(']')
+                .collect::<Vec<&str>>()
+                .pop()
+                .ok_or(DescriptorKeyParseError("No key after origin."))
+        }?;
+
+        if key_part.starts_with("xpub") {
```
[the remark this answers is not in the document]
darosior: I don't think we should ever support zpub ypub, that's kind of nonsensical even more as part of descriptors ?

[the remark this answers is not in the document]
sgeisler: I don't know how controversial this standard is … I just know that I found myself converting zpubs to xpubs recently to use rust-bitcoin. Imo we should at least support testnet keys.

.github/workflows/rust.yml
```diff
@@ -58,15 +58,17 @@ jobs:
...
+    steps:
+      - name: Checkout Crate
         uses: actions/checkout@v2
       - name: Checkout Toolchain
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
           toolchain: ${{ matrix.rust }}
           override: true
+      - name: Pin cc if rust 1.22
```
dr-orlovsky, line 70: Not sure I am getting why do we need this once we moved MSRV to 1.29

Cargo.toml
```diff
@@ -14,7 +14,7 @@ unstable = []
 default = []
 
 [dependencies]
-bitcoin = "0.23"
+bitcoin = "0.24"
```
dr-orlovsky, line 17: I think we can move now to 0.25

src/descriptor/satisfied_constraints.rs
```diff
@@ -25,6 +25,9 @@ use {BitcoinSig, ToPublicKey};
 /// Detailed Error type for Interpreter
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub enum Error {
+    /// An uncompressed public key was encountered in a context where it is
+    /// disallowed (e.g. in a Segwit script or p2wpkh output)
+    UncompressedPubkey,
```
dr-orlovsky, line 30: Shouldn't we provide an information on which key is uncompressed?

src/descriptor/mod.rs
dr-orlovsky: `DescriptorPublicKey` is not struct, and it will require to add external parameter to each of the enum options, which may be not that good

src/descriptor/mod.rs
```diff
@@ -68,6 +75,258 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
+pub struct DescriptorXPub {
+    origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+    xpub: bip32::ExtendedPubKey,
+    derivation_path: bip32::DerivationPath,
+    is_wildcard: bool,
+}
+
+#[derive(Debug, PartialEq)]
```
dr-orlovsky, line 98: +Clone +Copy

src/descriptor/mod.rs
```diff
@@ -68,6 +75,258 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+                Ok(())
+            }
+        }
+    }
+}
+
+/// Writes the fingerprint of the origin, if there is one.
+fn maybe_fmt_master_id(
+    f: &mut fmt::Formatter,
+    origin: &Option<(bip32::Fingerprint, bip32::DerivationPath)>,
```
dr-orlovsky, line 131: I am thinking of introducing this pair of `(bip32::Fingerprint, bip32::DerivationPath)` which we have all around throughout code of PSBTs to rust-bitcoin as a new type. Hard to pick a name for it...

src/descriptor/mod.rs
```diff
@@ -68,6 +75,258 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+) -> fmt::Result {
+    if let Some((ref master_id, ref master_deriv)) = *origin {
+        fmt::Formatter::write_str(f, "[")?;
+        for byte in master_id.into_bytes().iter() {
+            write!(f, "{:02x}", byte)?;
+        }
+        fmt_derivation_path(f, master_deriv)?;
+        fmt::Formatter::write_str(f, "]")?;
```
dr-orlovsky, line 139: It seems that
```suggestion
        fmt::Formatter::write_str(f, "[{:x?}]", master_id.into_bytes())?;
        fmt_derivation_path(f, master_deriv)?;
```
will be more idiomatic.

src/descriptor/mod.rs
darosior: Yes, that's a better name, thanks.

.github/workflows/rust.yml
darosior: It's not ?

src/descriptor/mod.rs
sgeisler: Maybe `KeySource`?

src/descriptor/mod.rs
darosior: Yeah i wanted to include it into our global `Error` enum at first but figured it'd be OK (we have per-file error eg for the policy compiler too).. On the other hand i think it's the only non-enum one. Will give it another look.

src/descriptor/mod.rs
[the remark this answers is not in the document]
darosior: Good point

src/descriptor/mod.rs
darosior: Will give a look

src/descriptor/satisfied_constraints.rs
darosior: This commit is part of #130

Cargo.toml
darosior: This commit is part of #130

src/descriptor/mod.rs
darosior: Is it idiomatic to rely on `Debug` to implement `Display` ? To be honest i'm fairly new to Rust but that sounds awkward to me, even though the diff is tempting to me !

dr-orlovsky: You are right, I missed that this fn is part of Display implementation

darosior:
> I noticed you based your PR on #93 instead of #116. I guess this makes reviewing/merging quicker, I just hope it won't be too annoying for @afilini.

I took #116 into consideration as well, i actually implemented the second item in [the list](https://github.com/rust-bitcoin/rust-miniscript/pull/116#issuecomment-668845773) you all seem to have agreed upon. A rebased and cleaned up version of generalistic descriptor keys seems to also be a requirement for it ?

src/descriptor/mod.rs
```diff
@@ -68,6 +75,258 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+
+            if origin_id_hex.len() != 8 {
+                return Err(DescriptorKeyParseError(
+                    "Master fingerprint should be 8 characters long",
+                ));
+            }
+            let origin_raw_id: [u8; 4] = FromHex::from_hex(origin_id_hex).map_err(|_| {
+                DescriptorKeyParseError("Malformed master fingerprint, expected 8 hex chars")
+            })?;
+            let parent_fingerprint = bip32::Fingerprint::from(origin_raw_id.as_ref());
```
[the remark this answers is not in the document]
darosior: Nice, thanks!

src/descriptor/mod.rs
darosior: Hmm or maybe we need to have the context as part of the struct ?..

src/descriptor/mod.rs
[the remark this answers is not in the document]
darosior: Ok so finally branched on `contains("pub")` to support testnet, this support your usecase as a consequence but i still don't think we should implement it :).

src/descriptor/mod.rs
sgeisler: We are working with public data, there is no reason not to use the global context imo.

darosior: Ok, but it then makes me introduce a new `secp256k1` dependency separated from the `rust-bitcoin` one

sgeisler: As long as it's the same as in the current rust-bitcoin version that seems ok, doesn't it? Cargo will just enable the feature, not add a whole new version of the crate.

src/descriptor/mod.rs
darosior: Ok so i tried implementing that as a proper `DescriptorParseError` enum. And i actually end up with less contextualized errors and useless enum variants.

We might want at some point to have it as a variant of a generalistic DescriptorError placed in mod.rs (there is already the InterpreterError in satisfied_constraints.rs). Don't want to drastically increase this PR's diff though.

src/descriptor/mod.rs
darosior: > Also the derivation could be used from ToPublicKey.

Oh, definitely. Will swap the two commits then.

> This could be extracted into an implementation of MiniscriptKey for DescriptorXPub.

Do we really want that ? If we do we probably want it for SinglePub too..

src/descriptor/mod.rs
darosior: Added

src/descriptor/mod.rs
sgeisler: > Do we really want that ? If we do we probably want it for SinglePub too.

You are right, let's implement it once there is an actual need for it.

src/descriptor/mod.rs
darosior: Ok so rust-secp256k1 v0.18 finally does not expose global-context https://github.com/rust-bitcoin/rust-secp256k1/blob/0.18.0/Cargo.toml, and v0.19 is a MSRV bump (and a rust-bitcoin bump as well).

afilini:
Makes sense, considering that I would have to rebase on master anyways, I might as well rebase on top of this PR.

Working on it now..

src/descriptor/mod.rs
dr-orlovsky: Done: https://github.com/rust-bitcoin/rust-bitcoin/pull/480

sgeisler:
@darosior do you consider the PR ready for another round of review?

darosior:
Yes

src/descriptor/mod.rs
```diff
@@ -553,13 +678,18 @@ serde_string_impl_pk!(Descriptor, "a script descriptor");
...
     use bitcoin::blockdata::opcodes::all::{OP_CLTV, OP_CSV};
     use bitcoin::blockdata::script::Instruction;
     use bitcoin::blockdata::{opcodes, script};
     use bitcoin::hashes::hex::FromHex;
     use bitcoin::hashes::{hash160, sha256};
+    use bitcoin::util::bip32;
     use bitcoin::{self, secp256k1, PublicKey};
+    use descriptor::{DescriptorPublicKey, DescriptorXPub};
     use miniscript::satisfy::BitcoinSig;
+    use policy;
```
sanket1729, line 827: only import with compiler flag

src/descriptor/mod.rs
sanket1729: Agreed. This can be done later

src/descriptor/mod.rs
```diff
@@ -68,6 +73,126 @@ pub enum Descriptor<Pk: MiniscriptKey> {
     ShWsh(Miniscript<Pk, Segwitv0>),
 }
 
+#[derive(Debug, Eq, PartialEq)]
+pub enum DescriptorPublicKey {
+    PukKey(bitcoin::PublicKey),
```
sanket1729: typo? Pubkey instead of PukKey?

src/descriptor/mod.rs
```diff
@@ -164,30 +189,34 @@ impl FromStr for DescriptorPublicKey {
...
                 })?;
+            origin = Some((parent_fingerprint, origin_path));
+        }
 
-            let key_deriv = parts
+        let key_part = if origin == None {
+            Ok(s)
+        } else {
+            s.split(']')
+                .rev()
```
sanket1729, line 199: Would this change accept incorrect strings with multiple `[` `]` as valid DescriptorPubkeys?

src/descriptor/mod.rs
```diff
@@ -225,6 +231,34 @@ impl DescriptorPublicKey {
             ))
         }
     }
+
+    /// Derives a new key using the path if self is a wildcard xpub. Otehrwise returns a copy of
+    /// self.
+    ///
+    /// Panics if derivation path contains a hardened child number
+    pub fn derive(&self, path: &[bip32::ChildNumber]) -> DescriptorPublicKey {
```
sanket1729, line 268: There is too much cloning going on here, should we consume `self` and let the caller decide whether to clone or not?

src/descriptor/mod.rs
```diff
@@ -613,6 +647,17 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
     }
 }
 
+impl Descriptor<DescriptorPublicKey> {
+    /// Derives all wildcard keys in the descriptor using the supplied `path`
+    pub fn derive(&self, path: &[bip32::ChildNumber]) -> Descriptor<DescriptorPublicKey> {
```
sanket1729, line 681: I am not that familiar with HD wallets, but does it make sense to all pubkeys be derived using the same path? For example, `sh( and (pk/1/*, pk/2/*))`, why should the * be the same for all pubkeys?

sanket1729 (commented):
Hi @darosior, I agree with the general idea of keeping things minimal in this PR so that easier for review. 

I reviewed commit by commit, so of the issues had been addressed in later commits and you can safely ignore those.

src/descriptor/mod.rs
darosior: Definitely is.. Thanks!

src/descriptor/mod.rs
darosior: No, a duplicated `[` would be counted as part of the size of the fingerprint: either the fingerprint is already valid (8 chars) and you get a "too long fingerprint, not 8 chars", or it eventually fails to parse [ as an hex char.

A duplicated `]`.... Would be accepted. Argh, nice catch correcting now. Tried with:
```rust
diff --git a/src/descriptor/mod.rs b/src/descriptor/mod.rs
index 91a0149..4eddec5 100644
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -1474,6 +1474,14 @@ mod tests {
                 "Error while parsing simple public key"
             ))
         );
+
+        let desc = "[78412e3a]]03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "Error while parsing simple public key"
+            ))
+        );
     }
 
     #[test]
```

darosior: Fixed
```patch
diff --git a/src/descriptor/mod.rs b/src/descriptor/mod.rs
index 91a0149..af5d470 100644
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -161,10 +161,11 @@ impl FromStr for DescriptorPublicKey {
             ));
         }
 
+        let mut parts = s[1..].split(']');
+
         // They may specify an origin
         let mut origin = None;
         if s.chars().next().unwrap() == '[' {
-            let mut parts = s[1..].split(']');
             let mut raw_origin = parts
                 .next()
                 .ok_or(DescriptorKeyParseError("Unclosed '['"))?
@@ -195,8 +196,7 @@ impl FromStr for DescriptorPublicKey {
         let key_part = if origin == None {
             Ok(s)
         } else {
-            s.split(']')
-                .rev()
+            parts
                 .next()
                 .ok_or(DescriptorKeyParseError("No key after origin."))
         }?;
@@ -1474,6 +1474,15 @@ mod tests {
                 "Error while parsing simple public key"
             ))
         );
+
+        // ..or invalid separators
+        let desc = "[78412e3a]]03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "Error while parsing simple public key"
+            ))
+        );
     }
 
     #[test]
```

src/descriptor/mod.rs
darosior: Yeah, definitely seems a better API.

src/descriptor/mod.rs
darosior: Corrected for the sake of clean history.

src/descriptor/mod.rs
sgeisler: I'd generally agree, but the primary use case is to do multiple derivations e.g. from a `DerivationPathIterator`, not a single derivation. So you'd nearly always clone anyway. Maybe a key type with a lifetime that only borrows the additional derivation path would be a solution? I just tried to keep it simple for now by not over-optimizing it.

src/descriptor/mod.rs
sgeisler: The `*` only stands for the last two derivation steps in my use case (internal/external, address number in wallet). So ideally you just specify your already-derived xpubs for that particular multisig wallet and these keys are derived in lockstep (which is what you want to do imo). You never want to generate an address from `sh(and(A/1/12, B/0/15))` imo.

darosior: Agreed with Sebastian on my specific usecase as well, but i have no particular opinion regarding the API.

src/descriptor/mod.rs
darosior: Came up with
```rust
diff --git a/src/descriptor/mod.rs b/src/descriptor/mod.rs
index 49ecae2..00a780c 100644
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -101,7 +101,7 @@ impl fmt::Display for DescriptorKeyParseError {
 impl fmt::Display for DescriptorPublicKey {
         match *self {
-            DescriptorPublicKey::PukKey(ref pk) => pk.fmt(f),
+            DescriptorPublicKey::PubKey(ref pk) => pk.fmt(f),
             DescriptorPublicKey::XPub(ref xpub) => {
                 if let Some((ref master_id, ref master_deriv)) = xpub.origin {
                     fmt::Formatter::write_str(f, "[")?;
@@ -232,21 +232,22 @@ impl DescriptorPublicKey {
         }
     }
 
-    /// Derives a new key using the path if self is a wildcard xpub. Otehrwise returns a copy of
+    /// Derives a new key using the path if self is a wildcard xpub. Otherwise returns a copy of
     /// self.
     ///
     /// Panics if derivation path contains a hardened child number
-    pub fn derive(&self, path: &[bip32::ChildNumber]) -> DescriptorPublicKey {
+    pub fn derive(self, path: &[bip32::ChildNumber]) -> DescriptorPublicKey {
         assert!(path.into_iter().all(|c| c.is_normal()));
 
-        match *self {
-            DescriptorPublicKey::PukKey(ref pk) => DescriptorPublicKey::PukKey(*pk),
-            DescriptorPublicKey::XPub(ref xpub) => {
+        match self {
+            DescriptorPublicKey::PubKey(_) => self,
+            DescriptorPublicKey::XPub(xpub) => {
                 if xpub.is_wildcard {
                     DescriptorPublicKey::XPub(DescriptorXPub {
-                        origin: xpub.origin.clone(),
-                        xpub: xpub.xpub.clone(),
-                        derivation_path: (&xpub.derivation_path)
+                        origin: xpub.origin,
+                        xpub: xpub.xpub,
+                        derivation_path: xpub
+                            .derivation_path
                             .into_iter()
                             .chain(path.iter())
                             .cloned()
@@ -254,7 +255,7 @@ impl DescriptorPublicKey {
                         is_wildcard: false,
                     })
                 } else {
-                    self.clone()
+                    DescriptorPublicKey::XPub(xpub)
                 }
             }
         }
@@ -272,7 +273,7 @@ impl MiniscriptKey for DescriptorPublicKey {
 impl ToPublicKey for DescriptorPublicKey {
     fn to_public_key(&self) -> bitcoin::PublicKey {
         match *self {
-            DescriptorPublicKey::PukKey(ref pk) => *pk,
+            DescriptorPublicKey::PubKey(ref pk) => *pk,
             DescriptorPublicKey::XPub(ref xpub) => {
                 let ctx = secp256k1::Secp256k1::verification_only();
                 xpub.xpub
@@ -651,7 +652,7 @@ impl Descriptor<DescriptorPublicKey> {
     /// Derives all wildcard keys in the descriptor using the supplied `path`
     pub fn derive(&self, path: &[bip32::ChildNumber]) -> Descriptor<DescriptorPublicKey> {
         self.translate_pk(
-            |pk| Result::Ok::<DescriptorPublicKey, ()>(pk.derive(path)),
+            |pk| Result::Ok::<DescriptorPublicKey, ()>(pk.clone().derive(path)),
             |pkh| Result::Ok::<hash160::Hash, ()>(*pkh),
         )
         .expect("Translation fn can't fail.")
```
For bf13c51df93399f0115c6d72e1f4cecf00219701, applying it for now but we can still discuss it.

darosior:
Thanks for the review, pushed the fixes.

src/descriptor/mod.rs
sanket1729: I was thinking more of a multi-party setting, but I guess in that case the users can call `translate_fn` directly. Since that is not common, we can have this API as is or perhaps rename it `derive_all_keys` to make it more explicit?

sanket1729:
tAck 51a21363a0bc71a7d85b03c920bab3c82549c991. Can merge after feedback from other reviewers as this already contains the commit from #130

src/descriptor/mod.rs
sgeisler: > Otherwise returns ~a copy of~ self.

src/descriptor/mod.rs
sgeisler: But it only derives keys with an asterisk at the end, so that wouldn't be all that accurate either.

src/descriptor/mod.rs
darosior: Thanks, corrected

darosior:
Made the members of the DescriptorXXX structs public as i actually tried to use this :-)

src/descriptor/mod.rs
```diff
@@ -420,6 +676,17 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
...
+        self.translate_pk(
+            |pk| Result::Ok::<DescriptorPublicKey, ()>(pk.clone().derive(path)),
+            |pkh| Result::Ok::<hash160::Hash, ()>(*pkh),
+        )
```
darosior, line 685: Actually *this* is wrong: it won't translate hashes for the new public key. This prevents the implementation of `lookup_pkh_sig` for the satisfier.

darosior: The only way around i see is to store the actual public key as well as the pkh in the enum variant.

sgeisler: I don't see how this breaks anything. Keys that are referred to by hash just won't be derived (because we can't), just as with non-wildcard xpubs and normal pubs. Can you point me to the part in the satisfier code where you think this derive impl breaks invariants please? Maybe I'm just missing something.

src/descriptor/mod.rs
```diff
@@ -231,18 +231,27 @@ impl MiniscriptKey for DescriptorPublicKey {
...
+        self.to_public_key().to_pubkeyhash()
+    }
+}
+
+impl ToPublicKey for DescriptorPublicKey {
+    fn to_public_key(&self) -> bitcoin::PublicKey {
         match *self {
-            DescriptorPublicKey::PubKey(ref pk) => pk.to_pubkeyhash(),
+            DescriptorPublicKey::PubKey(ref pk) => *pk,
             DescriptorPublicKey::XPub(ref xpub) => {
```
sanket1729, line 305: Is this is the expected behaviour for wildcard xpubs?

sgeisler: I think so. All we do on derivation is appending to the internal derivation path (the one that still has to be applied) iif it's an xpub which is also marked as wildcard. So when calculating the actual pub key for an xpub we need to apply this derivation path (nothing to do in case of normal pubs, they might be some static backup key, idk, they definitely don't need any derivation conceptually).

src/descriptor/mod.rs
darosior: Actually i came across this using it on my end.

I have those scripts for revault in which the policy compiler is optimizing part of the pubkeys to pubkey hashes.
I use a `descriptor_with_master_xpubs.derive([some, path])` kind of logic, which does not derive the hashes.

I could re-generate a descriptor each time i want to use a new depth, but that's... killing the purpose of having a derive() method.
I think that deriving only part of the keys is also an unexpected api: we could instead store the `Pk` as well as the `Pkh` when translating pubkeys to pubkey hashes.

sgeisler: Yeah, I mean even the `p2pkh` descriptor uses the key type, there's now way to construct this directly from a hash, so why not also keep the key in miniscript? The conversion to a hash can happen when needed. It should be clear from the context (`pkh`) anyway.

src/descriptor/mod.rs
darosior: Was a good idea finally, this allows to remove the `clone()` in `derive`.

darosior:
Pushed 2 commits adding the `Option<Pk>` to `PkH`es and using it to implement the translation in ExtendedPubkey descriptors derivation.

The API for derive() was changed: it now returns a `Result` as the derivation *can* fail if the Miniscript was parsed (Pk is None in this case because we can't guess the preimage).
I thought it was better to do this rather than silently falling back to not translating `PkH`es.

sanket1729:
Concept Ack @darosior. Thanks for doing this. Indeed, script is not the only way to create and having the Option<Pk> is a very nice idea. 
tAck e0e8b682ecc4475f65ac223e12534c36062e8efe . 

cc @apoelstra. This change is a breaking API change as it changes the signature of the `translate_fn` method which I think is used in liquid.

src/descriptor/mod.rs
```diff
@@ -68,6 +73,126 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+    PubKey(bitcoin::PublicKey),
+    XPub(DescriptorXPub),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct DescriptorXPub {
+    pub origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+    pub xpub: bip32::ExtendedPubKey,
+    pub derivation_path: bip32::DerivationPath,
+    pub is_wildcard: bool,
```
apoelstra, line 95: According to https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md we can end with `/*` or `/*'` ... so I think this should be an enum rather than a bool, with values `Fixed`, `Wildcard`, `WildcardHardened` say

sgeisler: As long as we don't support private keys in descriptors that doesn't make a lot of sense imo and would only add error cases.

But the stronger argument against this might be the current API design: it allows the asterisk to be replaced by arbitrary `DerivationPath`s, which may contain hardened and unhardened `ChildNumber`s. This design is intentional, it allows one descriptor to describe both the external and the internal wallet by deriving with `m/0/*` and `m/1/*` respectively. As long as we want to support derivation paths longer than 1 the `Wildcard`/`HardenedWildcard` distinction doesn't make any sense (what about e.g. `m/1'/42`? it's both hardened and unhardened).

darosior: I think we can for now explicitly fail because of hardened wildcard during parsing. What about:
```patch
diff --git a/src/descriptor/mod.rs b/src/descriptor/mod.rs
index 4abfcb2..07f0bb9 100644
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -240,6 +240,10 @@ impl DescriptorPublicKey {
                 if !is_wildcard && p == "*" {
                     is_wildcard = true;
                     None
+                } else if !is_wildcard && p == "*'" {
+                    Some(Err(DescriptorKeyParseError(
+                        "Hardened derivation is currently not supported.",
+                    )))
                 } else if is_wildcard {
                     Some(Err(DescriptorKeyParseError(
                         "'*' may only appear as last element in a derivation path.",
@@ -1457,6 +1461,15 @@ mod tests {
             ))
         );
 
+        // And ones which claim to be able to derive hardened child
+        let desc = "[78412e3a/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/42/*'";
+        assert_eq!(
+            DescriptorPublicKey::from_str(desc),
+            Err(DescriptorKeyParseError(
+                "Hardened derivation is currently not supported."
+            ))
+        );
+
         // And ones with misplaced wildcard
         let desc = "[78412e3a/44'/0'/0']xpub6ERApfZwUNrhLCkDtcHTcxd75RbzS1ed54G1LkBUHQVHQKqhMkhgbmJbZRkrgZw4koxb5JaHWkY4ALHY2grBGRjaDMzQLcgJvLJuZZvRcEL/1/*/44";
         assert_eq!(
```

sgeisler: Yeah, maybe we should also describe the deviation from core's behavior in the docs. It feels intuitive for me, but might not for others.

darosior: Definitely :-)
Would be great to have it in the release that Sanket and Andrew have been talking about on IRC as well

darosior:
Rebased on master, applied the explicitly-refuse-hardened-wildcards patch.
I've been using it downstream without issue.

src/descriptor/mod.rs
```diff
@@ -427,6 +680,31 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
...
+                    Result::Ok::<(Option<DescriptorPublicKey>, hash160::Hash), ()>((
+                        Some(child),
+                        child_hash,
+                    ))
+                }
+                None => Err(()),
+            },
+        )
+        .map_err(|_| Error::BadDescriptor)
+        //.expect("Translation fn can't fail.")
```
darosior, line 704: // Whoops

src/descriptor/mod.rs
```diff
@@ -427,6 +680,31 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
...
 }
 
+impl Descriptor<DescriptorPublicKey> {
+    /// Derives all wildcard keys in the descriptor using the supplied `path`
+    pub fn derive(
+        &self,
+        path: &[bip32::ChildNumber],
+    ) -> Result<Descriptor<DescriptorPublicKey>, Error> {
+        self.translate_pk(
+            |pk| Result::Ok::<DescriptorPublicKey, ()>(pk.clone().derive(path)),
```
darosior, line 690: Should it take `self` and not `clone()` ?

darosior (commented):
So, @sgeisler had a discussion with Sipa yesterday and it seems preferable for us to stick to `bitcoind`'s behaviour (and to original descriptors' purpose).

I'll change `derive(self..)` to take a single child number and not an entire derivation path.
However we could add in a follow-up a `WalletDescriptor` which would emulate [the behaviour described by Sebastian](https://github.com/rust-bitcoin/rust-miniscript/pull/131/files#r491191615) using multiple `Descriptor`s.

src/descriptor/mod.rs
```diff
@@ -68,6 +75,252 @@ pub enum Descriptor<Pk: MiniscriptKey> {
...
+                "Hardened derivation is currently not supported.",
+            ))
+        }
+    }
+
+    /// Derives a new key using the path if self is a wildcard xpub. Otherwise returns self.
+    ///
+    /// Panics if derivation path contains a hardened child number
+    pub fn derive(self, path: &[bip32::ChildNumber]) -> DescriptorPublicKey {
+        assert!(path.into_iter().all(|c| c.is_normal()));
```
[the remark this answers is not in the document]
sgeisler: Yeah, but there will still be an assert checking that either the child number is hardened or not too big (if you'd provide it as `u32` which I think would be reasonable).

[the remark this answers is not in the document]
darosior: Made it a debug assert as it's already "checked" at derivation and this line's only aim is to me being explicit for the reviewer

darosior:
Now we only derive descriptors from an actual child number and not a path anymore.

sgeisler (approved):
I think we reaches a pretty good state now (even though it doesn't do all I wanted anymore).

apoelstra:
With this fuzztesting code
```rust
fn do_test(data: &[u8]) {
    let data_str = String::from_utf8_lossy(data);
    if let Ok(dpk) = DescriptorPublicKey::from_str(&data_str) {
        let output = dpk.to_string();
        assert_eq!(data_str.to_lowercase(), output.to_lowercase());
    }
}
```
I can quickly get a crash by indexing into UTF-8 data which isn't ASCII.

Also, it would be good to re-export the `DescriptorPublicKey` type in lib.rs

apoelstra:
re adding a public key to `Miniscript`, I think this solution is really invasive and will break all existing code using `PkH`.

Why not just make the `MiniscriptKey::Hash` type for `DescriptorPublicKey` be `DescriptorPublicKey` again?

sgeisler:
That sounds like an awesome idea. The current version is a bit convoluted indeed, but I didn't see another option. That trick might work though.

darosior:
> That sounds like an awesome idea. The current version is a bit convoluted indeed, but I didn't see another option. That trick might work though.

It does work with a new placeholder type, is way shorter and nicer than my previous hack. Cleaning up, will push soon.

darosior:
For what it worth, still running on this branch downstream without issue :)

apoelstra (approved):
ack 28158a399978dc0c69ab1258ed0b4025d305c615

PR #136, opened 2020-09-29
[doc] psbt: bip is bip-0174

darosior:
micro-nit i just noticed while reading the implem.


sanket1729:
This was changed in this PR too, but that may take a while merge.

 https://github.com/rust-bitcoin/rust-miniscript/blob/c60f158343ddf87ba5dfe5df857b8c4dad748a3d/src/psbt/mod.rs#L18

sanket1729 (approved):

PR #137, opened 2020-09-30
psbt: satisfy any ToPublicKey key

darosior:
The satisfier was only implemented for `bitcoin::PublicKey` and could not be reimplemented for a more generalistic key downstream.

Also, it makes sense after #131 (i have a rebased version for my own needs here https://github.com/darosior/rust-miniscript/commits/satisfy_xpubs). All signatures are looked for by raw public key, and thus should be added to the mapping with `to_public_key()` (which derives the one pointed by the `DescriptorPublicKey` in #131).

Could have rebased it on #131 to add a test xpub-descriptor, but seemed overkill as the change is pretty straightforward.

### DIFF
--- a/src/psbt/mod.rs
+++ b/src/psbt/mod.rs
@@ -25,10 +25,10 @@ use bitcoin::util::psbt;
 use bitcoin::util::psbt::PartiallySignedTransaction as Psbt;
 use bitcoin::{self, secp256k1};
 
-use miniscript::{Legacy, Segwitv0};
 use BitcoinSig;
 use Miniscript;
 use Satisfier;
+use {Legacy, MiniscriptKey, Segwitv0, ToPublicKey};
 
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub enum Error {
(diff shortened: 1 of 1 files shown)


dr-orlovsky (approved):

sanket1729 (commented):
As discussed on IRC, we probably need to implement Satisfier on a structure that also has information for (nSequence, nLocktime) etc. The current implementation on `psbt::Input` does not have sufficient information to implement the satisfier for `after` and `older`. 

We require the complete psbt to get the nSequence, nLockTime fields.

sanket1729:
I guess this PR is independent of another issue, we can merge this

sanket1729 (approved):

apoelstra:
I'm actually gonna concept NACK this, because it plays badly with Rust's trait coherency rules. It prevents users from implementing `Satisfier` themselves on their types if those types also implement `MiniscriptKey`.

apoelstra:
@darosior why do you say it "could not be reimplemented for a more generalistic key downstream"?

darosior:
> @darosior why do you say it "could not be reimplemented for a more generalistic key downstream"?

IIRC: As it's implemented for `bitcoin::PublickKey` which i use and is `MiniscriptKey + ToPublicKey`, i had an error trying to implement it for a `<Pk: Miniscript + ToPublicKey>` as it was considered to reimplement a foreign trait.

apoelstra:
Right, if you try to implement it for all `Pk: Miniscript + ToPublicKey` the compiler will prevent it because it may cause breakage with your other dependencies which might have conflicting implementations ...it's legal for us to do it in rust-miniscript but I think it's still likely to be a problem for users.

sanket1729:
@apoelstra Why do you think that that downstream user would need another implementation of `Satisfier` for `psbt`? As far as I can see, there can only be one way of doing it?

apoelstra:
Oh, oops, I misread this PR. I thought we were implementing on `Pk` itself, not on `psbt::Input`.

darosior:
> concept ACK then :)

:)

More generally, what about we either:
- implement the default satisfier for every key, if downstream users want an exotic implementation they should use a newtype
- don't implement the satisfier at all, leaving the implementation design up to the user

As the middle ground seems confusing to me (you likely don't need a specific implem for just one key type but rather for all `MiniscriptKey + ToPublicKey` keys).

darosior:
Conceptually, i think so. Was not the original motivation though.

PR #138, opened 2020-10-01
Add Trivial and Unsatisfiable types to Concrete Policy

JeremyRubin:
It makes sense to have a corresponding type for trivial/unsatisfiable in the concrete compiler as it's quite useful when generating Policy from a higher level. This patch attempts to add these types.

See IRC:

```
[9/30/20 18:00] <jeremyrubin> Would anyone be opposed to adding a Unsatisifiable and Satisfied type to Policy. Context: sometimes when writing generic code to generate conditions, a component want to return "no further condition required" or "we don't want to authorize ever on this path". This is, of course, emulable with a higher order wrapper around the Policy type, but it seems simpler to add the types into Policy itself. There would be some identities, e.g.
[9/30/20 18:00] <jeremyrubin> , Or([A, Unsat]) = A, Or([A, Sat]) = Sat, And([A, Unsat]) = Unsat, And([A, Sat]) = A
[9/30/20 18:07] <jeremyrubin> In particular, it's inconvenient if these types don't exist because it's tricky to propagate an Unsat/Sat type -- e.g., if I construct an And tree, and one branch is Unsat, I should return an Unsat. But if I was expecting to plug the result of that And tree clause into e.g., an Or, I need to always manually handle these identities rather than having them baked in to Policy
[9/30/20 18:14] <jeremyrubin> In theory, you *could* use Sha256(0) to represent an unsatisfiable, and PK(G) to represent satisfiable BTW
[9/30/20 18:14] <jeremyrubin> maybe PK(G) is not good given the miniscript key abstraction
[9/30/20 18:15] <jeremyrubin> perhaps sha256(0) and sha256(H(0)) are better constants
[9/30/20 18:16] <jeremyrubin> My point is then I guess more that it might be nice to define these as constants and add a special compilation pass that simplifies them when detected.



[10/1/20 05:33] <andytoshi> jeremyrubin: i thought we had such a thing
[10/1/20 05:34] <andytoshi> https://docs.rs/miniscript/1.0.0/miniscript/policy/semantic/enum.Policy.html unsatisfiable and trivial
[10/1/20 05:35] <andytoshi> oh they don't exist on the compiler policy type
[10/1/20 05:36] <andytoshi> yeah i think they probably should, it does make composeability hard for those if there is no initial/terminal object
```

### DIFF
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -161,8 +161,11 @@ struct CompilerExtData {
 
 impl Property for CompilerExtData {
     fn from_true() -> Self {
-        // only used in casts. should never be computed directly
-        unreachable!();
+        CompilerExtData {
+            branch_prob: None,
+            sat_cost: 0.0,
+            dissat_cost: None,
+        }
     }
 
     fn from_false() -> Self {
@@ -811,6 +814,12 @@ where
     }
 
     match *policy {
+        Concrete::Unsatisfiable => {
+            insert_wrap!(AstElemExt::terminal(Terminal::False));
+        }
+        Concrete::Trivial => {
+            insert_wrap!(AstElemExt::terminal(Terminal::True));
+        }
         Concrete::Key(ref pk) => {
             insert_wrap!(AstElemExt::terminal(Terminal::PkH(
                 pk.to_pubkeyhash().clone()
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -37,6 +37,10 @@ use {Error, MiniscriptKey};
 /// to assist the compiler
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum Policy<Pk: MiniscriptKey> {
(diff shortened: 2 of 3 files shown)


sanket1729 (commented):
Looks good. Reached unreachable code. 
Should be changed to 
https://github.com/rust-bitcoin/rust-miniscript/blob/6af9cf7e55bcb2694d979401317e4331b75e2cbf/src/policy/compiler.rs#L165
```
        CompilerExtData {
            branch_prob: None,
            sat_cost: 0.0,
            dissat_cost: None,
        }
```

JeremyRubin:
fixed!

sanket1729 (approved):
Looks good. I wonder if we should have "1"/"0" instead of TRIVIAL and UNSATISFIABLE. 

But that is independent of this PR as we probably would need change that semantic policy too.

JeremyRubin:
rebased

apoelstra (approved):
ack da6ae53501c8e0e7182f80df15c95da74b17cf78

Issue #132, opened 2020-09-15
And()s compilation can produce invalid Script


darosior:
The translation of `thresh(n, n*policy)` to a concatenation of `and()` uncovered that we can actually compile policy that are creating invalid (non-standard) Scripts.

A small test to reproduce the behaviour with a 100-of-100 multisig (which cannot be spent by a standard transaction as the number on the stack would excess the `MAX_STANDARD_P2WSH_STACK_ITEMS`).
We can create (eg) 128-of-128 p2wsh as well, which are also invalid by the witness script size.

```patch
diff --git a/src/policy/compiler.rs b/src/policy/compiler.rs
index 31ea530..28d0bec 100644
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -1436,6 +1436,17 @@ mod tests {
             };
         }
     }
+
+    #[test]
+    fn test_limits() {
+        let (keys, _) = pubkeys_and_a_sig(100);
+
+        let pubkeys: Vec<Concrete<bitcoin::PublicKey>> =
+            keys.iter().map(|pubkey| Concrete::Key(*pubkey)).collect();
+        let big_thresh_ms: Result<SegwitMiniScript, _> =
+            Concrete::Threshold(100, pubkeys).compile();
+        assert_eq!(big_thresh_ms, Err(CompilerError::MaxOpCountExceeded));
+    }
 }
 
 #[cfg(all(test, feature = "unstable"))]
```

sanket1729:
I think we considered this but assumed that in Miniscript this limit cannot be reached as it would always exceed the op_code limit first. 
Anycase, the above reasoning should be revisited and documented.

sanket1729:
Opened an issue in #134 . Keeping this open until it resolved. 

The current documentation is unsatisfactory, I plan to do more documentation for (resource limits)#134 and (compiler)#128 next week.

darosior:
FWIW just got my new record: a 186 public keys script (with a path requiring a signature by 182 of them).

<details><summary>Click to expand the Script</summary>
<p>

```
[...]

PR #148, opened 2020-10-06
WIP: Error if we reach max stack items

darosior:
Had to break the API unfortunately, as if i'm not mistaking these are valid P2SH.

Note to self: we should also check for the max witness script size while at it.

Fixes https://github.com/rust-bitcoin/rust-miniscript/issues/132

### DIFF
--- a/src/miniscript/context.rs
+++ b/src/miniscript/context.rs
@@ -12,7 +12,9 @@
 // If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
 //
 
+use miniscript::types::extra_props::MAX_STANDARD_P2WSH_STACK_ITEMS;
 use std::fmt;
+use std::sync::Arc;
 use {Miniscript, MiniscriptKey, Terminal};
 
 /// Error for Script Context
@@ -31,6 +33,9 @@ pub enum ScriptContextError {
     /// Only Compressed keys allowed under current descriptor
     /// Segwitv0 fragments do not allow uncompressed pubkeys
     CompressedOnly,
+    /// At least one satisfaction path in the Miniscript fragment has more than
+    /// `MAX_STANDARD_P2WSH_STACK_ITEMS` (100) witness elements.
+    MaxWitnessItemssExceeded,
 }
 
 impl fmt::Display for ScriptContextError {
@@ -44,6 +49,11 @@ impl fmt::Display for ScriptContextError {
             ScriptContextError::CompressedOnly => {
                 write!(f, "Uncompressed pubkeys not allowed in segwit context")
             }
+            ScriptContextError::MaxWitnessItemssExceeded => write!(
+                f,
+                "At least one spending path in the Miniscript fragment has more \
+                 witness items than MAX_STANDARD_P2WSH_STACK_ITEMS.",
+            ),
         }
     }
 }
@@ -66,6 +76,8 @@ pub trait ScriptContext:
     /// Depending on script Context, some of the Terminals might not be valid.
     /// For example, in Segwit Context with MiniscriptKey as bitcoin::PublicKey
     /// uncompressed public keys are non-standard and thus invalid.
+    /// Requiring a too high number of stack elements for a satisfaction path
+    /// is another invalidity cause.
     /// Post Tapscript upgrade, this would have to consider other nodes.
     /// This does not recursively check
     fn check_frag_validity<Pk: MiniscriptKey, Ctx: ScriptContext>(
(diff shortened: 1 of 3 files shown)


sanket1729:
Thanks for tackling at this :) . 
Miniscript fragments may take a different number of elements for satisfaction and dissatisfaction. I think this will require keeping track of maximum possible satisfaction witness stack size and maximum possible dissatisfaction witness stack size at every node.

darosior:
I've a branch that does this, i can push it tomorrow.

Do you have a comment on the API break, it seems unevitable to me but i'm all hear if you have an idea :)

sanket1729:
Hi @darosior, I don't think we need any API changes for this. There are already context-specific checks in the compiler. The check for WSH scripts exceeding limits can go there too. 

https://github.com/rust-bitcoin/rust-miniscript/blob/b70042e5b29b077ad123b1af5f855a02d8a4b0f9/src/policy/compiler.rs#L654-L659. 

We can do this by adding checks to the ScriptContext trait and implementing those only for Segwitv0.

sanket1729:
Whenever the policy compilation fails, some possible compilations can be discarded because of opcode limit and some could be discarded because of stack size limit. 

I think this is a drawback as we cannot know whether the policy compilation failed due to stack size limit only, op-code limit only or a combination of both.

darosior:
I've pushed a new version which accounts for the sat cost in number of witness elements and check it for Segwitv0 in the compiler.

However, i wonder why not just using https://github.com/rust-bitcoin/rust-miniscript/blob/b70042e5b29b077ad123b1af5f855a02d8a4b0f9/src/miniscript/mod.rs#L179-L189

sanket1729:
We want to save information in types because that allows for efficient memoization and on the fly calculation for the values. That function is poor complexity for the uses of the compiler as it traverses the entire tree again.

I think in future the correct way might be to have a single function like `check_resource_limits(..)` in context.rs. And whenever we create a miniscript(either from compiler/from string/ or decoding a script), we invoke that check to check whether resource limits are respected. That way, we can be sure that miniscripts we are dealing with miniscript in the library they are spendable.

darosior:
Modified my last commit in a *much* nicer way (collect() on a map of `Result<>`) but which is invalid under 1.22 . Compiles fine under 1.29 (current rust-bitcoin MSRV) so i wonder if we'd like to do a MSRV-bumping release as in rust-bitcoin in which case i could use 1.29 here ?

darosior:
The bounds check (as well as a few other improvements as discussed on irc) is implemented in https://github.com/rust-bitcoin/rust-miniscript/pull/149 . Still refining the cost-tracking commits. Leaving this open only for the above questions :-)

sanket1729:
> Modified my last commit in a much nicer way (collect() on a map of Result<>) but which is invalid under 1.22 . Compiles fine under 1.29 (current rust-bitcoin MSRV) so i wonder if we'd like to do a MSRV-bumping release as in rust-bitcoin in which case i could use 1.29 here?

Yes, we should be getting 1.29 MSRV soon. So there is no need to push special effort into resolving this. 

> Why is the Multi fragment (di)satisfaction cost relative to the threshold (k) and not the number of keys (n) here?

The satisfactions require 1 dummy + `k` sigs for sure.
I did not test dissatisatisfaction one yet, but my review of the code tells me that it should be `k+1` instead of `n+1`. https://github.com/bitcoin/bitcoin/blob/283a73d7eaea2907a6f7f800f529a0d6db53d7a6/src/script/interpreter.cpp#L1109-L1116 .
I think the miniscript website has this written incorrectly. 

> Can the dissat cost of ZeroNotEqual ([X] 0NOTEQUAL) be of 1 ? Just push OP_0 ?

No, dissatisfaction means we need `0` on the top of the stack. To dissatisfy ZeroNotEqual, we need to feed it whatever it requires to dissat [X] as that will output zero, and 0NOTEQUAL will keep the `0` as is thus completely dissatisfying the expression. For example, consider [X] = thresh_m(k, ....), the dissat here would require k+1(as mentioned above) 0's.

> Why is the dissat cost of OrC None ? (And not like OP_1)

`or_c` is `[X] NOTIF [Z] ENDIF` and the typing rules require [Z] to be of type `V`. `V` is "verify" like expressions that do not put anything on the stack and cannot be dissatisfied. (e.g CHECKSIGVERIFY). And thus this expression as a whole can never be dissatisfied. 

> Why does the current implementation panics if both branches of a fragment are dissatisfiable ? For example with OrI:

This is a tricky one. I can explain try to the rationale for panic, but reasoning a bit involved. `or_i` is a like a conditional in regular programming languages we only execute either `X` or `Z` of `or_i(X, Z)`. If any satisfaction of miniscripts, requires a dissatisfaction of or_i whose both children can be dissatisfied, then there is a malleability possibility. Consider the following (invalid)miniscript, we will get into a bit why that is invalid. 
`or_d(or_i(pk(A),pk(B)),pk(C))` representing the semantics of 1 of 3 multisig. If we want to spend using the key C, we have two possible satisfactions `[1 , empty_sig_A, sig_C]` or `[0, empty_sig_B, C]` thereby breaking the malleability guarantees of miniscript. 

At a technical level, the typing rules make sure that if `or_i()` wants to dissatisfied, exactly one of the children should be dissatisfiable. 
This is governed by the rule,  `e = ex fz or ez fx`. `f` denotes forced (dissatisfaction requires a signature). Since, anyone can dissatisfy `pk(A)` and `pk(B)`(dissatisfaction is empty vector),  neither of them are f. Therefore, the expression does not `e` property where `e` means unique dissatisfaction without knowledge of sigs etc. Finally, the `s` rule (malleability rule) rule for `or_d` requires that first fragment to be `e` which is not. Therefore, such a miniscript is not safe and thus invalid. 

We make sure to check the top-level malleability whenever we create miniscripts in the library, therefore users should not be able to create such miniscripts(parser will error, compiler will not find them and fromStr will fail). Thus, it is safe to panic here!.

darosior:
Thank you for the detailed answer.

>  The satisfactions require 1 dummy + k sigs for sure.

Yeah, but `n` witness stack elements (empty byte arrays for non-needed ones).

> I did not test dissatisatisfaction one yet, but my review of the code tells me that it should be k+1 instead of n+1. https://github.com/bitcoin/bitcoin/blob/283a73d7eaea2907a6f7f800f529a0d6db53d7a6/src/script/interpreter.cpp#L1109-L1116 .

Yeah, that what i thought as well. Will go this way.

>  No, dissatisfaction means we need 0 on the top of the stack. To dissatisfy ZeroNotEqual, we need to feed it whatever it requires to dissat [X] as that will output zero, and 0NOTEQUAL will keep the 0 as is thus completely dissatisfying the expression. For example, consider [X] = thresh_m(k, ....), the dissat here would require k+1(as mentioned above) 0's.

I don't get it. How comes that feeding `0` to `0NOTEQUAL` doesn't dissatisfy it ?

> Therefore, such a miniscript is not safe and thus invalid.

Ok, got it. Thanks!

> We make sure to check the top-level malleability whenever we create miniscripts in the library, therefore users should not be able to create such miniscripts(parser will error, compiler will not find them and fromStr will fail). Thus, it is safe to panic here!.

I wonder how this plays out with the new checks introduced in https://github.com/rust-bitcoin/rust-miniscript/pull/149 .

darosior:
Closing this, think i'll still be mentioned if you answer anyways.

sanket1729:
> How comes that feeding 0 to 0NOTEQUAL doesn't dissatisfy it ?
We want to dis-satisfy the entire wrapped miniscript not only the fragment. 

How would you dissatisfy 
` 2 pk1 pk2 2 CHECKMULSIG 0NOTEQUAL`, clearly, you need two empty sigs.

darosior:
> How would you dissatisfy 2 pk1 pk2 2 CHECKMULSIG 0NOTEQUAL, clearly, you need two empty sigs.

But in this case that's `2 pk1 pk2 2 CHECKMULSIG` dissatisfaction ?? Moreover the two empty sigs you mention would make CMS push 0 on the stack.

Sorry to insist, but that's interesting and i want to be sure with my understanding.

PR #149, opened 2020-10-07
Check for segwit standardness limits

### LINKED ISSUE #132: And()s compilation can produce invalid Script

darosior:
This is the first part of #148, in order to fix #132.

This comports:
- a few drive-by cleanup commits added as i was reading the cost computation internals
- an API break: change max_witness_elements() to return an Option instead of unwrap()ing internally
- an addition of bounds-check for segwit standardness rules under the Segwitv0 context
- a move of the OPs count check to context in order to also check it at parse time (not only compilation) 

The second part is -coming soon :tm: and- storing  the cost in witness elements in the ExtraData struct in order to avoid the suboptimal recursion at each check.

### DIFF
--- a/src/miniscript/astelem.rs
+++ b/src/miniscript/astelem.rs
@@ -848,70 +848,79 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
     ///
     /// This number does not include the witness script itself, so 1 needs
     /// to be added to the final result.
-    pub fn max_satisfaction_witness_elements(&self) -> usize {
+    pub fn max_satisfaction_witness_elements(&self) -> Option<usize> {
         match *self {
-            Terminal::PkK(..) => 1,
-            Terminal::PkH(..) => 2,
-            Terminal::After(..) | Terminal::Older(..) => 0,
+            Terminal::PkK(..) => Some(1),
+            Terminal::PkH(..) => Some(2),
+            Terminal::After(..) | Terminal::Older(..) => Some(0),
             Terminal::Sha256(..)
             | Terminal::Hash256(..)
             | Terminal::Ripemd160(..)
-            | Terminal::Hash160(..) => 1,
-            Terminal::True => 0,
-            Terminal::False => 0,
+            | Terminal::Hash160(..) => Some(1),
+            Terminal::True => Some(0),
+            Terminal::False => None,
             Terminal::Alt(ref sub) | Terminal::Swap(ref sub) | Terminal::Check(ref sub) => {
                 sub.node.max_satisfaction_witness_elements()
             }
-            Terminal::DupIf(ref sub) => 1 + sub.node.max_satisfaction_witness_elements(),
+            Terminal::DupIf(ref sub) => sub.node.max_satisfaction_witness_elements().map(|x| x + 1),
             Terminal::Verify(ref sub)
             | Terminal::NonZero(ref sub)
             | Terminal::ZeroNotEqual(ref sub) => sub.node.max_satisfaction_witness_elements(),
-            Terminal::AndV(ref l, ref r) | Terminal::AndB(ref l, ref r) => {
-                l.node.max_satisfaction_witness_elements()
-                    + r.node.max_satisfaction_witness_elements()
-            }
+            Terminal::AndV(ref l, ref r) | Terminal::AndB(ref l, ref r) => l
+                .node
+                .max_satisfaction_witness_elements()
+                .and_then(|l| r.node.max_satisfaction_witness_elements().map(|r| l + r)),
             Terminal::AndOr(ref a, ref b, ref c) => cmp::max(
-                a.node.max_satisfaction_witness_elements()
-                    + c.node.max_satisfaction_witness_elements(),
-                a.node.max_dissatisfaction_witness_elements().unwrap()
-                    + b.node.max_satisfaction_witness_elements(),
+                a.node
+                    .max_satisfaction_witness_elements()
+                    .and_then(|a| c.node.max_satisfaction_witness_elements().map(|c| c + a)),
+                a.node
+                    .max_dissatisfaction_witness_elements()
+                    .and_then(|a| b.node.max_satisfaction_witness_elements().map(|b| a + b)),
             ),
             Terminal::OrB(ref l, ref r) => cmp::max(
-                l.node.max_satisfaction_witness_elements()
-                    + r.node.max_dissatisfaction_witness_elements().unwrap(),
-                l.node.max_dissatisfaction_witness_elements().unwrap()
-                    + r.node.max_satisfaction_witness_elements(),
+                l.node
+                    .max_satisfaction_witness_elements()
+                    .and_then(|l| r.node.max_dissatisfaction_witness_elements().map(|r| l + r)),
+                l.node
+                    .max_dissatisfaction_witness_elements()
+                    .and_then(|l| r.node.max_satisfaction_witness_elements().map(|r| r + l)),
             ),
             Terminal::OrD(ref l, ref r) | Terminal::OrC(ref l, ref r) => cmp::max(
                 l.node.max_satisfaction_witness_elements(),
-                l.node.max_dissatisfaction_witness_elements().unwrap()
-                    + r.node.max_satisfaction_witness_elements(),
+                l.node
+                    .max_dissatisfaction_witness_elements()
+                    .and_then(|l| r.node.max_satisfaction_witness_elements().map(|r| r + l)),
             ),
-            Terminal::OrI(ref l, ref r) => {
-                1 + cmp::max(
-                    l.node.max_satisfaction_witness_elements(),
-                    r.node.max_satisfaction_witness_elements(),
-                )
-            }
+            Terminal::OrI(ref l, ref r) => cmp::max(
+                l.node.max_satisfaction_witness_elements(),
+                r.node.max_satisfaction_witness_elements(),
+            )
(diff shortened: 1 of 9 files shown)


src/miniscript/astelem.rs
```diff
@@ -848,70 +848,79 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
...
+            Terminal::After(..) | Terminal::Older(..) => Some(0),
             Terminal::Sha256(..)
             | Terminal::Hash256(..)
             | Terminal::Ripemd160(..)
-            | Terminal::Hash160(..) => 1,
-            Terminal::True => 0,
-            Terminal::False => 0,
+            | Terminal::Hash160(..) => Some(1),
+            Terminal::True => Some(0),
+            Terminal::False => Some(0),
```
sanket1729, line 861: This was a bug in previous code too. I think this should be None.

darosior: Yeah definitely, it is in my "track costs" branch and in Sipa's implementation.

darosior:
I still need to add some tests. Some manual testing uncovered a panic on Descriptor's `max_witness_elements()`.

darosior:
Amended to make `Descriptor`'s `max_satisfaction_weight` return an Option as well as mentionned by @sanket1729 [here](https://github.com/rust-bitcoin/rust-miniscript/pull/150#discussion_r502440813). It makes sense as we can't really know if we are given a valid Miniscript.

sanket1729:
@darosior, the typechecking rules for and_or would make sure to accept only fragments such that the first part is dissatifiable. I don't know if the above comment is resolved and I am not able to follow the exact question there.

darosior:
> I don't know if the above comment is resolved and I am not able to follow the exact question there.

Looks like you just answered it. Thanks. This is now ready for review.

src/miniscript/context.rs
```diff
@@ -111,12 +134,31 @@ impl ScriptContext for Segwitv0 {
...
+                    return Err(ScriptContextError::CompressedOnly);
+                }
+
+                Ok(())
             }
             _ => Ok(()),
         }
     }
+
+    fn check_ms_validity<Pk: MiniscriptKey, Ctx: ScriptContext>(
```
sanket1729, line 172: As discussed in #148, we should combine check_ms_validity and check_frag_validity

src/miniscript/context.rs
```diff
@@ -53,6 +56,11 @@ impl fmt::Display for ScriptContextError {
                 "At least one spending path in the Miniscript fragment has more \
                  witness items than MAX_STANDARD_P2WSH_STACK_ITEMS.",
             ),
+            ScriptContextError::MaxOpCountExceeded => write!(
+                f,
+                "At least one satisfaction path in the Miniscript fragment contains \
+                 more than `MAX_OPS_PER_SCRIPT`(201) opcodes."
```
sanket1729, line 67: nit: use variable MAX_OPS_PER_SCRIPT variable in format string instead of hardcoded 201.

sanket1729: Or you can remove 201 completely. Anything works

sanket1729 (commented):
I think this is almost good to go. Left some minor feedback

src/miniscript/context.rs
darosior: Just implemented it and moving `check_frag_validity` to take a Miniscript reduces the number of call sites. Eg we cannot call it anymore in https://github.com/rust-bitcoin/rust-miniscript/blob/b70042e5b29b077ad123b1af5f855a02d8a4b0f9/src/miniscript/astelem.rs#L63-L75

darosior: Actually, for all Terminal-only methods. That's the reason why i added another method in the first place: check_frag is really Terminal-only while check_ms is Miniscript-only.
At very first (#148), i combined both into check_frag and took a Terminal: it was *really* dirty.

src/miniscript/context.rs
darosior: I don't expect this value to change much anyways :p
That said, removed it and the backticks from the error message as it makes it easier to match errors against it.

darosior:
Fixed the error messages.

src/miniscript/context.rs
sanket1729: There is a cheap function to get Miniscript from terminals. `Miniscript::from_ast`. 
I made a commit on top of this PR that combines the checks:
https://github.com/sanket1729/rust-miniscript/commit/4311108698c3590d0f587391be5f5eee8e5e5cee

darosior: Thanks!

darosior:
Cherry-picked @sanket1729 commit on top of this.

sanket1729:
@darosior, looks like I did not test my commit with compiler feature and forgot to update `compiler.rs` that is causing the test failures. Can you fix the commit to remove the `Ctx::check_frag..` from `compiler.rs`?

darosior:
Amended..

sanket1729 (approved):
tAck e2baa3c9b484d0168eb1d7310c8d5474f892126a

src/miniscript/types/extra_props.rs
```diff
@@ -93,9 +93,8 @@ impl TimeLockInfo {
     }
 }
 
-/// Structure representing the extra type properties of a fragment which are
-/// relevant to legacy(pre-segwit) safety and fee estimation. If a fragment is
-/// used in pre-segwit transactions it will only be malleable but still is
+/// Structure representing the extra type properties of a fragment. If a fragment
+/// is used in pre-segwit transactions it will only be malleable but still is
```
apoelstra, line 102: I think we should delete the second sentence. I can't see how it's related to this structure.

src/miniscript/mod.rs
```diff
@@ -186,8 +186,8 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
     /// This function may panic on misformed `Miniscript` objects which do
     /// not correspond to semantically sane Scripts. (Such scripts should be
     /// rejected at parse time. Any exceptions are bugs.)
```
apoelstra, line 188: We should fix the comment now that this doesn't panic

src/miniscript/context.rs
```diff
@@ -111,12 +134,31 @@ impl ScriptContext for Segwitv0 {
         frag: &Terminal<Pk, Ctx>,
```
apoelstra: If this is only checking a terminal, the function name should have `terminal` rather than `frag` in it.

src/miniscript/context.rs
apoelstra, line 142: ...and this one should be named `check_frag_validity`

apoelstra: I wonder if the sanity checks in `from_ast` should be moved into `check_frag_validity`, and then `from_ast` can be replaced by an `impl From<Terminal> for Miniscript`

apoelstra:
ack e2baa3c9b484d0168eb1d7310c8d5474f892126a

We can fix comments/renames in a followup PR

apoelstra:
Not sure what's up with CI here

apoelstra:
Ah, it's because the PR is based on `master` prior to us updating the github actions.

I'll go ahead and merge this, since it passed the 1.22.0 check, so clearly it must've passed the 1.29.0 one.

darosior:
> ack e2baa3c
> We can fix comments/renames in a followup PR

Will do a follow-up, or add it at part of  #150

src/miniscript/decode.rs
```diff
@@ -173,6 +173,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> TerminalStack<Pk, Ctx> {
         F: FnOnce(Arc<Miniscript<Pk, Ctx>>) -> Terminal<Pk, Ctx>,
     {
         let top = self.pop().unwrap();
+        Ctx::check_ms_validity(&top)?;
```
apoelstra: Oops! It turns out that this can panic if we call it on malformed input, because it tries to count the max_dissatisfaction_witness_elements but we still have a panic there https://github.com/rust-bitcoin/rust-miniscript/blob/master/src/miniscript/astelem.rs#L774

This caused a fuzztest failure on another PR.

Issue #56, opened 2019-10-08
Feature request: malleable signer


apoelstra:
If the non-malleable signer fails but you've gotta move your coins, you shouldn't be stuck

sanket1729:
Fixed by #167

PR #150, opened 2020-10-08
Non-recursive bound checks for Miniscript fragments

darosior:
This is based on #149 and keeps track of the maximum number of stack elements and the size used by both a satisfaction and a dissatisfaction directly in the Miniscript fragment extra data.
This avoids the recursive methods currently called *each time* we check for limits.

TODO:
- [x] implement it for satisfaction size as well (i'm using a tuple for the cost differing between witness and scriptSig)
- [x] replace callers to use these instead of recursive methods
- [x] fix the (likely compiler-related) error in #149
- [x] cleanup the FIXMEs which are basically reminder for me to ask you 

Moved from TODO to "can be done in a follow-up this one is large enough":
- add new method to Miniscript to get both the dissatisfaction costs in size and stack elements


src/miniscript/types/extra_props.rs
```diff
@@ -340,6 +382,9 @@ impl Property for ExtData {
             ops_count_static: self.ops_count_static + 1,
             ops_count_sat: self.ops_count_sat.map(|x| x + 1),
             ops_count_nsat: self.ops_count_nsat.map(|x| x + 1),
+            stack_elem_count_sat: self.stack_elem_count_sat,
+            // FIXME: Why not Some(1) ?
+            stack_elem_count_dissat: self.stack_elem_count_dissat,
```
darosior, line 387: This should be None.

src/miniscript/types/extra_props.rs
```diff
@@ -351,6 +396,8 @@ impl Property for ExtData {
             ops_count_static: self.ops_count_static,
             ops_count_sat: self.ops_count_sat,
             ops_count_nsat: None,
+            stack_elem_count_sat: self.stack_elem_count_sat,
+            stack_elem_count_dissat: self.stack_elem_count_dissat,
```
darosior, line 400: This too.

src/miniscript/types/extra_props.rs
```diff
@@ -404,6 +461,12 @@ impl Property for ExtData {
             ops_count_static: l.ops_count_static + r.ops_count_static,
             ops_count_sat: l.ops_count_sat.and_then(|x| r.ops_count_sat.map(|y| x + y)),
             ops_count_nsat: None,
+            stack_elem_count_sat: l
+                .stack_elem_count_sat
+                .and_then(|l| r.stack_elem_count_sat.map(|r| l + r)),
+            stack_elem_count_dissat: l
+                .stack_elem_count_dissat
+                .and_then(|l| r.stack_elem_count_dissat.map(|r| l + r)),
```
darosior, line 469: This too !

src/descriptor/mod.rs
```diff
@@ -619,6 +619,7 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
     /// transaction. Assumes all signatures are 73 bytes, including push opcode
     /// and sighash suffix. Includes the weight of the VarInts encoding the
     /// scriptSig and witness stack length.
+    /// Panics on malformed Miniscript.
```
sanket1729, line 622: I think we need to change this to option too. There are miniscripts that have no satisfactions. ex: False, or and(pk, False).

darosior: It's changed in #152, maybe i should merge the 2 PRs as it also "fixes" (hides?) the compiler test failure.

src/miniscript/context.rs
```diff
@@ -94,6 +111,12 @@ impl ScriptContext for Legacy {
     ) -> Result<(), ScriptContextError> {
         Ok(())
     }
+
+    fn check_ms_validity<Pk: MiniscriptKey, Ctx: ScriptContext>(
```
sanket1729, line 138: I think `check_ms_validity` and `check_frag_validity` can be merged. The only thing the `check_frag_validity` is doing is checking uncompressed public keys in Segwit Context.

src/miniscript/decode.rs
```diff
@@ -195,6 +196,9 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> TerminalStack<Pk, Ctx> {
         let left = self.pop().unwrap();
         let right = self.pop().unwrap();
 
+        Ctx::check_ms_validity(&left)?;
+        Ctx::check_ms_validity(&right)?;
+
```
sanket1729, line 201: We should have checks on the wrapped_ms instead of children. It is possible that the children did not reach stack limit, but the combination did. 

Similar to how it is done for check_frag_validity.

src/miniscript/context.rs
sanket1729: There also needs to be check in the FromStr for Miniscript. But I think if we combine both of these, the call sites are automatically covered.

darosior: > I think check_ms_validity and check_frag_validity can be merged. The only thing the check_frag_validity is doing is checking uncompressed public keys in Segwit Context.

I thought too and actually that's how i implemented it at first. Turns out it's quite messy (and costy) as we need to match the terminal and duplicate the logic, better to call it on the actual top miniscript.

src/miniscript/types/extra_props.rs
```diff
@@ -179,6 +191,8 @@ impl Property for ExtData {
             ops_count_static: 1,
             ops_count_sat: Some(n + 1),
             ops_count_nsat: Some(n + 1),
+            stack_elem_count_sat: Some(n + 1),
+            stack_elem_count_dissat: Some(n + 1),
```
sanket1729, line 212: Should be `k+1` for sat and dissat both?

src/miniscript/types/extra_props.rs
```diff
@@ -344,6 +382,8 @@ impl Property for ExtData {
             ops_count_static: self.ops_count_static + 1,
             ops_count_sat: self.ops_count_sat.map(|x| x + 1),
             ops_count_nsat: self.ops_count_nsat.map(|x| x + 1),
+            stack_elem_count_sat: self.stack_elem_count_sat,
+            stack_elem_count_dissat: None,
```
sanket1729, line 430: should be self.stack_elem_count_dissat

src/miniscript/types/extra_props.rs
```diff
@@ -458,6 +531,13 @@ impl Property for ExtData {
                     .and_then(|x| l.ops_count_nsat.map(|y| y + x + 2)),
             ),
             ops_count_nsat: None,
+            stack_elem_count_sat: cmp::max(
+                l.stack_elem_count_sat,
+                l.stack_elem_count_dissat
+                    .and_then(|l_dis| r.stack_elem_count_sat.map(|r_sat| r_sat + l_dis)),
+            ),
+            // FIXME: why??
+            stack_elem_count_dissat: None,
```
sanket1729, line 620: nit: FIXME can be removed. The typing rules ensure that output of or_c is V type. (cannot be dissatisfied)

src/miniscript/types/extra_props.rs
```diff
@@ -476,6 +556,18 @@ impl Property for ExtData {
                 (_, Some(x)) | (Some(x), _) => Some(x + 3),
                 (None, None) => None,
             },
+            stack_elem_count_sat: match (l.stack_elem_count_sat, r.stack_elem_count_sat) {
+                (Some(l), Some(r)) => Some(1 + cmp::max(l, r)),
```
sanket1729, line 646: Currently, the library only has support for non-malleable scripts. So, it is safe to transform this into unreachable! . In future(I doubt) if we want to support malleable miniscripts, we would need this. 

In any case, no harm in having this.

src/miniscript/types/extra_props.rs
```diff
@@ -494,6 +586,18 @@ impl Property for ExtData {
...
+                    .and_then(|a| c.stack_elem_count_sat.map(|c| c + a)),
+                a.stack_elem_count_dissat
+                    .and_then(|a_dis| b.stack_elem_count_sat.map(|b| b + a_dis)),
+            ),
+            stack_elem_count_dissat: cmp::max(
+                a.stack_elem_count_sat
+                    .and_then(|a| c.stack_elem_count_dissat.map(|c| c + a)),
+                a.stack_elem_count_dissat
+                    .and_then(|a_dis| b.stack_elem_count_dissat.map(|b| b + a_dis)),
+            ),
```
sanket1729, line 699: We should not add non-cannonical solutions here. We should remove first element in max because our signer never takes that path. Maybe if there is some demand for #56 (I doubt), we should keep things simple and only allow canonical options.  

See section malleability from http://bitcoin.sipa.be/miniscript/ for details for canonical vs non-cannonical if you are interested(not required for this PR).

src/miniscript/types/extra_props.rs
```diff
@@ -553,6 +668,10 @@ impl Property for ExtData {
             ops_count_sat: ops_count_sat
                 .map(|x: usize| (x + (n - 1) + 1 + (sum + ops_count_nsat_sum as i32) as usize)), //adds and equal
             ops_count_nsat: ops_count_nsat.map(|x| x + (n - 1) + 1), //adds and equal
+            // For simplicity we only multiply the worse cost seen
```
sanket1729, line 789: I think we should calculate the explicit values.
Sort by max_elem_sat - max_elem_dissat and pick the indices corresponding to max k. Add those k satisfaction costs with other `n-k` dissat costs. The same algorithm is used `max_satisfaction_witness_elements()`

src/miniscript/types/extra_props.rs
```diff
@@ -553,6 +668,10 @@ impl Property for ExtData {
...
+            stack_elem_count_sat: worse_stack_elem_count_sat.and_then(|elems| Some(k * elems)),
```
sanket1729, line 790: For dissat, it is the sum of all dissat.

src/miniscript/types/extra_props.rs
```diff
@@ -193,6 +210,9 @@ impl Property for ExtData {
             ops_count_nsat: Some(n + 1),
             stack_elem_count_sat: Some(n + 1),
             stack_elem_count_dissat: Some(n + 1),
+            // FIXME: What about the cost of the empty byte array ?
```
sanket1729, line 213: nit: remove fixme

src/miniscript/types/extra_props.rs
```diff
@@ -117,6 +117,15 @@ pub struct ExtData {
     pub stack_elem_count_sat: Option<usize>,
     /// The worst case number of stack elements for dissatisfying this Miniscript fragment.
     pub stack_elem_count_dissat: Option<usize>,
+    /// Maximum size, in bytes, of a satisfying witness. First elements is the cost for the
+    /// witness stack, the second one is the cost for scriptSig.
+    /// All signatures are assumed to be 73 bytes in size, including the
+    /// length prefix (segwit) or push opcode (pre-segwit) and sighash
+    /// postfix.
+    pub max_sat_size: Option<(usize, usize)>,
```
sanket1729, line 125: Putting out some random thought here. I have not completely evaluated the solution. Since we force the user to choose Miniscript Context (via descriptor) or by explicitly specifying it, we have the information whether the Miniscrpit is Segwit or not. Would it be possible to have Ctx information in Property and a function one_cost() in ScriptContext trait? That should simplify this code.

src/miniscript/types/extra_props.rs
```diff
@@ -211,6 +231,8 @@ impl Property for ExtData {
             ops_count_nsat: None,
             stack_elem_count_sat: Some(1),
             stack_elem_count_dissat: Some(1),
+            max_sat_size: Some((33, 33)),
+            max_dissat_size: None,
```
sanket1729, line 235: Should be 33. Same for other hashes

src/miniscript/types/extra_props.rs
```diff
@@ -224,6 +246,8 @@ impl Property for ExtData {
...
```
sanket1729, line 250: Should be 33. Same for other hashes

src/miniscript/types/extra_props.rs
```diff
@@ -237,6 +261,8 @@ impl Property for ExtData {
...
```
sanket1729, line 265: Should be 33. Same for other hashes

src/miniscript/types/extra_props.rs
```diff
@@ -384,6 +428,8 @@ impl Property for ExtData {
             ops_count_nsat: self.ops_count_nsat.map(|x| x + 1),
             stack_elem_count_sat: self.stack_elem_count_sat,
             stack_elem_count_dissat: None,
+            max_sat_size: self.max_sat_size,
+            max_dissat_size: None,
```
sanket1729, line 432: should be self.max_dissat_size

src/miniscript/types/extra_props.rs
```diff
@@ -415,6 +463,8 @@ impl Property for ExtData {
             ops_count_nsat: Some(self.ops_count_static + 3),
             stack_elem_count_sat: self.stack_elem_count_sat.map(|x| x + 1),
             stack_elem_count_dissat: self.stack_elem_count_dissat.map(|x| x + 1),
+            max_sat_size: self.max_sat_size.map(|(w, s)| (w + 2, s + 1)),
+            max_dissat_size: self.max_dissat_size.map(|(w, s)| (w + 2, s + 1)),
```
sanket1729, line 467: w + 1 because it is dis-satisfied by 0.

src/miniscript/types/extra_props.rs
```diff
@@ -428,6 +478,9 @@ impl Property for ExtData {
             ops_count_nsat: Some(self.ops_count_static + 3),
             stack_elem_count_sat: self.stack_elem_count_sat.map(|x| x + 1),
             stack_elem_count_dissat: self.stack_elem_count_dissat.map(|x| x + 1),
+            // +1 for the witness too as it's an empty byte array!
+            max_sat_size: self.max_sat_size.map(|(w, s)| (w + 1, s + 1)),
+            max_dissat_size: self.max_dissat_size.map(|(w, s)| (w + 1, s + 1)),
```
sanket1729, line 483: should be w + 2

src/miniscript/types/extra_props.rs
```diff
@@ -672,6 +787,9 @@ impl Property for ExtData {
             stack_elem_count_sat: worse_stack_elem_count_sat.and_then(|elems| Some(k * elems)),
             stack_elem_count_dissat: worse_stack_elem_count_dissat
                 .and_then(|elems| Some(k * elems)),
+            // Same for cost in size
```
sanket1729, line 793: Again, this can be improved to as mentioned previously.

src/miniscript/types/extra_props.rs
```diff
@@ -659,10 +659,13 @@ impl Property for ExtData {
                 r.max_sat_size.and_then(|(w, s)| Some((w + 1, s + 1))),
             ),
             max_dissat_size: match (l.max_dissat_size, r.max_dissat_size) {
+                (Some(l), Some(r)) => {
+                    let max = cmp::max(l, r);
+                    Some((1 + max.0, 1 + max.1))
+                }
                 (None, Some(r)) => Some((1 + r.0, 1 + r.1)),
-                (Some(l), None) => Some((2 + l.0, 1 + l.1)),
+                (Some(l), None) => Some((1 + l.0, 1 + l.1)),
```
sanket1729, line 667: Should this not be 2?

sanket1729 (commented):
Thanks for doing this :)  There are some breaking API changes here, but they are worth it for sure.

sanket1729:
It looks like one of tests was broken :) . https://github.com/rust-bitcoin/rust-miniscript/pull/150/checks?check_run_id=1226708928

src/miniscript/types/extra_props.rs
darosior: Fun thing is that it makes a lot of tests panic in the current (not correct) state. Will try out after applying all the corrections.

src/miniscript/types/extra_props.rs
darosior: Definitely interested, will look into it.

src/miniscript/types/extra_props.rs
darosior: Yeah, it would greatly simplify this commit full of nested `map()` and `and_then()`.

Familiarizing with the internals i thought that we could start stuffing quite a lot of things in the Ctx :sweat_smile:

src/miniscript/types/extra_props.rs
darosior: I thought hashes were not dissatisfiable ? Sipa implemented this https://github.com/sipa/miniscript/blob/027b422495a8ee3ea3fa7787150389bb65d1b9de/bitcoin/script/miniscript.h#L535-L538 and iirc we did too.

There is, however, a mistake on the `stack_elem_count_dissat` in this case.

src/miniscript/types/extra_props.rs
darosior: Yep, as you pass on the first branch of the IF to dissatisfy (?) (same as your comment wrt to `cast_likely` above).

src/miniscript/decode.rs
darosior: Actually, we can't. We need to call it on a Miniscript not a Terminal.

src/miniscript/types/extra_props.rs
darosior: Why do you compute the OP count cost this way ?
```rust
match (sub.ops_count_sat, sub.ops_count_nsat) {
                (Some(x), Some(y)) => {
                    ops_count_sat_vec.push(Some(x as i32 - y as i32));
```

darosior: > I think we should calculate the explicit values.

FWIW i don't think it's tractable to have an accurate value, but yeah we can have a better approximation..

src/miniscript/types/extra_props.rs
darosior: I dropped a fixme for this :/

src/miniscript/types/extra_props.rs
darosior: Ok, even after the fix it still makes everything panic.. Didn't look into it yet.

src/miniscript/types/extra_props.rs
darosior: Ok, so i think your hash dissatisfaction below are a clear example of non-canonical dissatisfaction :p

src/descriptor/mod.rs
darosior: Added in #149

src/miniscript/types/extra_props.rs
darosior: Made `stack_elem_count_dissat` None, as i really think the mistake is this way around.

src/miniscript/types/extra_props.rs
sanket1729: Sorry, I commented at the wrong place. This should be dissat

src/miniscript/types/extra_props.rs
```diff
@@ -472,6 +639,31 @@ impl Property for ExtData {
...
                 (None, None) => None,
             },
+            stack_elem_count_sat: match (l.stack_elem_count_sat, r.stack_elem_count_sat) {
+                (Some(l), Some(r)) => Some(1 + cmp::max(l, r)),
+                (Some(l), None) => Some(1 + l),
+                (None, Some(r)) => Some(1 + r),
+                (None, None) => None,
+            },
+            stack_elem_count_dissat: match (l.stack_elem_count_dissat, r.stack_elem_count_dissat) {
+                (Some(l), Some(r)) => Some(1 + cmp::max(l, r)),
```
sanket1729, line 649: This can be changed to unreachable!(), my bad!

src/miniscript/types/extra_props.rs
sanket1729: This should be as it is. https://github.com/rust-bitcoin/rust-miniscript/pull/150/files#r502990941 can be changed

src/miniscript/types/extra_props.rs
sanket1729: Those are not non-cannonical dissatisfactions. Non-cannonical ones are the ones require satisfaction of one fragment to dissatisfy another fragment. dissatisfaction hash check does not require any satisfaction, so it is not non-cannonical. 

But you are right, if any top level satisfaction of miniscript requires dissatisfaction of hashpreimages, then it is malleable. So, those can be safely changed to `None`.

I will bring up the issue about "How much effort we want to put into dealing with malleable scripts" with Andrew to also get his feedback. But that is an independent issue, and you are correct to change hash to None.

src/miniscript/types/extra_props.rs
sanket1729: Agreed

src/miniscript/decode.rs
sanket1729: Terminal and Miniscript are equivalent. Miniscript is Arc to Terminal. It is cheap to create and clone Arcs

sanket1729: The return values of all functions `reduce0/reduce1/reduce2` are all Miniscripts corresponding to the terminals.
We can change the existing check_frag_validity to take in &Miniscript as an argument and use the `node` field to get the terminals. I think this will simplify the code.

src/miniscript/types/extra_props.rs
sanket1729: See : https://github.com/rust-bitcoin/rust-miniscript/pull/148#issuecomment-706815941

`0NOTEQUAL` is added at the end of miniscript. Maybe that is causing the confusion?

src/miniscript/types/extra_props.rs
```diff
@@ -529,10 +645,27 @@ impl Property for ExtData {
                     sat_count = sat_count + 1;
                     ops_count_sat = ops_count_sat.map(|y| x + y);
                     ops_count_nsat = None;
+                    stack_elem_count_dissat = None;
```
sanket1729, line 763: Misplaced line?

sanket1729 (commented):
Addressed all of the comments in #148, #149, #150. Let me know if there is anything unanswered.

src/miniscript/types/extra_props.rs
darosior: Yeah i understood your first comment (as in done in the current recursive function). However this still causes most tests to panics to do so.

src/miniscript/types/extra_props.rs
darosior: Commenting there, resolving this.

src/miniscript/types/extra_props.rs
darosior: Ouch definitely, thanks.

darosior:
> Addressed all of the comments in #148, #149, #150. Let me know if there is anything unanswered.

Think i fixed everything (apart from the "both dissat are available" case). It's ready for review (after #149 i presume)

src/miniscript/decode.rs
darosior: Right. This should be part of #149 anyways.

src/miniscript/types/extra_props.rs
darosior: I don't get your comment ? This was a note to myself on a (very) previous version :) (And ofc Github moved the comment around).

src/miniscript/types/extra_props.rs
sanket1729: As discussed on IRC, this is incorrect for top-level scripts. We can keep this as it is for now without the panic

sanket1729:
Will review again after Andrew merges #149.

darosior:
Rebased and added a commit as followup to @apoelstra review in https://github.com/rust-bitcoin/rust-miniscript/pull/149

src/miniscript/types/extra_props.rs
apoelstra: I think this is fine. In non-malleable scripts this is actually unreachable; in malleable scripts (or incomplete script fragments) it is correct.

apoelstra: Ah, I see, "correct" is maybe in the eye of the beholder. You could argue that, if both dissatisfactions are available, then we should actually report the minimum here rather than the maximum, otherwise we may incorrectly consider a script to be unsatisfiable.

But then we run into complicated issues ... there could be other reasons that we can't actually take one path, that we can't see from this part of the code. Maybe the dissatisfaction somehow requires a key/hash preimage that the user is missing; maybe it's blocked because of some other resource limit, or because it turns out to mix timelocks, or something crazy like that.

So best to return the maximum, and then when we're wrong, we'll say something is unsatisfiable when it's not...rather than the other way around.

And in any case, we'll be correct for non-malleable scripts, and users know that our analysis will not be very tight for malleable ones.

src/miniscript/types/extra_props.rs
```diff
@@ -513,12 +616,25 @@ impl Property for ExtData {
...
             pk_cost += sub.pk_cost;
             ops_count_static += sub.ops_count_static;
             timelocks.push(sub.timelock_info);
+
+            if let Some(n_items) = sub.stack_elem_count_dissat {
+                stack_elem_count_dissat = stack_elem_count_dissat.map(|x| x + n_items);
+            } else {
+                // The thresh is dissatifiable iff all sub policies are dissatifiable
+                stack_elem_count_dissat = None;
+            }
```
apoelstra, line 747: sucks that we don't have `Option::zip` or `Option::zip_with`

apoelstra (approved):
ack 58817339989b9072b48533c8a8db4886f757dffa

apoelstra:
Actually can you add a commit which removes `max_dissatisfaction_witness_elements` etc from astelem.rs? These methods are no longer needed.

darosior:
Added

sanket1729:
Based on the comment from @apoelstra, I will re-review the entire PR assuming we are allowing malleable Miniscripts. I think my first review assumed that we were dealing with malleable miniscripts.

src/miniscript/types/extra_props.rs
```diff
@@ -217,6 +235,8 @@ impl Property for ExtData {
             ops_count_static: 4,
             ops_count_sat: Some(4),
             ops_count_nsat: None,
+            stack_elem_count_sat: Some(1),
+            stack_elem_count_dissat: None,
```
sanket1729, line 260: If we allow malleable dissat, this should be Some(1). Same thing for rest of the hashes

sanket1729: Note that hash fragment is 

`SIZE <32> EQUALVERIFY SHA256 <h> EQUAL`

src/miniscript/types/extra_props.rs
sanket1729: I am certain that this is wrong in the sipa implementation. We should have that as 33 for reasons in the previous comment.

sanket1729 (commented):
Left a minor discussion comment about dissatisfying hash fragments. 
If that requires more discussion, we can merge this PR and possibly have that as a small followup if it is required.

src/miniscript/types/extra_props.rs
sanket1729: Okay, that is bug in the sipa implementation as he explicitly considers them being dissatifyable here:

https://github.com/sipa/miniscript/blob/027b422495a8ee3ea3fa7787150389bb65d1b9de/bitcoin/script/miniscript.h#L657 .

apoelstra: Heh, good catch. There is definitely no canonical dissatisfaction for an unwrapped hash.

@darosior I think we want the estimation functions to return the worst-case value that is possible under standardness rules. If the fragment is supposed not to be satisfiable (for malleability checks or other reasons) then we should use the `d` flag to indicate that. But if we make it imposible to even estimate these values, it will become harder to work with incomplete fragments or other malleable scripts.

sanket1729 (commented):
nit: removing the last commit gives a warning about unused std::cmp

apoelstra:
Will need rebase on #169

src/miniscript/types/extra_props.rs
darosior: > @darosior I think we want the estimation functions to return the worst-case value that is possible under standardness rules.

Does this mean we want this not to be None even if the dissat is malleable ? If so we could have a dissat cost for virtually all fragments here..

> But if we make it imposible to even estimate these values, it will become harder to work with incomplete fragments or other malleable scripts.

Hmm.. That's how [it was before this PR though](https://github.com/rust-bitcoin/rust-miniscript/pull/150/files#diff-0b3a60807b8797a0bce33355b60f14c996d2e7e783359d47f3d97d81eda66737L746-L788) ? Just to be clear (i didn't follow the last IRC discussions very closely): do we want to not ignore malleable dissatisfaction anymore ?

darosior:
Rebased on master, left a comment [in the inner discussion](https://github.com/rust-bitcoin/rust-miniscript/pull/150#discussion_r518702489) regarding the dissat cost computation.

src/miniscript/types/extra_props.rs
sanket1729: > Does this mean we want this not to be None even if the dissat is malleable ? If so we could have a dissat cost for virtually all fragments here..

I don't think there are any non-canonical ones that we are missing here. As you point out, it was a bug in the previous code too.

sanket1729:
@darosior , In case you missed I added a comment here. https://github.com/rust-bitcoin/rust-miniscript/pull/150#discussion_r518770701 . We should have hashes be dissatisfyable

darosior:
We discussed it on IRC and you told me it should be part of another PR, so i discarded the patch...

Did i misunderstand something ?
Also, this patch iirc created some new tests failure that i didn't look into. Seemed a good idea to have it as part of another PR since it clearly is a different feature ?

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

sanket1729:
@darosior , I meant we can fix the opcode stuff .

```
<darosior> But checking right now and we don't :)
<darosior> Hmm so i should fill_ops_count_nsat as well..
<darosior> Will add a new commit rather than amending
<sanket1729_> we can keep that as separate PR. I think this is almost good to go
<darosior> Ok, will open a new PR with the commit then
```

apoelstra (approved):
ack eb5e04b025008da0844f96741a0b16e1fd60d93d

We can fix the dissat thing later

darosior:
Sorry Sanket i never got to answer your comment, pretty busy right now...

>
 ```
<darosior> Hmm so i should fill_ops_count_nsat as well..
<darosior> Will add a new commit rather than amending
<sanket1729_> we can keep that as separate PR. I think this is almost good to go```

Yea, right. I misunderstood that it only concerned the ops.

sanket1729:
Sure, no worries :)

PR #152, opened 2020-10-09
Updating to bitcoin 0.25 and MSRV 1.29

dr-orlovsky:
This is trivial, but urgently required for the projects already using bitcoin 0.25 due to secp256k1 library version conflict (you can't use miniscript v2.0 with bitcoin 0.25 in the same crate for now)


apoelstra:
It look like the conflict is caused by a bug in rust-secp (seems secp256k1-sys 0.3.0 has `link = "rustsecp256k1_v0_2_0"` instead of `v0_3_0`)

apoelstra:
We can fix this as part of https://github.com/rust-bitcoin/rust-secp256k1/pull/237 -- once we get that PR merged I'll quickly release secp-sys 0.3.1 and there should no longer be a conflict.

apoelstra:
You cannot update bitcoin to 0.25 without also increasing the MSRV of this library.

dr-orlovsky:
@apoelstra MSRV updated as a part of this PR

sanket1729 (approved):
tAck 7f2482a3acb868b8c584739ed72790d08cc20dd5. Why do we still need the pinning for cc version? It does not seem to work without it.

dr-orlovsky:
I tried to have a build w/o it and it did not pass

dr-orlovsky:
Is it possible to release Miniscript 2.1.0 supporting bitcoin 0.25?

sanket1729:
Yes, that is the current plan. We should have the release soon. cc @apoelstra

apoelstra (approved):

apoelstra:
Released 3.0.0 which uses 0.25

PR #157, opened 2020-10-13
Psbt bip fixes

### LINKED ISSUE #122: Change Satisfier trait for consistency with BIP 65/112
The current implementation of `satisfier` trait is not enforcing other rules strictly from BIP65/BIP 112. 

TODO:

- [x] Check timelocks values can be compared if both are either below or above 500 million

- [x] In pbst finalizer, also check additional rules in BIP 65(transaction version) and BIP 112(whether inputs are finalized etc)

sanket1729:
Fixes #122 . Thanks to Russell and simplicity implementation of timelocks for pointing out the details in BIP65/112.

### DIFF
--- a/src/miniscript/types/extra_props.rs
+++ b/src/miniscript/types/extra_props.rs
@@ -17,6 +17,18 @@ pub const MAX_STANDARD_P2WSH_SCRIPT_SIZE: usize = 3600;
 // https://github.com/bitcoin/bitcoin/blob/9ccaee1d5e2e4b79b0a7c29aadb41b97e4741332/src/script/script.h#L39
 pub const HEIGHT_TIME_THRESHOLD: u32 = 500_000_000;
 
+/* If nSequence encodes a relative lock-time and this flag
+ * is set, the relative lock-time has units of 512 seconds,
+ * otherwise it specifies blocks with a granularity of 1. */
+// https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki
+pub const SEQUENCE_LOCKTIME_TYPE_FLAG: u32 = 1 << 22;
+
+/* Below flags apply in the context of BIP 68*/
+/* If this flag set, nSequence is NOT interpreted as a
+ * relative lock-time. For future soft-fork compatibility*/
+// https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki
+pub const SEQUENCE_LOCKTIME_DISABLE_FLAG: u32 = 1 << 31;
+
 /// Helper struct Whether any satisfaction of this fragment contains any timelocks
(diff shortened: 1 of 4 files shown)


src/psbt/mod.rs
```diff
@@ -249,8 +250,19 @@ impl<'psbt, Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for PsbtInputSatisfie
     }
 
     fn check_older(&self, n: u32) -> bool {
-        let csv = self.psbt.global.unsigned_tx.input[self.index].sequence;
-        <Satisfier<Pk>>::check_older(&Older(csv), n)
+        let seq = self.psbt.global.unsigned_tx.input[self.index].sequence;
+        // https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki
+        // Disable flag set. return true
+        if n & SEQUENCE_LOCKTIME_DISABLE_FLAG != 0 {
+            true
```
apoelstra, line 257: I don't see this in the BIP

sanket1729: See the reference implementation and the c++ codebase.
We check the highest bit before doing any comparison checks.
```
    // To provide for future soft-fork extensibility, if the
    // operand has the disabled lock-time flag set,
    // CHECKSEQUENCEVERIFY behaves as a NOP.
    if ((nSequence & CTxIn::SEQUENCE_LOCKTIME_DISABLE_FLAG) != 0)
        break;

    // Compare the specified sequence number with the input.
    if (!checker.CheckSequence(nSequence))
        return set_error(serror, SCRIPT_ERR_UNSATISFIED_LOCKTIME);

    break;
```

src/miniscript/types/extra_props.rs
```diff
@@ -257,8 +269,8 @@ impl Property for ExtData {
             ops_count_sat: Some(1),
             ops_count_nsat: None,
             timelock_info: TimeLockInfo {
-                csv_with_height: t < HEIGHT_TIME_THRESHOLD,
-                csv_with_time: t >= HEIGHT_TIME_THRESHOLD,
+                csv_with_height: (t & SEQUENCE_LOCKTIME_TYPE_FLAG) == 0,
+                csv_with_time: (t & SEQUENCE_LOCKTIME_TYPE_FLAG) == 0,
```
sanket1729, line 273: Should be !=

apoelstra:
roundtrip_miniscript_script fuzzer fails on `"009251870051515151000000515151515151510051b20051875193006300670068518751"` which appears to be a bug in master

apoelstra:
Kicked CI so that maybe this PR will pass but we need to circle back and fix the fuzz failure it found.

apoelstra (approved):
Cool, lgtm

sanket1729:
The fuzz failure was because of `tried to dissatisfy or_i with both branches being dissatisfiable` which should be fixed in #150

PR #160, opened 2020-10-27
Descriptor pubkeys fixups

darosior:
Just noticed i didn't implement Error for the descriptor pubkey parsing error (happens to be problematic downstream)...
Also re-exported `DescriptorPublicKey` which [was suggested previously](https://github.com/rust-bitcoin/rust-miniscript/pull/131#issuecomment-699036746) but i overlooked


sanket1729 (approved):
utACK 6b0e0ea261a00eac09d6326a2a8531a56eaff4db.

PR #161, opened 2020-10-28
Misc bugs

sanket1729:
Miscellaneous bugs in the current codebase from integration tests. Each commit has some brief information about the bug

### DIFF
--- a/src/descriptor/satisfied_constraints.rs
+++ b/src/descriptor/satisfied_constraints.rs
@@ -627,7 +627,7 @@ where
                                     .map(|sig| *sig == StackElement::Dissatisfied)
                                     .filter(|empty| *empty)
                                     .count();
-                                if nonsat == *k {
+                                if nonsat == *k + 1 {
                                     self.stack.push(StackElement::Dissatisfied);
                                 } else {
                                     return Some(Err(Error::MissingExtraZeroMultiSig));
--- a/src/miniscript/decode.rs
+++ b/src/miniscript/decode.rs
@@ -240,17 +240,45 @@ pub fn parse<Ctx: ScriptContext>(
                         tokens,
                         Tk::Equal => match_token!(
                             tokens,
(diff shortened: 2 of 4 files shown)


src/miniscript/satisfy.rs
```diff
@@ -503,6 +503,14 @@ pub struct Satisfaction {
 
 impl Satisfaction {
     fn minimum(sat1: Self, sat2: Self) -> Self {
+        // If there is only one available satisfaction, we must choose that
+        // regardless of has_sig marker.
+        // This handles the case where both are unavailable.
+        match (&sat1.stack, &sat2.stack) {
```
apoelstra, line 509: as discussed on the phone, this allows too much; need to split `Unavailable` to distinguish between "we won't satisfy this" and "we can't satisfy this" and only skip the hassig check in the "can't" case

sanket1729:
Removed the satisfaction commit into a separate PR. This is now ready for review.

src/miniscript/satisfy.rs
sanket1729: This makes sense, but it goes against our assumptions for Miniscript. From Pieter's website.

> The attacker only has access to hash preimages that honest users have access to as well. This is a reasonable assumption because hash preimages are revealed once globally, and then available to everyone.

sanket1729: Based on this I think it is reasonable to have the previous commit as-is. I will be opening a separate PR for it in case there is more discussion to be had about it.

apoelstra: I never really agreed with that paragraph, but in any case, it does not preclude some honest users from having access to hash preimages that others don't.

apoelstra: Resolving, moving discussion to #167

apoelstra (approved):
ack 1252d3497a4b82efaaee9a041b336c949d17a981

PR #162, opened 2020-10-28
Integration tests

### LINKED ISSUE #124: Consensus testing of minscripts
#123 brought an interesting point that we should consensus test our miniscript with rust-libbitcoinconsensus. 

Create transactions using rust-miniscripts and check the transaction validity/satisfaction logic/weight calculation etc and other features with https://github.com/rust-bitcoin/rust-bitcoinconsensus. We can also grammar that Pieter had to generate random miniscirpts. 

Not completely sure of how we are going to do this, but I think we should have this.

sanket1729:
Added integration tests based for consensus and standardness checks for bitcoin core 0.20.0. Downloads the `bitcoind` binary form bitcoin.org and uses `rust-bitcoin-core-rpc` crate to make rpc queries to it. This serves as a test of our current satisfier logic as well as the rust-bitcoin psbt against the bitcoin core consensus and the policy rules. 

Note that we could not use the rust wrapper of `libbitcoinconsensus` because it does not expose the standardness flags. The current integration test setup should also help us test upcoming consensus changes in Taproot upgrade. Currently, we only test consensus/policy rules against bitcoin core 0.20.0 but I think that should be sufficient. 

Fixes #124 .

Marked as draft because some cases are silently failing because we cannot yet satisfy psbt hash preimages.


sanket1729:
The CI fails should be resolved once #161 is merged.

contrib/test.sh
```diff
@@ -66,3 +66,20 @@ if [ "$DO_BENCH" = true ]
 then
     cargo bench --features="unstable compiler"
 fi
+
+# Run Integration tests if told so
+if [ -n "$BITCOINVERSION" ]; then
+    set -e
+    cd integration_test
+    rm -rf bitcoin-$BITCOINVERSION-x86_64-linux-gnu.tar.gz  
+    wget https://bitcoincore.org/bin/bitcoin-core-$BITCOINVERSION/bitcoin-$BITCOINVERSION-x86_64-linux-gnu.tar.gz
```
RCasatta, line 75: with curl you can pipe directly to avoid saving an external file (and removing it), something like:
https://bitcoincore.org/bin/bitcoin-core-$BITCOINVERSION/bitcoin-$BITCOINVERSION-x86_64-linux-gnu.tar.gz | tar -xvz 

optionally to speed up a bit you can also extract only the bitcoind executable, but in that case you have to change the shasum verification part so maybe not worthy

sanket1729: Thanks, updated the PR with both suggestions.

integration_test/src/test_random_miniscripts.rs
```diff
@@ -0,0 +1,57 @@
...
+use std::str::FromStr;
+
+type MsString = Miniscript<String, Segwitv0>;
+type Ms = Miniscript<bitcoin::PublicKey, Segwitv0>;
+
+// parse 1000 miniscripts from file
+pub fn parse_miniscripts(pks: &[bitcoin::PublicKey], hash: sha256::Hash) -> Vec<Ms> {
+    // File must exist in current path before this produces output
+    let mut ms_vec = vec![];
+    if let Ok(lines) = read_lines("./random_ms.txt") {
```
apoelstra, line 24: i think we should read from stdin

integration_test/src/test_random_miniscripts.rs
```diff
@@ -0,0 +1,57 @@
...
+use std::fs::File;
+use std::io::{self, BufRead};
+use std::path::Path;
+use std::str::FromStr;
+
+type MsString = Miniscript<String, Segwitv0>;
+type Ms = Miniscript<bitcoin::PublicKey, Segwitv0>;
+
+// parse 1000 miniscripts from file
+pub fn parse_miniscripts(pks: &[bitcoin::PublicKey], hash: sha256::Hash) -> Vec<Ms> {
```
apoelstra, line 21: should be an iterator rather than reading everything at once into a vec

integration_test/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "integration_tests_miniscript"
+version = "0.1.0"
+authors = ["Steven Roose <steven@stevenroose.org>", "Sanket K <sanket1729@gmail.com>"]
+
+[dependencies]
+miniscript = {path = "../"}
+bitcoincore-rpc = "0.12.0"
```
apoelstra, line 8: I wonder .... if we can use the p2p layer rather than RPC here

contrib/test.sh
shesek: This could also be done with `wget -O - ... | tar ...`

sanket1729:
Resuming work on this. Now that we have rust-bitcoin 0.26

integration_test/src/test_random_miniscripts.rs
sanket1729: So, the tradeoff of avoiding allocations is that we will need to mine a single transaction per block to see that it is getting confirmed. If we collect all the miniscripts into a `Vec`. We can do the following. 
1) Create transactions spending to these Miniscripts.
2) Wait for confirmation
3) Spend from the Miniscripts
4) More confirmations

I guess I can mine 1 tx per block. Not a big deal.

sanket1729:
Based on #227. Still awaiting a release of `rust-bitcoincore-rpc` with 0.26 types.

sanket1729:
This is also ready for review. The CI would show incomplete because this PR changes the CI script

contrib/test.sh
```diff
@@ -49,3 +49,16 @@ if [ "$DO_BENCH" = true ]
 then
     cargo bench --features="unstable compiler"
 fi
+
+# Run Integration tests if told so
+if [ -n "$BITCOINVERSION" ]; then
+    set -e
+    cd integration_test
```
apoelstra, line 56: nit: trailing whitespace here

contrib/test.sh
```diff
@@ -49,3 +49,16 @@ if [ "$DO_BENCH" = true ]
 then
     cargo bench --features="unstable compiler"
 fi
+
+# Run Integration tests if told so
+if [ -n "$BITCOINVERSION" ]; then
+    set -e
+    cd integration_test  
+    curl https://bitcoincore.org/bin/bitcoin-core-$BITCOINVERSION/bitcoin-$BITCOINVERSION-x86_64-linux-gnu.tar.gz | tar xvzf - bitcoin-$BITCOINVERSION/bin/bitcoind    # will abort if the check fails.
+    sha256sum --check bitcoin-core-$BITCOINVERSION.sha256sum
```
apoelstra, line 58: I think you're forgetting to download the sha256sum? At least,when I try running this with 0.21.1 or 0.20.1 it doesn't work here.

sanket1729: The file would have to be manually downloaded, sha256sum checked locally, and then finally committed here. Unfornutealy, this will be a manual process for every release of bitcoin core. 

I don't see the point of downloading the sha2 sum from bitcoincore.org if we are downloading the binary from there :P .

apoelstra: Ah, perfect, I understand. I don't know how I didn't see the sha256sum file checked in for 0.20.0 (now 0.21.0).

contrib/test.sh
apoelstra: This nit is still here

integration_test/src/main.rs
```diff
@@ -2,6 +2,272 @@
...
+    fn flush(&self) {}
+}
+
+static LOGGER: StdLogger = StdLogger;
+
+static mut VERSION: usize = 0;
+/// Get the version of the node that is running.
+fn version() -> usize {
+    unsafe { VERSION }
+}
```
apoelstra, line 52: As near as I can tell there is no reason for this function or the `static mut` global to exist

contrib/test.sh
sanket1729: Will update the version to 0.21.0

sanket1729 (commented):
This would be blocked on https://github.com/rust-bitcoin/rust-bitcoincore-rpc/pull/196 for the time being

integration_test/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "integration_tests_miniscript"
+version = "0.1.0"
+authors = ["Steven Roose <steven@stevenroose.org>", "Sanket K <sanket1729@gmail.com>"]
+
+[dependencies]
+miniscript = {path = "../"}
+
+# Until 0.26 support is released on rust-bitcoincore-rpc
```
RCasatta, line 9: outdated comment

RCasatta (commented):
FYI, maybe for future PR, I think [`bitcoind`](https://crates.io/crates/bitcoind) crate as a dev dep could help to:

* avoid shell scripts
* potentially use different bitcoind version by just changing a feature
* make multiple independent, parallel tests (in bdk for example every #[test] spawn its own node)
* have guarantees bitcoind process are terminated whether the test terminate correctly or fail, while avoiding the killall that on local machine could terminate other bitcoind process on the dev machine

I think there is an issue with bitcoind MSRV which is now 1.41.1, but I may look to lower that if there is interest to use it

Disclaimer: I did the crate so maybe I am biased :)

sanket1729:
@RCasatta I read through the features and I really like it. I definitely want to migrate to this because of all the reasons you listed. 

> I think there is an issue with bitcoind MSRV which is now 1.41.1, but I may look to lower that if there is interest to use it

This is fine because at least for our CI, we are planning to run integration tests on rust stable. I will be on my TODO list after we have taproot support in the rust bitcoin ecosystem.

integration_test/run.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
+TESTDIR=/tmp/rust_miniscript_test
```
apoelstra, line 3: In dd322eb8fb9d7732b4e6ec7ca8c0213763ba4200:

This is fine for now but we should use a real single-use temp name. This is going to conflict e.g. whenever you and I are simultaneously running these tests on my big server.

apoelstra (approved):
ACK 712ffbe907529605fcc9d0757012e4bdf1b4017d

We can leave the nits to a followup PR.

sanket1729:
Merging this. Created an issue to address Andrew's comment. #287

Issue #163, opened 2020-10-30
Speed of descriptor derivation


RCasatta:
I noticed deriving a bunch of addresses from a descriptor containing a single xpub was quite slow compared to `ExtendedPubKey::derive`, the reason is that secp context is initialized at every derivation here:
https://github.com/rust-bitcoin/rust-miniscript/blob/5714f091bc30676b6e6140b23cee4bdd559a4f78/src/descriptor/mod.rs#L330

I am not sure which is the best way to avoid initialization at every derivation, since passing secp as parameter "pollute" the API and I don't see an easy way to add secp as field of `DescriptorXPub` since there would be problem with the derived trait of `DescriptorXPub` such `Clone`. Also, I don't think `lazy_static!` globally is a viable option.

Are there plan to improve this?

sanket1729:
I think adding a parameter is the way to go. Similar API's in rust-bitcoin take secp context as a parameter.
`sk.public_key(&secp)` . 

Thanks for bringing this up, I think pollution is unavoidable and we should probably prefer speed, cc @apoelstra .

afilini:
Alternatively I guess you could add a "global" context initialized with something like `lazy_static`

apoelstra:
Yeah, we definitely can't put a static variable into a general purpose library.

I'm not thrilled about polluting the API but it probably does make sense to have a secp context for converting to public keys. An interesting question though is what context capabilities should we require?

sanket1729:
To address this specific issue, it is possible to modify the function 
https://github.com/rust-bitcoin/rust-miniscript/blob/5714f091bc30676b6e6140b23cee4bdd559a4f78/src/descriptor/mod.rs#L713

to use a secp reference instead of calling `to_public_key`. 
But other functions in the library that rely on `ToPublicKey` for `witness_script/address/etc` would still be slow.

apoelstra:
We're going to have a similar problem with Taproot/p2c derivation ... I think we should change the `ToPublicKey` method to take a secp context. And I guess parameterize `ToPublicKey` over the required secp functionality.

I guess in rust-secp we also ought to reintroduce a dummy context so users without any context, who don't need one, would still be able to comfortably use the library.

sanket1729:
Maybe an Option<SecpContext>?

apoelstra:
Then `to_public_key` would need to be able to fail, which kinda sucks.

BTW what happens now when you parse a wildcard-containing path and you call `to_public_key` on it?

[...]

PR #166, opened 2020-11-04
Update ToPublicKey API to take in context param

sanket1729:
Although the diff is large, most of it just compiler guiding me on how to fix the errors. As can be seen from the diff the change is pretty invasive.  

On the other hand, we can have the user use the translate functions to get a `Miniscript<bitcoin::PublicKey, _>` and then use the API operations on those. Considering the API changes required here, I am leaning toward the other option of not implementing `ToPublicKey` to DescriptorKey at all. Even for tweaked keys where we need cryptographic operations, we should use the translate functions. 

I am starting to think that having additional parameters for address, max_satisfaction_wieght, etc might not be worth it. 
Does not currently build/test for compiler feature. I can fix that depending on what we decide. 

w.r.t #163, I think the best way to not implement `ToPublicKey` at all and have users use the translate function to get a `Miniscript<bitcoin::PublicKey, _>`. We can even provide a simple utility function that does it incase the user finds it hard to use the `traslate_pk` function.  

@apoelstra What do you think?

### DIFF
--- a/examples/htlc.rs
+++ b/examples/htlc.rs
@@ -20,6 +20,7 @@ extern crate miniscript;
 use bitcoin::Network;
 use miniscript::policy::{Concrete, Liftable};
 use miniscript::Descriptor;
+use miniscript::NullCtx;
 use std::str::FromStr;
 
 fn main() {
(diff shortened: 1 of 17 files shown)


apoelstra (approved):
yolo

PR #167, opened 2020-11-04
Add malleable signer

sanket1729:
Based on the comment: 
https://github.com/rust-bitcoin/rust-miniscript/pull/161#discussion_r517353815

> The attacker only has access to hash preimages that honest users have access to as well. This is a reasonable assumption because hash preimages are revealed once globally, and then available to everyone. On the other hand, making the assumption that attackers may have access to more preimages than honest users makes a large portion of scripts impossible to satisfy in a non-malleable way.

from bitcoin.sipa.be/miniscript

### DIFF
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -141,7 +141,7 @@ impl<Ctx: ScriptContext> Miniscript<bitcoin::PublicKey, Ctx> {
     /// script into a Miniscript representation.
     /// Use this to parse scripts with repeated pubkeys, timelock mixing, malleable
     /// scripts without sig or scripts that can exceed resource limits.
-    /// Some of the analysis gurantees of miniscript are lost when dealing with
+    /// Some of the analysis guarantees of miniscript are lost when dealing with
     /// insane scripts. In general, in a multi-party setting users should only
     /// accept sane scripts.
     pub fn parse_insane(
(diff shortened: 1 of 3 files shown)


apoelstra:
I never really agreed with that paragraph, but in any case, it does not preclude some honest users from having access to hash preimages that others don't.

sanket1729:
Because of our assumption 
> The attacker only has access to hash preimages that honest users have access to as well. This is a reasonable assumption because hash preimages are revealed once globally, and then available to everyone. On the other hand, making the assumption that attackers may have access to more preimages than honest users makes a large portion of scripts impossible to satisfy in a non-malleable way.

out current signer was not able to sign many interesting miniscripts which increased the need for malleable signer. This PR updates the satisfy logic to add `Witness::Impossible` and fixes some bugs in stack size calculation. Finally, it adds malleable signer as a fallback mode incase we fail in malleable signing.

src/miniscript/mod.rs
```diff
@@ -263,6 +247,42 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
     }
 }
 
+impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
+    /// Attempt to produce a satisfying witness for the
+    /// witness script represented by the parse tree
+    /// First tries to create a non-malleable satisfaction for
+    /// for the miniscript. If there is no guranteed non-malleable
```
apoelstra, line 254: nit "guaranteed" has two as

src/miniscript/satisfy.rs
```diff
@@ -440,7 +455,8 @@ impl Witness {
                 ret.push(hashtype.as_u32() as u8);
                 Witness::Stack(vec![ret])
             }
-            None => Witness::Unavailable,
+            // Signatures cannot be forged
+            None => Witness::Impossible,
```
apoelstra, line 459: This should stay as `Unavailable`

src/miniscript/satisfy.rs
```diff
@@ -562,7 +584,209 @@ pub struct Satisfaction {
 }
 
 impl Satisfaction {
+    // produce a non-malleable satisafaction for thesh frag
+    fn thresh<ToPkCtx, Pk, Ctx, Sat, F>(
+        k: usize,
+        subs: &Vec<Arc<Miniscript<Pk, Ctx>>>,
```
apoelstra, line 590: shouldn't take an `&Vec`

src/miniscript/satisfy.rs
apoelstra: after talking on IRC I think you're right actually. Our assumption is that a given satisfier knows all the signatures, and no others exist. (We assume no duplicated keys and assume our counterparties are not trying to malleate us.)

apoelstra:
Fuzztest is failing due to https://github.com/rust-bitcoin/rust-bitcoin/issues/520

We could patch it here but hopefully we can resolve it upstream. In any case it's unrelated to this PR and we can just kick CI when we're ready to merge.

apoelstra (approved):

PR #168, opened 2020-11-05
Complete Resource limits

### LINKED ISSUE #134: Miniscript Resource Limit calculations
Review the codebase to re-verify which of the resource limits are tested and document how are tested/calculated.

In general, we only enforce limits of standardness of the bitcoin-core node policies. Taken from https://github.com/rust-bitcoin/rust-miniscript
- [ ] Scripts over 10000 bytes are invalid by consensus (bare, P2SH, P2WSH, P2SH-P2WSH).

- [ ] Scripts over 520 bytes are invalid by consensus (P2SH).

- [ ] Script satisfactions where the total number of non-push opcodes plus the number of keys participating in all executed thresh_ms, is above 201, are invalid by consensus (bare, P2SH, P2WSH, P2SH-P2WSH).

- [ ] Anything but c:pk(key) (P2PK), c:pk_h(key) (P2PKH), and thresh_m(k,...) up to n=3 is invalid by standardness (bare).

- [ ] Scripts over 3600 bytes are invalid by standardness (P2WSH, P2SH-P2WSH).

- [ ] Script satisfactions with a serialized scriptSig over 1650 bytes are invalid by standardness (P2SH).

- [ ] Script satisfactions with a witness consisting of over 100 stack elements (excluding the script itself) are invalid by standardness (P2WSH, P2SH-P2WSH).

sanket1729:
Completes the resource limit detection of miniscript. Marking as draft, because we should merge this after #150 .

Fixes #134


src/miniscript/analyzable.rs
```diff
@@ -0,0 +1,151 @@
...
+/// 2. It contains a unspendable path because of either
+///     a. Resource limitations
+///     b. Timelock Mixing
+/// 3. The script is malleable and thereby some of satisfaction weight
+///    gurantees are not satisfied.
+/// 4. It has repeated publickeys
+#[derive(Debug)]
+pub enum AnalysisError {
+    /// Top level is not safe.
+    TopLevelNonSafe,
```
apoelstra, line 37: should rename

src/miniscript/analyzable.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// Check whether the underlying Miniscript is safe under the current context
+    /// Lifting these polices would create a semantic representation that does
+    /// not represent the underlying semantics when miniscript is spent.
+    /// Signing logic may not find satisfaction even if one exists.  
+    ///
+    /// For most cases, users should be dealing with safe scripts.
+    /// Use this function to check whether the guarantees of library hold.
+    /// Most functions of the library like would still
+    /// work, but results cannot be relied upon
+    pub fn check_safety(&self) -> Result<(), AnalysisError> {
```
apoelstra, line 136: should rename

src/miniscript/analyzable.rs
apoelstra: `SiglessBranch` maybe?

apoelstra: `BranchWithNoSignature` is ok actually, it's not crazy long

src/miniscript/analyzable.rs
```diff
@@ -0,0 +1,151 @@
...
+///    gurantees are not satisfied.
+/// 4. It has repeated publickeys
+#[derive(Debug)]
+pub enum AnalysisError {
+    /// Top level is not safe.
+    TopLevelNonSafe,
+    /// Repeated Pubkeys
+    RepeatedPubkeys,
+    /// Miniscript contains atleast one path that exceeds resource limits
+    ResourceLimitsExceeded,
```
apoelstra, line 41: also should rename to indicate that it's maybe only one branch

apoelstra: `BranchExceedsResourceLimits` lol this is getting a bit long..

src/miniscript/analyzable.rs
apoelstra: Maybe `sanity_check`

And then we could have `parse_insane` and `from_str_insane` :P

src/descriptor/mod.rs
```diff
@@ -732,6 +732,34 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
+            | Descriptor::Pkh(ref _pk)
+            | Descriptor::Wpkh(ref _pk)
+            | Descriptor::ShWpkh(ref _pk) => {}
+            Descriptor::Wsh(ref ms) | Descriptor::ShWsh(ref ms) => ms.sanity_check()?,
+            Descriptor::Sh(ref ms) => ms.sanity_check()?,
+            Descriptor::WshSortedMulti(ref svm) | Descriptor::ShWshSortedMulti(ref svm) => {
+                // extra allocation using clone allows us to reuse
+                // check safety function from Miniscript fragment instead
+                // of implemneting more code for safety checks
+                svm.clone().multi_ms().sanity_check()?
```
apoelstra, line 757: I think you should pull this clone-and-sanity-check logic into its own method on `SortedMultiVec`. Even though it's internal I think it's confusing to have a method `multi_ms` which does not sort the keys (and which therefore does not match the actual conversion of the SVM to a miniscript)

src/miniscript/analyzable.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// Whether the miniscript has repeated Pk or Pkh
+    pub fn has_repeated_keys(&self) -> bool {
+        // Simple way to check whether all of these are correct is
+        // to have an iterator
+        let all_pkhs_len = self
+            .iter_pk_pkh()
+            .map(|pk_pkh| match pk_pkh {
+                PkPkh::PlainPubkey(pk) => pk.to_pubkeyhash(),
+                PkPkh::HashedPubkey(h) => h,
+            })
```
apoelstra, line 112: you can drop the `.map` here entirely

apoelstra:
utack, except nits, and also there is an IRC discussion about naming for the validity constraints

src/descriptor/mod.rs
sanket1729: I tried doing this. The odd part about that I don't have access to `Ctx` to create a new miniscript which I do not have inside the SortedMultiVec

src/miniscript/analyzable.rs
sanket1729: And substitute it with?

src/descriptor/mod.rs
apoelstra: You can parameterize the sanity-check function over `Ctx`, which will be a bit awkward to call but should work

apoelstra (approved):

PR #169, opened 2020-11-05
Added lints from rust-bitcoin

sanket1729:
Added lints from rust-bitcoin.
changed [forbid(unsafe_code)] to [deny(unsafe_code)] as we are actually using unsafe code in ScriptContext


apoelstra (approved):
Thank you!!

PR #170, opened 2020-11-05
Only compile polcies containing non repeated keys

sanket1729:
Some of the compiler guarantees are lost if we allow repeated keys and therefore we should detect and error on these policies. 
For miniscripts, we have functions to detect these but our library would still parse and allow such miniscripts. Such miniscripts are treated as non-safe and the guarantees of miniscript no longer apply.


sanket1729:
Adding as a part of #168

PR #171, opened 2020-11-06
Add "sortedmulti" descriptor support

### LINKED ISSUE #164: Support for sortedmulti descriptor script expression?
Any plans to add support for the `sortedmulti` script expression? Some wallets like Spectre and hww's like Coldcard produce / expect BIP67 format.

This would improve compatibility for bdk, which uses rust-miniscript as a dependency and doesn't easily allow for loading descriptors with sortedmulti.

justinmoon:
Closes #164

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -23,8 +23,11 @@
 //! these with BIP32 paths, pay-to-contract instructions, etc.
 //!
 
-use std::str::{self, FromStr};
 use std::{error, fmt};
+use std::{
+    marker::PhantomData,
+    str::{self, FromStr},
+};
 
 use bitcoin::blockdata::{opcodes, script};
 use bitcoin::hashes::hash160;
@@ -35,8 +38,11 @@ use bitcoin::{self, Script};
 
 use expression;
 use miniscript;
-use miniscript::context::ScriptContextError;
-use miniscript::{Legacy, Miniscript, Segwitv0};
+use miniscript::context::{ScriptContext, ScriptContextError};
+use miniscript::{decode::Terminal, satisfy, Legacy, Miniscript, Segwitv0};
+use policy;
+use push_opcode_size;
+use script_num_size;
 use Error;
 use MiniscriptKey;
 use Satisfier;
@@ -70,6 +76,12 @@ pub enum Descriptor<Pk: MiniscriptKey> {
     Wsh(Miniscript<Pk, Segwitv0>),
     /// P2SH-P2WSH with Segwitv0 context
     ShWsh(Miniscript<Pk, Segwitv0>),
+    /// Sortedmulti under P2SH
+    ShSortedMulti(SortedMultiVec<Pk, Legacy>),
+    /// Sortedmulti under P2WSH
+    WshSortedMulti(SortedMultiVec<Pk, Segwitv0>),
+    /// Sortedmulti under P2SH-P2WSH
+    ShWshSortedMulti(SortedMultiVec<Pk, Segwitv0>),
 }
 
 /// The MiniscriptKey correponding to Descriptors. This can
@@ -394,6 +406,15 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
             Descriptor::ShWsh(ref ms) => Ok(Descriptor::ShWsh(
                 ms.translate_pk(&mut translatefpk, &mut translatefpkh)?,
             )),
+            Descriptor::ShSortedMulti(ref smv) => Ok(Descriptor::ShSortedMulti(
+                smv.translate_pk(&mut translatefpk)?,
+            )),
+            Descriptor::WshSortedMulti(ref smv) => Ok(Descriptor::WshSortedMulti(
+                smv.translate_pk(&mut translatefpk)?,
+            )),
+            Descriptor::ShWshSortedMulti(ref smv) => Ok(Descriptor::ShWshSortedMulti(
+                smv.translate_pk(&mut translatefpk)?,
+            )),
         }
     }
 }
@@ -421,6 +442,15 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             }
             Descriptor::ShWsh(ref miniscript) => {
                 Some(bitcoin::Address::p2shwsh(&miniscript.encode(), network))
+            }
+            Descriptor::ShSortedMulti(ref smv) => {
+                Some(bitcoin::Address::p2sh(&smv.encode(), network))
+            }
+            Descriptor::WshSortedMulti(ref smv) => {
+                Some(bitcoin::Address::p2wsh(&smv.encode(), network))
+            }
+            Descriptor::ShWshSortedMulti(ref smv) => {
+                Some(bitcoin::Address::p2shwsh(&smv.encode(), network))
             }
         }
     }
@@ -451,6 +481,9 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             Descriptor::Sh(ref miniscript) => miniscript.encode().to_p2sh(),
             Descriptor::Wsh(ref miniscript) => miniscript.encode().to_v0_p2wsh(),
             Descriptor::ShWsh(ref miniscript) => miniscript.encode().to_v0_p2wsh().to_p2sh(),
+            Descriptor::ShSortedMulti(ref smv) => smv.encode().to_p2sh(),
+            Descriptor::WshSortedMulti(ref smv) => smv.encode().to_v0_p2wsh(),
+            Descriptor::ShWshSortedMulti(ref smv) => smv.encode().to_v0_p2wsh().to_p2sh(),
         }
     }
 
@@ -468,9 +501,12 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             Descriptor::Bare(..)
             | Descriptor::Pk(..)
             | Descriptor::Pkh(..)
-            | Descriptor::Sh(..) => Script::new(),
+            | Descriptor::Sh(..)
+            | Descriptor::ShSortedMulti(..) => Script::new(),
             // pure segwit, empty scriptSig
-            Descriptor::Wsh(..) | Descriptor::Wpkh(..) => Script::new(),
+            Descriptor::Wsh(..) | Descriptor::WshSortedMulti(..) | Descriptor::Wpkh(..) => {
+                Script::new()
+            }
             // segwit+p2sh
             Descriptor::ShWpkh(ref pk) => {
                 let addr = bitcoin::Address::p2wpkh(&pk.to_public_key(), bitcoin::Network::Bitcoin)
@@ -482,6 +518,12 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             }
             Descriptor::ShWsh(ref d) => {
                 let witness_script = d.encode();
+                script::Builder::new()
+                    .push_slice(&witness_script.to_v0_p2wsh()[..])
+                    .into_script()
+            }
+            Descriptor::ShWshSortedMulti(ref smv) => {
+                let witness_script = smv.encode();
                 script::Builder::new()
                     .push_slice(&witness_script.to_v0_p2wsh()[..])
                     .into_script()
@@ -506,6 +548,10 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             }
             Descriptor::Sh(ref d) => d.encode(),
             Descriptor::Wsh(ref d) | Descriptor::ShWsh(ref d) => d.encode(),
+            Descriptor::ShSortedMulti(ref smv) => smv.encode(),
+            Descriptor::WshSortedMulti(ref smv) | Descriptor::ShWshSortedMulti(ref smv) => {
+                smv.encode()
+            }
         }
     }
 
@@ -626,12 +672,44 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
                 witness.push(witness_script.into_bytes());
                 Ok((witness, script_sig))
             }
+            Descriptor::ShSortedMulti(ref smv) => {
+                let mut script_witness = match smv.satisfy(satisfier) {
+                    Some(wit) => wit,
+                    None => return Err(Error::CouldNotSatisfy),
+                };
+                script_witness.push(smv.encode().into_bytes());
+                let script_sig = witness_to_scriptsig(&script_witness);
+                let witness = vec![];
+                Ok((witness, script_sig))
+            }
+            Descriptor::WshSortedMulti(ref smv) => {
+                let mut witness = match smv.satisfy(satisfier) {
+                    Some(wit) => wit,
+                    None => return Err(Error::CouldNotSatisfy),
+                };
+                witness.push(smv.encode().into_bytes());
+                let script_sig = Script::new();
+                Ok((witness, script_sig))
+            }
+            Descriptor::ShWshSortedMulti(ref smv) => {
+                let witness_script = smv.encode();
+                let script_sig = script::Builder::new()
+                    .push_slice(&witness_script.to_v0_p2wsh()[..])
+                    .into_script();
+
+                let mut witness = match smv.satisfy(satisfier) {
+                    Some(wit) => wit,
+                    None => return Err(Error::CouldNotSatisfy),
+                };
+                witness.push(witness_script.into_bytes());
+                Ok((witness, script_sig))
+            }
         }
     }
     /// Attempts to produce a satisfying witness and scriptSig to spend an
     /// output controlled by the given descriptor; add the data to a given
     /// `TxIn` output.
-    pub fn satisfy<S: Satisfier<Pk>>(
+    pub fn satisfy<S: Satisfier<Pk> + Satisfier<bitcoin::PublicKey>>(
         &self,
         txin: &mut bitcoin::TxIn,
         satisfier: S,
@@ -662,17 +740,8 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             Descriptor::ShWpkh(ref pk) => 4 * 24 + 1 + 73 + pk.serialized_len(),
             Descriptor::Sh(ref ms) => {
                 let ss = ms.script_size();
-                let push_size = if ss < 76 {
-                    1
-                } else if ss < 0x100 {
-                    2
-                } else if ss < 0x10000 {
-                    3
-                } else {
-                    5
-                };
-
-                let scriptsig_len = push_size + ss + ms.max_satisfaction_size()?;
+                let ps = push_opcode_size(ss);
+                let scriptsig_len = ps + ss + ms.max_satisfaction_size()?;
                 4 * (varint_len(scriptsig_len) + scriptsig_len)
             }
             Descriptor::Wsh(ref ms) => {
@@ -690,6 +759,28 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
                     + script_size
                     + varint_len(ms.max_satisfaction_witness_elements()?)
                     + ms.max_satisfaction_size()?
+            }
+            Descriptor::ShSortedMulti(ref smv) => {
+                let ss = smv.script_size();
+                let ps = push_opcode_size(ss);
+                let scriptsig_len = ps + ss + smv.max_satisfaction_size(1);
+                4 * (varint_len(scriptsig_len) + scriptsig_len)
+            }
+            Descriptor::WshSortedMulti(ref smv) => {
+                let script_size = smv.script_size();
+                4 +  // scriptSig length byte
+                    varint_len(script_size) +
+                    script_size +
+                    varint_len(smv.max_satisfaction_witness_elements()) +
+                    smv.max_satisfaction_size(2)
+            }
+            Descriptor::ShWshSortedMulti(ref smv) => {
+                let script_size = smv.script_size();
+                4 * 36
+                    + varint_len(script_size)
+                    + script_size
+                    + varint_len(smv.max_satisfaction_witness_elements())
+                    + smv.max_satisfaction_size(2)
             }
         })
     }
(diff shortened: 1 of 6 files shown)


src/descriptor/mod.rs
```diff
@@ -506,13 +547,17 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
...
+                smv.encode()
+            }
         }
     }
 
     /// Returns satisfying witness and scriptSig to spend an
     /// output controlled by the given descriptor if it possible to
     /// construct one using the satisfier.
-    pub fn get_satisfication<S: Satisfier<Pk>>(
+    pub fn get_satisfication<S: Satisfier<Pk> + Satisfier<bitcoin::PublicKey>>(
```
sanket1729, line 560: This should not be required. See the comments below for clarification.

src/descriptor/mod.rs
```diff
@@ -848,7 +978,153 @@ impl<Pk: MiniscriptKey> fmt::Display for Descriptor<Pk> {
...
+            k: self.k,
+            pks: pks?,
+            phantom: PhantomData,
+        })
+    }
+}
+
+impl<Pk: MiniscriptKey + ToPublicKey, Ctx: ScriptContext> SortedMultiVec<Pk, Ctx> {
+    /// Create Terminal::Multi containing sorted pubkeys
+    pub fn sorted_node(&self) -> Terminal<bitcoin::PublicKey, Ctx> {
```
sanket1729, line 1051: We should change this to return `Terminal<Pk, Ctx>` while still sorting using the internal public keys.
`        pks.sort_by(|a, b| a.to_public_key().partial_cmp(&b.to_public_key()).unwrap())`

src/descriptor/mod.rs
```diff
@@ -848,7 +978,153 @@ impl<Pk: MiniscriptKey> fmt::Display for Descriptor<Pk> {
...
+        let mut pks: Vec<bitcoin::PublicKey> =
+            self.pks.iter().map(|pk| pk.to_public_key()).collect();
+        // Sort pubkeys lexigraphically according to BIP 67
+        pks.sort_by(|a, b| a.to_string().partial_cmp(&b.to_string()).unwrap());
+        Terminal::Multi(self.k, pks)
+    }
+
+    fn encode(&self) -> script::Script {
+        self.sorted_node()
+            .encode(script::Builder::new())
```
sanket1729, line 1061: Here, we can translate the `Pk` to `bitcoin::PublicKeys` as 
` 
        self.sorted_node()
            .translate_pk(
                &mut |pk| Result::Ok::<_, ()>(pk.to_public_key()),
                &mut |_| unreachable!("Sorted multi cannot have Pkh"),
            ). 
.....
`

src/descriptor/mod.rs
```diff
@@ -848,7 +978,153 @@ impl<Pk: MiniscriptKey> fmt::Display for Descriptor<Pk> {
...
+            .into_script()
+    }
+
+    // FIXME: is this trait bound right?
+    /// Attempt to produce a satisfying witness for the
+    /// witness script represented by the parse tree
+    fn satisfy<S: Satisfier<bitcoin::PublicKey>>(&self, satisfier: S) -> Option<Vec<Vec<u8>>> {
```
sanket1729, line 1068: I think than we can change this to `Satisfier<Pk>`

src/descriptor/mod.rs
```diff
@@ -691,6 +768,37 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
                     + varint_len(ms.max_satisfaction_witness_elements()?)
                     + ms.max_satisfaction_size(2)
             }
+            Descriptor::ShSortedMulti(ref smv) => {
+                let ss = smv.script_size();
+                let push_size = if ss < 76 {
```
sanket1729, line 773: Since, this logic is now repeated twice. Can we make a function `push_size(usize) -> usize`?

src/descriptor/mod.rs
```diff
@@ -833,6 +960,9 @@ impl<Pk: MiniscriptKey> fmt::Debug for Descriptor<Pk> {
             Descriptor::Sh(ref sub) => write!(f, "sh({:?})", sub),
             Descriptor::Wsh(ref sub) => write!(f, "wsh({:?})", sub),
             Descriptor::ShWsh(ref sub) => write!(f, "sh(wsh({:?}))", sub),
+            Descriptor::ShSortedMulti(ref smv) => write!(f, "sh({})", smv),
+            Descriptor::WshSortedMulti(ref smv) => write!(f, "wsh({})", smv),
+            Descriptor::ShWshSortedMulti(ref smv) => write!(f, "sh(wsh({}))", smv),
```
sanket1729, line 965: Should be ":?"

src/descriptor/mod.rs
```diff
@@ -848,7 +978,153 @@ impl<Pk: MiniscriptKey> fmt::Display for Descriptor<Pk> {
...
+    /// signatures required
+    pub k: usize,
+    /// public keys inside sorted Multi
+    pub pks: Vec<Pk>,
+    /// The current ScriptContext for sortedmulti
+    pub(crate) phantom: PhantomData<Ctx>,
+}
+
+impl<Pk: MiniscriptKey, Ctx: ScriptContext> SortedMultiVec<Pk, Ctx> {
+    fn new(k: usize, pks: &[Pk]) -> Result<Self, Error> {
```
sanket1729, line 1000: Although we only expect users to `SortedMultiVec` via descriptors. I think we should also make this pub for advanced users? cc @apoelstra

src/descriptor/satisfied_constraints.rs
```diff
@@ -337,6 +337,34 @@ where
                     has_errored: false,
                 }
             }
+            &Descriptor::WshSortedMulti(ref smv) | &Descriptor::ShWshSortedMulti(ref smv) => {
+                SatisfiedConstraints {
+                    verify_sig: verify_sig,
+                    public_key: None,
+                    state: vec![NodeEvaluationState {
+                        node: &Any::from_smv_segwitv0(smv),
```
sanket1729, line 345: Here we should first create a new miniscript as 
`let ms = &Miniscript::from_ast(smv.sorted_node()).expect("sorted node is also a valid ms");` and them call Any::from_segwitv0 and so on.

That should avoid editing the context.rs file

sanket1729 (commented):
Awesome, thanks for doing this. There had been lot for support of this descriptor. 

The PR looks clean, left a couple of remarks. 
The CI failures are because of rust MSRV 1.29

src/descriptor/mod.rs
justinmoon: Thanks for help on these trait bounds. This `sort_by` statement will compare the 64-byte representation of the pubkeys from secp256k1, not lexigraphic representation as required by BIP 67. Calling `to_pubkey().to_string()`  does do lexigraphic comparison, however.

src/descriptor/satisfied_constraints.rs
sanket1729: As discussed on IRC, this does not work because of lifetime errors. I will think of a better way to do this, this can be left unaddressed.

src/descriptor/mod.rs
sanket1729: Ah, good catch.

sanket1729: We can add comment about bip 67 comparison here.

src/descriptor/mod.rs
justinmoon: We decided to ignore this comment over IRC

src/descriptor/mod.rs
justinmoon: Done

src/descriptor/mod.rs
justinmoon: I made them all `pub` and copied over the doc strings from comparable `Miniscript` methods

src/descriptor/mod.rs
justinmoon: Is the `Sort pubkeys lexigraphically according to BIP 67` comment enough?

src/descriptor/mod.rs
justinmoon: Done

src/descriptor/mod.rs
justinmoon: Done

src/lib.rs
```diff
@@ -506,6 +506,19 @@ pub fn script_num_size(n: usize) -> usize {
     }
 }
 
+/// FIXME
```
justinmoon, line 509: I need some help here. I don't totally understand what this does!

justinmoon:
@sanket1729 Thanks for your review. Addressed the comments that I could and got CI passing.

src/lib.rs
apoelstra: This should be renamed to `push_opcode_size` (and arguably should live in rust-bitcoin rather than here, but that's a problem for another day). An explanation and doccomment might be
```
/// Returns the size of the smallest push opcode used to push a given number of bytes onto the stack
///
/// For sizes ≤ 75, there are dedicated single-byte opcodes, so the push size is one. Otherwise, if the
/// size can fit into 1, 2 or 4 bytes, we use the `PUSHDATA{1,2,4}` opcode respectively, followed
/// by the actual size, encoded in that many bytes.
```

apoelstra: Let's also not make this `pub`, since we'll want to remove it if/when it moves into rust-bitcoin

src/descriptor/mod.rs
```diff
@@ -773,6 +890,11 @@ where
                         }
                     }
                     _ => {
+                        if top.args[0].name == "sortedmulti" {
```
apoelstra, line 876: I would rather you moved this up a level, and match on `("sortedmulti", _)`, rather than adding an if-check to the `_` branch.

src/descriptor/mod.rs
apoelstra: Yeah I generally lean in the direction of making everything `pub` in this library.

src/descriptor/mod.rs
```diff
@@ -848,7 +978,153 @@ impl<Pk: MiniscriptKey> fmt::Display for Descriptor<Pk> {
...
+        let term: miniscript::decode::Terminal<Pk, Ctx> = Terminal::Multi(k, Vec::from(pks));
+        let ms = Miniscript::from_ast(term)?;
+
+        // This would check all the consensus rules for p2sh/p2wsh and
+        // even tapscript in future
+        Ctx::check_ms_validity(&ms)?;
+
+        Ok(Self {
+            k,
+            pks: Vec::from(pks),
```
apoelstra, line 996: I'm gonna let this slide in the interest in code clarity (and because it's hard to imagine anyone constructing sortedmulti descriptors in a tight loop), but we could avoid the second allocation here by fishing the `pks` vector out of `ms` :P

src/descriptor/mod.rs
```diff
@@ -848,7 +978,153 @@ impl<Pk: MiniscriptKey> fmt::Display for Descriptor<Pk> {
...
+            phantom: PhantomData,
+        })
+    }
+    /// Parse an expression tree into a SortedMultiVec
+    fn from_tree(tree: &Tree) -> Result<Self, Error>
+    where
+        <Pk as FromStr>::Err: ToString,
+    {
+        let k = tree.args[0].name.parse::<usize>().unwrap(); // FIXME: unwrap
```
apoelstra, line 1005: You can use `expression::parse_num` rather than `usize::parse` which will return a compatible error type so you can just use `?`.

src/descriptor/mod.rs
```diff
@@ -848,7 +978,153 @@ impl<Pk: MiniscriptKey> fmt::Display for Descriptor<Pk> {
...
+            Descriptor::ShWshSortedMulti(ref smv) => write!(f, "sh(wsh({}))", smv),
+        }
+    }
+}
+
+/// Contents of a "sortedmulti" descriptor
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SortedMultiVec<Pk: MiniscriptKey, Ctx: ScriptContext> {
+    /// signatures required
+    pub k: usize,
```
apoelstra, line 975: I think we should use `u32` rather than `usize`. It's a bit less ergonomic but keeps us consistent across platforms, and we can't support such large numbers anyway.

src/descriptor/mod.rs
apoelstra, line 983: Let's take a `Vec<Pk>` here rather than taking a slice that we immediately promote to a vector

src/descriptor/mod.rs
apoelstra: We should fix the `Ord` impl upstream to not sort on the internal representation, because that's crazy and has burned multiple projects.

But in the meantime we can use `.encode` rather than `to_string` and avoid an allocation.

apoelstra: Also we can use `sort_by_key` rather than `sort_by`.

I'd prefer `sort_by_cached_key` but unfortunately we don't have that with our MSRV..

apoelstra: I meant `serialize` rather than `encode`, which returns a `[u8; 33]`, and because Rust is stupid we'll have to further wrap that in a type because it doesn't come with its own `Ord` impl.

So my new proposal is to define, within this module,
```
pub struct Bip67Sorted<'a, Pk: ToPublicKey>(&'a Pk);
impl<'a, Pk: ToPublicKey> Ord for Bip67Sorted<'a, Pk> {
    fn cmp(&self, other: &Self) -> cmp::Ordering {
        (&self.to_public_key().serialize()[..]).cmp(&other.to_public_key().serialize()[..])
    }
}
impl<'a, Pk: ToPublicKey> PartialOrd for Bip67Sorted<'a, Pk> {
    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
        Some(self.cmp(other))
    }   
}
```
with suitable modifications to make that compile since I just typed it into github without testing :).

We should also move it to the top-level `lib.rs`, unless Sanket can think of a better place for it. I'd like to expose this since I want to use it in Liquid :P.

src/descriptor/mod.rs
```diff
@@ -848,7 +978,153 @@ impl<Pk: MiniscriptKey> fmt::Display for Descriptor<Pk> {
...
+    }
+}
+
+impl<Pk: MiniscriptKey + ToPublicKey, Ctx: ScriptContext> SortedMultiVec<Pk, Ctx> {
+    /// Create Terminal::Multi containing sorted pubkeys
+    pub fn sorted_node(&self) -> Terminal<bitcoin::PublicKey, Ctx> {
+        let mut pks: Vec<bitcoin::PublicKey> =
+            self.pks.iter().map(|pk| pk.to_public_key()).collect();
+        // Sort pubkeys lexigraphically according to BIP 67
+        pks.sort_by(|a, b| a.to_string().partial_cmp(&b.to_string()).unwrap());
```
apoelstra: Then with the above, this should just become `pks.sort_by_key(Bip67Sorted)`.

src/descriptor/mod.rs
```diff
@@ -848,7 +978,153 @@ impl<Pk: MiniscriptKey> fmt::Display for Descriptor<Pk> {
...
+    fn max_satisfaction_witness_elements(&self) -> usize {
+        2 + self.k
+    }
+
+    fn max_satisfaction_size(&self, _: usize) -> usize {
+        1 + 73 * self.k
+    }
+}
+
+impl<Pk: MiniscriptKey, Ctx: ScriptContext> crate::policy::Liftable<Pk>
```
apoelstra: I'm not convinced that we should implement `lift` directly on `SortedMultiVec`, but if we do, we should definitely sort the keys before lifting.

src/descriptor/satisfied_constraints.rs
apoelstra: One way to approach this may be to always store the sorted vector in the `SortedMultiVec`.

src/miniscript/context.rs
```diff
@@ -221,6 +223,39 @@ impl Any {
...
+        // The fields Miniscript<Pk, Legacy> and Miniscript<Any, Legacy> only
+        // differ in PhantomData. This unsafe assumes that the unlerlying byte
+        // representation of the both should be the same. There is a Miri test
+        // checking the same.
+        unsafe {
+            use std::mem::transmute;
+            transmute::<&Miniscript<bitcoin::PublicKey, Legacy>, &Miniscript<bitcoin::PublicKey, Any>>(
+                ms,
+            )
+        }
```
apoelstra, line 238: Can you add a miri test which detects the fact that you are returning a pointer to dead memory?

Also, you are returning a pointer to dead memory :P.

src/descriptor/satisfied_constraints.rs
justinmoon: But then we lose the ability to even print out the original descriptor, right?

src/miniscript/context.rs
sanket1729: The current implementation that creates a new miniscript returns a reference to a value from the function which is unsafe.  The correct way would be have additional information in the iterator about this. 

I think I can add this as a follow up PR as it more related to `SatisfiedConstraints` than the sortedmulti part. If you are okay, we can have these are unimplemented!() for now?

apoelstra: To be clear - this is UB and we need to come up with a different solution

apoelstra: Yeah I'm ok with just making them `unimplemented!()` for now. cc @justinmoon

src/descriptor/satisfied_constraints.rs
sanket1729: We can't do that without `ToPublicKey` and is descriptor is `Descriptor<MiniscriptKey>`

sanket1729: As discussed later, I will address this in a separate PR and we can leave this as `unimplemented!` for now

src/descriptor/mod.rs
justinmoon: `Terminal::Multi` uses `usize`. Would it make sense to change that, too? 

It would be odd to use different types in these two places because they mean the same thing and we convert between them.

src/descriptor/mod.rs
justinmoon: Did you mean `pks.sort_by_key(|pk| Bip67Sorted(pk))`?

According to [the docs](https://doc.rust-lang.org/beta/std/vec/struct.Vec.html#method.sort_by_key), the argument to `sort_by_key` needs to be `FnMut`, which we haven't implemented.

If I use a lambda, then it complains that return value of `Bip67Sorted(pk)` may outlive `pk` ... not sure how to fix that.

justinmoon: I think it would be best to just use this for now:
```
pks.sort_by(|a, b| {
    a.to_public_key()
        .serialize()
        .partial_cmp(&b.to_public_key().serialize())
        .unwrap()
});
```

But get [the upstream change](https://github.com/rust-bitcoin/rust-secp256k1/pull/248) merged and switch to `pks.sort()`  as soon as upstream change is released.

justinmoon: Doesn't make sense to make a new public struct to implement behavior that should be in `rust-libsecp256k1`

src/descriptor/mod.rs
justinmoon: Responded here: https://github.com/rust-bitcoin/rust-miniscript/pull/171/files/079955beb9c469b0c5c1aae43d3bfd0061176807#r520250242

src/descriptor/mod.rs
justinmoon: I don't fully understand what `lift` does so I'll defer to you guys on that.

IIRC @sanket1729 said that pubkeys didn't need to be sorted for this.

Also, I just realized that [the `lift` impl for `Descriptor` doesn't have the `ToPublicKey` trait bound](https://github.com/rust-bitcoin/rust-miniscript/blob/master/src/policy/mod.rs#L119). Would we want to add it?

src/descriptor/satisfied_constraints.rs
justinmoon: Leaving this `unimplemented!`

src/lib.rs
justinmoon: Removed `pub`

src/descriptor/mod.rs
justinmoon: Done

src/descriptor/mod.rs
justinmoon: Done

src/descriptor/mod.rs
justinmoon: Done

src/miniscript/context.rs
justinmoon: I left this `unimplemented!()`

justinmoon:
Thanks for the review @apoelstra. I added a few commits addressing most of your comments.

src/descriptor/mod.rs
apoelstra: `pks.sort_by_key(|pk| Bip67Sorted(pk))` is identical to `pks.sort_by_key(Bip67Sorted)` except the latter uses fewer symbols. And struct constructors always implement all of the `Fn` traits.

Using a lambda or not should have no effect on this. What is the error you are getting?

src/descriptor/mod.rs
apoelstra: We definitely do not want to add a `ToPublicKey` bound to the lift feature of descriptor...I think the existing code is fine actually, after talking to Sanket.

src/descriptor/mod.rs
apoelstra: Yeah, you're right, usize is fine

src/descriptor/mod.rs
```diff
@@ -29,12 +29,15 @@ use std::{
...
 
-use bitcoin::blockdata::{opcodes, script};
 use bitcoin::hashes::hash160;
 use bitcoin::hashes::hex::FromHex;
 use bitcoin::secp256k1;
 use bitcoin::util::bip32;
 use bitcoin::{self, Script};
+use bitcoin::{
+    blockdata::{opcodes, script},
+    util::psbt::serialize::Serialize,
```
apoelstra, line 39: You definitely do not want this serialize function!

src/descriptor/mod.rs
apoelstra: I can reproduce the lifetime error. The issue is that `sort_by_key` and `sort_key_cached_key` have their lifetimes set incorrectly by stdlib, so we do indeed need to use `sort_by`

justinmoon: Ahh sorry, should have posted the full error message.

src/descriptor/mod.rs
justinmoon: EEK! I blame Rust Analyzer!!!

Pushed a commit to serialize public keys using `pubkey.key.serialize()` and removed that import.

apoelstra: Perfect!

apoelstra:
Ok @justinmoon can you rebase on master and squash everything? You can break up all the commits however you want ... one big commit is fine ... but please make sure that every commit compiles by itself.

justinmoon:
Done

apoelstra (approved):

src/descriptor/mod.rs
apoelstra: Unresolving this for now since there is opposition from upstream to including a nonsense Core ordering in rust-bitcoin.

PR #172, opened 2020-11-06
Change psbt APIs to take in secp parameter

sanket1729:
Each time psbt APIs accept a secp parameter reference
so that we don't have to create a new one which is the expensive
process.


apoelstra (approved):
utack

PR #175, opened 2020-11-13
Sorted Multi explaination for satisfied constraint

sanket1729:
Added a panic for SortedMulti satisfied constraints
along with some explaination.


apoelstra (approved):
Yeah, I think this is reasonable.

PR #177, opened 2020-11-13
descriptor: Accept a generic Secp256k1 ctx instead of just `SignOnly`

afilini:
Without this patch two different secp contexts are needed in order to interact with the library: one that has the `Verification` flag to use with `ToPublicKey` and one with `SignOnly` to call methods on extended keys.

By accepting a generic parameter one can just make an `All` instance and use it anywhere.


sanket1729 (approved):
tAck b9e24c24e1c8b8dbd2a946b79c13eb60ca0308ad. 

Also checked the entire codebase for other instances where are not suing generic secp ctx. All instances of (non-test) code are addressed by this PR.

apoelstra (approved):
utack, thank you!

PR #180, opened 2020-11-13
Fix Display/Debug For SortedMultiVec

shesek:
The keys part was encoded with quotes around it due to the use of `{:?}`,
like so: `wsh(sortedmulti(1,"xpub66..,xpub69.."))`.

While at it, I also took the liberty to avoid unnecessary allocations (similarly to how `Display` is implemented for `Terminal::Multi`) and to DRY things by delegating `Debug` to `Display`.


apoelstra (approved):
ack, though we should definitiely have a unit test for this

justinmoon:
Thanks @shesek!

PR #184, opened 2020-11-17
descriptor: Make `SortedMultiVec::new()` public

afilini:
Considering that the fields of the struct are already public keeping it private
doesn't really prevent anybody from creating an instance with some weird arguments.

Even worse, it forces the user to manually instantiate the struct, bypassing all
the necessary checks present in the constructor.


afilini:
Actually, I've just noticed that the `phantom` field is private, so an external user has no way of building an instance of the struct without making the constructor public

darosior (approved):
ACK e45563646dd4ee40d2834a9e65c3b4f9c95c1a82

sanket1729:
utAck e455636

sanket1729 (approved):

PR #188, opened 2020-11-18
update fuzz dependencies

sanket1729:


apoelstra (approved):
sgtm.

To be honest I've never used afl and I'm not sure how.

PR #189, opened 2020-11-18
Fail satisfy when resource limits exceed

sanket1729:
Satisfy can fail if the resource limits are exceeded

### DIFF
--- a/src/miniscript/context.rs
+++ b/src/miniscript/context.rs
@@ -18,6 +18,7 @@ use miniscript::types::extra_props::{
     MAX_STANDARD_P2WSH_SCRIPT_SIZE, MAX_STANDARD_P2WSH_STACK_ITEMS,
 };
 use std::fmt;
+use util::{witness_size, witness_to_scriptsig};
 use Error;
 use {Miniscript, MiniscriptKey, Terminal};
 /// Error for Script Context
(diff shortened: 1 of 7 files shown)


fuzz/Cargo.toml
```diff
@@ -15,7 +15,7 @@ honggfuzz_fuzz = ["honggfuzz"]
 honggfuzz = { version = "0.5", optional = true }
 afl = { version = "0.8", optional = true }
 regex = { version = "1.4"}
-miniscript = { path = "..", features = ["fuzztarget", "compiler"] }
```
apoelstra, line 18: Why are we changing this? The fuzzer won't be able to come up with hash preimages or signatures without the fuzztarget on.

sanket1729: What does the `fuzztarget` feature do in rust-bitcoin? I think the only use we have in rust-miniscript is pass through for `rust-bitcoin`.

sanket1729: I think I disabled them because I was not able to come up with secp Signature required for satisfier.

apoelstra (approved):
ack 70b987fa12f9c637f2688b1e96ac30353f7f6298

src/miniscript/context.rs
```diff
@@ -294,6 +318,17 @@ impl ScriptContext for Segwitv0 {
         Ok(())
     }
 
+    fn check_witness<Pk: MiniscriptKey, Ctx: ScriptContext>(
+        witness: &[Vec<u8>],
+    ) -> Result<(), ScriptContextError> {
+        if witness_size(witness) > MAX_STANDARD_P2WSH_SCRIPT_SIZE {
+            return Err(ScriptContextError::MaxScriptSigSizeExceeded);
```
darosior, line 325: Hmm shouldn't it return a `MaxWitnessSizeExceeded` Error instead ?

PR #191, opened 2020-11-19
overhaul interpreter ("satisfied constraints") API

apoelstra:


src/descriptor/create_descriptor.rs
```diff
@@ -1,391 +0,0 @@
...
-            .collect();
-        if !witness.is_empty() {
-            return Err(Error::NonEmptyWitness);
-        }
-        let ms = Miniscript::<bitcoin::PublicKey, Bare>::parse(script_pubkey)?;
-        Ok((Descriptor::Bare(ms), Stack::from(stack?)))
-    }
-}
-
-#[cfg(test)]
```
sanket1729, line 243: We are losing these tests. I don't know if it is worth fixing these in this PR, but these can be adapted to the new `from_txdata` API in the future.

src/interpreter/inner.rs
```diff
@@ -0,0 +1,209 @@
...
+    } else if spk.is_p2sh() {
+        match ssig_stack.pop() {
+            Some(elem) => {
+                // ** p2sh-wrapped wsh **
+                if let stack::Element::Push(slice) = elem {
+                    let scripthash = hash160::Hash::hash(slice);
+                    if *spk != bitcoin::Script::new_p2sh(&scripthash.into()) {
+                        return Err(Error::IncorrectScriptHash);
+                    }
+                    if slice.len() == 34 && slice[0] == 0 && slice[1] == 32 {
```
sanket1729, line 197: Mssing ShWpkh case.

src/interpreter/mod.rs
```diff
@@ -100,6 +99,39 @@ impl<'txin> Interpreter<'txin> {
...
+    /// If you are using the interpreter as a sanity check on a transaction,
+    /// it is worthwhile to try to parse this as a descriptor using `from_str`
+    /// which will check standardness and consensus limits, which the interpreter
+    /// does not do on its own. Or use the `inferred_descriptor` method which
+    /// does this for you.
+    pub fn inferred_descriptor_string(&self) -> String {
+        match self.inner {
+            inner::Inner::PublicKey(ref pk, inner::PubkeyType::Pk) => format!("pk({})", pk),
+            inner::Inner::PublicKey(ref pk, inner::PubkeyType::Pkh) => format!("pkh({})", pk),
+            inner::Inner::PublicKey(ref pk, inner::PubkeyType::Wpkh) => format!("wpkh({})", pk),
```
sanket1729, line 126: The public key one is lacking ShWpkh. I think we are not parsing it in the `from_txdata` function probably a few commits back.

src/interpreter/mod.rs
```diff
@@ -123,6 +125,19 @@ impl<'txin> Interpreter<'txin> {
         }
     }
 
+    /// Whether this is a pre-segwit spend
+    pub fn is_legacy(&self) -> bool {
```
sanket1729, line 135: Should we also have a function for segwitv0? Should this be a part of `Descriptor` struct?

sanket1729: Maybe descriptor is not the place, but interpreter.is_legacy() does not sound right.

src/interpreter/mod.rs
```diff
@@ -132,6 +147,64 @@ impl<'txin> Interpreter<'txin> {
...
         Descriptor::from_str(&self.inferred_descriptor_string())
     }
+
+    /// Returns a sighash over the entire transaction which can be used to verify signatures
+    /// in the descriptor
+    ///
+    /// Not all fields are used by legacy descriptors; if you are sure this is a legacy
+    /// spend (you can check with the `is_legacy` method) you can provide dummy data for
+    /// the amount.
+    pub fn sighash_message(
```
sanket1729, line 163: Same here. It is not sounding right to call this from  interpreter. Maybe all these methods do belong together but belong in the `Inner`struct. Maybe, we can think of a better name for `Inner`. Maybe `TxComponents`

src/interpreter/mod.rs
```diff
@@ -132,6 +147,64 @@ impl<'txin> Interpreter<'txin> {
...
+    /// Returns a closure which can be given to the `iter` method to check all signatures
+    pub fn sighash_verify<'a, C: secp256k1::Verification>(
+        &self,
+        secp: &'a secp256k1::Secp256k1<C>,
+        unsigned_tx: &'a bitcoin::Transaction,
+        input_idx: usize,
+        amount: u64,
+    ) -> impl Fn(&bitcoin::PublicKey, BitcoinSig) -> bool + 'a {
+        // Precompute all sighash types because the borrowck doesn't like us
+        // pulling self into the closure
```
sanket1729, line 190: Ouch

src/psbt/finalizer.rs
```diff
@@ -216,44 +211,20 @@ pub fn interpreter_check<C: secp256k1::Verification>(
             .as_ref()
             .unwrap_or(&empty_witness);
 
+        println!("spk : {}", spk);
+        println!("ssi : {}", script_sig);
```
sanket1729, line 215: Maybe you forgot to remove debug print

sanket1729 (commented):
Wow, this is much cleaner. In a followup commit/PR, we also remove the miri tests from the CI.

sanket1729:
CI failures are rust 1.29.

src/descriptor/satisfied_constraints.rs
```diff
@@ -42,7 +42,7 @@ pub enum Error {
     /// MultiSig requires 1 extra zero element apart from the `k` signatures
     MissingExtraZeroMultiSig,
     /// Script abortion because of incorrect dissatisfaction for multisig.
-    /// Any input witness apart from sat(0 sig ...) or nsat(0 0 ..) leads to
+    /// NoChecks input witness apart from sat(0 sig ...) or nsat(0 0 ..) leads to
```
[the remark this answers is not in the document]
apoelstra: lol, yes, thanks

src/interpreter/mod.rs
apoelstra: Oh, yeah, I totally forgot that was a thing...yeah, we were not parsing it in the original code.

src/psbt/finalizer.rs
apoelstra: Oh, lol, I saw those but totally forgot they were mine

src/interpreter/mod.rs
apoelstra: The `Inner` struct is not part of the public API

src/interpreter/mod.rs
apoelstra: We can rename the interpreter...but this is the object that you get when you try to take a txin and analyze it using Miniscript. Determining legacy or not seems perfectly reasonable.

src/descriptor/create_descriptor.rs
apoelstra: Added a commit with replacement for these tests

src/interpreter/inner.rs
```diff
@@ -278,3 +281,449 @@ pub fn from_txdata<'txin>(
...
+        .expect("parse txdata");
+        assert_eq!(inner, Inner::PublicKey(fixed.pk_comp, PubkeyType::ShWpkh));
+        assert_eq!(stack, Stack::from(vec![]));
+        assert_eq!(script_code, comp.pkh_spk);
+
+        // sh_wpkh, right pubkey, signature
+        let (inner, stack, script_code) = from_txdata(
+            &comp.sh_wpkh_spk,
+            &comp.sh_wpkh_sig,
+            &comp.sh_wpkh_stack_justkey,
```
sanket1729, line 580: Needs sig here instead fo justkey

sanket1729 (commented):
tACk 3f23956d644836e3e1f50ad782680ea6e96fa0c1. I think we are good, just one more change in the test.

sanket1729:
tAck 8c35525eca0f560e4671a814da1fc3ebade55114

sanket1729 (approved):

PR #196, opened 2020-11-24
Satisfy error: return a max witness size error for Segwitv0

darosior:
We returned a ScriptSig-related error for the Witness.
Also added a commit to move the constraints to their own module since their number grew up significantly :)

### DIFF
--- a/src/miniscript/context.rs
+++ b/src/miniscript/context.rs
@@ -12,11 +12,11 @@
 // If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
 //
 
-use miniscript::types;
-use miniscript::types::extra_props::{
+use miniscript::limits::{
     MAX_OPS_PER_SCRIPT, MAX_SCRIPTSIG_SIZE, MAX_SCRIPT_ELEMENT_SIZE, MAX_SCRIPT_SIZE,
     MAX_STANDARD_P2WSH_SCRIPT_SIZE, MAX_STANDARD_P2WSH_STACK_ITEMS,
 };
+use miniscript::types;
 use std::fmt;
 use util::{witness_size, witness_to_scriptsig};
 use Error;
--- a/src/miniscript/context.rs
+++ b/src/miniscript/context.rs
@@ -18,7 +18,7 @@ use miniscript::limits::{
 };
 use miniscript::types;
 use std::fmt;
-use util::{witness_size, witness_to_scriptsig};
+use util::witness_to_scriptsig;
 use Error;
 use {Miniscript, MiniscriptKey, Terminal};
 /// Error for Script Context
@@ -321,9 +321,7 @@ impl ScriptContext for Segwitv0 {
     fn check_witness<Pk: MiniscriptKey, Ctx: ScriptContext>(
         witness: &[Vec<u8>],
     ) -> Result<(), ScriptContextError> {
-        if witness_size(witness) > MAX_STANDARD_P2WSH_SCRIPT_SIZE {
-            return Err(ScriptContextError::MaxScriptSigSizeExceeded);
-        } else if witness.len() > MAX_STANDARD_P2WSH_STACK_ITEMS {
+        if witness.len() > MAX_STANDARD_P2WSH_STACK_ITEMS {
             return Err(ScriptContextError::MaxWitnessItemssExceeded);
         }
         Ok(())
--- a/doc/resource_limitations.md
+++ b/doc/resource_limitations.md
@@ -3,12 +3,12 @@ TODO: Rust-miniscript behaviour for resource limitations:
 # Safe vs Valid vs Sanity/Analyzable/Liftable
 This document refers to bitcoin consensus and standardness rules as of bitcoin core release 0.20.
 
-One of Miniscript’s goals are to make advanced Script functionality accommodate both machine and human analysis. However such a analysis is not possible in all cases.
+One of Miniscript’s goals are to make advanced Script functionality accommodate both machine and human analysis. However such an analysis is not possible in all cases.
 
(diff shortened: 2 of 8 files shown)


src/miniscript/context.rs
```diff
@@ -322,7 +322,7 @@ impl ScriptContext for Segwitv0 {
         witness: &[Vec<u8>],
     ) -> Result<(), ScriptContextError> {
         if witness_size(witness) > MAX_STANDARD_P2WSH_SCRIPT_SIZE {
-            return Err(ScriptContextError::MaxScriptSigSizeExceeded);
+            return Err(ScriptContextError::MaxWitnessScriptSizeExceeded);
```
sanket1729, line 325: Oops, this check should not be here in the first place. This was originally introduced to do satisfaction time checks after we create a witness. This is script creation time check that should have been already been checked. 

Can you update the commit to remove this if check entirely?

sanket1729 (commented):
Wondering if we should rename constraints to constants because constraints can be mean constraints for the miniscript(Pk, hashes, timelocks) etc.

apoelstra:
I think it should be named `limits`. Agree that `constraints` is unclear (though I'd also like the interpreter constraints to be renamed to "conditions" :))

src/miniscript/context.rs
darosior: >  Can you update the commit to remove this if check entirely?

So i should remove both, as if MAX_STANDARD_P2WSH_SCRIPT_SIZE is already checked so is MAX_STANDARD_P2WSH_STACK_ITEMS. Will remove the method entirely

darosior:
Agreed, 'limits' is definitely a better name. Also added a commit to add some clarifications to the limits doc

src/miniscript/context.rs
sanket1729: Script size is known at script creation time. So, any scripts that exceed those limits would not be parsed. 
But we still allow parsing `insane` scripts(https://github.com/rust-bitcoin/rust-miniscript/blob/c8cd42b096a62ec8fc3276f61f0e114273fbe63e/src/miniscript/mod.rs#L147) 

that can contain a spend path that can exceed resource limits. The satisfaction logic could still find satisfaction that does exceed the stack limit, so we need to fail that over here. However, if it finds another satisfaction(less than 100 elements) it should succeed.

darosior: Then.. Why did you say i should remove the size check in the first place ?
If you want to check the satisfaction of script created with parse_insane, you do want to check both the size and number elements don't you ? I don't see why you'd only want to check the number of elements...

apoelstra (dismissed):

apoelstra:
@sanket1729 i'll let you merge if you're ok with the current state

src/miniscript/context.rs
sanket1729: Sorry, I got confused because of the constant name. The left side of the equality is indeed checking size with `witness_size()`, the right side should be a new constant which I don't think we don't have defined currently. It should be only be checked for `Legacy` scripts and should be `MAX_STANDARD_SCRIPTSIG_SIZE`. 

As far I can see, there are no total witness size rules in standard/consensus rules. 
`https://github.com/bitcoin/bitcoin/pull/20497`

darosior: > As far I can see, there are no total witness size rules in standard/consensus rules.

There are, but should we be checking them at satisfaction time ? If so i'll just remove the current commit and replace it with my original one (which was just correcting the error name). In this case we should also add some tests as just removing the check it did not make any fail.

sanket1729: > There are,

I think we only have rules for total ScriptSig size and not for witness size(Segwitv0). The only I see is that witness script is 10,000. This does not include the witness. 

> but should we be checking them at satisfaction time? 

In script creation time, we can only check whether they can be possible spend path. Normal methods like parse, from_str won't allow such scripts but the corresponding insane versions would allow them. For insane scripts, after we satisfy the witness, we know whether the current witness found by the satisfier is within limits or not and this is where this method is useful.

darosior: So, should i keep this commit or revert to the precedent version ?

darosior: >  I think we only have rules for total ScriptSig size and not for witness size(Segwitv0). The only I see is that witness script is 10,000. This does not include the witness.

We do (size and number of elements), but i think we aren't talking about the same thing ?

sanket1729: You can keep the current commit. I can suggest a modification later :). 

From the BIP 
> The witnessScript (≤ 10,000 bytes) is popped off the initial witness stack. SHA256 of the witnessScript must match the 32-byte witness program.

So only the redeem script(10,000) has a rule. The number of elements(100) and the total size is implicitly bound by 520 bytes per element. 
So, there are three things
1. Redeem Script (10,000) 
2. stack elems(100)
3. Total witness size(100 * 520 + 10,000) // implicit limit should never be reached.

My original intent was to check 3. with the PR, but turns out we don't need to Segwitv0.

darosior: Remove the witness script size check, left the n elements check untouched

sanket1729 (approved):
tAck 0c9661ff107a203850631c375edaf8953b0e4011

apoelstra (approved):
tACK a5619a8a4e0e994c40fec56c2bcbc183551d8bfc

apoelstra (approved):
tACK 0c9661ff107a203850631c375edaf8953b0e4011 even

src/miniscript/context.rs
Pragmatic2021: > Oops, this check should not be here in the first place. This was originally introduced to do satisfaction time checks after we create a witness. This is script creation time check that should have been already been checked. 
> 
> 
> 
> Can you update the commit to remove this if check entirely?

I'm really sorry moving it to the developer. So far it's been insane and I don't think I'll be able to get a dime..,that's what my repoBitcoin core.

PR #199, opened 2020-11-30
lift failure bug fixes

sanket1729:
Lifting should fail for miniscripts/descriptors
- timelock combination
- branch exceeds resource limits

Lifting fails for concrete policies:
- timelock combination
- If there is a branch that can exceed resource limits, our compiler would not create miniscripts anyway. And it general, it is difficult to check the resource limit checks without compiling.

### DIFF
--- a/src/policy/mod.rs
+++ b/src/policy/mod.rs
@@ -21,6 +21,7 @@
 //! The format represents EC public keys abstractly to allow wallets to replace
 //! these with BIP32 paths, pay-to-contract instructions, etc.
 //!
+use {error, fmt};
 
 #[cfg(feature = "compiler")]
 pub mod compiler;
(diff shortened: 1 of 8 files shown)


src/policy/mod.rs
```diff
@@ -43,22 +44,78 @@ const ENTAILMENT_MAX_TERMINALS: usize = 20;
...
+}
+
+impl error::Error for LiftError {
+    fn description(&self) -> &str {
+        ""
+    }
+    fn cause(&self) -> Option<&error::Error> {
+        None
+    }
+}
```
apoelstra, line 79: I just checked, and if you replace this whole block with
```
impl error::Error for LiftError {}
```
it compiles even on 1.29.0.

apoelstra (approved):
ack e7eeb6c24894bc2908d2ab11687d5e21b1cfa23e

PR #200, opened 2020-11-30
Satisfy fixes

sanket1729:
More bug fixes from our integration tests. Looks like I messed up #157 again with `older/after` again. I think we should probably release a patch version with this fix.


apoelstra (approved):
ack. will do a release after #199

PR #204, opened 2020-12-10
Minor cleanup for psbt finalizer

### LINKED ISSUE #202: fuzz failure when sortedmulti() has zero args

sanket1729:
Does the check for all SigHashTypes. Earlier this only checked if SigHash was explicitly set.
The second commit fixes #202

### DIFF
--- a/src/psbt/finalizer.rs
+++ b/src/psbt/finalizer.rs
@@ -251,42 +251,41 @@ pub fn finalize<C: secp256k1::Verification>(
 
     // Check well-formedness of input data
     for (n, input) in psbt.inputs.iter().enumerate() {
-        if let Some(target) = input.sighash_type {
-            for (key, rawsig) in &input.partial_sigs {
-                if rawsig.is_empty() {
+        let target = input.sighash_type.unwrap_or(bitcoin::SigHashType::All);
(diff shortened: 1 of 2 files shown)


src/psbt/finalizer.rs
```diff
@@ -251,42 +251,41 @@ pub fn finalize<C: secp256k1::Verification>(
...
-        if let Some(target) = input.sighash_type {
-            for (key, rawsig) in &input.partial_sigs {
-                if rawsig.is_empty() {
+        let target = input.sighash_type.unwrap_or(bitcoin::SigHashType::All);
+        for (key, rawsig) in &input.partial_sigs {
+            if rawsig.is_empty() {
+                return Err(Error::InputError(
+                    InputError::InvalidSignature {
+                        pubkey: *key,
+                        sig: rawsig.clone(),
```
apoelstra, line 260: you shouldn't need this clone here (not that it matters for an empty signature)

apoelstra: oh, no, I see, you're converting the borrow to owned

apoelstra (approved):
ack 99868ce025dbcd493528c1f1afb38b87915bcce4

PR #206, opened 2020-12-10
Return Err on None in context checks

sanket1729:
Thanks to @darosior for pointing out this bug. In our context checks, I implemented a short circuit branch to error out when the resource limit exceeded. But when the check function(e.g max_satisfaction_witness_elements) itself returned `None`, we had a bug which returned Ok(()). 

I think it is worth changing the API's that return None to return a result so that they are more descriptive and b) these types of bugs don't occur. ( '?' is easy to write and review. Someone I missed to match `None`).


darosior (approved):
ACK bab678b220c272d2f7b5b1d1cbfde4722390cdfd

Thanks!

apoelstra (approved):
ack bab678b220c272d2f7b5b1d1cbfde4722390cdfd

Issue #181, opened 2020-11-14
Idea: add a PhantomData tag to differentiate `DescriptorPublicKey` before and after derivation


afilini:
I am starting to realize that most operations one might want to do on a `Descriptor` are only meaningful either if the descriptor has all the keys in their "extended" form or if all of them are "derived", not both: generally speaking, all the operations that need to call `to_public_key()` somewhere want the derived keys (and there are actual assertions in the code to ensure that), while other functions that work on the full paths of the keys like `matches()` want the original descriptor with the wildcard keys.

I think it would be nice to have the compiler check that the right methods are called on the right descriptor, and one way to do that would be to add a PhatomData "tag" to the `Pk` type that basically carries this information. For instance, the `MiniscriptKey` trait could be changed to take a type argument which we could call, let's say, "KeyState" and there would be two possibile "states", one called "Extended" and one called "Derived".

For key types that cannot be derived further by definition, only the "Derived" type would be implemented (i.e. `impl MiniscriptKey<Derived> for bitcoin::PublicKey`), while for types that can be derived like `DescriptorPublicKey` both would be implemented with a generic. Then, `ToPublicKey` would only be implemented for the `Derived` version, which would ensure that all the methods that need `to_public_key()` cannot possibly be called on a non-derived key. Other methods like `derive()` and `matches()` would instead only be available for `Descriptor`s where their key type is `Extended`, like: `impl<Pk: MiniscriptKey<Extended>> Descriptor<Pk> ... `. The `derive()` method would obviously be responsible for translating the keys from the `Extended` state to the `Derived` state.

I haven't actually tried implementing this concept (yet), so there's a good chance that this is either not useful at all, completely broken, or just not the best way to do it, but I'm curious to hear your feedback on this topic!

sanket1729:
> (and there are actual assertions in the code to ensure that)
Are there any assertions in the current master that could be causing panic? I couldn't find any in my first pass of the latest master. Or do you mean that you have some assertions downstream?. In any case, I like the overall idea on the surface to separate the operations on `Descriptor` before and after derivation.

Some discussion starter questions
> In this case, what are the advantages of having a generic instead of a separate return type, and are they worth the code complexity? `derive()` takes in an `ExtendedDescriptorPubkey` type and outputs `DerviedDescriptorPubKey` type. We implement `MiniscriptKey` for both and implement `ToPublicKey` for `DerivedDescriptorPubkey`.

[...]

PR #209, opened 2020-12-11
Remove ToPkCtx from length/satisfaction_weight

LLFourn:
To improve ergonomics and efficiency. You don't need to do an ecmult to figure out the length of your public key! Also should be possible on string descriptors now I guess?

### DIFF
--- a/examples/sign_multisig.rs
+++ b/examples/sign_multisig.rs
@@ -83,15 +83,15 @@ fn main() {
     let my_descriptor =
         BitcoinDescriptor::from_str(&descriptor_str[..]).expect("parse descriptor string");
 
-    // Sometimes it is necessary to have additional information to get the bitcoin::PublicKey
-    // from the MiniscriptKey which can supplied by `to_pk_ctx` parameter. For example,
-    // when calculating the script pubkey of a descriptor with xpubs, the secp context and
-    // child information maybe required.
-
     // Check weight for witness satisfaction cost ahead of time.
     // 4(scriptSig length of 0) + 1(witness stack size) + 106(serialized witnessScript)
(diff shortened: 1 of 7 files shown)


sanket1729 (approved):
We can this PR as is as there is a strict improvement over what we currently do. If we go along this same route, we will have to the same for XonlyPubkeys with methods like is_xonly(). Maybe we will need that to disallow 32-byte pks in other ScriptContext anyway.

sanket1729:
But strictly speaking, it is impossible to know the size without actually converting to a public key. 

Since many of the APIs do a conversion internally call a possibly heavy `to_public_key()` operation. Calling them multiple times can be inefficient without translating to `bitcoin::PublicKey` first can be inefficient. The following can be considered to avoid recomputation.

1) Convert the String/Generic Descriptor to Bitcoin::PublicKey(using `translate_fn` or `derive`) once and then call all the methods on it. (relates to #181 by @afilini )
2) Create some sort of cache pattern to avoid the repeated ecmult operation.

sanket1729:
Each of the functions and the doc example in lib.rs have a comment explanation for the parameter `ToPkCtx`. Can you also remove those?

LLFourn:
@sanket1729 I changed the position of one comment: https://github.com/rust-bitcoin/rust-miniscript/pull/209/commits/6b373378219b7ade954344f404fe0e1fe500fcf1 . Do you see others? All the ones I see  still make sense (they are documenting functions which take the parameter).

sanket1729:
Sorry, that was all :)

apoelstra (approved):
ack 6b373378219b7ade954344f404fe0e1fe500fcf1

PR #213, opened 2020-12-16
Descriptor struct overhaul

sanket1729:
As discussed, this refactors descriptor module. We have a `DescriptorTrait` which we implement for all the basic descriptors. The internal descriptor structs are also more organized into separate files and clean implementations with code reuse where things make sense. 

Only the following should change for library user:
1) Instead of creating a Descriptor from an enum, the user now has to call the `new` functions for creating descriptors. 
2) Users would have to import the trait `DescriptorTrait` to use the functions.


src/descriptor/mod.rs
```diff
@@ -66,6 +66,170 @@ pub use self::key::{
...
+    /// from MiniscriptKey using [ToPublicKey].
+    /// If MiniscriptKey is already is [bitcoin::PublicKey], then the context
+    /// would be [NullCtx] and [descriptor.DescriptorPublicKeyCtx] if MiniscriptKey is [descriptor.DescriptorPublicKey]
+    ///
+    /// In general, this is defined by generic for the trait [trait.ToPublicKey]
+    fn address<ToPkCtx: Copy>(
+        &self,
+        to_pk_ctx: ToPkCtx,
+        network: bitcoin::Network,
+    ) -> Option<bitcoin::Address>
```
darosior, line 102: When would we ever return None ? Especially since below we don't return an Option for the scriptPubKey.

darosior (commented):
Concept ACK, even if that means YA big refactor downstream :( 

Quickly skimmed through it and left a few comment. Will review it more in depth soon :tm:. Overall it really is clean and neat

src/descriptor/mod.rs
sanket1729: In the case of Bare Descriptor, we return `None`. But as another PR, after this is merged I plan to change all option returning APIs to return Results.

dr-orlovsky:
Looks nice to me, as a concept! (and yes, yet another big refactoring, but it worth it)

Have you thought about providing a simple enumeration with descriptor types? Some software may present with options which descriptors are supported, and having an enumeration w/o embedded descriptor data may be useful. Also a `DesctiptorTrait::descriptor_type` method returning that contentless type may be useful

sanket1729:
@dr-orlovsky, I like the idea of the descriptor type enum without the content. I think we should do something like this.

dr-orlovsky:
@sanket1729 let me work on the PR (I will make it against the master, but it will be easy to rebase on top of it or merge it into your PR). I had a lot of requirements for the functionality such descriptor types and I have already did a few attempts in different projects like [Bitcoin Pro](https://github.com/pandoracore/bitcoin-pro)

src/descriptor/mod.rs
```diff
@@ -66,6 +66,170 @@ pub use self::key::{
...
+/// A general trait for Bitcoin descriptor.
+/// It should also support FromStr, fmt::Display and should be liftable
+/// to bitcoin Semantic Policy.
+/// Offers function for witness cost estimation, script pubkey creation
+/// satisfaction using the [Satisfier] trait.
+// Unfortunately, the translation function cannot be added to trait
+// because of traits cannot know underlying generic of Self.
+// Thus, we must implement additional trait for translate function
+pub trait DescriptorTrait<Pk: MiniscriptKey>:
+    FromStr + Display + Debug + Clone + Eq + PartialEq + PartialOrd + Ord + Liftable<Pk>
```
apoelstra, line 79: We can probably drop most (all?) of these requirements.

src/descriptor/mod.rs
```diff
@@ -66,6 +66,170 @@ pub use self::key::{
...
+    /// script before any hashing is done. For `Bare`, `Pkh` and `Wpkh` this
+    /// is the scriptPubkey; for `ShWpkh` and `Sh` this is the redeemScript;
+    /// for the others it is the witness script.
+    /// `to_pk_ctx` denotes the ToPkCtx required for deriving bitcoin::PublicKey
+    /// from MiniscriptKey using [ToPublicKey].
+    /// If MiniscriptKey is already is [bitcoin::PublicKey], then the context
+    /// would be [NullCtx] and [descriptor.DescriptorPublicKeyCtx] if MiniscriptKey is [descriptor.DescriptorPublicKey]
+    ///
+    /// In general, this is defined by generic for the trait [ToPublicKey]
+    fn witness_script<ToPkCtx: Copy>(&self, to_pk_ctx: ToPkCtx) -> Script
```
apoelstra, line 145: Should name this differently, say `explicit_script` or something ... and it should also be a `Result` since Taproot descriptors won't have anything meaningful to put here.

src/descriptor/mod.rs
```diff
@@ -66,6 +66,170 @@ pub use self::key::{
...
+    where
+        Pk: ToPublicKey<ToPkCtx>;
+}
+
+/// Convert a descriptor using abstract keys to one using specific keys
+/// This will panic if translatefpk returns an uncompressed key when
+/// converting to a Segwit descriptor. To prevent this panic, ensure
+/// translatefpk returns an error in this case instead.
+pub trait PkTranslate<P: MiniscriptKey, Q: MiniscriptKey> {
+    /// The associated output type. This must be Self<Q>
```
apoelstra, line 215: s/must/should/

src/descriptor/mod.rs
sanket1729: I think we should leave this to other changes in Taproot descriptors. We may understand more things as we understand and implement them.

src/descriptor/mod.rs
sanket1729: I removed the bounds. But unfortunately, removing them did not help that much with code ergonomics.

apoelstra (approved):
ack f8a9478349ad660fdae1c7d4931daeb6deb0250e

PR #216, opened 2021-01-02
remove `ToPkCtx`, allow hardened derivation steps (and wildcards) in `DescriptorPublicKey`

apoelstra:


src/miniscript/astelem.rs
```diff
@@ -25,6 +25,7 @@ use bitcoin::blockdata::{opcodes, script};
 use bitcoin::hashes::hex::FromHex;
 use bitcoin::hashes::{hash160, ripemd160, sha256, sha256d, Hash};
 
+use descriptor::PkTranslate;
```
sanket1729: If we are using PkTranslate for underlying Miniscript and Terminal, maybe we should just move the trait to lib.rs from descirptor/mod.rs

src/descriptor/mod.rs
```diff
@@ -234,13 +233,68 @@ pub trait PkTranslate<P: MiniscriptKey, Q: MiniscriptKey> {
...
+    }
+}
+impl<P: MiniscriptKey, Q: MiniscriptKey<Hash = P::Hash>, T: PkTranslate<P, Q>> PkTranslate1<P, Q>
+    for T
+{
+}
+
+/// Variant of `PkTranslate` where P's hash is P, so the hashes
+/// can be converted by reusing the key-conversion function
+pub trait PkTranslate2<P: MiniscriptKey<Hash = P>, Q: MiniscriptKey>: PkTranslate<P, Q> {
```
sanket1729: I know you probably thought of this :P , but should we give these full names? I think everything we write translate downstream, we will wonder what 1 or 2 stands for.

src/descriptor/key.rs
```diff
@@ -112,7 +125,7 @@ pub struct DescriptorXKey<K: InnerXKey> {
     /// The derivation path
     pub derivation_path: bip32::DerivationPath,
     /// Whether the descriptor is wildcard
-    pub is_wildcard: bool,
+    pub is_wildcard: Wildcard,
```
sanket1729, line 126: nit: Change to `wildcard` instead of `is_wildcard` since this is no longer a bool?

src/descriptor/mod.rs
```diff
@@ -224,8 +224,28 @@ pub trait PkTranslate<P: MiniscriptKey, Q: MiniscriptKey> {
...
+        Fpk: FnMut(&P) -> Q,
+        Fpkh: FnMut(&P::Hash) -> Q::Hash,
+        Q: MiniscriptKey,
+    {
+        self.translate_pk::<_, _, ()>(|pk| Ok(translatefpk(pk)), |pkh| Ok(translatefpkh(pkh)))
+            .expect("infallible translation function")
+    }
 }
 
+// There are some additional convenience functions we could add to  `PkTranslate`
```
apoelstra: should remove this comment, it's outdated (I did figure out how to add the convenience functions after all)

src/descriptor/key.rs
```diff
@@ -560,10 +576,10 @@ impl<'secp, C: secp256k1::Verification> Copy for DescriptorPublicKeyCtx<'secp, C
 
 impl<'secp, C: secp256k1::Verification> DescriptorPublicKeyCtx<'secp, C> {
     /// Create a new context
-    pub fn new(secp_ctx: &'secp secp256k1::Secp256k1<C>, child_number: bip32::ChildNumber) -> Self {
+    pub fn new(secp_ctx: &'secp secp256k1::Secp256k1<C>, index: u32) -> Self {
         Self {
             secp_ctx: secp_ctx,
-            child_number: child_number,
+            index: index,
```
sanket1729: Should we panic here or return a result in order to avoid the panic of index >= 2^31 when calling derive.

src/descriptor/mod.rs
```diff
@@ -661,9 +660,9 @@ where
 }
 
 impl Descriptor<DescriptorPublicKey> {
-    /// Derives all wildcard keys in the descriptor using the supplied `child_number`
-    pub fn derive(&self, child_number: bip32::ChildNumber) -> Descriptor<DescriptorPublicKey> {
-        self.translate_pk2_infallible(|pk| pk.clone().derive(child_number))
+    /// Derives all wildcard keys in the descriptor using the supplied index
+    pub fn derive(&self, index: u32) -> Descriptor<DescriptorPublicKey> {
```
sanket1729, line 562: We should add a doccoment about index panic since this will function will be used a lot downstream.

sanket1729 (commented):
Reviewed the first few commits. Only minor nits so far. Will review the rest later. I think based on ToPkCtx, I will also have to update the elements miniscript.

src/lib.rs
```diff
@@ -186,40 +186,30 @@ impl MiniscriptKey for String {
...
-pub trait ToPublicKey<ToPkCtx: Copy>: MiniscriptKey {
+pub trait ToPublicKey: MiniscriptKey {
     /// Converts an object to a public key
     /// C represents additional context information that maybe
     /// required for deriving a bitcoin::PublicKey from MiniscriptKey
     /// You may require secp context for crypto operations
     /// or additional information for substituting the wildcard in
     /// extended pubkeys
-    fn to_public_key(&self, to_pk_ctx: ToPkCtx) -> bitcoin::PublicKey;
+    fn to_public_key(&self) -> bitcoin::PublicKey;
```
sanket1729, line 197: nit: remove comment above this function

sanket1729 (dismissed):
tACK b7d453647012011b176953f99aedfdd29c4b5489 . 
Surprisingly easy to review despite the large diff :). I like that cleaner API over all the ToPkCtx stuff we had.

sanket1729:
The title of the PR says `WalletPublicKey` but codebase has `DescriptorPublicKey`. Did you intend to rename it?

src/descriptor/key.rs
apoelstra: I think we should panic. It's very easy for the caller to sanity-check that their numbers are in range, and very annoying to have to unwrap in the 99% of cases when you know you're in range.

src/descriptor/mod.rs
apoelstra: My reasoning was roughly:
* There is readability value in having short names, that you have to trade off against the readability of being explicit
* Usually, the specific `translate_pk` method being called is irrelevant or clear from context

apoelstra:
lol, I'll fix the title. In my head I constantly think of `WalletPublicKey` as the more obvious name, but no, I don't intend to change it.

sanket1729 (approved):
tAck 9646ad8878480000e602891b4b679f30e3309020

PR #217, opened 2021-01-03
Change options to result in all APIs

sanket1729:
~Build on top of 216. Will rebase after #216 is merged.~
Ready for review

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -81,7 +81,8 @@ pub trait DescriptorTrait<Pk: MiniscriptKey> {
     fn sanity_check(&self) -> Result<(), Error>;
 
     /// Computes the Bitcoin address of the descriptor, if one exists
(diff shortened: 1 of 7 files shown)


apoelstra (approved):
ack e340ad88342476bb7e12f2ccbab6634fd2c7e16f

Checked that all of the remaining `Option` returns appear to be legitimately options.

src/descriptor/mod.rs
```diff
@@ -668,7 +670,12 @@ mod tests {
                 "512102000000000000000000000000000000000000000000000000000000000000000251ae"
             )
         );
-        assert_eq!(bare.address(bitcoin::Network::Bitcoin,), None);
+        assert_eq!(
+            bare.address(bitcoin::Network::Bitcoin)
+                .unwrap_err()
+                .to_string(),
+            "Bare descriptors don't have address"
```
RCasatta, line 677: maybe comparing to `Error::BareDescriptorAddr.to_string()` is slightly better than repeating the string

PR #220, opened 2021-01-08
remove FromStr bound on MiniscriptKey and MiniscriptKey::Hash

apoelstra:
Without this bound you can't deserialize, but there are lots of in-memory things you might like to do with weird keytypes (e.g. looking up a bunch of aux data from a database and tagging keys with them, or even attaching an `&'a wallet`) where the requirement to impl `FromStr` is extremely onerous.


sanket1729 (approved):
tACK e5b8a73de403426a05977e227ad229ffe614d416

PR #222, opened 2021-01-11
Pretty print a policy

sanket1729:
Adds methods to pretty print policies similar to directory tree printing in Linux. Prints diff in colors which highlights the differences between the two policies. 

Our representation of Semantic policies in only in Thresh, so we combine `or(pk,or(pk,pk))` into `thresh(1,pk,pk,pk)`. And so we use the latter in comparing two policies that complicate some of the implementation but is still fairly small.

### DIFF
--- a/src/policy/semantic.rs
+++ b/src/policy/semantic.rs
@@ -57,6 +57,52 @@ pub enum Policy<Pk: MiniscriptKey> {
     Threshold(usize, Vec<Policy<Pk>>),
 }
 
+/// Compute the Policy difference between two policies.
+/// This is useful when trying to find out the conditions
+/// under which the two policies are different.
+#[derive(Debug, Clone)]
+pub struct PolicyDiff<Pk: MiniscriptKey> {
+    /// The first policy
+    pub a: Vec<Policy<Pk>>,
+    /// The second policy
+    pub b: Vec<Policy<Pk>>,
+}
+
+impl<Pk: MiniscriptKey> PolicyDiff<Pk> {
+    /// Combine two policy differences.
+    // Policies should not generally contain repeated conditions,
+    // therefore we do not many attempt to deal with those.
+    pub fn combine(&mut self, second: Self) {
+        self.a.extend(second.a);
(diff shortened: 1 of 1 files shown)


src/policy/semantic.rs
```diff
@@ -187,6 +216,67 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+                            j = j + 1;
+                        }
+                        sub_a._pprint_diff(&subs2[j], prefix.to_string(), i == last_child);
+                    }
+                }
+            }
+            (x, y) => {
+                let red = "\x1b[0;31m";
+                let nc = "\x1b[0m";
+                let green = "\x1b[0;32m";
```
apoelstra, line 265: I think we need to come up with a different encoding scheme. It doesn't have to be pretty, we could provide a shellscript or something that converted it into a colorized version.

But using shell escape sequences means that we won't work on windows and are likely to confuse other command line tools.

src/policy/semantic.rs
```diff
@@ -78,6 +124,42 @@ impl<Pk: MiniscriptKey> ForEachKey<Pk> for Policy<Pk> {
...
+            let last_child = subs.len() - 1;
+
+            for (i, child) in subs.iter().enumerate() {
+                child._pprint_tree(prefix.to_string(), i == last_child);
+            }
+        }
+    }
+
+    /// Pretty Print a tree
+    pub fn pprint_tree(&self) {
```
apoelstra, line 189: In 4d23b799f6d066358f2bbe3792981e43db82f417:

Can we make this `pub(crate)`? I think we can come up with a much better API here and I'd rather not expose this one.

portlandhodl:
My shot at this PR if anyone is interested. Though I'm having a hard time seeing a solid use case.

https://github.com/rust-bitcoin/rust-miniscript/compare/master...portlandhodl:rust-miniscript:policy-diff

apoelstra:
e82943462513bcbc6699c9e120a708c7aad48717 needs rebase

PR #225, opened 2021-01-14
release 5.0, increase rust-bitcoin dep to 0.26

apoelstra:


src/descriptor/key.rs
```diff
@@ -360,7 +360,10 @@ impl DescriptorPublicKey {
                     fingerprint
                 } else {
                     let mut engine = XpubIdentifier::engine();
-                    single.key.write_into(&mut engine);
+                    single
+                        .key
+                        .write_into(&mut engine)
+                        .expect("engines don't error");
                     bip32::Fingerprint::from(&XpubIdentifier::from_engine(engine)[..])
```
dr-orlovsky, line 367: correct me if I'm wrong, but if we do a single write it is the same as using `::hash()` directly without engine. This will collapse these 6 lines into a single one

src/lib.rs
```diff
@@ -165,7 +165,7 @@ impl MiniscriptKey for bitcoin::PublicKey {
 
     fn to_pubkeyhash(&self) -> Self::Hash {
         let mut engine = hash160::Hash::engine();
-        self.write_into(&mut engine);
+        self.write_into(&mut engine).expect("engines don't error");
         hash160::Hash::from_engine(engine)
```
dr-orlovsky, line 169: Ibid

dr-orlovsky (commented):
Not sure, but it seems this could be simplified with direct use of `HashType::hash` function instead of engines.

src/descriptor/key.rs
apoelstra: yeah, good call

apoelstra: actually no, because `hash` takes a &[u8] not a key

dr-orlovsky: with `.serialize()` you can convert a key into a `&[u8]` w/o allocation

dr-orlovsky: (and this reminds me of the fact that it may be nice to have `::hash(data: impl AsRef<[u8]>)` and do a `impl AsRef<[u8]>` for keys)

sanket1729: We would require a branch for `serialize()` or `serialize_uncompressed()`. Note that serialize always serializes to 33 bytes, and SinglePub can be uncompressed.

apoelstra: there is no such method `.serialize()` on bitcoin::PublicKey https://docs.rs/bitcoin/0.26.0/bitcoin/util/key/struct.PublicKey.html

sanket1729:
tACK a59bd5e5077c5e8c2075fa2f260f249a727d7475. We can wait sometime for @dr-orlovsky to resolve his conversation, but I don't think we can do better.

sanket1729 (approved):

src/descriptor/key.rs
dr-orlovsky: @apoelstra correct, I confused it with `secp156k1` version. I've being laready two years into this and still got confused about API difference between both these versions, which may probably mean that they are really confusing and non-intuitive :)

@sanket1729: wrtie_into does exactly the same: it takes branch for serialize() or serialize_uncompressed() on secp_key. So at the end of the day we still use `serialize` method of secp256k1, and why not to have it exposed at `bitcoin::Pubkey` than to avoid such situations like here?

Anyway, I think there is not much that can be done in miniscript, so I will open an issue in rust-bitcoin

dr-orlovsky (approved):

PR #227, opened 2021-01-14
Add psbt preimage satisfaction logic

sanket1729:
Required for Integration tests in #162 .

### DIFF
--- a/src/psbt/mod.rs
+++ b/src/psbt/mod.rs
@@ -22,6 +22,7 @@
 use std::{error, fmt};
 
 use bitcoin;
+use bitcoin::hashes::{hash160, ripemd160, sha256, sha256d};
 use bitcoin::secp256k1::{self, Secp256k1};
 use bitcoin::util::psbt::PartiallySignedTransaction as Psbt;
 use bitcoin::Script;
@@ -29,8 +30,8 @@ use bitcoin::Script;
 use interpreter;
 use miniscript::limits::SEQUENCE_LOCKTIME_DISABLE_FLAG;
 use miniscript::satisfy::{bitcoinsig_from_rawsig, After, Older};
-use BitcoinSig;
 use Satisfier;
+use {BitcoinSig, Preimage32};
 use {MiniscriptKey, ToPublicKey};
 
 mod finalizer;
@@ -278,6 +279,44 @@ impl<'psbt, Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for PsbtInputSatisfie
         } else {
             <Satisfier<Pk>>::check_older(&Older(seq), n)
         }
+    }
+
+    fn lookup_hash160(&self, h: hash160::Hash) -> Option<Preimage32> {
+        self.psbt.inputs[self.index]
+            .hash160_preimages
+            .get(&h)
+            .and_then(try_vec_as_preimage32)
+    }
+
+    fn lookup_sha256(&self, h: sha256::Hash) -> Option<Preimage32> {
+        self.psbt.inputs[self.index]
+            .sha256_preimages
+            .get(&h)
+            .and_then(try_vec_as_preimage32)
+    }
+
+    fn lookup_hash256(&self, h: sha256d::Hash) -> Option<Preimage32> {
+        self.psbt.inputs[self.index]
+            .hash256_preimages
+            .get(&h)
+            .and_then(try_vec_as_preimage32)
+    }
+
+    fn lookup_ripemd160(&self, h: ripemd160::Hash) -> Option<Preimage32> {
+        self.psbt.inputs[self.index]
+            .ripemd160_preimages
+            .get(&h)
+            .and_then(try_vec_as_preimage32)
+    }
+}
+
+fn try_vec_as_preimage32(vec: &Vec<u8>) -> Option<Preimage32> {
+    if vec.len() == 32 {
+        let mut arr = [0u8; 32];
+        arr.copy_from_slice(&vec);
+        Some(arr)
+    } else {
+        None
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -127,7 +127,7 @@ pub use descriptor::{Descriptor, DescriptorPublicKey, DescriptorTrait};
 pub use interpreter::Interpreter;
 pub use miniscript::context::{BareCtx, Legacy, ScriptContext, Segwitv0};
 pub use miniscript::decode::Terminal;
-pub use miniscript::satisfy::{BitcoinSig, Satisfier};
+pub use miniscript::satisfy::{BitcoinSig, Preimage32, Satisfier};
 pub use miniscript::Miniscript;
 
 ///Public key trait which can be converted to Hash type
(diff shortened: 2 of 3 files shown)


src/psbt/mod.rs
```diff
@@ -279,6 +280,37 @@ impl<'psbt, Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for PsbtInputSatisfie
             <Satisfier<Pk>>::check_older(&Older(seq), n)
         }
     }
+
+    fn lookup_hash160(&self, h: hash160::Hash) -> Option<[u8; 32]> {
```
dr-orlovsky, line 284: In fact I think we need an introduction of a preimage type like in https://github.com/LNP-BP/rust-lnpbp/blob/master/src/bp/hlc.rs#L25-L60 — but I assume this is outside of the scope of the present PR

dr-orlovsky: It's a pity we do not have a const generics with our MSRV, otherwise we can generalize the preimage type.

src/psbt/mod.rs
```diff
@@ -279,6 +280,37 @@ impl<'psbt, Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for PsbtInputSatisfie
...
+    }
+
+    fn lookup_ripemd160(&self, h: ripemd160::Hash) -> Option<[u8; 32]> {
+        lookup_preimage(&self.psbt.inputs[self.index].ripemd160_preimages, &h)
+    }
+}
+
+// Lookup a [u8;32] preimage from a psbt hashmap
+// Returns Some() only if the preimage is 32 bytes
+fn lookup_preimage<T: Eq + Ord>(map: &BTreeMap<T, Vec<u8>>, key: &T) -> Option<[u8; 32]> {
```
dr-orlovsky, line 303: Not sure why we do not use here types available in the current PSBT implementation

dr-orlovsky (commented):
No particular opinion: this looks like a rather temporary solution

src/psbt/mod.rs
sanket1729: I don't follow this, Which type are you referring to? 

This is an internal function to avoid repetition of code in 4 satisfier definitions.

src/psbt/mod.rs
sanket1729: Yeah, right now these function signatures are decided by the satisfier Trait. We could have a new type for PreImage.

src/psbt/mod.rs
dr-orlovsky: What I am trying to say is that a hash preimage is a _type_. `[u8; 32]` is much more general (and used in many other circumstances than hash preimages), and a newtype can make a better job here — however it is not highly important if this is just a temporary stub.

sanket1729: With the new BIP 174, we added fields for looking up preimages given hash. This is not a proprietary key. This function tells the Satisfier where to lookup for Preimages which it can use to construct satisfaction for generic Miniscripts. 

This is a generic lookup function cover the cases for `BTreeMap<sha256::Hash, Vec<u8>>`, `BTreeMap<sha256d::Hash, Vec<u8>>`, `BTreeMap<ripemd160::Hash, Vec<u8>>`, `BTreeMap<has160::Hash, Vec<u8>>`, into a single lookup function to avoid code repeatation.

dr-orlovsky: Ok, I was really confused b/c `map` in terms of PSBT/BIP 174 is either global, input or output map — and not a "map" inside a specific key values from PSBT map. Also "key" is the PSBT map key and not a hash of some secret.

dr-orlovsky: Can we do

```suggestion
fn lookup_preimage<T: bitcoin::Hash>(hash_preimages: &BTreeMap<T, Vec<u8>>, hash: &T) -> Option<Preimage> {
```

with `type Preimage = [u8; 32]` at least to help the reader/user to distinguishing `[u8; 32]` preimage from hash of the same dimension?

However, if I am right in a way I understand this function, I do not get what it does aside from the fact of looking up value for a key in a really odd fashion...

sanket1729: Yes, it does lookup in a really weird fashion. The reason for that is 
1) PSBT spec does not enforce len = 32 on preimage, it is a constraint of miniscript. 
2) We get a `&[u8]` slice from the PSBT lookup, which we need to convert to [u8;32] if it has 32 length. If there is a more idiomatic MSRV compatible way to do this, please let me know. 

> However, if I am right in a way I understand this function, I do not get what it does aside from looking up value for a key in a really odd fashion...

This code of creating the 32 byte array from `&[u8]` slice if was len=32 was repeated 4 functions, so I forced it into a function. 

Nice idea with the Preimage32 type, I will add to the trait definition.

sanket1729: Add a commit changing `[u8; 32]` to `Preimage32` in the `Satisfier` trait and in the impl for `psbt::InputSatisfier` for more readability.

dr-orlovsky: Ok, now much clearer.

Well, there is a function to copy slice into array: https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice

So what I would do here is to get rid of this function and use pretty simple line for looking up hashmap + copying from slice right inline. Moving that single line into a function makes code more obscure and complicated.

dr-orlovsky: This is my take on the problem: https://github.com/sanket1729/rust-miniscript/pull/2/files

Seems like we still need a separate fn to avoid duplication, but it is rather tiny and trivial in this case

sanket1729: Thanks, this is approach is much cleaner. Picked and squashed your commit and added as co-author.

dr-orlovsky (approved):
tACK with a single note: new type is always better than the type alias, however considering that here solution is temporary (32-byte hash preimage is just a single case of all possible preimages) and we do not need to avoid using normal `[u8; 32]` arrays as `Preimage32`, the choice of type alias over newtype is appropriate.

sanket1729:
> however considering that here solution is temporary (32-byte hash preimage is just a single case of all possible preimages)

Miniscript by design is based on assuming that preimages are 32 bytes. Otherwise, all of the estimations of satisfaction weight are hard to do as we can't know anything about the preimage. 
I doubt that this library will ever support satisfaction for non-32 byte preimages.

dr-orlovsky:
Than, making `Preimage32` a newtype sounds more native - but I will ask @apoelstra to decide upon this matter

apoelstra:
I think I'm okay leaving it as a type alias. Agreed that a newtype would be better for type safety, but given that a preimage is just an unstructured pile of bytes (and in certain applications, may have meaning beyond just "being a preimage"), I'm not sure that we gain a **lot** in exchange for the reduced ergonomics and all the extra code that we'd have to write.

I also worry about winding up with something similar to rust-secp's `Message` type which is incredibly annoying to use and pollutes basically every call to sign or verify with an (IMO) pointless conversion from a hash.

apoelstra (approved):
ack b9895119e1442422656768efa65b0b07b252d6f2

apoelstra:
oops, I mean ack c267633798015995e0363b1e4a29d1711e8ea22b

PR #228, opened 2021-01-19
Fix checksum for nested sh(wpkh())

shesek:
Only include it for the outer descriptor and not for the inner one.

This was already considered for `sh(wsh(..))`, but not for `sh(wpkh(..))`, which resulted in `sh(wpkh(<key>)#<checksum1>)#<checksum2>`.

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -1388,4 +1388,15 @@ pk(03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8))";
         // try to turn it back into a string with the secrets
         assert_eq!(descriptor_str, descriptor.to_string_with_secret(&keymap));
     }
+
+    #[test]
+    fn checksum_for_nested_sh() {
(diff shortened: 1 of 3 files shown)


shesek:
I added a `to_string_no_checksum()` method to nestable descriptors, to avoid repeating it in the `Sh` implementation as [currently done](https://github.com/rust-bitcoin/rust-miniscript/blob/48c0ff8cf333f1eeea9389d2506ff2364388e267/src/descriptor/sh.rs#L85-L90) for `sh(wsh())`. Let me know if you prefer the previous way this was handled.

sanket1729 (approved):
Thanks for doing this. I missed it for the nested `wpkh` case. 

I think we can add `to_string_no_checksum` to the `DescriptorTrait` itself. I can do that as a part of another PR, this PR is good to go.

tACK ee0add31e0be934bbd12581f01fcdb733fc61d2e

apoelstra (approved):
ack ee0add31e0be934bbd12581f01fcdb733fc61d2e

I'm not thrilled about the multiple `String` creations only to be `format!`ed, but there were a bunch of these before, and anyway we can revisit this later whenever we want to do some optimizations.

PR #229, opened 2021-01-22
Satisfaction bugs

sanket1729:
The first one is a serious one which caused incorrect results. 

~~While the second one is a relaxation of rules that allow dissatisfactions to be malleable. Because of this, earlier we were not able to satisfy a miniscript that required ORI dissatisfaction where both of the available dissatisfactions were malleable. (which is almost always the case).~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "miniscript"
-version = "5.0.0"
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ack 5192b8cd5260f89c569cadc5177cafd0839e384c

Issue #165, opened 2020-11-04
Support for Non Standard Miniscripts


sanket1729:
Miniscript language is designed to be standardness complete, however in certain applications like block explorer may require the ability to parse non-standard Miniscripts. 

However, at this point, the current solution to change Miniscript ScriptContext from Generic Type to a parameter in every function seems overly invasive and is not worth the change as lose the strong type guarantees of Miniscript. (For example, using uncompressed keys in segwit context). 

Considering the tradeoff that we maybe gaining very little compared by supporting non-standard miniscripts, and that there would be scripts on the blockchain that are not miniscripts which we would not be able to parse anyways, it is probably not worth it to support for those.

Marking this as brainstorming and a low priority issue for now. If there are additional usecases for converting `Miniscript<MiniscriptKey, ScriptContext>` to `Miniscript<MiniscriptKey>` we can consider implementing the changes.

mmm0171:
0x515B0941aFc4Db1F2D20f2379323a44Edb3c4460

PR #233, opened 2021-02-03
Explicitly error on non-standard sighash types

darosior:
I forgot to append the ANYONECANPAY flag with my boutique signer for hand-testing and spent some time before realising we were actually treating the last byte of the sig as a valid SIGHASH_ALL in the finalizer !
I believe it's cleaner to error early on too short signatures.

EDIT (made after the comments below): actually if i was just using SIGHASH_ALL i could have messed up my sighash type and not notifying it. Even though checking this is on the user of the lib, i believe it would be much nicer to not silently accept invalid-by-standardness signatures in rust-miniscript.

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -757,7 +757,8 @@ where
     F: FnOnce(&bitcoin::PublicKey, BitcoinSig) -> bool,
 {
     if let Some((sighash_byte, sig)) = sigser.split_last() {
-        let sighashtype = bitcoin::SigHashType::from_u32(*sighash_byte as u32);
+        let sighashtype = bitcoin::SigHashType::from_u32_standard(*sighash_byte as u32)
+            .map_err(|_| Error::NonStandardSigHash([sig, &[*sighash_byte]].concat().to_vec()))?;
         let sig = secp256k1::Signature::from_der(sig)?;
         if verify_sig(pk, (sig, sighashtype)) {
             Ok(sig)
--- a/src/psbt/finalizer.rs
+++ b/src/psbt/finalizer.rs
@@ -262,7 +262,14 @@ pub fn finalize<C: secp256k1::Verification>(
                 ));
             }
             let (flag, sig) = rawsig.split_last().unwrap();
-            let flag = bitcoin::SigHashType::from_u32(*flag as u32);
+            let flag = bitcoin::SigHashType::from_u32_standard(*flag as u32).map_err(|_| {
+                super::Error::InputError(
+                    InputError::Interpreter(interpreter::Error::NonStandardSigHash(
+                        [sig, &[*flag]].concat().to_vec(),
+                    )),
+                    n,
+                )
+            })?;
             if target != flag {
                 return Err(Error::InputError(
                     InputError::WrongSigHashFlag {
--- a/src/interpreter/error.rs
+++ b/src/interpreter/error.rs
@@ -12,7 +12,7 @@
 // If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
 //
 
-use bitcoin::hashes::hash160;
+use bitcoin::hashes::{hash160, hex::ToHex};
 use bitcoin::{self, secp256k1};
 use std::{error, fmt};
 
@@ -39,6 +39,8 @@ pub enum Error {
     InsufficientSignaturesMultiSig,
     /// Signature failed to verify
     InvalidSignature(bitcoin::PublicKey),
+    /// Last byte of this signature isn't a standard sighash type
+    NonStandardSigHash(Vec<u8>),
     /// Miniscript error
     Miniscript(::Error),
     /// MultiSig requires 1 extra zero element apart from the `k` signatures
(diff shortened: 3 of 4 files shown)


src/psbt/finalizer.rs
```diff
@@ -252,7 +252,9 @@ pub fn finalize<C: secp256k1::Verification>(
     for (n, input) in psbt.inputs.iter().enumerate() {
         let target = input.sighash_type.unwrap_or(bitcoin::SigHashType::All);
         for (key, rawsig) in &input.partial_sigs {
-            if rawsig.is_empty() {
+            // If it comports a SIGHASH flag, it must at least be 71-bytes long.
+            // If it doesn't, it's invalid. FIXME: it's only valid for ECDSA sigs.
+            if rawsig.len() < 71 {
```
sanket1729, line 257: I think it is also possible to have short signatures in bitcoin. I think it is possible to have shorter signatures than 71 bytes. ECDSA.  
https://github.com/bitcoin/bitcoin/blob/ea5a50f92a6ff81b1d2dd67cdc3663e0e66733ac/src/secp256k1/src/ecdsa_impl.h#L189-L190

I think with a probability of 1/1024 we can have sigs of 70 bytes. 1/(1024*256) 69 bytes and so on. Infact, Rusell from Blockstream has used sigs as short as `300602010102010101`   on testnet.

darosior: It's valid by consensus but i'd have thought it's non standard ? Would have to check though.

sanket1729: I don't think there should be a standardness rule against it. 70 bytes sigs can occur naturally with a 0.1% chance. I could not find one in bitcoin core `policy.h`, but I have had cases something was not included `policy.h` .

apoelstra: There is no standardness rule about the length of signatures. Why would it be illegal to try to save some bytes?

Without parsing DER it is impossible to tell what the "real" length of an ECDSA signature is.

apoelstra:
An alternate strategy would be to just try and parse the signature with the "sighash byte" removed and fail early if that doesn't work.

darosior:
Yes, can try this approach. Alternatively we could check the hashtype is standard: what do you think about adding a new method to `SigHashType` upstream which would fail on non-standard types (something like `from_u32_standard`)?

sanket1729:
> An alternate strategy would be to just try and parse the signature with the "sighash byte" removed and fail early if that doesn't work.

We already do that in the current finalizer code. Maybe this catches all errors as it unlikely that we form another valid der formatted secp signature if we remove one byte. 

Treating unknown flags as all must have a backward compatibility good reason, that I can't think of right now. 
https://github.com/rust-bitcoin/rust-bitcoin/blob/5bd61967b231939626dd847fd82cc5da33052dc1/src/blockdata/transaction.rs#L676-L689
 
I think for this case, it is useful to have a method like from_u32_standard.

darosior:
> Treating unknown flags as all must have a backward compatibility good reason

Yep, this is consensus behaviour (but not standard!).

> I think for this case, it is useful to have a method like from_u32_standard.

Will PR it upstream, should be trivial.

sanket1729:
Yeah, bitcoin core has a`IsDefinedHashtypeSignature`, but I think this may be context dependant because, in tapscript, we have a SIGHASH_DEFAULT too. But we can think about that later.

darosior:
Done here https://github.com/rust-bitcoin/rust-bitcoin/pull/573

> Yeah, bitcoin core has aIsDefinedHashtypeSignature, but I think this may be context dependant because, in tapscript, we have a SIGHASH_DEFAULT too. But we can think about that later.

Yes. Bitcoin-Core has `EvalChecksigPreTapscript` which indirectly call this check, for Tapscript it's checked in `SignatureHashSchnorr`.

sanket1729:
Should we close this PR?

darosior:
iirc i was waiting for a release of rust-bitcoin. Will check and close if needed

darosior:
No, yeah i'm waiting for a rust-bicoin release since https://github.com/rust-bitcoin/rust-miniscript/pull/233#issuecomment-779695097

apoelstra:
Is the recent 0.26.2 release sufficient?

darosior:
Yes, just pushed an update that leverages the new `from_u32_standard` from https://github.com/rust-bitcoin/rust-bitcoin/pull/573 instead.

src/interpreter/mod.rs
```diff
@@ -757,7 +757,8 @@ where
     F: FnOnce(&bitcoin::PublicKey, BitcoinSig) -> bool,
 {
     if let Some((sighash_byte, sig)) = sigser.split_last() {
-        let sighashtype = bitcoin::SigHashType::from_u32(*sighash_byte as u32);
+        let sighashtype = bitcoin::SigHashType::from_u32_standard(*sighash_byte as u32)
```
sanket1729, line 760: I wonder if the interpreter should use the `from_u32_consensus`. It is used to interpret the transactions that are already present in the blockchain. The main use is for block explorers which should accept non-standard transactions

darosior: Hmm, but i thought we shouldn't be parsing invalid-by-standardness Script? Or is it called by `parse_insane_*`?

sanket1729: The goal of `parse_insane` was to parse non-safe miniscripts or miniscripts containing duplicate keys or miniscripts possibly exceeding resource limits while satisfaction. 

`parse_insane` does not currently support parsing non-standard scripts, so what you are doing in this PR seems correct. Having a non-standard miniscript is out of scope for now. This is can be addressed later(while we address #165)

sanket1729 (approved):
utACk b8432e11ae824ca2a906e80fd414cdb6f1e9e1a3 . We should not waste review cycles for non-standard support because
1) It is a low-priority feature
2) It is currently inconsistently supported/partially implemented currently.

apoelstra:
ACK b8432e11ae824ca2a906e80fd414cdb6f1e9e1a3

PR #234, opened 2021-02-16
Workaround rust compiler goof in Policy::translate_pk

LLFourn:
These previously untested methods result in a compiler error if you ever tried to call them. This happens because if you put &mut in front of the type when you recursively pass it to another call you are actually creating a &mut &mut since the argument may already be a &mut. This ends up in infinite &muts prefix. The compiler only figures this out when you call the method (not at compile time of the library).

I fixed this by having an inner function take a &mut FnMut as the argument and so breaking the chain of &mut. The public functions remains the same. There might be a better way to fix this but I tried for more than an hour to do this without an inner and outer function and nothing else worked.

### DIFF
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -166,7 +166,34 @@ impl<Pk: MiniscriptKey> ForEachKey<Pk> for Policy<Pk> {
 impl<Pk: MiniscriptKey> Policy<Pk> {
     /// Convert a policy using one kind of public key to another
     /// type of public key
+    ///
+    /// # Example
+    ///
+    /// ```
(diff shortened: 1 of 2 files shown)


sanket1729 (approved):
Thanks, good catch. I confirmed that the previous code did not work and only errored when the function was called. 

ACK ef1d69c81e3acc34ed2ce71b6484d719cc631ab6

apoelstra:
Oops, good catch!

And yeah, we've done the same workaround elsewhere in the code. I also cannot find a good solution :(.

apoelstra (approved):
ack ef1d69c81e3acc34ed2ce71b6484d719cc631ab6

PR #236, opened 2021-02-26
Thresh bug with equal verify

sanket1729:
Parsing an `EQUALVERIFY` ending thresh fragment failed roundtrip.

### DIFF
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -874,6 +874,12 @@ mod tests {
              OP_PUSHNUM_1\
              )"
         );
+
+        // Thresh bug with equal verify roundtrip
+        roundtrip(
+            &ms_str!("tv:thresh(1,pk(02d7924d4f7d43ea965a465ae3095ff41131e5946f3c85f79e44adbcf8e27e080e))", ),
+            "Script(OP_PUSHBYTES_33 02d7924d4f7d43ea965a465ae3095ff41131e5946f3c85f79e44adbcf8e27e080e OP_CHECKSIG OP_PUSHNUM_1 OP_EQUALVERIFY OP_PUSHNUM_1)",
+        );
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


src/miniscript/mod.rs
```diff
@@ -874,6 +874,12 @@ mod tests {
              OP_PUSHNUM_1\
              )"
         );
+
+        // Thresh bug with equal verify roundtrip
+        roundtrip(
+            &ms_str!("tv:thresh(1,pk(02d7924d4f7d43ea965a465ae3095ff41131e5946f3c85f79e44adbcf8e27e080e))", ),
+            "Script(OP_PUSHBYTES_33 02d7924d4f7d43ea965a465ae3095ff41131e5946f3c85f79e44adbcf8e27e080e OP_CHECKSIG OP_PUSHNUM_1 OP_EQUALVERIFY OP_PUSHNUM_1)",
+        );
```
darosior, line 882: I seem to only be able to reproduce it with a single subpolicy, which is not a valid Miniscript (see https://github.com/rust-bitcoin/rust-miniscript/pull/242 ).

Also, a broader Miniscript question: why isn't the Script here `1 <pk> CHECKSIG ADD 1 EQUALVERIFY 1` (as i would have expected of a `thresh` wrapped in a `v`, wrapped in a `and_or(X, 1)`) ?

sanket1729: Sorry, I have been away for a while. Will catch up with all the miniscript stuff this week.

sanket1729: `and_or` is a ternary fragment(3 args) and this is binary fragment(2 args). Why should we have `and_or` here?

darosior: Sorry, i meant `and_v`. Editing now.

darosior: And i think i answered the question myself, `ADD` is only appended if `k > 1`.

apoelstra (approved):
ack 865bf4badb77ba3ee8ba3132ae9504077a2ea9d0

PR #239, opened 2021-03-12
export ConversionError

RCasatta:
Without this, I cannot wrap ConversionError thrown from derive_public_key in a downstream library.


apoelstra (approved):
utack

PR #242, opened 2021-03-18
Exclusive bounds for thresh: 1 < k < n

### LINKED ISSUE #126: Optimize 1-of-N thresholds to or()s
https://github.com/rust-bitcoin/rust-miniscript/pull/113 left out a `FIXME` that we are unsure to fix at all, opening this to clarify if we want it or not.

We now optimize N-of-N thresholds to `and()`, and i propose to do the same for 1-of-N thresholds (using `or()`s). After discussing this with sipa (who implemented this in C++) and @sanket1729 the main issue with the latter seems to be the exponential complexity with the number of `or()`s.

I personally think (user point of view) that if the complexity growth is only applied on the policy compilation it's fine, as we only ever compile once to miniscript per user and then store the Miniscript Descriptor while the Script size is felt onchain For Ever (:tm:).

### LINKED ISSUE #237: thresh(N, n subs) is not optimized to And() if N == 1
`thresh(1, pk)` will be compiled to `pk OP_CHECKSIG 1 OP_EQUAL`. It seems strange at first sight but i believe it'd actually make us panic:
 https://github.com/rust-bitcoin/rust-miniscript/blob/42367cf75b0db6c82cabd1343613a5457404aa07/src/policy/compiler.rs#L1003
 
 So maybe it was on purpose (i can't recall, though i introduced this code). Will take a look soon :tm: 
 [self-assignment]

### LINKED ISSUE #241: thresh checks are different from specs
According to the specs (http://bitcoin.sipa.be/miniscript/) `thresh` operator in miniscript (not in policy) should only accept `1 < k < n`, but in this implementation `thresh` is allowed with `1 < k <= n`.

Can you clarify what should be changed - the specs or the implementation?

Examples:
```
wsh(thresh(5,pk(036070b1f2995d8ffda8478ef55affd39795689a3982d54b12180397b1ad1f5f75),s:pk(026515fa7603c10c44f6d316ae7592b5899d46d87ac1e574ec53de8b59f95efad6),s:pk(038c8f919f70062c084376223fd8b4f0c08958e70499df496411dde83a1bb64b0d),s:pk(02d0ea7084e344b56625277b074d15a15301b9d96b0b2dd9fc905e01fc3de408e1),sdv:after(100)))
```
is converted to the following raw bitcoin script:
```
21036070b1f2995d8ffda8478ef55affd39795689a3982d54b12180397b1ad1f5f75ac7c21026515fa7603c10c44f6d316ae7592b5899d46d87ac1e574ec53de8b59f95efad6ac937c21038c8f919f70062c084376223fd8b4f0c08958e70499df496411dde83a1bb64b0dac937c2102d0ea7084e344b56625277b074d15a15301b9d96b0b2dd9fc905e01fc3de408e1ac937c76630164b16968935587
```
and `desc.sanity_check().is_ok()` returns `true`

If I use `thresh(1,...)` it also can be converted to raw script, but at least `desc.sanity_check().is_ok()` returns `false`.

darosior:
~~Currently a draft as i'm unsure how to proceed with the compiler. For `k == n`, we already translate to `and()`s, but for `k == 1` we stepped by because of the exponential behaviour (see https://github.com/rust-bitcoin/rust-miniscript/issues/126).~~

(Credits to @stepansnigirev for finding the inconsistency)
Fixes #241 
Fixes #237
Fixes #126

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -455,6 +455,9 @@ pub enum Error {
     Script(script::Error),
     /// A `CHECKMULTISIG` opcode was preceded by a number > 20
     CmsTooManyKeys(u32),
+    /// A threshold with either a single subpolicy or as many subpolicies as the threshold value.
+    /// (threhshold, number of policies)
+    ThreshLaxBound(usize, usize),
     /// Encountered unprintable character in descriptor
     Unprintable(u8),
     /// expected character while parsing descriptor; didn't find one
@@ -588,6 +591,11 @@ impl fmt::Display for Error {
             Error::InvalidPush(ref push) => write!(f, "invalid push {:?}", push), // TODO hexify this
             Error::Script(ref e) => fmt::Display::fmt(e, f),
             Error::CmsTooManyKeys(n) => write!(f, "checkmultisig with {} keys", n),
+            Error::ThreshLaxBound(k, n) => write!(
+                f,
+                "threshold is 1 or equal to the number of subexpressions (k: '{}', n: '{}')",
+                k, n
+            ),
             Error::Unprintable(x) => write!(f, "unprintable character 0x{:02x}", x),
             Error::ExpectedChar(c) => write!(f, "expected {}", c),
             Error::UnexpectedStart => f.write_str("unexpected start of script"),
--- a/src/miniscript/astelem.rs
+++ b/src/miniscript/astelem.rs
@@ -543,6 +543,9 @@ where
                 if k > n - 1 {
                     return Err(errstr("higher threshold than there are subexpressions"));
                 }
+                if k == 1 || k == n - 1 {
+                    return Err(Error::ThreshLaxBound(k, n));
+                }
                 if n == 1 {
                     return Err(errstr("empty thresholds not allowed in descriptors"));
                 }
--- a/src/miniscript/decode.rs
+++ b/src/miniscript/decode.rs
@@ -452,6 +452,10 @@ pub fn parse<Ctx: ScriptContext>(
                 });
             }
             Some(NonTerm::ThreshW { n, k }) => {
+                if k == 1 || k == n {
+                    return Err(Error::ThreshLaxBound(k, n));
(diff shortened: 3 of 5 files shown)


darosior:
See #241 (it was actually opened before my PR but i did not see it..)

darosior:
@apoelstra @sanket1729 do you have an opinion on how to proceed to compile `thresh(1, a, b, c, ..)` after this fix? Should we just bear the compilation time, have a hard limit on the number of subs in this case (iirc that's what the C++ implementation does) ?

darosior:
So i went ahead and implemented the Policy compilation part. Added a benchmark for the compilation of nested 'or's too. Wonder if we should add a limit on the number of nested `or()` possible in a policy (it basically starts taking a very long time at `9` subs and runs forever at `10`, so it could be `< 10` nested `or()`s).

sanket1729 (commented):
@darosior, I should have seen this earlier. Sorry for the delay in response, but I think we decided to allow these scripts for easier way to write miniscripts. It seems natural to write a miniscript with thresh `1` and thresh `k = n`, so I think we should allow it. This was discussed with Pieter and I was supposed to update the c++ version. 

https://github.com/sipa/miniscript/issues/39

sanket1729:
>  Added a benchmark for the compilation of nested 'or's too. Wonder if we should add a limit on the number of nested or() possible in a policy (it basically starts taking a very long time at 9 subs and runs forever at 10, so it could be < 10 nested or()s).

Since we are allowing miniscripts thresh `k = 1`, we can compile policies with > 6(or some number that is not too long) children into a thresh fragment and use the `or` fragment for less children

apoelstra:
Should we close this PR as we have decided to update the spec to allow these?

darosior:
Yes, i planned to do this as well as maybe opening a PR for the last 3 commits that are not dependent on the spec change. Closing for now :)

PR #243, opened 2021-04-02
Derive std::hash::Hash for Descriptor et al

stevenroose:
Actually, I'm quite confused by this. The `Miniscript` type derives `Hash` without the `ScriptContext` trait requiring `Hash`, but when I tried for `Descriptor` it would complain about that. Is there some magic that made it work in `Miniscript`? Ideally we wouldn't require the `Hash` on `ScriptContext` there, but it doesn't hurt either.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -66,8 +66,12 @@ jobs:
           profile: minimal
           toolchain: ${{ matrix.rust }}
           override: true
-      - name: Pin cc if rust 1.29
+      - name: Pin deps if rust 1.29
         if: matrix.rust == '1.29.0'
-        run: cargo generate-lockfile --verbose && cargo update -p cc --precise "1.0.41" --verbose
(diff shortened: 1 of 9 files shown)


sanket1729:
utACK, looks like there are some rust 1.29 errors

stevenroose:
I pushed some CI fixes here, if you prefer them separate, I can take them apart.

apoelstra:
@stevenroose when we derive `Hash` for `Miniscript`, it only implements `Hash` in the case that the generic `ScriptContext` also implements `Hash`....in other words, never. Oops.

apoelstra (approved):
ack 8e52efbe44cb1c6be2516818bd283a856ba6607b

apoelstra:
(To clarify my above comment -- this PR fixes the issue by putting a `std::Hash` requirement on the `ScriptContext` trait.)

PR #244, opened 2021-04-26
miniscript: remove a redundant check for non-ASCII chars

darosior:
A small nit i noticed while reviewing #236 :)


apoelstra (approved):
nice. utack

PR #246, opened 2021-06-16
miniscript/types: check that timelock values aren't too large

darosior:
Wonder if we should check that the timelock value is minimal too, what do you think?

### DIFF
--- a/src/miniscript/types/extra_props.rs
+++ b/src/miniscript/types/extra_props.rs
@@ -1,7 +1,9 @@
 //! Other miscellaneous type properties which are not related to
 //! correctness or malleability.
 
-use miniscript::limits::{HEIGHT_TIME_THRESHOLD, SEQUENCE_LOCKTIME_TYPE_FLAG};
+use miniscript::limits::{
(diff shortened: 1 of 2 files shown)


src/miniscript/types/extra_props.rs
```diff
@@ -851,20 +853,19 @@ impl Property for ExtData {
                 Ok(Self::from_multi(k, pks.len()))
             }
             Terminal::After(t) => {
-                // FIXME check if t > 2^31 - 1
-                if t == 0 {
+                if t == 0 || (t & SEQUENCE_LOCKTIME_DISABLE_FLAG) == 1 {
```
sanket1729, line 856: I don't think that there are any rules applicable for after fragment. The disable flag is only for CSV.

darosior: Right, i was confused by the FIXME comment and didn't double check.. FIxed, thanks!

darosior: Reverted post discussion on IRC and added a rationale comment so we don't forget about it :)

sanket1729 (approved):
utACK a3dd9fe840c8cdae84e9d586cd9a9d6c195f0092 . Thanks for doing this.

apoelstra (approved):
ACK a3dd9fe840c8cdae84e9d586cd9a9d6c195f0092

PR #251, opened 2021-07-13
Bump bitcoin dependency to 0.26.2

stevenroose:
Needed because some method from .2 is used now.


sanket1729 (approved):
utACK 7823e9b19114ee3fc707083916f729dc699650fd

Issue #223, opened 2021-01-13
Why ForEachKey when we have iterators?


dr-orlovsky:
Sorry for potentially dumb question, but why do we need `ForEachKey` if we already have iterators? I checked trait implementations, they seem to replicate the code from iterators — and iterators should be more idiomatic than custom trait...

apoelstra:
I did consider implementing this as a new iterator (it actually does not match the existing iterators exactly because it combines `Pk` and `PkH` but does not also include hashlocks or timelocks) and then letting users do `.for_each` on this.

My reasons were
* the implementation was a fair bit simpler and also more efficient
* it's a bit of a PITA to make iterators generic, although to be honest I didn't try very hard ... you'd have to do something like adding `type PkIter; type PkhIter; type TimelockIter;` etc to the `DescriptorTrait` trait, and then every implementor needs to create an explicit type for this (they can't use `impl Iterator` because Rust is annoying)

The efficiency thing I'm not too worried about ... I have some plans to overhaul our internal representation of trees to improve cache efficiency and iteration logic anyway ... but the other issues, I believe, are why the existing iterators are implemented for `Miniscript` but not any of the descriptor or policy types.

dr-orlovsky:
Yeah, I see the reasoning... Will try to think and play whether there can be anything done in practice with iterators, or we will be limited by a exponentially growing complexity

stevenroose:
Yeah I agree that a `iter_keys()` method that returns an `impl Iterator<Item=&Pk>` would be great. I gave it a quick shot, but decided it involved a bit more thinking so I'm just gonna create HashSets using `translate_pks` (`for_each_key` doesn't guarantee that it passes each key because it stops as soon as the predicate returns false, so that's a bit subtle there, the name breaks with the `Iterator::for_each`, it's more an `Iterator::all` implementation).

PR #252, opened 2021-07-14
Add IntoIterator<Item = &Pk> for various types

stevenroose:
Related to https://github.com/rust-bitcoin/rust-miniscript/issues/223. Might make `ForEachKey` obsolete. However, these impls use the heap with `Box`es.


sgeisler:
I feel quite neutral towards this particular implementation, it's probably good enough. But I'm wondering how hard it would be to, instead of essentially re-building the miniscript tree as an interator tree, to build a custom iterator that can traverse the miniscript tree. I'll take a more thorough look once the fuzz test runs through.

stevenroose:
I never really understand the fuzzer error outputs.. Anyone that does? https://github.com/rust-bitcoin/rust-miniscript/runs/3076686285

About your comment, @sgeisler I thought about a custom iterator struct that takes keeps a descriptor reference and some "incrementor metadata" to see where it is. You'd kinda need inspection into the descriptor, but I think it'd be possible. Not really sure how the different `Terminal` vs `Policy` vs `Miniscript` vs `Descriptor` implementations would interact, though.

sanket1729:
@stevenroose, the fuzzer bug is not a bug in miniscript, but a bug in rust-secp-zkp handwritten crypto for fuzzing.

src/miniscript/astelem.rs
```diff
@@ -815,3 +815,45 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
...
+    type IntoIter = Box<dyn Iterator<Item = &'a Pk> + 'a>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        use std::iter;
+
+        match self {
+            Terminal::PkK(ref pk) => Box::new(iter::once(pk)),
+            Terminal::True
+            | Terminal::False
+            | Terminal::PkH(_)
```
sgeisler, line 830: Afaik there exists the quite common case that the key "hash" is actually also a key in our representation of miniscript (e.g. when using a descriptor to derive child descriptors). For the key mapping functions there exist very ugly-named different functions for that (`TranslatePk1`-`TranslatePk3`). It might be worth thinking about this here too, because a user may not care about if it is a p2pkh or p2pk node and just wants to treat them all the same.

mmm0171:
0x515B0941aFc4Db1F2D20f2379323a44Edb3c4460

mmm0171 (commented):
0x515B0941aFc4Db1F2D20f2379323a44Edb3c4460

sgeisler:
@mmm0171 what are you trying to tell us?

RCasatta:
First I thought it was an hash of a CVE

Then I recognized the Ethereum address mixed case style

mmm0171 (commented):
0xa36fD8fF0524f287422337d99bABdf05923f581B

mmm0171:
0xa36fD8fF0524f287422337d99bABdf05923f581B

mmm0171 (approved):

src/miniscript/astelem.rs
stevenroose: Hmm, I think only for `TranslatePk2` that could be relevant here. Though if the "pubkey hash" type is secretly a key type, that's something pretty messy. It seems that indeed for `DescriptorPublicKey` that is the case and the `Self::Hash` type is set to `Self`. That's pretty neat.

I wonder if we can do some conditional on the `K::Hash` type also being K or something.

stevenroose: Hmm, it would create conflicting implementations and also it's not possible to do equality conditions in where clauses. So I can't see how it can be done without creating messy alternative traits and the thing I liked about this impl is that it doesn't need a trait.

As an alternative, I thought about having an IntoIterator where the output iter type is an iterator of Pk::Hash items and we'd call hash on all keys, so that you'd get all the key hashes. But I think if Pk and Pk::Hash are identical, it might again cause problems of duplicate implementations. Because then the type system won't figure out which iterator you want.

stevenroose: We could yield an iterator of `miniscript::ForEach` elements. That name is a bit unfortunately chosen though because it's named with the use case it's used in instead of actually naming what it is.

stevenroose:
So I added fixup commits that would change the implementation to yield an iterator of `ForEach` elements so that you can also get easy access to key hashes.

However the `ForEach` type is named quite specifically for the `for_each` method use-case, so perhaps it makes sense to find a more general-purpose name for it like `KeyRef` or `KeyId` or something?

src/miniscript/astelem.rs
stevenroose: See my comment here: https://github.com/rust-bitcoin/rust-miniscript/pull/252#issuecomment-900497210

sanket1729 (commented):
CR ack 3a73c7d9c75db7067ba3cfa65d08d7973a21312f. This looks good to me.

 I will wait on Andrew to comment before proceeding on this because he primarily implemented the `ForEachKey` related stuff.

apoelstra:
I don't think it makes sense to have the iterator return only public keys. Miniscripts are not public key containers, they contain different kinds of data. It might make sense to iterate over all the sub-miniscripts, or over all the terminals, or over all the pk/hash/timelocks.

stevenroose:
Hmm, I mean keys are the most meaningful things miniscripts contain. Other than keys there's timelocks and hashes.

I'd be ok trying to make this into a more general iterator over interesting things, `MiniscriptIterator` or so that iterators over `MiniscriptItem::{PublicKey, Hash256, Sha256, Hash160, Timelock}` or something. If we have a newtype on it, we can even define filters on them `fn keys(self) -> impl Interator<Item = &PublicKey>`.

apoelstra:
I like the idea of having newtypes that can wrap (references to) things! Would be cool if we could genericize this so it worked on all the objects in the crate.

dr-orlovsky:
Just as a reminder: we already have iterators for miniscript items: https://github.com/rust-bitcoin/rust-miniscript/blob/master/src/miniscript/iter.rs

PR #254, opened 2021-07-24
Add timelock mix tests from miniscript alloy spec

sanket1729:
There found one bug in the current type system! The last commit disables the test while I work on fuzzer fix for rust-secp256k1.


apoelstra:
Can you rebase this? Github is confused and is still showing two commits from #257 in this.

sanket1729:
Rebased

sanket1729:
But this still has bugs, I am yet to address the changes to fuzz tests and removing the `#[cfg(not(fuzzing))]` thing

apoelstra:
Do these tests need to be ignored? On my system they run in less than a second.

sanket1729:
No, these run really fast. I don't think the ignored is needed.

apoelstra (approved):
ACK 003226ee34cd13048eda165404bddda644a736be

PR #255, opened 2021-07-26
[DONOTMERGE] Add TapCtx

sanket1729:
This is still a draft WIP for taproot PR. There are several API changes and multiple design decisions, and I am open to other possible design opinions. In particular, feedback about changes to `ToPublicKey` trait is appreciated.

This does **not** include support for P2TR descriptors, multi_a fragment in miniscript.


sanket1729 (commented):
Ready for review now

src/miniscript/context.rs
```diff
@@ -402,6 +436,97 @@ impl ScriptContext for Segwitv0 {
     }
 }
 
+/// Tap ScriptContext
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub enum Tap {}
```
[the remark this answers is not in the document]
apoelstra: I think `Tap` is fine, given that it needs to be typed inside `<>` all the time

src/lib.rs
```diff
@@ -183,6 +192,24 @@ pub trait ToPublicKey: MiniscriptKey {
...
     fn to_public_key(&self) -> bitcoin::PublicKey;
 
+    /// Convert an object to x-only pubkey
+    fn to_x_only_pubkey(&self) -> bitcoin::schnorr::PublicKey {
+        let pk = self.to_public_key();
+        bitcoin::schnorr::PublicKey::from(pk.key)
+    }
+
+    /// Serialize the key as bytes based on script context. Used when encoding miniscript into bitcoin script
+    // / - Why do we need both [`MiniscriptKey::is_xonly()`] and [`ScriptContext::Tap`]?
```
apoelstra, line 202: I think this comment is out of date .. I can't find any mention of `is_xonly` outside of this comment

src/lib.rs
```diff
@@ -183,6 +192,24 @@ pub trait ToPublicKey: MiniscriptKey {
...
+    // /   Is it possible to have
+    // Named as `serialize_bytes` instead of `serialize` because we already have many functions named serialize
+    fn serialize_bytes<Ctx: ScriptContext>(&self) -> Vec<u8> {
```
apoelstra, line 205: I would suggest a method like

```
fn push_to_builder<Ctx: ScriptContext>(&self, script::Builder) -> script::Builder
```

which would avoid an allocation and also avoid overloading the name `serialize` any more.

src/lib.rs
```diff
@@ -145,8 +144,10 @@ pub trait MiniscriptKey: Clone + Eq + Ord + fmt::Debug + fmt::Display + hash::Ha
 
     /// Computes the size of a public key when serialized in a script,
     /// including the length bytes
-    fn serialized_len(&self) -> usize {
-        if self.is_uncompressed() {
+    fn serialized_len<Ctx: ScriptContext>(&self) -> usize {
```
apoelstra, line 147: What if we inverted this function, put it on the `Ctx` object like
```
fn pk_len<Pk: MiniscriptKey>(_pk: &Pk) -> usize
```
and have the Segwit context return 34 always, Tap context return 33 always, and the bare context would look at `pk.is_compressed()`.

?

src/miniscript/context.rs
```diff
@@ -400,6 +441,105 @@ impl ScriptContext for Segwitv0 {
...
+                if pk.is_uncompressed() {
+                    return Err(ScriptContextError::UncompressedKeysNotAllowed);
+                }
+                Ok(())
+            }
+            Terminal::Multi(..) => {
+                return Err(ScriptContextError::TaprootMultiDisabled);
+            }
+            // What happens to the Multi node in tapscript? Do we use it, create
+            // a new fragment?
```
apoelstra, line 495: I think this is the right approach.

src/miniscript/context.rs
```diff
@@ -271,6 +271,9 @@ pub trait ScriptContext:
     fn is_tapctx() -> bool {
         return false;
     }
+
+    /// Local helper function to display error messages with context
+    fn to_string() -> String;
```
apoelstra, line 276: We shouldn't have our own method for this. We should implement the `fmt::Display` trait and require it for `Ctx` objects.

src/miniscript/lex.rs
```diff
@@ -54,28 +53,22 @@ pub enum Token {
     Sha256,
     Hash256,
     Num(u32),
-    Hash20([u8; 20]),
-    Hash32([u8; 32]),
-    Pubkey(PublicKey),
+    Hash20(&'s [u8]),
+    Bytes32(&'s [u8]),
+    Bytes33(&'s [u8]),
+    Bytes65(&'s [u8]),
```
apoelstra, line 59: Do you recall if you tried using `[u8; 65]` here to avoid the `'s` lifetime?

src/miniscript/decode.rs
```diff
@@ -213,325 +213,348 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> TerminalStack<Pk, Ctx> {
...
 
-/// Parse a script fragment into an `Terminal`
-#[allow(unreachable_patterns)]
-pub fn parse<Ctx: ScriptContext>(
-    tokens: &mut TokenIter,
-) -> Result<Miniscript<bitcoin::PublicKey, Ctx>, Error> {
-    let mut non_term = Vec::with_capacity(tokens.len());
-    let mut term = TerminalStack(Vec::with_capacity(tokens.len()));
+macro_rules! parse_fn {
+    ($ret_ty: ty, $fn_name: ident, $pk: ident) => {
```
apoelstra, line 217: I think here we could've rather parameterized the function over `ret_ty`, like
```
￼pub fn parse<Ctx: ScriptContext, Pk: ParseableKey>(
￼    tokens: &mut TokenIter,
￼) -> Result<Miniscript<Pk, Ctx>, Error> {
```
then defined a simple trait (private to this module even) `ParseableKey` which has the `from_slice` method on it.

src/miniscript/lex.rs
sanket1729: That would be inefficient. Because we would create an owned `[u8; 65]`. And then pass it to `Pk::from_slice(&sl)` which would create another copy. This avoids it. 

This was not needed earlier because we created the public key object directly while lexing, but now that is awkward for 32-byte keys and would require special-case handling in decode when we would get a `bitcoin::PublicKey` object while parsing taproot keys. The method `Pk::from_slice` fits it nicely with the generic function

src/miniscript/decode.rs
sanket1729: Agreed, this is cleaner

darosior:
CC'ing myself as i'd like to (hopefully) review this soon :)

src/miniscript/context.rs
sanket1729: These are not created, so I don't know if this approach can work. Right now, I can use `Ctx::to_string()` without a need for `self.`

src/miniscript/decode.rs
sanket1729: The trait can not be private because `parse` is a public interface. Not adding the seal for now, but I can add it if you think it's necessary

src/miniscript/context.rs
apoelstra: Ah, I see!

src/miniscript/decode.rs
apoelstra: Yeah, if you can seal it that'd be great. Or we can do it in a later PR, it's not a huge deal.

src/expression.rs
```diff
@@ -206,4 +222,12 @@ mod tests {
         assert!(parse_num("+6").is_err());
         assert!(parse_num("-6").is_err());
     }
+
+    #[test]
+    #[ignore]
```
apoelstra, line 227: This `#[ignore]` can be removed. (The `heavy_nest` one cannot.)

src/expression.rs
```diff
@@ -37,82 +35,104 @@ pub trait FromTree: Sized {
...
-                        return Err(Error::ExpectedChar(')'));
+                Found::Comma(pos) | Found::Rparen(pos) => {
+                    // Deal with the case where RParen gives no new children
+                    // Ex: or(A,or(B,C))
+                    // stack is [or(A) , or(B,C)]
+                    let node = if pos == 0 && start > 0 && &sl[start - 1..start] == ")" {
+                        // top will or(B,C)
+                        stk.pop().ok_or(errstr(&format!(
+                            "Incomplete miniscript: Unmatched '{}' character",
+                            ')'
```
apoelstra, line 113: These should all use the existing error types e.g. `Error::ExpectedChar(')')`

src/lib.rs
```diff
@@ -182,6 +172,21 @@ pub trait ToPublicKey: MiniscriptKey {
...
+        let pk = self.to_public_key();
+        bitcoin::schnorr::PublicKey::from(pk.key)
+    }
+
+    /// Serialize the key as bytes based on script context. Used when encoding miniscript into bitcoin script
+    fn push_to_builder<Ctx: ScriptContext>(&self, builder: script::Builder) -> script::Builder {
+        if Ctx::is_tapctx() {
+            builder.push_slice(&self.to_x_only_pubkey().serialize())
+        } else {
+            builder.push_slice(&self.to_public_key().to_bytes())
```
apoelstra, line 194: This can be `builder.push_key(&self.to_public_key())` which avoids any allocations

src/miniscript/decode.rs
```diff
@@ -34,6 +35,44 @@ fn return_none<T>(_: usize) -> Option<T> {
...
+    /// Parse a key from slice
+    fn from_slice(sl: &[u8]) -> Result<Self, KeyParseError>;
+}
+
+/// Decoding error while parsing keys
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum KeyParseError {
+    /// Bitcoin PublicKey parse error
+    FullKeyParseError(bitcoin::util::key::Error),
+    /// Wrong Input Count
```
apoelstra, line 49: I don't know what this comment means :)

sanket1729: Copy paste error :facepalm:

src/miniscript/types/extra_props.rs
```diff
@@ -119,6 +119,14 @@ pub struct ExtData {
...
     /// The timelock info about heightlocks and timelocks
     pub timelock_info: TimeLockInfo,
+    /// Maximum stack + alt stack size during satisfaction execution
+    /// This does **not** include initial witness elements. This element only captures
+    /// the additional elements that are pushed during execution.
+    pub exec_stack_elem_count_sat: Option<usize>,
+    /// Maximum stack + alt stack size during dissat execution
+    /// This does **not** include initial witness elements. This element only captures
+    /// the additional elements that are pushed during execution.
+    pub exec_stack_elem_count_dissat: Option<usize>,
```
apoelstra, line 129: Why not include initial witness elements? I **think** every fragment has a fixed number of initial stack elements that it consumes before pushing new things onto the stack

src/miniscript/types/extra_props.rs
```diff
@@ -269,6 +293,8 @@ impl Property for ExtData {
             max_sat_size: Some((33, 33)),
             max_dissat_size: Some((33, 33)),
             timelock_info: TimeLockInfo::default(),
+            exec_stack_elem_count_sat: Some(2), // either size <32> or <hash256> <32 byte>
```
apoelstra, line 296: nit: some of these comments are slightly wrong (wrong hash, 20 bytes vs 32 bytes)

src/miniscript/types/extra_props.rs
```diff
@@ -455,6 +506,10 @@ impl Property for ExtData {
             stack_elem_count_dissat: self.stack_elem_count_dissat.map(|x| x + 1),
             max_sat_size: self.max_sat_size.map(|(w, s)| (w + 2, s + 1)),
             max_dissat_size: self.max_dissat_size.map(|(w, s)| (w + 1, s + 1)),
+            // TODO: fix dissat stack elem counting above in a later commit
+            // Technically max(1, self.exec_stack_elem_count_sat), same rationale as cast_dupif
```
apoelstra, line 510: I'm curious why you defer this to a later commit. You have `cmp::max` :)

apoelstra:
Reviewed 9ae4d7b34e1cb9b7687a4bfa8c8db39f4f6737d9

Mostly good, just a few nits.

src/expression.rs
sanket1729: Why? are they not the same thing?

src/miniscript/types/extra_props.rs
sanket1729: We are already tracking `max_elem_stack_count` for initial witness elements in a separate field. 

> I think every fragment has a fixed number of initial stack elements that it consumes before pushing new things onto the stack.

ors/ threshes can have multiple satisfactions which can different heights.

src/miniscript/types/extra_props.rs
sanket1729: That is a bug fix in existing code about `max_stack_elem_count` (different from exec count) #268 which should have a separate commit. Opened #268 for it

src/miniscript/context.rs
```diff
@@ -521,8 +529,15 @@ impl ScriptContext for Tap {
         // own unique signature. That is there is no way to re-use signatures for another
         // checksig. Therefore, for each successfully executed checksig, we will have
         // 64 bytes signature and thus sigops budget is always covered.
-        // There is overall limit of consensus
         // TODO: track height during execution
+        if let (Some(s), Some(h)) = (
+            ms.ext.exec_stack_elem_count_sat,
+            ms.ext.stack_elem_count_sat,
+        ) {
+            if s + h > MAX_STACK_SIZE {
```
sanket1729, line 537: The check with initial elements + exec elements is done here.

src/expression.rs
apoelstra: The `heavy_nest` one stack overflows on my system while this one does not.

I assumed the other stack overflow would be fixed in a future PR.

src/miniscript/types/extra_props.rs
apoelstra: Derp, yes, thank you!

apoelstra: Wait, no, I'm still confused .. `or`s have multiple satisfactions but the max stack depth will just be the maximum of the children. Similar for `thresh`, you just take the sum of the `k` maximum-depth subitems.

src/expression.rs
sanket1729: Interesting, does it show an overflow error? Or just take a long time because there is no recursion here.

src/miniscript/types/extra_props.rs
sanket1729: > but the max stack depth will just be the maximum of the children

If we include initial elements. The stack depth while executing `A` in and_b(A,B) would be `max_exec _A` + initial witness `A` + initial witness `B`. But for `B` would `max_exec_B` + `initial_witness_B` because A witness would have been consumed

src/expression.rs
sanket1729: The reason for #[ignore] is that all ignored tests are run in release mode in CI. In debug mode, the rustc keeps lot of debug information which makes the program really slow to execute. 

In the first commit of the PR, both of the tests should pass in a few seconds in release mode. there is no bug fix for a future commit

src/miniscript/decode.rs
sanket1729: Done

src/expression.rs
sanket1729: Based on a discussion in IRC, I have reverted the commits that switched the implementation back to recursive ones.

sanket1729:
@apoelstra this is also ready for review

src/lib.rs
```diff
@@ -182,6 +172,21 @@ pub trait ToPublicKey: MiniscriptKey {
...
 
+    /// Convert an object to x-only pubkey
+    fn to_x_only_pubkey(&self) -> bitcoin::schnorr::PublicKey {
+        let pk = self.to_public_key();
+        bitcoin::schnorr::PublicKey::from(pk.key)
+    }
+
+    /// Serialize the key as bytes based on script context. Used when encoding miniscript into bitcoin script
+    fn push_to_builder<Ctx: ScriptContext>(&self, builder: script::Builder) -> script::Builder {
+        if Ctx::is_tapctx() {
```
JeremyRubin, line 191: it might be nicer if the ScriptContext or Builder were more directly responsible for this logic than the MiniscriptKey.

src/miniscript/astelem.rs
```diff
@@ -750,6 +750,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
                     .push_opcode(opcodes::all::OP_EQUAL)
             }
             Terminal::Multi(k, ref keys) => {
+                debug_assert!(!Ctx::is_tapctx());
```
JeremyRubin, line 753: nit:

this is a pretty big type smell -- if Multi should not be permitted during Ctx tapctx, perhaps one thing that could be done is to give Multi a type tag which is defined as `enum Void{}`for Taproot and `enum Inhabited {By}` for non Taproot, which disables Multi at the type level.

This is -- by all accounts -- an annoying-ish API to work with, but would ensure that it's impossible to use Taproot with Multi which is superior to debug_assert (especially if we want to hanlde user generated code...)

src/miniscript/context.rs
```diff
@@ -353,6 +394,12 @@ impl ScriptContext for Segwitv0 {
                 }
                 Ok(())
             }
+            Terminal::Multi(_k, ref pks) => {
```
JeremyRubin, line 410: This seems like a new rule for segwit v0... separate PR?

(side note: I'm a bit confused as to why these functions are all parametrized by Ctx vs. Self?)

src/miniscript/context.rs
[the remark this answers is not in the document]
JeremyRubin: I think it's OK... but not robust IMO if there are future leaf versions. TapL0?

src/miniscript/context.rs
```diff
@@ -400,6 +447,105 @@ impl ScriptContext for Segwitv0 {
...
+    }
+
+    fn check_global_consensus_validity<Pk: MiniscriptKey, Ctx: ScriptContext>(
+        ms: &Miniscript<Pk, Ctx>,
+    ) -> Result<(), ScriptContextError> {
+        // No script size checks for global consensus rules
+        // Should we really check for block limits here.
+        // When the transaction sizes get close to block limits,
+        // some guarantees are not easy to satisfy because of knapsack
+        // constraints
```
JeremyRubin, line 501: I think it'd be reasonable to enforce here 50Kb limit, which is 1/2 the max transaction size according to mempool standardness rules (100kb) so saves healthy margin for other interesting stuff. "50kb should be enough for everybody". Miniscript doesn't care to target non standard txs, right?

JeremyRubin: ah "global consensus" v.s. local.

src/miniscript/context.rs
JeremyRubin: Should we fail if we have one of these properties and not the other?

src/miniscript/astelem.rs
apoelstra: We cannot give Multi a type tag without giving `Terminal` a type tag which would infect `Miniscript` and then `Ctx`. We have enough type tags.

src/lib.rs
apoelstra: `Builder` belongs to rust-bitcoin and it would require an extra dummy trait to extend it here.

`ScriptContext` doesn't make a lot more sense to me than `MiniscriptKey`

src/miniscript/context.rs
apoelstra: It's fine for it to be in this PR.

src/miniscript/context.rs
```diff
@@ -276,6 +276,8 @@ pub trait ScriptContext:
     /// 34/66 for Bare/Legacy based on key compressedness
     /// 34 for Segwitv0, 33 for Tap
     fn pk_len<Pk: MiniscriptKey>(pk: &Pk) -> usize;
+    /// Local helper function to display error messages with context
+    fn to_string() -> String;
```
apoelstra, line 288: I think this could return a `&'static str` (and similarly the `String`s in all the error variants could be `&'static str`s). The function should probably be called `to_str` then.

src/miniscript/astelem.rs
JeremyRubin: the type can be an associated type to Ctx, no?

src/miniscript/context.rs
```diff
@@ -66,7 +73,17 @@ impl fmt::Display for ScriptContextError {
                 write!(f, "DupIf is malleable under Legacy rules")
             }
             ScriptContextError::CompressedOnly => {
-                write!(f, "Uncompressed pubkeys not allowed in segwit context")
+                write!(
+                    f,
+                    "Only Compressed pubkeys not allowed in segwit context. X-only and uncompressed keys are forbidden"
```
dr-orlovsky, line 80: should read:
```suggestion
                    "Only Compressed pubkeys are allowed in segwit context. X-only and uncompressed keys are forbidden"
```

src/miniscript/context.rs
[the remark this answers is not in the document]
dr-orlovsky: One will be able to use it as `context::Tap`, which is fine.

@JeremyRubin: new leaf version will require softfork, i.e. major API change, so it should not be a problem to rename the type at that point in time

src/miniscript/context.rs
```diff
@@ -400,6 +447,105 @@ impl ScriptContext for Segwitv0 {
...
+        _ms: &Miniscript<Pk, Ctx>,
+    ) -> Result<(), ScriptContextError> {
+        // Taproot introduces the concept of sigops budget.
+        // In all possible valid miniscripts satisfy the given sigops constraint
+        // Whenever we add new fragment that uses pk(pk() or multi based on checksigadd)
+        // miniscript typing rules ensure that pk when executed successfully has it's
+        // own unique signature. That is there is no way to re-use signatures for another
+        // checksig. Therefore, for each successfully executed checksig, we will have
+        // 64 bytes signature and thus sigops budget is always covered.
+        // There is overall limit of consensus
```
dr-orlovsky: Comment is hardly readable. I can't understand what it says/means

`In all possible valid miniscripts satisfy the given sigops constraint` - not a proper sentense.
`That is there is no way to re-use signatures` - two "is" make hard to guess the meaning
`There is overall limit of consensus` - dropped end of the sentence

src/miniscript/context.rs
```diff
@@ -400,6 +447,105 @@ impl ScriptContext for Segwitv0 {
...
+    }
+
+    fn max_satisfaction_size<Pk: MiniscriptKey, Ctx: ScriptContext>(
+        ms: &Miniscript<Pk, Ctx>,
+    ) -> Option<usize> {
+        // The witness stack cost is the first element of the tuple
+        ms.ext.max_sat_size.map(|x| x.0)
+    }
+
+    fn is_tapctx() -> bool {
```
dr-orlovsky, line 564: Not sure that `is_tapctx` a good name. First, the context name is `Tap`, not `TapCtx`. Second, Rust API naming says we should use `_` between words, i.e. this should probably be `is_tap_ctx` or `is_tap`.

src/lib.rs
```diff
@@ -159,6 +159,14 @@ impl MiniscriptKey for bitcoin::PublicKey {
     }
 }
 
+impl MiniscriptKey for bitcoin::schnorr::PublicKey {
+    type Hash = hash160::Hash;
+
+    fn to_pubkeyhash(&self) -> Self::Hash {
```
dr-orlovsky, line 165: The question is why we need to support pubkey hashes in Taproot - and why it is of hash160::Hash type, when we have specific hash types for other cases (like `PubkeyHash`, `WPubkeyHash`)?

src/lib.rs
```diff
@@ -206,6 +214,20 @@ impl ToPublicKey for bitcoin::PublicKey {
     }
 }
 
+impl ToPublicKey for bitcoin::schnorr::PublicKey {
+    fn to_public_key(&self) -> bitcoin::PublicKey {
+        // This code should never be used.
+        // But is implemented for completeness
+        let mut data: Vec<u8> = vec![0x02];
+        data.extend(self.serialize().iter());
+        bitcoin::PublicKey::from_slice(&data).expect("Cannot fail")
```
dr-orlovsky, line 223: IMO `expect("Cannot fail")` is in fact `unwrap()`. The point of the message in `expect` is to provide debugging information to see which specific unreachable statement was executed. Maybe something like `"failed to construct public key from 33-byte slice"` will be more informative.

Overall, I think that we need to introduce `PublicKey::from_bytes` which takes fixed-length array in `rust-bitcoin` and can't error.

src/lib.rs
```diff
@@ -525,6 +547,8 @@ pub enum Error {
     ImpossibleSatisfaction,
     /// Bare descriptors don't have any addresses
     BareDescriptorAddr,
+    /// PubKey invalid under current context
+    PubKeyCtxError(String, String),
```
dr-orlovsky, line 551: Why we have to use `String` here? Because we do not have a type enumerating all possible public keys in `rust-bitcoin`? But we can still use a specific context type with `dyn ScriptContext` I assume...

Another thing: I prefer to use named associated fields when it is not obvious which type contains which informtion:

```suggestion
    PubKeyCtxError { pubkey: String, context: String },
```

src/miniscript/context.rs
dr-orlovsky: It is also not clear why we can't just require `ScriptContext: Display`, implement `Display` for specific context types and get rid from this method here

src/miniscript/decode.rs
```diff
@@ -34,6 +35,54 @@ fn return_none<T>(_: usize) -> Option<T> {
...
+    }
+}
+
+impl ParseableKey for bitcoin::schnorr::PublicKey {
+    fn from_slice(sl: &[u8]) -> Result<Self, KeyParseError> {
+        bitcoin::schnorr::PublicKey::from_slice(sl).map_err(KeyParseError::XonlyKeyParseError)
+    }
+}
+
+impl error::Error for KeyParseError {}
```
dr-orlovsky, line 65: Better to implement `source` methods and return inner error type here

src/miniscript/decode.rs
```diff
@@ -34,6 +35,54 @@ fn return_none<T>(_: usize) -> Option<T> {
...
+
+impl fmt::Display for KeyParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            KeyParseError::FullKeyParseError(e) => write!(f, "FullKey Parse Error {}", e),
+            KeyParseError::XonlyKeyParseError(e) => write!(f, "XonlyKey Parse Error {}", e),
```
dr-orlovsky, line 71: If you will move the inner error type to `source` (see my comment above) that we should not print inner error here and just use `FullKey parse error` etc

src/miniscript/context.rs
sanket1729: Indeed this is a separate bug fix, I will make it under a separate commit. This prevents the usage of uncompressed keys in segwitv0 context.

> I'm a bit confused as to why these functions are all parametrized by Ctx vs. Self

This is because we never have an instance of Self in our usage of ScriptContext, but is always used as a generic in Miniscript type itself. For example, if the user writes `let ms = Miniscript::<String, Tap>::from_str(multi(...))` we can sanely error. And we prevent using miniscripts across contexts. I don't see how we can do that if we have a self parameter.

src/miniscript/lex.rs
dr-orlovsky: Wish we had upper MSRV with support for const generics - in this case we can have a generic over different key lengths and also can avoid use of `Pk::from_slice` and have `Pk::from_bytes` with different byte array length per each type...

src/miniscript/mod.rs
```diff
@@ -30,7 +30,7 @@ use std::{fmt, str};
 use bitcoin;
 use bitcoin::blockdata::script;
 
-pub use self::context::{BareCtx, Legacy, Segwitv0};
+pub use self::context::{BareCtx, Legacy, Segwitv0, Tap};
```
dr-orlovsky, line 33: It is really strange to have types named `BareCtx` and `Legacy`/`Tap` in the same time. IMO either should be `Bare`, `Legacy`, `Tap` - or `BareCtx`, `LegacyCtx` and `TapCtx`.

src/miniscript/mod.rs
```diff
@@ -168,7 +171,38 @@ impl<Ctx: ScriptContext> Miniscript<bitcoin::PublicKey, Ctx> {
     /// This function will fail parsing for scripts that do not clear
     /// the [Miniscript::sanity_check] checks. Use [Miniscript::parse_insane] to
     /// parse such scripts.
-    pub fn parse(script: &script::Script) -> Result<Miniscript<bitcoin::PublicKey, Ctx>, Error> {
+    /// ## Decode/Parse a miniscript from script hex
```
dr-orlovsky, line 174: Seems we need an empty doc comment line for `rustdoc` to handle `##` title properly

dr-orlovsky, line 174: Why from "script hex" when we provide the method with a `Script` object, not a hex string?

src/miniscript/astelem.rs
sanket1729: > This is -- by all accounts -- an annoying-ish API to work with, but would ensure that it's impossible to use Taproot with Multi which is superior to debug_assert 

We still ensure that wherever we have any user input(1) FromStr 2) TranslatePk 3) compiler output or 4) From decoded script or 5) Miniscript::from_ast()), we pass it through the context checks, thats check the validity of miniscript rules. If you ever create any instance of `Miniscript<Pk: MiniscriptKey, Tap>` with invalid context rules, the library would error.

```
let ms = Miniscript<Pk, Tap>::from_str("multi(..)").unwrap_err();
let ms = Miniscript<Pk, Segwitv0>::from_str("multi(..)").unwrap();
```

The debug_assert here is merely a sanity check, and if this PR(this commit infact) is correctly implemented should never be triggered.

src/miniscript/mod.rs
```diff
@@ -433,6 +467,7 @@ mod tests {
     use std::sync::Arc;
 
     type Segwitv0Script = Miniscript<bitcoin::PublicKey, Segwitv0>;
+    type Tapscript = Miniscript<bitcoin::schnorr::PublicKey, Tap>;
```
dr-orlovsky, line 470: I think we need somehow prevent users from using `Miniscript<bitcoin::schnorr::PublicKey, SegwitV0>`-like types...

Non-clean route will be to provide a set of type aliases for the supported combinations, but this still will not prevent users from constructing incorrect onces.

The cleaner way is much more complex though: something like creating more traits, one per context, for public keys; and moving first `Miniscript` generic parameter to be a generic parameter of the context, restricted by the public key trait...

src/miniscript/types/extra_props.rs
```diff
@@ -119,6 +119,14 @@ pub struct ExtData {
     pub max_dissat_size: Option<(usize, usize)>,
     /// The timelock info about heightlocks and timelocks
     pub timelock_info: TimeLockInfo,
+    /// Maximum stack + alt stack size during satisfaction execution
+    /// This does **not** include initial witness elements. This element only captures
+    /// the additional elements that are pushed during execution.
+    pub exec_stack_elem_count_sat: Option<usize>,
```
dr-orlovsky, line 125: We have very distinct naming across methods and fields: some places use `stackelem`, some `stack_elem`. According to Rust API guidelines, `stack_elem` is more preferable

dr-orlovsky (commented):
Made initial code read. Concept ACK, left some questions, nits & suggestions. Have not checked against BIPs/consensus rules.

src/miniscript/context.rs
sanket1729: I will change to `is_tap`. I was also confused between `Tap` and `TapCtx` naming for the PR, so this is a remaining artifact of it.

src/lib.rs
sanket1729: Good call. Basically, when this was originally implemented I was not aware of specific hash types in rust-bitcoin. I will change bitcoin::schnorr::PublicKey in this PR and change the `impl MiniscriptKey for bitcoin::PublicKey` in a separate PR.

src/lib.rs
sanket1729: Yes, indeed. Will add a message "failed to construct 33 Publickey from 0x02 appended x-only key".

> Overall, I think that we need to introduce PublicKey::from_bytes which takes a fixed-length array in rust-bitcoin and can't error.
Not all 33-byte arrays are valid PublicKeys. About slightly less than half of are valid x-only keys. and 0x02 and 0x03 are valid first bytes for 33 byte publickey

src/miniscript/mod.rs
sanket1729: Initially everything was without `Ctx`. But `Bare` could have been confused with the `Bare/Raw` descriptor. That is why for `Bare`, I added `BareCtx`. I agree that we should be consisted, but given the limited usage of `Bare`/`BareCtx`, I think saving to type `Ctx` argument wing.

src/miniscript/types/extra_props.rs
sanket1729: Yes, I will do this as a separate PR.

src/lib.rs
sanket1729: > It is also not clear why we can't just require ScriptContext: Display, implement Display for specific context types and get rid from this method here

Replying to the above comment first because GitHub is not letting me reply there. :shrug: 
`ScriptContext` is never implemented on any object that can be created. Therefore, this is no `Self` to impl `fmt::Display` on. In our usage, `ScriptContext` is just used as a generic parameter with no objects ever created. Similar to `secp256k1::Signing` and `secp256k1::Verification`. 

```
// empty enum. No object can be created for this
pub enum Legacy {}
```

> Why we have to use String here? Because we do not have a type enumerating all possible public keys in rust-bitcoin? But we can still use a specific context type with dyn ScriptContext I assume...

The first pubkey string is my laziness. I don't see any reason why I can't return `KeyParseError`. The second one has answers in this comment and other comments.

src/lib.rs
dr-orlovsky: Thanks!But in whic cases do we need pubkey hashes in Taproot/script? I understand that nothing prohibits from adding key hash to arbitrary tapscript leaf, but I think we should discourage this nonstandard practice in API and just handle that as any other hash, unrelated to pubkey

src/miniscript/mod.rs
sanket1729: Agreed, this can be a problem and can be confusing. But even if users do that, they cannot shoot themselves in foot. All the current APIs would work correctly. The encoding API would give a 33-byte public key when the user calls `ms.encode()` and decoding would be looking for 33 byte pubkey keys and would error on `schnorr::PublicKey`. All the APIs would output correctly assuming the context is `Segwitv0`. 

All `MiniscriptKey` is "provide me some basic bounds(FromStr/Clone/etc) way to convert to `bitoin::PublicKey`". Users can free to any `impl MiniscriptKey` the user wants. If the wants to only use only a enum ```
enum AllowedKeys
{
AliceKey,
BobKey,
CarolKey,
}
```
and provide a `impl MiniscriptKey` for it, the library would just work. Similarly, for x-only keys, the output would not be wrong, we are not breaking any safety rules, just that the encoded output might be in `SegwitV0` context with 33-byte keys which might be confusing. 

If the user then goes on to the encoded bytes into a tapscript script, I think it is a user error. Only advanced users should be dealing with the `Miniscript` object, and users that deal with descriptors directly should would very clearly know that they are dealing with `tr` descriptor or `wsh` one.

File: src/miniscript/context.rs (hunk 19 above)
Comment (JeremyRubin) [reply src/miniscript/context.rs:410]: it's off topic for this pr, but:

not a self parameter a Self type. As in

```
pub trait ScriptContext:
    fmt::Debug + Clone + Ord + PartialOrd + Eq + PartialEq + hash::Hash + private::Sealed
{
    fn check_terminal_non_malleable<Pk: MiniscriptKey, Ctx: ScriptContext>(
        _frag: &Terminal<Pk, Ctx>,
    ) -> Result<(), ScriptContextError>;
```

should be

```rust
pub trait ScriptContext:
    fmt::Debug + Clone + Ord + PartialOrd + Eq + PartialEq + hash::Hash + private::Sealed
{
    fn check_terminal_non_malleable<Pk: MiniscriptKey>(
        _frag: &Terminal<Pk, Self>,
    ) -> Result<(), ScriptContextError>;
```

It seems like otherwise the type of ScriptContext permittable into  any of the functions is *any* ScriptContext, not the specific one we just defined it for?

src/miniscript/astelem.rs
JeremyRubin: I've implemented this approach here:

https://github.com/sanket1729/rust-miniscript/pull/3

I beleive with this approach I also found a bug in this PR w.r.t. Policy Compilations (Multi Could be emitted from a Tap context it seems?).

src/miniscript/context.rs
sanket1729: Indeed, this is an improvement.

JeremyRubin: Ok I can do this as a follow up commit after this goes through -- hopefully not too annoying.

src/miniscript/astelem.rs
sanket1729: > I believe with this approach I also found a bug in this PR w.r.t. Policy Compilations (Multi Could be emitted from a Tap context it seems?).

No. Every compiler created `Terminal` calls `Ctx::check_local_validity` (which implies global_validity). 
local validity => check if all possible spend paths are valid spends
global validity => Check if any spend path is valid

https://github.com/rust-bitcoin/rust-miniscript/blob/8ad083a2e4f67e32c6b05ca278cf398089d43cc5/src/policy/compiler.rs#L650 
.

One thing worth considering is if we should really bother adding type safety to the `Terminal` enum. It is a fairly internal thing that users should never really have to deal with. This PR already implements the check for the `Miniscript` struct when creating it from `Terminal` (Miniscript::<Pk, Ctx>::from_ast). 

Most users should always use `Descriptor`, and the few advanced users that use `Miniscript` are also safe. So, we are protecting users that interact with `Terminal` directly. Unfortunately, we have to expose it because users might want `match` the `Terminal` object.

Currently, it is possible to create Miniscript directly from members, but we can do the standard fix.
i.e Add a `new()` method to `Miniscript`, make its members private, provide accessor method, add global checks to new(). This will make sure that there is only one way to create `Minsicript` objects and we implement checks at that point. 

I initially strongly ACKed your suggestion, but now I am considering if it is worth the new code diff. Particularly considering that we might want to stick to this and add another `Ctx::gen_multi_a` with new a new associated type for tapscript `multi_a` (multi-sig with CHECKSIGADD).

src/miniscript/mod.rs
dr-orlovsky: I think we should need API users ability to restrict that their custom `impl MiniscriptKey` keys like `AllowedKeys` should be restricted only for Taproot or other context. For instance, this may be required when they are derived under certain BIP32 path requiring Taproot or SegWit usage. And the second (more complex) way I proposed should work in this case as well

src/lib.rs
dr-orlovsky: Then, the question is why we use non-instantiatable enum `pub enum Legacy {}` instead of unit struct `pub struct Legacy;`? In the second case the form will remain the same (`MiniscriptKey<..., Tap>`) and we will be able to add `where Self: ` conditions

src/miniscript/context.rs
sanket1729: `None` here means that there is no satisfaction for this miniscript. In all cases, if one of them is Some, the other must also be Some

src/lib.rs
sanket1729: This is more a spec question. I don't see why we should disallow `pkh`. It's more efficient in case the fragment is not satisfied.

src/lib.rs
sanket1729: I think Andrew liked this cannot be instantiated because it does not make sense to have a context object. I don't really mind changing it, but @apoelstra has some opinions on this.

src/miniscript/context.rs
JeremyRubin: Maybe should be Some((x,y)) instead of Some(x), Some(y) then? Or assert that `!a.xor(b).is_some()`?

src/miniscript/mod.rs
sanket1729: How about the following which is less invasive, but yet another hack. 
- We add a method to MiniscriptKey `is_full_key` with a default impl that returns True.
- We implement `is_full_key` as false for `schnorr::PublicKey`
- Inside the global checks, we add check for keys in PkK, MultiA.

This is not clean because this does not disable this at type level, nor it allows downstream users to restrict implementation based on context. But I think this can fall within the scope of this PR if we think it is the right thing to do.

However, I am still not completely convinced that `Miniscript<bitcoin::schnorr::PublicKey, Tap>` should be prevented. In my head, MiniscriptKey is still "provide me some basic bounds(FromStr/Clone/etc) way to convert to bitoin::PublicKey" which is still fits for `schnorr::PublicKey`. cc @apoelstra

src/lib.rs
sanket1729: I can add a DummyTrait, one less method in `ToPublicKey`

src/miniscript/mod.rs
JeremyRubin: Maybe we can add an associate type to ScriptContext which is "ConesnusKey"?

And then you can have methods that are only defined when `Miniscript<K, C> /*...*/  where K : IsEqual<C::ConsensusKey>`, or perhaps `Miniscript<K, C> /*...*/  where C::ConsensusKey : From<K>`, or `Miniscript<K, C> /*...*/  where C::ConsensusKey : From<(K, Option<KeyDerivationStruct>)>`. There's obviously some design flexibility here, but the end outcome is that ScriptContexts know what kind of key they need for consensus purposes.

src/miniscript/astelem.rs
JeremyRubin: One note is if you don't like gen_multi, it could be replaced with something like...

```rust
trait EnableDisable : TryFrom<ScriptContextError, Error=ScriptContextError>;
```

and then

```rust
impl TryFrom<ScriptContextError> for Enabled {
   fn try_from(v:ScriptContextError) -> Result<Self, ScriptContextError> {
   Ok(Enabled {})
   }
}
impl TryFrom<ScriptContextError> for Disabled {
   fn try_from(v:ScriptContextError) -> Result<Self, ScriptContextError> {
   Err(v)
   }
}
```

And then you never need any special functions, you would just need to do something like:

```rust
Terminal::Multi(n, k, ScriptContextError::MultiDisabled.try_into()?)
```

The try_into would infer if, based on the current active Ctx, we need to fill Terminal with Enabled or Disabled and appropriately Error out at this point as the type would not be constructible.


Then you would not need any `gen_multi_a` or anything...

src/miniscript/mod.rs
sanket1729: Sounds nice. Let me fiddle around with this.

src/miniscript/astelem.rs
sanket1729: My primary concern we never had the safety guarantee for `Terminal` w.r.t to consensus rules. For example, we don't have guarantees about height/time combination in the `Terminal` struct. Or you could create a `Terminal<bitcoin::PublicKey, BareCtx>` with Multisig (with n >= 4) or Legacy scripts of more than 520 bytes. Terminal need not even be of type `B`. 

All of these are permitted representations of the `Terminal` object, but are unsafe for consensus. Why does `Multi` deserve special detection? As long as the `Miniscript` struct is type-safe, I think we are good.

src/lib.rs
sanket1729: As in `DUP HASH160 <hash> EQUALVERIFY CHECKSIG` is still a valid script and has less size than `<x_only_pk> CHECKSIG`. So, there is one use-case

dr-orlovsky: But would this smaller size matter for the branches of the tapscript which are never used? It will never hit blockchain. But while spent, it will have a larger size for `witness` in the spending tx input.

sanket1729: Consider the following:
`thresh(11,S1,S2,S2......,S15)` where `S15` is `or(pk(likely),pkh(unlikely))`

It is difficult to always decompose scripts into ORs(DNF). Therefore, it might be necessary to a leaf miniscript that contains a or

src/miniscript/decode.rs
sanket1729: source is MSRV 1.30 :( . I will make an issue for better error handing to change everything to source. I can use the now deprecated `cause` for now, will make an issue to change everything to source once we change MSRV

src/miniscript/context.rs
sanket1729: I will add the assert for now.

src/miniscript/decode.rs
dr-orlovsky: Oh yes. Ok, probably it is better not to change than and have an issue for `source` adoption once we will bump MSRV

src/miniscript/mod.rs
sanket1729: Oops, just realized that the example below shows decoding from hex. That is why it was labelled as such

src/miniscript/mod.rs
sanket1729: Thanks for the suggestion. Ended up adding an associated `Key` type to the `ScriptContext` trait.

sanket1729:
Pushed another version. Added an associated type `Key` to `ScriptContext` trait and addressed all comments but:
- Changing the Terminal object to include type tag. Rationale and discussion: https://github.com/rust-bitcoin/rust-miniscript/pull/255#discussion_r712519014
- Decision on whether to disallow creation of `Miniscript<bitcoin::schnorr::PublicKey, Segwitv0>`. This is independent of this PR. And we can implement restrictions if required in a later PR.

src/miniscript/astelem.rs
JeremyRubin: I think it's an understandable perspective, but generally I fall on the side of wanting to encode as much safety as possible in the type system, to the extent it doesn't prevent a usable API / leaks too much complexity. This seemed like a win and a step in that direction at low code complexity, but perhaps not.

src/miniscript/mod.rs
JeremyRubin: looks good

src/miniscript/context.rs
```diff
@@ -66,7 +73,17 @@ impl fmt::Display for ScriptContextError {
...
+                write!(
+                    f,
+                    "Only Compressed pubkeys are allowed in segwit context. X-only and uncompressed keys are forbidden"
+                )
+            }
+            ScriptContextError::UncompressedKeysNotAllowed => {
+                write!(
+                    f,
+                    "Only x-only keys are allowed in tapscript checksig. \
+                    Compressed keys maybe specified in descriptor."
```
apoelstra, line 90: In 5d61bfea0cc7c078da8b81dae04190821c9d2860

I think this error would be clearer just as "uncompressed keys cannot be used in Taproot descriptors".

src/miniscript/context.rs
```diff
@@ -39,6 +41,9 @@ pub enum ScriptContextError {
     /// Only Compressed keys allowed under current descriptor
     /// Segwitv0 fragments do not allow uncompressed pubkeys
     CompressedOnly,
+    /// Tapscript descriptors cannot contain uncompressed keys
+    /// Tap context can contain compressed or xonly
+    UncompressedKeysNotAllowed,
     /// At least one satisfaction path in the Miniscript fragment has more than
     /// `MAX_STANDARD_P2WSH_STACK_ITEMS` (100) witness elements.
     MaxWitnessItemssExceeded,
```
apoelstra, line 52: I know this isn't part of this PR, but `MaxWitnessItemssExceeded` has too many `s`s :)

src/miniscript/astelem.rs
apoelstra: I tend to agree that restricting `Terminal` would be hard in general (and potentially undesireable as it would turn many types of analysis errors into parsing errors, which for a user could mean less information or ability to recover), especially as it is mostly an internal-only type.

apoelstra (dismissed):
ACK e4dcba28123eca0e2d8605f7533077d01014ecb5

(Will not merge as I'm not sure what more of your TODOs you want to address in this PR, or if any of the ongoing conversations with Jeremy have actionable results.)

sanket1729:
There are no more todos that I want to address in this PR. I will work on the taproot descriptor PR(s), while we await reviews from Jeremy and dr-orlovsky.

src/miniscript/decode.rs
```diff
@@ -213,11 +271,11 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> TerminalStack<Pk, Ctx> {
     }
 }
 
-/// Parse a script fragment into an `Terminal`
+/// Parse a script fragment into an `Miniscript`
 #[allow(unreachable_patterns)]
 pub fn parse<Ctx: ScriptContext>(
     tokens: &mut TokenIter,
-) -> Result<Miniscript<bitcoin::PublicKey, Ctx>, Error> {
+) -> Result<Miniscript<Ctx::Key, Ctx>, Error> {
```
dr-orlovsky, line 278: I am still not getting why Miniscript needs two generic parameters and does not automatically use new `Ctx::Key`

dr-orlovsky (commented):
Have not finished review yet, but left several comments/questions

src/miniscript/decode.rs
JeremyRubin: i think it's becasue we want to be able to reason about miniscripts with abstract keys that we will later translate, but we only ever want to e.g. parse/compile consensus keys.

dr-orlovsky: Correct. Sorry for forgetting

src/miniscript/astelem.rs
JeremyRubin: My overall feeling is that we should aim to embed as much correctness properties in the type system as we can. Certain things are hard to do (e.g., checking size limits) in the type system, but other things (valid clause definitions) should be easier to do. That includes things like "does this have a multisig".

The largest benefit of this, from a reviewers perspective, is that I know (just by looking at the types!) that the Tap Policy implementation is correct with respect to Multi clauses, whereas before I was unsure since I had to understand the behavior of a bunch of macros in the compilation code of Policy. The types serve to lighten that review burden by making it clear that it would be a compiler error. When making other refactorings in the future, we can be sure that anything that affects the e.g. Policy compiler will not introduce errors of creating clauses that should not be made. And I think that's fantastic!

W.r.t. better or worse errors at parse time v.s. analysis time, I think we can return the appropriate error during Parse (e.g., no multi allowed) and the patch I sent over does that.

fuzz/fuzz_targets/roundtrip_miniscript_script.rs
```diff
@@ -8,7 +8,7 @@ fn do_test(data: &[u8]) {
     // Try round-tripping as a script
     let script = script::Script::from(data.to_owned());
 
-    if let Ok(pt) = Miniscript::<_, Segwitv0>::parse(&script) {
+    if let Ok(pt) = Miniscript::<miniscript::bitcoin::PublicKey, Segwitv0>::parse(&script) {
```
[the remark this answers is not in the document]
sanket1729: When using miniscript fuzzing, I generally preferred reusing the bitcoin from miniscript instead of importing a separate crate into fuzzer

src/miniscript/context.rs
```diff
@@ -99,6 +121,15 @@ impl fmt::Display for ScriptContextError {
                     "Impossible to satisfy Miniscript under the current context"
                 )
             }
+            ScriptContextError::TaprootMultiDisabled => {
+                write!(f, "No Multi node in taproot context")
+            }
+            ScriptContextError::StackSizeLimitExceeded => {
+                write!(
+                    f,
+                    "Stack limit can exceed in atleast one script path during script execution"
```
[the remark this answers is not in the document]
sanket1729: The exact stack limit calculations can be done at execution time whereas we do some static analysis about the execution time at script creation time. It is possible that the script path that exceeds the limit is not taken at satisfaction time and thus this message would be incorrect. 

Would split `atleast` into `at least`

src/miniscript/context.rs
```diff
@@ -353,6 +418,12 @@ impl ScriptContext for Segwitv0 {
                 }
                 Ok(())
             }
+            Terminal::Multi(_k, ref pks) => {
+                if pks.iter().any(|pk| pk.is_uncompressed()) {
+                    return Err(ScriptContextError::CompressedOnly);
```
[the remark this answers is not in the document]
sanket1729: Unfortunately, the keys here are abstract `MiniscriptKey` objects, and propagating a generic argument through the Error type would be annoying. I will add a String type here to instead have a `CompressedOnly(String)` for better error reporting

src/miniscript/context.rs
```diff
@@ -506,17 +707,27 @@ impl ScriptContext for NoChecks {
...
+    use super::{BareCtx, Legacy, NoChecks, Segwitv0, Tap};
 
     pub trait Sealed {}
 
     // Implement for those same types, but no others.
     impl Sealed for BareCtx {}
     impl Sealed for Legacy {}
     impl Sealed for Segwitv0 {}
+    impl Sealed for Tap {}
     impl Sealed for NoChecks {}
```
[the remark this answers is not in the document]
sanket1729: Agreed in principle with the naming change for SegwitV0, but that would be a big breaking change for everything downstream. I think we should bite it as it is, or we can discuss that after this PR.

src/miniscript/context.rs
```diff
@@ -109,7 +142,11 @@ impl fmt::Display for ScriptContextError {
 /// For example, disallowing uncompressed keys in Segwit context
 pub trait ScriptContext:
     fmt::Debug + Clone + Ord + PartialOrd + Eq + PartialEq + hash::Hash + private::Sealed
+where
+    Self::Key: MiniscriptKey<Hash = bitcoin::hashes::hash160::Hash>,
```
JeremyRubin, line 146: is there a reason for splitting the trait bound in a where / type clause (below)?

src/miniscript/context.rs
```diff
@@ -400,6 +481,126 @@ impl ScriptContext for Segwitv0 {
...
+pub enum Tap {}
+
+impl ScriptContext for Tap {
+    type Key = bitcoin::schnorr::PublicKey;
+    fn check_terminal_non_malleable<Pk: MiniscriptKey, Ctx: ScriptContext>(
+        _frag: &Terminal<Pk, Ctx>,
+    ) -> Result<(), ScriptContextError> {
+        // No fragment is malleable in tapscript context.
+        // Certain fragments like Multi are invalid, but are not malleable
+        Ok(())
```
JeremyRubin, line 505: does that include fragments that have a repeated key?

src/miniscript/context.rs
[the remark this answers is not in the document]
JeremyRubin: suggest: rename as `pub use xCtx as x` for re-export while fixing internal naming.

src/miniscript/decode.rs
```diff
@@ -30,10 +31,67 @@ use std::sync::Arc;
...
 use MiniscriptKey;
 
+use ToPublicKey;
+
 fn return_none<T>(_: usize) -> Option<T> {
     None
 }
 
+/// Trait for parsing keys from byte slices
+pub trait ParseableKey: Sized + ToPublicKey + private::Sealed {
```
JeremyRubin, line 41: should parseablekey require miniscriptkey?

src/miniscript/lex.rs
JeremyRubin: these don't need to be unsized slices though.

you could do &[u8; 32] instead and not copy.

just polyfill this code for casting a slice to an array.,

```rust
#[stable(feature = "try_from", since = "1.34.0")]
impl<T, const N: usize> TryFrom<&[T]> for [T; N]
where
    T: Copy,
{
    type Error = TryFromSliceError;

    fn try_from(slice: &[T]) -> Result<[T; N], TryFromSliceError> {
        <&Self>::try_from(slice).map(|r| *r)
    }
}

#[stable(feature = "try_from", since = "1.34.0")]
impl<'a, T, const N: usize> TryFrom<&'a [T]> for &'a [T; N] {
    type Error = TryFromSliceError;

    fn try_from(slice: &[T]) -> Result<&[T; N], TryFromSliceError> {
        if slice.len() == N {
            let ptr = slice.as_ptr() as *const [T; N];
            // SAFETY: ok because we just checked that the length fits
            unsafe { Ok(&*ptr) }
        } else {
            Err(TryFromSliceError(()))
        }
    }
}
```

src/policy/semantic.rs
```diff
@@ -14,6 +14,7 @@
 
 //! Abstract Policies
 
+use std::collections::HashSet;
```
JeremyRubin, line 17: prefer using BTreeSet since it gives you reproducible results, although less critical for abstract policies than concrete.

JeremyRubin:
Overall looking pretty fantastic, awesome work @sanket1729. A few minor points of cleanup, but nothing that IMO feels blocking for merge (although would be worth it to fix before release).

src/miniscript/types/extra_props.rs
```diff
@@ -121,11 +121,32 @@ pub struct ExtData {
...
+            }
+        }
+        debug_assert!(!xor(
+            self.stack_elem_count_sat,
+            self.exec_stack_elem_count_sat
+        ));
+        debug_assert!(!xor(
+            self.stack_elem_count_dissat,
+            self.exec_stack_elem_count_dissat
+        ));
```
apoelstra, line 149: nit: this could probably be more clearly written as
```
debug_assert_eq!(self.stack_elem_count_stat.is_some(), self.exec_stack_elem_count_sat.is_some());
```
etc. without the `xor` helper.

PR #256, opened 2021-07-28
Bump rust-bitcoin dependency to v0.27

sandipndev:
Needed because `rust-bitcoin v0.27` has added Bech32m address support, see the complete changelog [here](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/CHANGELOG.md#027---2020-07-21).


sanket1729:
The CI would probably fail on the fuzz test, but that should be fixed after we merge #257.

apoelstra (approved):
nice. glad there were no changes required

PR #257, opened 2021-07-28
fix miniscript typesystem bug

sanket1729:
And disables a fuzz test which will be after I fix upstream issue in rust-secp256k1


apoelstra (approved):
utack

PR #265, opened 2021-08-19
descriptor/key: fix master fingerprint when there are none

darosior:
It's clearer to return `0x00000000` if we don't know the origin. This is arguably not an API break since, well, this code path could not be taken without panic'ing before.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -345,7 +345,7 @@ impl fmt::Display for ConversionError {
 impl error::Error for ConversionError {}
 
 impl DescriptorPublicKey {
-    /// The fingerprint of the master key associated with this key
+    /// The fingerprint of the master key associated with this key, `0x00000000` if none.
     pub fn master_fingerprint(&self) -> bip32::Fingerprint {
         match *self {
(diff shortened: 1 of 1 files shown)


darosior:
Hmm, fwiw this is not compatible with Bitcoin Core's behaviour (which the previous code tried to emulate). Is it problematic? I mean both versions are pretty explicit about "there is no parent".

sanket1729:
I think we should just stick to how core does it and take the first 4 bytes. My rationale is that we don't want users to rely on master fingerprint being 0's for checking that there is no parent.

apoelstra:
Where was the previous panic?

darosior:
In `Fingerprint::from` which asserts a len of 4 bytes (iirc)

darosior:
FWIW i'll update this soon (tm)

darosior:
Finally got to update this :)

sanket1729 (approved):
ACK 775da51094620dc1d607d1db4154d7cd1fdcb5cb

apoelstra:
ACK 775da51094620dc1d607d1db4154d7cd1fdcb5cb

Will leave it to Sanket whether we should go ahead and merge this now.

PR #266, opened 2021-08-19
policy: minimum_n_keys returns an Option

### LINKED ISSUE #264: Weird behavior for Policy::minimum_n_keys
`minimum_n_keys` returns 0 when it encounters `Policy::Unsatisfiable`, this leads to some strange behavior with policies such as:
```
or(UNSATISFIABLE,and(pkh(alice_pkh),pkh(bob_pkh)))
```
where `minimum_n_keys` returns 0, even though the minimum number of keys for satisfying that policy is 2.

danielabrozzoni:
minimum_n_keys returns None if the policy is not satisfiable.
It would previously (erroneously) return 0.

Fixes #264

### DIFF
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -791,16 +791,16 @@ mod tests {
 
         let mut abs = miniscript.lift().unwrap();
         assert_eq!(abs.n_keys(), 5);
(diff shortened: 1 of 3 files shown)


sanket1729 (approved):
ACK 427222e46498dae13d527e133d97af63c6e9328a

PR #267, opened 2021-08-30
Taproot descriptor

SarcasticNastik:
This defines tree structure for *TapTree* and implements parsing of **Taproot Descriptors** as a part of #255 . The purpose is to get basic functionalities for the same and  integrate it with the Miniscript compiler.

Feedback about changes to [tr.rs](https://github.com/SarcasticNastik/rust-miniscript/blob/c0930e623802fafbfacccb8cb062bf16ffb3ee52/src/descriptor/tr.rs) and [expression.rs](https://github.com/SarcasticNastik/rust-miniscript/blob/c0930e623802fafbfacccb8cb062bf16ffb3ee52/src/expression.rs) would be really appreciated. 
- Currently helper functions specific to `Tr` are required to parse taproot descriptor into `Tree` (as defined in [expression.rs](https://github.com/SarcasticNastik/rust-miniscript/blob/c0930e623802fafbfacccb8cb062bf16ffb3ee52/src/expression.rs)).

## TODO

- [ ] Docs for `tr.rs`
- [x] Pass fuzz tests
- [x] Include hex checksum for taproot descriptor.

## Future Work

- [ ] Implement functionalities required to integrate it with `Descriptor` (in [mod.rs](https://github.com/SarcasticNastik/rust-miniscript/blob/c0930e623802fafbfacccb8cb062bf16ffb3ee52/src/descriptor/mod.rs))

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -45,6 +45,7 @@ mod bare;
 mod segwitv0;
 mod sh;
 mod sortedmulti;
+mod tr;
 // Descriptor Exports
 pub use self::bare::{Bare, Pkh};
 pub use self::segwitv0::{Wpkh, Wsh, WshInner};
@@ -53,6 +54,7 @@ pub use self::sortedmulti::SortedMultiVec;
 
 mod checksum;
 mod key;
+
 pub use self::key::{
     ConversionError, DescriptorKeyParseError, DescriptorPublicKey, DescriptorSecretKey,
     DescriptorSinglePriv, DescriptorSinglePub, DescriptorXKey, InnerXKey, Wildcard,
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -46,6 +46,7 @@ mod segwitv0;
 mod sh;
 mod sortedmulti;
 mod tr;
+
 // Descriptor Exports
 pub use self::bare::{Bare, Pkh};
 pub use self::segwitv0::{Wpkh, Wsh, WshInner};
@@ -168,6 +169,9 @@ pub enum Descriptor<Pk: MiniscriptKey> {
     Sh(Sh<Pk>),
     /// Pay-to-Witness-ScriptHash with Segwitv0 context
     Wsh(Wsh<Pk>),
+    // /// Pay-to-Taproot with Segwitv0 context
+    // /// TODO: Update context to Segwitv1
+    // Tr(Tr<Pk>)
 }
 
 /// Descriptor Type of the descriptor
@@ -193,6 +197,8 @@ pub enum DescriptorType {
     WshSortedMulti,
     /// Sh Wsh Sorted Multi
     ShWshSortedMulti,
+    // /// Tr Descriptor
+    // Tr
 }
 
 impl<Pk: MiniscriptKey> Descriptor<Pk> {
@@ -279,6 +285,12 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
         Ok(Descriptor::Wsh(Wsh::new_sortedmulti(k, pks)?))
     }
 
+    // /// Create new tr descriptor
+    // /// Errors when miniscript exceeds resource limits under Segwitv0 context
+    // pub fn new_tr(key: Pk, script: Option<tr::TapTree<Pk>>) -> Result<Self, Error> {
+    //     Ok(Descriptor::Tr(Tr::new(key, script)?))
+    // }
+
     /// Get the [DescriptorType] of [Descriptor]
     pub fn desc_type(&self) -> DescriptorType {
         match *self {
@@ -298,6 +310,7 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
                 WshInner::SortedMulti(ref _smv) => DescriptorType::WshSortedMulti,
                 WshInner::Ms(ref _ms) => DescriptorType::Wsh,
             },
+            // Descriptor::Tr(_) => DescriptorType::Tr,
         }
     }
 }
@@ -624,6 +637,7 @@ serde_string_impl_pk!(Descriptor, "a script descriptor");
 #[cfg(test)]
 mod tests {
     use super::checksum::desc_checksum;
+    use super::tr::Tr;
     use super::DescriptorTrait;
     use bitcoin::blockdata::opcodes::all::{OP_CLTV, OP_CSV};
     use bitcoin::blockdata::script::Instruction;
@@ -1086,6 +1100,44 @@ mod tests {
         let check = actual_instructions.last().unwrap();
 
         assert_eq!(check, &Ok(Instruction::Op(OP_CSV)))
+    }
+
+    #[test]
+    fn tr_roundtrip_key() {
+        let script = Tr::<DummyKey>::from_str("tr()").unwrap().to_string();
+        assert_eq!(script, format!("tr()#x4ml3kxd"))
+    }
+
+    #[test]
+    fn tr_roundtrip_script() {
+        let descriptor = Tr::<DummyKey>::from_str("tr(,{pk(),pk()})")
+            .unwrap()
+            .to_string();
+
+        assert_eq!(descriptor, "tr(,{pk(),pk()})#7dqr6v8r")
+    }
+
+    #[test]
+    fn tr_roundtrip_tree() {
+        let p1 = "020000000000000000000000000000000000000000000000000000000000000001";
+        let p2 = "020000000000000000000000000000000000000000000000000000000000000002";
+        let p3 = "020000000000000000000000000000000000000000000000000000000000000003";
+        let p4 = "020000000000000000000000000000000000000000000000000000000000000004";
+        let p5 = "f54a5851e9372b87810a8e60cdd2e7cfd80b6e31";
+        let descriptor = Tr::<PublicKey>::from_str(&format!(
+            "tr({},{{pk({}),{{pk({}),or_d(pk({}),pkh({}))}}}})",
+            p1, p2, p3, p4, p5
+        ))
+        .unwrap()
+        .to_string();
+
+        assert_eq!(
+            descriptor,
+            format!(
+                "tr({},{{pk({}),{{pk({}),or_d(pk({}),pkh({}))}}}})#fdhmu4fj",
+                p1, p2, p3, p4, p5
+            )
+        )
     }
 
     #[test]
--- a/src/descriptor/tr.rs
+++ b/src/descriptor/tr.rs
@@ -1,0 +1,260 @@
+// Tapscript
+
+use super::checksum::{desc_checksum, verify_checksum};
+use bitcoin::hashes::_export::_core::fmt::Formatter;
+use errstr;
+use expression::{self, FromTree, Tree};
+use miniscript::{limits::TAPROOT_MAX_NODE_COUNT, Miniscript};
+use std::cmp::max;
+use std::sync::Arc;
+use std::{fmt, str::FromStr};
+use Segwitv0;
+use {Error, MiniscriptKey};
+
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub enum TapTree<Pk: MiniscriptKey> {
+    Tree(Arc<TapTree<Pk>>, Arc<TapTree<Pk>>),
+    Leaf(Arc<Miniscript<Pk, Segwitv0>>),
+}
+
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub struct Tr<Pk: MiniscriptKey> {
+    internal_key: Pk,
+    tree: Option<TapTree<Pk>>,
+}
+
+impl<Pk: MiniscriptKey> TapTree<Pk> {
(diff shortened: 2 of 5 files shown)


SarcasticNastik:
@sanket1729 Kindly review this PR while I add the required docs for [tr.rs](https://github.com/SarcasticNastik/rust-miniscript/blob/c0930e623802fafbfacccb8cb062bf16ffb3ee52/src/descriptor/tr.rs).

src/descriptor/tr.rs
```diff
@@ -0,0 +1,201 @@
...
+// };
+use bitcoin::hashes::_export::_core::fmt::Formatter;
+use expression::{self, FromTree, Tree};
+use std::sync::Arc;
+use std::{fmt, str::FromStr};
+use Segwitv0;
+use {miniscript::Miniscript, Error, MiniscriptKey};
+
+// TODO: Update this to infer version from descriptor.
+const VER: u8 = 0xc0;
```
sanket1729: In 73dd23e247dc47fd4d6d50244bc30e27abd3b487,
The above two line should be removed.

src/descriptor/tr.rs
```diff
@@ -0,0 +1,201 @@
...
+
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub enum TapTree<Pk: MiniscriptKey> {
+    Tree(Arc<TapTree<Pk>>, Arc<TapTree<Pk>>),
+    Miniscript_(u8, Arc<Miniscript<Pk, Segwitv0>>),
```
sanket1729: In 73dd23e247dc47fd4d6d50244bc30e27abd3b487,
The leaf version should be removed from Miniscript, and I think we can name the enum variants as `Miniscript_` as `Leaf`

src/descriptor/tr.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub struct Tr<Pk: MiniscriptKey> {
+    key_path: Pk,
```
sanket1729, line 20: rename as `internal_key` and `tree`? `key_path` and `script_path` represent the ways in which we spend scripts, not the structures.

src/descriptor/tr.rs
```diff
@@ -0,0 +1,201 @@
...
+impl<Pk: MiniscriptKey> TapTree<Pk>
+where
+    Pk: MiniscriptKey + FromStr,
+    Pk::Hash: FromStr,
+    <Pk as FromStr>::Err: ToString,
+    <<Pk as MiniscriptKey>::Hash as FromStr>::Err: ToString,
+{
+    pub fn to_string_no_checksum(&self) -> String {
+        match self {
+            TapTree::Tree(ref left, ref right) => format!("{{{},{}}}", *left, *right),
```
sanket1729: Wow, surprised that this really works. I would expected that you might need `\` for `{`

src/descriptor/tr.rs
```diff
@@ -0,0 +1,201 @@
...
+impl<Pk: MiniscriptKey> fmt::Display for TapTree<Pk>
+where
+    Pk: MiniscriptKey + FromStr,
+    Pk::Hash: FromStr,
+    <Pk as FromStr>::Err: ToString,
+    <<Pk as MiniscriptKey>::Hash as FromStr>::Err: ToString,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let desc = self.to_string_no_checksum();
+        // let checksum = desc_checksum(&desc).map_err(|_| fmt::Error)?;
```
sanket1729: Commented code here. Either squash the checksum related to this commit or remove the comments from this commit.

src/descriptor/tr.rs
```diff
@@ -0,0 +1,201 @@
...
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub struct Tr<Pk: MiniscriptKey> {
+    key_path: Pk,
+    script_path: Option<TapTree<Pk>>,
+}
+
+impl<Pk: MiniscriptKey> TapTree<Pk>
+where
+    Pk: MiniscriptKey + FromStr,
+    Pk::Hash: FromStr,
```
sanket1729, line 27: I don't think we should need the `FromStr` bounds for this. Can you double-check if all of these bounds are really necessary.

sanket1729: Or the `::Err` bounds

src/descriptor/tr.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+impl<Pk: MiniscriptKey> fmt::Display for TapTree<Pk>
+where
+    Pk: MiniscriptKey + FromStr,
+    Pk::Hash: FromStr,
+    <Pk as FromStr>::Err: ToString,
+    <<Pk as MiniscriptKey>::Hash as FromStr>::Err: ToString,
+{
```
sanket1729, line 47: Same comment about the bounds here.

src/descriptor/tr.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+impl<Pk: MiniscriptKey> Tr<Pk>
+where
+    Pk: MiniscriptKey + FromStr,
+    Pk::Hash: FromStr,
+    <Pk as FromStr>::Err: ToString,
+    <<Pk as MiniscriptKey>::Hash as FromStr>::Err: ToString,
+{
```
sanket1729, line 60: Same comment about the bounds here. Try to separate out the functions that really need the bound into its own `impl`. For example, the new function `Tr::new()` should not need this bound.

sanket1729: Also make accessor function for fields in `Tr` as those are private

src/descriptor/tr.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub fn new(key_path: Pk, script_path: Option<TapTree<Pk>>) -> Result<Self, Error> {
```
sanket1729, line 61: Add this check and return an error when this does not hold. 
https://github.com/bitcoin/bitcoin/blob/81f4a3e84d6f30e7b12a9605dabc3359f614da93/src/script/interpreter.h#L229.

Introduce this constant in limits.rs

Right now, you can implement a simple depth/height function on `Taptree`. 

We can cache the height/depth in the future.

src/expression.rs
```diff
@@ -38,11 +38,29 @@ pub trait FromTree: Sized {
 }
 
 impl<'a> Tree<'a> {
-    fn from_slice(sl: &'a str) -> Result<(Tree<'a>, &'a str), Error> {
-        Self::from_slice_helper(sl, 0u32)
+    pub fn from_slice(sl: &'a str) -> Result<(Tree<'a>, &'a str), Error> {
```
sanket1729, line 41: In 73dd23e247dc47fd4d6d50244bc30e27abd3b487, the function signature should be changed in this commit itself.

src/lib.rs
```diff
@@ -95,9 +95,9 @@
 #![deny(non_camel_case_types)]
 #![deny(non_snake_case)]
 #![deny(unused_mut)]
-#![deny(dead_code)]
-#![deny(unused_imports)]
-#![deny(missing_docs)]
+// #![deny(dead_code)]
+// #![deny(unused_imports)]
+// #![deny(missing_docs)]
```
sanket1729, line 100: in 73dd23e247dc47fd4d6d50244bc30e27abd3b487, this is fine. 


Adding this change to check this is re-enabled again.

sanket1729 (commented):
This looks good. I just did a partial review and left some specific comments in the first commit.
Here are some general remarks:

In general, commits should be atomic (and they should just do one thing and the one thing completely in that commit) and they should do pass all tests/build. 

A good way to test this is via `-x` option in `git rebase`.
 
`git rebase -i HEAD~4 -x "cargo test --features=compiler"`. This will stop the code at the commit where the build/test fails, you can fix the changes there and `git rebase --continue`. 

It might take a while to get familiar with advanced git techniques, but it is an important skill in the long run. Every project uses this method so that it is easier for tools like `git bisect`(essentially a binary search on commits) to figure out errors when something goes wrong and figure out which commit things went bad.

I think for starters, you can separate this into two commits. Adding the code and later another commit for adding the test.

src/descriptor/mod.rs
```diff
@@ -45,6 +45,7 @@ mod bare;
 mod segwitv0;
 mod sh;
 mod sortedmulti;
+mod tr;
```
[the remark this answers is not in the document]
sanket1729: This is still not addressed

src/descriptor/tr.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+
+    // helper function for semantic parsing of script paths
+    pub fn tr_script_path(tree: &Tree) -> Result<TapTree<Pk>, Error> {
+        match tree {
+            Tree { name, args } if name.len() > 0 && args.len() == 0 => {
+                let script = name;
+                let script = Self::parse_miniscript(script)?;
+                let script = Arc::new(script);
```
sanket1729, line 114: nit: slight code preference for inlining some operations. You don't really need the function `parse_miniscript`. You can call 
```
Miniscript::<Pk, Ctx>::from_str()
```
to get the miniscript directly. The function already does all the checks that you implement here. Also, the method `parse_method` does make sense itself inside an impl for `TapTree`

src/descriptor/tr.rs
```diff
@@ -0,0 +1,279 @@
...
+                Ok(TapTree::Leaf(script))
+            }
+            Tree { name, args } if name.len() == 0 && args.len() == 2 => {
+                let left_branch = &args[0];
+                let right_branch = &args[1];
+                let left_tree = Self::tr_script_path(&left_branch)?;
+                let right_tree = Self::tr_script_path(&right_branch)?;
+                let left_ref = Arc::new(left_tree);
+                let right_ref = Arc::new(right_tree);
```
sanket1729, line 123: nit: Prefer inlining some code instead of 6 lines. 

```
let left = Self::tr_script_path(&args[0])?;
..
Ok(TapTree::Tree(Arc::new(left_ref), Arc::new(right_ref)))
```

src/descriptor/tr.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+fn parse_tr(s: &str) -> Result<Tree, Error> {
+    for ch in s.bytes() {
+        if ch > 0x7f {
+            return Err(Error::Unprintable(ch));
+        }
+    }
+
+    return if s.len() > 3 && &s[..3] == "tr(" && s.as_bytes()[s.len() - 1] == b')' {
```
sanket1729, line 223: assign let ret = expr; and do a return statement in the end

src/descriptor/tr.rs
```diff
@@ -0,0 +1,279 @@
...
+        for (idx, ch) in inp.chars().enumerate() {
+            if ch == delim {
+                found = idx;
+                break;
+            }
+        }
+        if found == inp.len() - 1 {
+            Some((&inp[..], &""))
+        } else {
+            Some((&inp[..found], &inp[found + 1..]))
```
sanket1729, line 276: Add a comment that this deals with the case when the `,` is the last character.

sanket1729 (commented):
Left some more remarks, we are getting closer :)

src/descriptor/tr.rs
SarcasticNastik: Will do! Besides this, refactoring this function to a method should be more idiomatic, right?

SarcasticNastik:
@sanket1729 I have made the suggested changes. Apart from that,
- In the [tr_roundtrip_tree](https://github.com/SarcasticNastik/rust-miniscript/blob/ff365fd19e5f5230e52d59d5812de45aa6bd97a9/src/descriptor/mod.rs#L1121) test, using DummyKey caused `AnalysisError::RepeatedPubKeys` error (I think we have to turn that off while using `DummyKey`) so I changed the test to use `PublicKey` accordingly.

src/descriptor/mod.rs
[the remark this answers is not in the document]
sanket1729: We still need a statement with `pub use self::tr::{TapTree}` and all other utilities that need to be exposed.

sanket1729 (commented):
I think this is good overall. I can take it over from here and update the PR once #255 is merged.

src/descriptor/tr.rs
```diff
@@ -0,0 +1,260 @@
...
+use std::cmp::max;
+use std::sync::Arc;
+use std::{fmt, str::FromStr};
+use Segwitv0;
+use {Error, MiniscriptKey};
+
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub enum TapTree<Pk: MiniscriptKey> {
+    Tree(Arc<TapTree<Pk>>, Arc<TapTree<Pk>>),
+    Leaf(Arc<Miniscript<Pk, Segwitv0>>),
```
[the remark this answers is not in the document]
sanket1729: @dr-orlovsky. This is being done in #278 now

sanket1729:
This commit was merged as a part of #278.

PR #269, opened 2021-09-17
Fix decoding bug

sanket1729:
Caught while reviewing c++ implementation(PR from meshcollider).

### DIFF
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -909,4 +909,24 @@ mod tests {
             .to_string()
             .contains("unprintable character"));
     }
+
+    #[test]
+    fn temp() {
+        let ms = Miniscript::<String, Segwitv0>::from_str_insane(
+            "and_b(1,s:and_v(v:older(9),c:pk_k(A)))",
+        )
+        .unwrap();
+        let ms_trans = ms.translate_pk_infallible(
+            |x| {
+                bitcoin::PublicKey::from_str(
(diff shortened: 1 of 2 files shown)


src/miniscript/mod.rs
```diff
@@ -909,4 +909,24 @@ mod tests {
             .to_string()
             .contains("unprintable character"));
     }
+
+    #[test]
+    fn temp() {
```
apoelstra, line 914: Can you rename this test? Otherwise ack cec1e08351d97cb5acb56fc3b8bfc505c5683fac

sanket1729: Oops yeah, this was for initial testing.

darosior (commented):
ACK 84e001c16e03b70e58020e1aa574f702d7c36d5b

apoelstra (approved):
ACK 87c65b65d28429fb09ad45551adfbf654a5f0546

JeremyRubin:
do you have a link to the issue?

sanket1729:
@JeremyRubin, I don't recall the exact details. This was related to incorrectly judging when an expression is swap. 
`and_b(1,s:and_v(v:older(9),c:pk_k(A)))` we are applying Swap to the entire thing previously because of incorrectly placed `MaybeSwap`. 

The new code is much cleaner in we have a cleanly marked `WExpression` where we know if we don't see `FromAtlStack`, we must see a Swap.

PR #275, opened 2021-10-21
Miniscript tapscript support (Pr 2 of 3)

sanket1729:
Adds support for MultiA script fragment. Builds on top of #255 .

Do not merge this until we have a rust-bitcoin release

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -27,7 +27,7 @@ use miniscript::context::NoChecks;
 use miniscript::ScriptContext;
 use Miniscript;
 use Terminal;
-use {BitcoinSig, Descriptor, ToPublicKey};
+use {Descriptor, ToPublicKey};
 
 mod error;
 mod inner;
@@ -82,7 +82,7 @@ impl<'txin> Interpreter<'txin> {
     ///
     /// Running the iterator through will consume the internal stack of the
     /// `Iterpreter`, and it should not be used again after this.
-    pub fn iter<'iter, F: FnMut(&bitcoin::PublicKey, BitcoinSig) -> bool>(
+    pub fn iter<'iter, F: FnMut(&bitcoin::PublicKey, bitcoin::EcdsaSig) -> bool>(
         &'iter mut self,
         verify_sig: F,
     ) -> Iter<'txin, 'iter, F> {
@@ -190,7 +190,7 @@ impl<'txin> Interpreter<'txin> {
         unsigned_tx: &'a bitcoin::Transaction,
         input_idx: usize,
         amount: u64,
-    ) -> Result<impl Fn(&bitcoin::PublicKey, BitcoinSig) -> bool + 'a, Error> {
+    ) -> Result<impl Fn(&bitcoin::PublicKey, bitcoin::EcdsaSig) -> bool + 'a, Error> {
         // Precompute all sighash types because the borrowck doesn't like us
         // pulling self into the closure
         let sighashes = [
@@ -232,19 +232,21 @@ impl<'txin> Interpreter<'txin> {
             )?,
         ];
 
-        Ok(move |pk: &bitcoin::PublicKey, (sig, sighash_type)| {
-            // This is an awkward way to do this lookup, but it lets us do exhaustiveness
-            // checking in case future rust-bitcoin versions add new sighash types
-            let sighash = match sighash_type {
-                bitcoin::EcdsaSigHashType::All => sighashes[0],
-                bitcoin::EcdsaSigHashType::None => sighashes[1],
-                bitcoin::EcdsaSigHashType::Single => sighashes[2],
-                bitcoin::EcdsaSigHashType::AllPlusAnyoneCanPay => sighashes[3],
-                bitcoin::EcdsaSigHashType::NonePlusAnyoneCanPay => sighashes[4],
-                bitcoin::EcdsaSigHashType::SinglePlusAnyoneCanPay => sighashes[5],
-            };
-            secp.verify_ecdsa(&sighash, &sig, &pk.key).is_ok()
-        })
+        Ok(
+            move |pk: &bitcoin::PublicKey, ecdsa_sig: bitcoin::EcdsaSig| {
+                // This is an awkward way to do this lookup, but it lets us do exhaustiveness
+                // checking in case future rust-bitcoin versions add new sighash types
+                let sighash = match ecdsa_sig.hash_ty {
+                    bitcoin::EcdsaSigHashType::All => sighashes[0],
+                    bitcoin::EcdsaSigHashType::None => sighashes[1],
+                    bitcoin::EcdsaSigHashType::Single => sighashes[2],
+                    bitcoin::EcdsaSigHashType::AllPlusAnyoneCanPay => sighashes[3],
+                    bitcoin::EcdsaSigHashType::NonePlusAnyoneCanPay => sighashes[4],
+                    bitcoin::EcdsaSigHashType::SinglePlusAnyoneCanPay => sighashes[5],
+                };
+                secp.verify_ecdsa(&sighash, &ecdsa_sig.sig, &pk.key).is_ok()
+            },
+        )
     }
 }
 
@@ -327,7 +329,7 @@ struct NodeEvaluationState<'intp> {
 ///
 /// In case the script is actually dissatisfied, this may return several values
 /// before ultimately returning an error.
-pub struct Iter<'intp, 'txin: 'intp, F: FnMut(&bitcoin::PublicKey, BitcoinSig) -> bool> {
+pub struct Iter<'intp, 'txin: 'intp, F: FnMut(&bitcoin::PublicKey, bitcoin::EcdsaSig) -> bool> {
     verify_sig: F,
     public_key: Option<&'intp bitcoin::PublicKey>,
     state: Vec<NodeEvaluationState<'intp>>,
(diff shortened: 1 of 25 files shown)


src/interpreter/mod.rs
```diff
@@ -189,34 +189,34 @@ impl<'txin> Interpreter<'txin> {
         unsigned_tx: &'a bitcoin::Transaction,
         input_idx: usize,
         amount: u64,
-    ) -> impl Fn(&bitcoin::PublicKey, BitcoinSig) -> bool + 'a {
+    ) -> Result<impl Fn(&bitcoin::PublicKey, BitcoinSig) -> bool + 'a, Error> {
```
apoelstra: I think, rather than returning an error here, we should return a closure that always returns `false`. But I'm not sure.

src/miniscript/satisfy.rs
```diff
@@ -947,6 +947,44 @@ impl Satisfaction {
...
+                let mut sigs = vec![vec![vec![]]; keys.len()];
+                for (i, pk) in keys.iter().enumerate() {
+                    match Witness::signature(stfr, pk) {
+                        Witness::Stack(sig) => {
+                            sigs[i] = sig;
+                            sig_count += 1;
+                            // This a privacy issue, we are only selecting the first available
+                            // sigs. Incase pk at pos 1 is not selected, we know we did not have access to it
+                            // bitcoin core also implements the same logic for MULTISIG, so I am not bothering
+                            // permuting the sigs for now
```
apoelstra, line 1126: In c05eb7c70d087269cc52c156b4ae4eadd2657133:

Ok to leave this comment here, but TBH I don't think there's any way we could eliminate this privacy leak. If the first signature is missing then the resulting witnesses will have a different distribution than if it's not ... which will become apparent over time.

Although, I guess, in practice users are unlikely to reuse keys in visible ways, so an adversary will have only a one-shot view of the signature distribution.

src/interpreter/mod.rs
JeremyRubin: maybe you want a closure that always returns a Result<bool, Error>?

src/descriptor/sh.rs
```diff
@@ -209,10 +209,18 @@ impl<Pk: MiniscriptKey> DescriptorTrait<Pk> for Sh<Pk> {
...
+            ShInner::Wsh(ref wsh) => Ok(bitcoin::Address::p2sh(&wsh.script_pubkey(), network)
+                .expect("Size checked in Miniscript")),
+            ShInner::Wpkh(ref wpkh) => Ok(bitcoin::Address::p2sh(&wpkh.script_pubkey(), network)
+                .expect("Size checked in Miniscript")),
+            ShInner::SortedMulti(ref smv) => {
+                Ok(bitcoin::Address::p2sh(&smv.encode(), network)
+                    .expect("Size checked in Miniscript"))
+            }
+            ShInner::Ms(ref ms) => {
+                Ok(bitcoin::Address::p2sh(&ms.encode(), network)
```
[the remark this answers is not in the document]
sanket1729: Agreed, it is much safer to do so. This does not rely on size check being the only error.

src/interpreter/mod.rs
sanket1729: This should return an error if the input_idx is out of bounds.

On related note, I think we need a major redesign of this for taproot support anyways. This is currently hacked together that supports segwitv0 and legacy sighash. For taproot, we need entire sets of prevouts, it may be worth considering redesigning this entire API (yet) again :P

src/miniscript/astelem.rs
```diff
@@ -554,7 +574,12 @@ where
                     .map(|sub| expression::terminal(sub, Pk::from_str))
                     .collect();
 
-                pks.map(|pks| Terminal::Multi(k, pks))
+                if frag_name == "multi" {
+                    pks.map(|pks| Terminal::Multi(k, pks))
+                } else {
+                    // must be multi_a
+                    pks.map(|pks| Terminal::MultiA(k, pks))
+                }
```
dr-orlovsky, line 582: I do believe this should be `match`; otherwise with future TapLeaf version-defined post-TapScripts we will not have a compiler error here

src/miniscript/satisfy.rs
```diff
@@ -34,28 +34,107 @@ use Miniscript;
...
 
-/// Type alias for a signature/hashtype pair
-pub type BitcoinSig = (secp256k1::Signature, bitcoin::SigHashType);
+/// type alias for legacy sighash type
+pub type LegacySigHashType = bitcoin::SigHashType;
+/// type alias for schnorr signature hash type (includes sighash default)
+pub type SchnorrSigHashType = bitcoin::util::sighash::SigHashType;
+/// Type for a EC signature/hashtype pair
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct BitcoinECSig {
```
dr-orlovsky, line 43: Btw we will have this type in `rust-bitcoin`: https://github.com/rust-bitcoin/rust-bitcoin/pull/669

As for the name, probably it should have `ECDSA` part (not simply `EC`)

src/miniscript/satisfy.rs
```diff
@@ -34,28 +34,107 @@ use Miniscript;
...
 use Terminal;
 
-/// Type alias for a signature/hashtype pair
-pub type BitcoinSig = (secp256k1::Signature, bitcoin::SigHashType);
+/// type alias for legacy sighash type
+pub type LegacySigHashType = bitcoin::SigHashType;
+/// type alias for schnorr signature hash type (includes sighash default)
+pub type SchnorrSigHashType = bitcoin::util::sighash::SigHashType;
+/// Type for a EC signature/hashtype pair
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
dr-orlovsky, line 42: `PartialOrd, Ord, Hash`

src/miniscript/satisfy.rs
```diff
@@ -34,28 +34,107 @@ use Miniscript;
...
+            // nothing to push for sighash type
+        } else {
+            ret.push(self.hash_ty as u8);
+        }
+        ret
+    }
 }
+
+/// Type for a schnorr signature/hashtype pair
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
dr-orlovsky, line 118: `PartialOrd, Ord, Hash`

src/miniscript/satisfy.rs
```diff
@@ -34,28 +34,107 @@ use Miniscript;
...
+pub struct BitcoinSchnorrSig {
```
dr-orlovsky, line 119: Related: https://github.com/rust-bitcoin/rust-bitcoin/pull/671

src/miniscript/satisfy.rs
```diff
@@ -401,15 +563,19 @@ impl Ord for Witness {
...
 impl Witness {
     /// Turn a signature into (part of) a satisfaction
-    fn signature<Pk: ToPublicKey, S: Satisfier<Pk>>(sat: S, pk: &Pk) -> Self {
-        match sat.lookup_sig(pk) {
-            Some((sig, hashtype)) => {
-                let mut ret = sig.serialize_der().to_vec();
-                ret.push(hashtype.as_u32() as u8);
-                Witness::Stack(vec![ret])
+    fn signature<Pk: ToPublicKey, S: Satisfier<Pk>, Ctx: ScriptContext>(sat: S, pk: &Pk) -> Self {
+        if Ctx::is_tap() {
```
dr-orlovsky, line 567: With more post-TapScript contexts in the future we will get a bug in this code and compiler will not warn/error. So I propose to use `match` here and not `if`

dr-orlovsky (commented):
utACK mod some code style comments and proposal to re-use types which will be introduced in `rust-bitcoin` (once they got merged).

Also think that `Tap` context should be renamed to `TapScript`, since taproot may have more scripts in the future and `Tap` context requirements cover only leaf hash version `0xC0`, but not other future versions with their own contexts

src/miniscript/astelem.rs
sanket1729: There is a match check that frag_name is either `multi` or `multi_a`. I don't see an issue with doing an if-else here. 

```rust
("multi", n) | ("multi_a", n) => {
  				// code..
                if frag_name == "multi" {
                    pks.map(|pks| Terminal::Multi(k, pks))
                } else {
                    // must be multi_a
                    pks.map(|pks| Terminal::MultiA(k, pks))
                }
            }
```

src/miniscript/satisfy.rs
sanket1729: Good catch

src/miniscript/satisfy.rs
sanket1729: Agreed.

sanket1729:
Pushed another branch. Some portions of this PR were merged in rust-bitcoin, so the review here should be (slightly) easy. Since this depends on a specific from rust-bitcoin, it is not possible to run integration tests on these because the dependencies operate on rust-bitcoin release. 

I think I should have considered this before merging the integration tests PR, but now we have to wait for a rust-bitcoincore-rpc release too.

src/miniscript/satisfy.rs
sanket1729: I think the last commit should make this cleaner. Let me know if you want me to reorder the commits.

sanket1729:
Rebased on ready for review.

apoelstra (approved):
ACK 42e4c50689ee96fc0a806f02c88834f24d78fd59

I'm a little unsure about the last commit - splitting `NoChecks`. In this PR, `NoChecksSchnorr` is not actually used and the interpreter continues to be ECDSA-only. It's not obvious how to taproot-enable the interpreter, so I trust you have a plan in mind that will make use of the new split `NoChecks`.

sanket1729:
> In this PR, NoChecksSchnorr is not actually used and the interpreter continues to be ECDSA-only. It's not obvious how to taproot-enable the interpreter, so I trust you have a plan in mind that will make use of the new split NoChecks. 

I do have a plan for it. Our test-cases actually fail if we don't have this. We do a fancy interpreter check after we finalize a psbt to see that everything is good before we spend it. In that we need to serialize a script, We need some way to figure out `Miniscript<Pk, NoChecks>` should serialize key as 33 byte or 32 byte one or which signature algorithm to use for verification. All other contexts have this information implicitly in them because of associated consensus key type.

This was also done partly to address https://github.com/rust-bitcoin/rust-miniscript/pull/275#discussion_r748818958 in this PR.

apoelstra:
Ah, I forgot about the PSBT sanity check. Makes sense!

PR #276, opened 2021-10-26
Fix over-generic parameterization of ScriptContext

JeremyRubin:
The previous genericization of the ScriptContext methods permitted some very surprising behaviors that are certainly not intended. The Ctx Parameters should in all cases be a Self parameter as far as I can tell. This PR fixes this issue.

fixes https://github.com/rust-bitcoin/rust-miniscript/pull/255#discussion_r711772814

Technically, this is a breaking change as the generic paremeter count decreases and it is not generally inferred I think.  It's easy to fix, and any code which could not be fixed trivially would be hopelessly broken/buggy.

### DIFF
--- a/src/miniscript/context.rs
+++ b/src/miniscript/context.rs
@@ -155,16 +155,14 @@ where
     /// This does NOT recursively check if the children of the fragment are
     /// valid or not. Since the compilation proceeds in a leaf to root fashion,
     /// a recursive check is unnecessary.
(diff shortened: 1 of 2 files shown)


dr-orlovsky (approved):
utACK 6a891562be49718afb8b7774e21bcb6e52c5b604

apoelstra:
lol great catch. post-merge concept ACK.

PR #278, opened 2021-11-02
Add tr descriptor

sanket1729:


src/descriptor/tr.rs
```diff
@@ -0,0 +1,671 @@
...
+// be simple to integrate those here, but it is best to wait on core for the exact syntax.
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub enum TapTree<Pk: MiniscriptKey> {
+    /// A taproot tree structure
+    Tree(Arc<TapTree<Pk>>, Arc<TapTree<Pk>>),
+    /// A taproot leaf denoting a spending condition
+    // A new leaf version would require a new Context, therefore there is no point
+    // in adding a LeafVersion with Leaf type here. All Miniscripts right now
+    // are of Leafversion::default
+    Leaf(Arc<Miniscript<Pk, Tap>>),
```
dr-orlovsky, line 36: I think the problem may be solved by having a dedicated enum variant for each Leaf version. So this will become `LeafTapScript`, the enum will become `#[non_exhaustive]` and there will be no problem to add more contexts with more enum variants with post-`0xC0` leafs

src/descriptor/bare.rs
```diff
@@ -63,6 +63,26 @@ impl<Pk: MiniscriptKey> Bare<Pk> {
     }
 }
 
+impl<Pk: MiniscriptKey + ToPublicKey> Bare<Pk> {
```
dr-orlovsky, line 66: I think all this non-failing methods can be re-organized into an additional trait, let's say `PreTaprootDescriptor: Descriptor`, such that all pre-taproot descriptor types may be used as a generalized function arguments

dr-orlovsky (commented):
Concept ACK. Left two API improvement suggestions

src/util.rs
```diff
@@ -46,3 +48,15 @@ impl MsKeyBuilder for script::Builder {
         }
     }
 }
+
+// This belongs in rust-bitcoin, make this upstream later
+pub(crate) fn script_is_v1_tr(s: &Script) -> bool {
+    let mut iter = s.instructions_minimal();
+    if s.len() != 32
```
dr-orlovsky, line 55: This is bug: the key is 32 bytes indeed, but the script s 34.

```suggestion
    if s.len() != 34
```

dr-orlovsky (changes requested):
Spotted one bug

src/util.rs
```diff
@@ -46,3 +48,15 @@ impl MsKeyBuilder for script::Builder {
...
+        || iter.next() != Some(Ok(Instruction::Op(opcodes::all::OP_PUSHNUM_1)))
+        || iter.next() != Some(Ok(Instruction::Op(opcodes::all::OP_PUSHBYTES_32)))
```
dr-orlovsky, line 57: This does not work neither, I do not know why. But using `spk.is_v1_p2tr()` makes things working

src/descriptor/tr.rs
sanket1729: Yeah, marking everything non-exhaustive is a big step. I suppose we will have a big breaking as soon as the MSRV lands

src/descriptor/bare.rs
sanket1729: We could do for the release. But I think that the effort is not worth it considering everything will go back to non-fallible once we have a new major breaking release of rust-secp that has static contexts

src/descriptor/tr.rs
```diff
@@ -0,0 +1,698 @@
...
+}
+
+impl<Pk: MiniscriptKey + ToPublicKey> Tr<Pk> {
+    /// Obtain the corresponding script pubkey for this descriptor
+    /// Same as[`DescriptorTrait::script_pubkey`] for this descriptor
+    pub fn spk(&self) -> Result<Script, Error> {
+        let spend_info = self
+            .spend_info
+            .as_ref()
+            .ok_or(Error::TaprootSpendInfoUnavialable)?;
```
LLFourn, line 230: This error case is the thing that forces as to make `script_pubkey()` return a `Result` right? Wouldn't it be preferable to make caching `spend_info` optional. So make a public API for caching it but if it hasn't been cached just compute it on the fly. For me, the risk of possibly doing some extra (usually minor) computation time seems better than giving me a `Result` that I may arrogantly `.unwrap()`.

sanket1729: This is not possible because we require a `secp` parameter to do the tweaking. Creating a new context inside a method is expensive and causes severe performance issues. This would go away once we get static verification tables in rust-secp, but still a while to go.

I am also unhappy with the current PR :( . Some rough thoughts, 

- I am considering more changes in the type system to distinguish between `Derived` and `UnDerived` descriptors. We would then implement `DescriptorTriat` only for `Derived` descriptors. Or perhaps having two traits.
- I also don't like that `address()` fails for raw descriptors which are not that common. Maybe `DescriptorTrait` methods are not that useful. 

I will make another candidate PR by tomorrow with the goal of avoiding unwraps() for common case like `script_pubkey` and `address` etc

sanket1729:
@apoelstra, the last commit is a suggested way to avoid calling `.compute_spend_info` everytime. The code naming/quality is not perfect, but this is an example design that would allow us to remove the `TaprootCacheNotPresent` error. It basically forces the user to call a function to convert `Descriptor<Pk, Uncached>` to `Descriptor<Pk, Cached>`. 

It would be great if we can get rust-secp release with static contexts, then we don't have to worry about caching at all. In the interest of moving things forward, I will be separating the PR into two parts. Adding a 1) native `Tr` descriptor and 2) Integrating it with the DescriptorTrait.

LLFourn:
hey @sanket1729 -- I've been testing out this branch. FWIW I found the `Cached` an `Uncached` distinction easy to work with. I just did the caching when going from abstract descriptor (e.g. with wildcards) to concrete descriptor. I think this allows you remove the `Result`s on the trait but you just haven't gotten around to it yet?

One thing that I found to be inconvenient was the `DescriptorTrait` was it was only implemented on `Cached` Descriptors but `max_satisfaction_weight` could theoretically be answered by by an uncached descriptor as well (I think). So to figure out the `max_satisfaction_weight` of a more abstract descriptor I had to first make it concrete and then cache the spending info and then ask it its max satisfaction weight.

sanket1729:
Hey @LLFourn, good news is that we now would not even have to do the `Cache` part. Creating new verification-contexts would be cheap as secp new release (https://github.com/rust-bitcoin/rust-secp256k1/pull/384 has support for static contexts). Eventually (maybe in the next release of a rust-bitcoin ecosystem cycle), we could rid of all verification context parameters altogether. Static Signing context is a bit more complicated, but fortunately, we don't really do it in this lib. 

> but you just haven't gotten around to it yet?

I stopped working on this, because the cheap context creation means that we should be back to the old API :)

sanket1729:
I did some benchmarking on the rust-bitcoin/rust-secp256k1#384 for the tweaking.

Currently

```
test benches::bench_context_create_and_tweak ... bench:   5,183,500 ns/iter (+/- 42,121)
```

After the update:

```
test benches::bench_context_create_and_tweak ... bench:      54,330 ns/iter (+/- 301)
```

Ideally (with full static context support), meaning no calls to context_create.
```
test benches::bench_context_create_and_tweak ... bench:      23,253 ns/iter (+/- 62)
```

A 100 times increase from 5ms to 50mirco s. I think (for now), it is acceptable to create a new context with a double overhead. This is a massive improvement over the previous 100 times overhead. This can be improved later when we have an API without secp context like `xpk.tweak(&tweak)`. This allows us to have a cleaner API for now, with a performance hit at verification time. 

I will be working on a follow-up PR where we use interior mutability to cache the spend_data so that only the first invocation would be costly. Subsequent invocations can use the cached value of `SpendInfo`

apoelstra:
The 50us delay should also go away in the near(ish) future -- I believe it is caused by our current logic re-randomizing even static contexts on creation, which (a) ought to be done at compile-time, it doesn't use fresh randomness, and (b) doesn't need to be done at all for verification-only contexts.

sanket1729 (commented):
@apoelstra, should we proceed with this PR?

 The current implementation in master would be inefficient and slow and until we update to the latest rust-bitcoin and rust-secp release. I think it would save some time in subsequent rust-miniscript release if we can proceed with this review meanwhile. 
 
 I am not sure if we should merge this or not? Part of me says yes that people should not be using `master` branch directly for development and they can expect to find bugs. I have two more follow-up PRs to this one, so might take some time before we get a rust-miniscript release even after we have a rust-bitcoin release.

nickfarrow:
Really like the removal of `Cached` and the unwraps. I've also been using this branch and found that the taproot [spend_info](https://github.com/sanket1729/rust-miniscript/pull/4/files#diff-93b7da5ab70ed3a0305ed89a166c4594ee548bff8d71afb6466a39b4df450dd9L215) reading from cache can hang. With idea for fix https://github.com/sanket1729/rust-miniscript/pull/4

sanket1729:
Thanks @nickfarrow, merged, credited, and squashed your commit into [7eac908](https://github.com/rust-bitcoin/rust-miniscript/pull/278/commits/7eac9080bd525c47b046e846848fb629e8607345).

apoelstra:
Done reviewing f5c8d7ff5dd885508a2d4fbed6a4876b972a7a1c

Overall looks great. A few comments, which I think should be addressed in future PRs:

* I think we should split off a new struct `RawTr` from the `Tr` descriptor, and require that `Tr` have a script path (or hidden tree or whatever). The current situation has `Tr::internal_key` confusingly returning the external key in the case that there is no script path.
* I think we should have a `PreTaprootDescrtiptor` and `TaprootDescriptor` trait which are subtraits of `Descriptor` and have additional methods, rather than making many of the existing methods fallible. Then the new `ecdsa_sighash_script_code` can just be called `script_code` again.
* I'm confused by the rename of `lookup_schnorr_sig` to `lookup_tap_leaf_script_sig`. Also what if we made the signature type a type parameter of `Satisfier` then we could just use `lookup_sig` for both signature types?
* I'm not thrilled with the new expression parsing code; it seems like we could reduce the duplicated code.

Finally, I have some fixups at https://github.com/apoelstra/rust-miniscript/tree/2022-02--278-fixups -- mostly minor stuff, but I also cleaned up the locking code.

sanket1729:
> The current situation has Tr::internal_key confusingly returning the external key in the case that there is no script path.

I don't follow this. Every `Tr` descriptor must have an internal key. There is no confusion with the external_key(or output_key) here.

> I think we should have a PreTaprootDescrtiptor and TaprootDescriptor trait which are subtraits of Descriptor and have additional methods, rather than making many of the existing methods fallible. Then the new ecdsa_sighash_script_code can just be called script_code again.

Agreed.

> I'm confused by the rename of lookup_schnorr_sig to lookup_tap_leaf_script_sig. Also what if we made the signature type a type parameter of Satisfier then we could just use lookup_sig for both signature types?

The naming was mostly done to mimic the psbt naming. This is slightly more complicated because there are two different uses of Schnorr sigs. Overall, we have three main types of sigs:
- Regular ECDSA sig lookup (pk -> ecdsa_sig)
- Schnorr keyspend lookup (this does not have any notion of leaf_hash) (xonly_pk -> schnorr_sig) 
- Schnorr leaf sig that has key (xonly_key, leaf_hash) to value -> schnorr_sig

So, really need different types here :( because the function signatures vastly differ for key_spend and  script_spend.

> I'm not thrilled with the new expression parsing code; it seems like we could reduce the duplicated code.

I am also not happy with the current state of parsing. I will dedup it in a follow-up PR.

> Finally, I have some fixups at [https://github.com/apoelstra/rust-miniscript/tree/2022-02--278-fixups](https://github.com/apoelstra/rust-miniscript/tree/2022-02--278-fixups?rgh-link-date=2022-02-08T19%3A28%3A05Z) -- mostly minor stuff, but I also cleaned up the locking code.

Any special reason why you prefer .clone() on `Arc` instead of `Arc::clone`. All examples prefer in rust std docs the latter.

apoelstra:
> I don't follow this. Every `Tr` descriptor must have an internal key. There is no confusion with the external_key(or output_key) here.

I may be confused. What happens if you create a `Tr` with an internal key but `None` as the script tree?

> Overall, we have three main types of sigs:

Can you elaborate on the difference between the latter two cases?

> Any special reason why you prefer .clone() on Arc instead of Arc::clone. All examples prefer in rust std docs the latter.

No special reason. The docs on `Arc` say [Some people prefer to use fully qualified syntax, while others prefer using method-call syntax.](https://doc.rust-lang.org/std/sync/struct.Arc.html) and this is how I've always done it. I'm happy to change it if you prefer the fully-qualified way.

sanket1729:
> I may be confused. What happens if you create a Tr with an internal key but None as the script tree?

The internal key would return the internal key correctly. The script_pubkey()/spk() API calculation would calculate the output key  as per BIP341.

apoelstra:
@sanket1729 ah! Ok, I missed the BIP341 logic. Alright, I'm happy.

If you cherry-pick my commits I'll ACK this. (Or I can change the `Arc::clone`s if you'd like me to do that first)

apoelstra:
@sanket1729 there are four extra commits on my branch

sanket1729:
@apoelstra, oops. Missed those

sanket1729:
Awesome. Working on fixups based on this

PR #282, opened 2021-12-02
Context: check for the maximum number of public keys in a `CHECKMULTISIG`

darosior:
We check at parsing time, but we would still allow to create a `Multi` with `> 20` keys and not error on sanity checks! I'd add a test but wanted to push a patch quickly first.

Reported by @danielabrozzoni .


apoelstra (approved):
ACK f365e236a1d2bbe2164f2e40d8b203ce9fc07879

It could use a unit test I suppose, but I think it's fine.

Issue #287, opened 2021-12-25
Use a random name for integration test directory


sanket1729:
More context: https://github.com/rust-bitcoin/rust-miniscript/pull/162#discussion_r772476659

PR #288, opened 2021-12-26
use a single-use temp name for integration test dir

z8674558:
Closes https://github.com/rust-bitcoin/rust-miniscript/issues/287

### DIFF
--- a/integration_test/run.sh
+++ b/integration_test/run.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
-TESTDIR=/tmp/rust_miniscript_test
+RAND_DIR=$(head -c 4 </dev/urandom | xxd -p)
+TESTDIR=/tmp/"rust_miniscript_test_$RAND_DIR"
 
 rm -rf ${TESTDIR}
 mkdir -p ${TESTDIR}/1


apoelstra:
Can you use `xxd` rather than `base64` so we get hex-encoding? 10 digits is still more than enough.

As an aside, do we care about support for systems without `/dev/urandom` or anything like it? I guess not.

z8674558:
> As an aside, do we care about support for systems without /dev/urandom or anything like it? I guess not.

Maybe it is better to use timestamp instead of random string, because what we want to do is to avoid conflict on the shared machine. We should be ok with just `date`

z8674558:
updated to use `xxd`

sanket1729:
Bitcoin core names it as `/tmp/bitcoin_func_test_o16z4jkv`. Anyways, I am okay with `xxd` or `base64`

sanket1729 (approved):
utACK e3dafb25284520a190ab50bc689b4eb76e15354a

apoelstra:
utACK the current proposal (use `/dev/urandom` into `xxd`, add prefix)

apoelstra (approved):
ACK e3dafb25284520a190ab50bc689b4eb76e15354a

This looks great!

PR #289, opened 2022-01-11
Update to a dev version latest bitcoin master 0e2e55971275da64ceb62e8991a0a5fa962cb8b1

sanket1729:
Update to a version of dev rust-bitcoin so that we can prepare for rust-miniscript release. The integration tests are updated to point a personal fork of rust-bitcoincore-rpc to have rust semver type issues. 

After we have this, we can make progress on tapscript while we await finalized rust-bitcoin release. After which, we can remove the git dependencies and have crates.io dependencies instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ description = "Miniscript: a subset of Bitcoin Script designed for analysis"
 license = "CC0-1.0"
 
 [features]
-fuzztarget = ["bitcoin/fuzztarget"]
+fuzztarget = []
 compiler = []
 trace = []
 unstable = []
@@ -16,7 +16,8 @@ use-serde = ["bitcoin/use-serde", "serde"]
 rand = ["bitcoin/rand"]
 
 [dependencies]
-bitcoin = "0.27"
+# bitcoin = "0.27"
+bitcoin = {git = "https://github.com/rust-bitcoin/rust-bitcoin", rev = "0e2e55971275da64ceb62e8991a0a5fa962cb8b1"}
 
 [dependencies.serde]
 version = "1.0"
--- a/examples/verify_tx.rs
+++ b/examples/verify_tx.rs
@@ -83,11 +83,11 @@ fn main() {
         0xa9, 0x14, 0x92, 0x09, 0xa8, 0xf9, 0x0c, 0x58, 0x4b, 0xb5, 0x97, 0x4d, 0x58, 0x68, 0x72,
         0x49, 0xe5, 0x32, 0xde, 0x59, 0xf4, 0xbc, 0x87,
     ]);
-
+    let wit = transaction.input[0].witness.to_vec();
     let mut interpreter = miniscript::Interpreter::from_txdata(
         &spk_input_1,
         &transaction.input[0].script_sig,
-        &transaction.input[0].witness,
+        &wit,
         0,
         0,
     )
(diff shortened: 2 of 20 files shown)


examples/verify_tx.rs
```diff
@@ -83,11 +83,11 @@ fn main() {
         0xa9, 0x14, 0x92, 0x09, 0xa8, 0xf9, 0x0c, 0x58, 0x4b, 0xb5, 0x97, 0x4d, 0x58, 0x68, 0x72,
         0x49, 0xe5, 0x32, 0xde, 0x59, 0xf4, 0xbc, 0x87,
     ]);
-
+    let wit = transaction.input[0].witness.to_vec();
     let mut interpreter = miniscript::Interpreter::from_txdata(
```
RCasatta, line 87: What do you think to change the parameter type in `from_txdata` (and in `inner::from_txdata`) to `&Witness` avoiding the step of converting to vec? 

If I am looking correctly it's used only for iteration and we have `Witness::iter` which doesn't allocate

sanket1729: Yes, this is an already planned change in a follow-up PR. Wanted to keep things simple here.

sanket1729: Does it make sense for psbt FINAL_SCRIPT_WITNESS to be `Witness` type instead of `Vec<Vec<u8>>`

sanket1729: Added a follow up commit and discovered that rust-miniscript needs https://github.com/rust-bitcoin/rust-bitcoin/pull/774 for a cleaner non-allocating API.

sanket1729:
@apoelstra, this can help move things forward on rust-miniscript

Cargo.toml
```diff
@@ -7,7 +7,7 @@ description = "Miniscript: a subset of Bitcoin Script designed for analysis"
 license = "CC0-1.0"
 
 [features]
-fuzztarget = ["bitcoin/fuzztarget"]
+fuzztarget = []
```
apoelstra, line 10: In f3c38b8cc04fed0a68f4d6074d8c30f6912d958f:

We should drop `fuzztarget` entirely.

src/psbt/finalizer.rs
```diff
@@ -268,26 +270,17 @@ pub fn finalize_helper<C: secp256k1::Verification>(
...
-                super::Error::InputError(
-                    InputError::Interpreter(interpreter::Error::NonStandardSigHash(
-                        [sig, &[*flag]].concat().to_vec(),
-                    )),
-                    n,
-                )
-            })?;
+        let target = input.sighash_type.unwrap_or(bitcoin::EcdsaSigHashType::All);
+        for (key, ecdsa_sig) in &input.partial_sigs {
+            let flag = bitcoin::EcdsaSigHashType::from_u32_standard(ecdsa_sig.hash_ty as u32)
```
apoelstra, line 277: In f3c38b8cc04fed0a68f4d6074d8c30f6912d958f:

lol, is this conversion just to check standardness? We should add a method to rust-bitcoin so we can call `ecdsa_sig.hash_tx.is_standard()'.

apoelstra (approved):
ACK aa36494f5ac2f2dcd5acbb9ae8ae933a3492fd5d

src/psbt/finalizer.rs
sanket1729: Actually, this line is not needed. ecdsa_sig.hash_ty is guaranteed to follow standardness rules.

sanket1729:
Resolving the comments in follow up PR

PR #290, opened 2022-01-12
Remove fuzztaget and cleanup psbt finalizer

sanket1729:


src/psbt/finalizer.rs
```diff
@@ -274,20 +274,12 @@ pub fn finalize_helper<C: secp256k1::Verification>(
...
-                    super::Error::InputError(
-                        InputError::Interpreter(interpreter::Error::NonStandardSigHash(
-                            ecdsa_sig.to_vec(),
-                        )),
-                        n,
-                    )
-                })?;
-            if target != flag {
+            if target != ecdsa_sig.hash_ty {
+                // hash_ty is guaranteed to follow standardness rules
```
apoelstra, line 278: In 7f4adf5e96a65adf78e8606be7ac86f3d272dad4:

I think this comment can be dropped -- standardness isn't actually relevant to this check, which is just that the signature hashtype matches the one required by the PSBT.

Also, I don't think that standardness is actually guaranteed here. See this code: https://github.com/rust-bitcoin/rust-bitcoin/pull/669#discussion_r780689066

sanket1729: I think that is probably a bug upstream because we will risk generating wrong signatures. Anything that is consensus valid sighashtype byte (0x01, 0x02, x03, 0x81, 0x82, and 0x83) is mapped to sighash all. Consider a case with 0x05 sighash byte, while signing, we sighash_u32 as `0x05u32`, but while remapping it, we check it as `0x01` because the information that sighash_all from `0x05` instead of `0x01` is lost.

apoelstra: oo good catch!

I think the bug is actually that we're losing information, rather than that we allow non-standard sighash types in PSBT. I can see how this happened, since this recent PSBT thing is the first time that we've ever deliberately supported non-standard sighash types..

apoelstra: I opened https://github.com/rust-bitcoin/rust-bitcoin/issues/777 about this.

For now I still think we should remove this comment, because standardness is irrelevant, although you could add a comment linking to that rust-bitcoin issue.

apoelstra (approved):
ACK bb9f2a0d4d02cff3c116d4badb629cd68b9dbc17

sanket1729:
The sighash issue was taken upstream to rust-bitcoin and resolved cleanly there. The current code only parses `PsbtSigHashType` using the new method `ecdsa_sighash_ty` correctly and was merged in #301 while updating to bitcoin 0.28-rc1. The scope of this PR is now only removing the `fuzztarget`

apoelstra (approved):
ACK a142ab40e40fd6f302bb59b04dc3441c2a724465

JeremyRubin:
ack

PR #291, opened 2022-01-16
Taproot Compiler

SarcasticNastik:
This PR builds on top of #278. 

This is one in a series of PRs aimed to implement a feature-complete *Pay-to-Taproot* **P2Tr** compiler.
Specifically, this introduces a basic compilation for a given policy by collecting the leaf nodes of the *tree* generated by considering root-level disjunctive `OR`and using this to generate a `TapTree`. 

> Assuming that _duplicate keys_ are **NOT** possible even in different branches of the TapTree.

# Follow Up PRs
 
- #342  - Uses heuristic for tree-generation/ _merging_ algorithm while buillding `TapTree` to optimize over the *expected average total cost*. 
- #352 
- A future PR implementing enumerative strategies for optimizing `thresh` `k`-of-`n` and similar structures.

### DIFF
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -36,7 +36,7 @@ type PolicyCache<Pk, Ctx> =
 
 ///Ordered f64 for comparison
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
-struct OrdF64(f64);
+pub(crate) struct OrdF64(pub f64);
 
 impl Eq for OrdF64 {}
 impl Ord for OrdF64 {
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -24,6 +24,7 @@ use std::marker::PhantomData;
 use std::sync::Arc;
 use std::{cmp, error, f64, fmt, hash, mem};
 
+use crate::miniscript::context::SigType;
 use crate::miniscript::limits::MAX_PUBKEYS_PER_MULTISIG;
 use crate::miniscript::types::{self, ErrorKind, ExtData, Property, Type};
 use crate::miniscript::ScriptContext;
@@ -987,18 +988,23 @@ where
                 })
                 .collect();
 
-            if key_vec.len() == subs.len() && subs.len() <= MAX_PUBKEYS_PER_MULTISIG {
-                insert_wrap!(AstElemExt::terminal(Terminal::Multi(k, key_vec)));
-            }
-            // Not a threshold, it's always more optimal to translate it to and()s as we save the
-            // resulting threshold check (N EQUAL) in any case.
-            else if k == subs.len() {
-                let mut policy = subs.first().expect("No sub policy in thresh() ?").clone();
-                for sub in &subs[1..] {
-                    policy = Concrete::And(vec![sub.clone(), policy]);
+            match Ctx::sig_type() {
+                SigType::Schnorr if key_vec.len() == subs.len() => {
+                    insert_wrap!(AstElemExt::terminal(Terminal::MultiA(k, key_vec)))
                 }
+                SigType::Ecdsa
+                    if key_vec.len() == subs.len() && subs.len() <= MAX_PUBKEYS_PER_MULTISIG =>
+                {
+                    insert_wrap!(AstElemExt::terminal(Terminal::Multi(k, key_vec)))
+                }
+                _ if k == subs.len() => {
+                    let mut it = subs.iter();
+                    let mut policy = it.next().expect("No sub policy in thresh() ?").clone();
+                    policy = it.fold(policy, |acc, pol| Concrete::And(vec![acc, pol.clone()]));
 
-                ret = best_compilations(policy_cache, &policy, sat_prob, dissat_prob)?;
+                    ret = best_compilations(policy_cache, &policy, sat_prob, dissat_prob)?;
+                }
+                _ => {}
             }
 
             // FIXME: Should we also optimize thresh(1, subs) ?
@@ -1548,6 +1554,17 @@ mod tests {
                 policy::concrete::PolicyError::DuplicatePubKeys
             ))
         );
+    }
+
+    #[test]
+    fn compile_tr_thresh() {
+        for k in 1..4 {
+            let small_thresh: Concrete<String> =
+                policy_str!("{}", &format!("thresh({},pk(B),pk(C),pk(D))", k));
+            let small_thresh_ms: Miniscript<String, Tap> = small_thresh.compile().unwrap();
+            let small_thresh_ms_expected: Miniscript<String, Tap> = ms_str!("multi_a({},B,C,D)", k);
+            assert_eq!(small_thresh_ms, small_thresh_ms_expected);
+        }
     }
 }
 
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -20,20 +20,19 @@ use std::{error, fmt, str};
 
 use bitcoin::hashes::hex::FromHex;
 use bitcoin::hashes::{hash160, ripemd160, sha256, sha256d};
+#[cfg(feature = "compiler")]
+use {
+    crate::descriptor::TapTree, crate::miniscript::ScriptContext, crate::policy::compiler,
+    crate::policy::compiler::CompilerError, crate::Descriptor, crate::Miniscript, crate::Tap,
+    std::sync::Arc,
+};
 
 use super::ENTAILMENT_MAX_TERMINALS;
 use crate::expression::{self, FromTree};
 use crate::miniscript::limits::{HEIGHT_TIME_THRESHOLD, SEQUENCE_LOCKTIME_TYPE_FLAG};
 use crate::miniscript::types::extra_props::TimeLockInfo;
-#[cfg(feature = "compiler")]
-use crate::miniscript::ScriptContext;
-#[cfg(feature = "compiler")]
-use crate::policy::compiler;
-#[cfg(feature = "compiler")]
-use crate::policy::compiler::CompilerError;
-#[cfg(feature = "compiler")]
-use crate::Miniscript;
 use crate::{errstr, Error, ForEach, ForEachKey, MiniscriptKey};
+
 /// Concrete policy which corresponds directly to a Miniscript structure,
 /// and whose disjunctions are annotated with satisfaction probabilities
 /// to assist the compiler
@@ -128,6 +127,21 @@ impl fmt::Display for PolicyError {
 }
 
 impl<Pk: MiniscriptKey> Policy<Pk> {
+    /// Single-Node compilation
+    #[cfg(feature = "compiler")]
+    fn compile_leaf_taptree(&self) -> Result<TapTree<Pk>, Error> {
+        let compilation = self.compile::<Tap>().unwrap();
+        Ok(TapTree::Leaf(Arc::new(compilation)))
+    }
+
+    /// Compile a [`Policy`] into a single-leaf [`TapTree`]
+    #[cfg(feature = "compiler")]
+    pub fn compile_tr(&self, unspendable_key: Option<Pk>) -> Result<Descriptor<Pk>, Error> {
+        let internal_key = unspendable_key.ok_or(errstr("No internal key found"))?;
+        let tree = Descriptor::new_tr(internal_key, Some(self.compile_leaf_taptree()?))?;
+        Ok(tree)
+    }
+
     /// Compile the descriptor into an optimized `Miniscript` representation
     #[cfg(feature = "compiler")]
     pub fn compile<Ctx: ScriptContext>(&self) -> Result<Miniscript<Pk, Ctx>, CompilerError> {
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -25,11 +25,13 @@ use {
     crate::descriptor::TapTree,
     crate::miniscript::ScriptContext,
     crate::policy::compiler::CompilerError,
+    crate::policy::compiler::OrdF64,
     crate::policy::{compiler, Concrete, Liftable, Semantic},
     crate::Descriptor,
     crate::Miniscript,
     crate::Tap,
-    std::collections::HashMap,
+    std::cmp::Reverse,
+    std::collections::{BinaryHeap, HashMap},
     std::sync::Arc,
 };
 
@@ -173,14 +175,20 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
         }
     }
 
-    /// Single-Node compilation
+    /// Compile [`Policy::Or`] and [`Policy::Threshold`] according to odds
     #[cfg(feature = "compiler")]
-    fn compile_leaf_taptree(&self) -> Result<TapTree<Pk>, Error> {
-        let compilation = self.compile::<Tap>().unwrap();
-        Ok(TapTree::Leaf(Arc::new(compilation)))
+    fn compile_tr_policy(&self) -> Result<TapTree<Pk>, Error> {
+        let leaf_compilations: Vec<_> = self
+            .to_tapleaf_prob_vec(1.0)
+            .into_iter()
+            .filter(|x| x.1 != Policy::Unsatisfiable)
+            .map(|(prob, ref policy)| (OrdF64(prob), compiler::best_compilation(policy).unwrap()))
+            .collect();
+        let taptree = with_huffman_tree::<Pk>(leaf_compilations).unwrap();
+        Ok(taptree)
     }
 
-    /// Extract the Taproot internal_key from policy tree.
+    /// Extract the internal_key from policy tree.
     #[cfg(feature = "compiler")]
     fn extract_key(self, unspendable_key: Option<Pk>) -> Result<(Pk, Policy<Pk>), Error> {
         let mut internal_key: Option<Pk> = None;
@@ -223,11 +231,28 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
         }
     }
 
-    /// Compile the [`Tr`] descriptor into optimized [`TapTree`] implementation
+    /// Compile the [`Policy`] into a [`Tr`][`Descriptor::Tr`] Descriptor.
+    ///
+    /// ### TapTree compilation
+    ///
+    /// The policy tree constructed by root-level disjunctions over [`Or`][`Policy::Or`] and
+    /// [`Thresh`][`Policy::Threshold`](1, ..) which is flattened into a vector (with respective
+    /// probabilities derived from odds) of policies.
+    /// For example, the policy `thresh(1,or(pk(A),pk(B)),and(or(pk(C),pk(D)),pk(E)))` gives the vector
+    /// `[pk(A),pk(B),and(or(pk(C),pk(D)),pk(E)))]`. Each policy in the vector is compiled into
+    /// the respective miniscripts. A Huffman Tree is created from this vector which optimizes over
+    /// the probabilitity of satisfaction for the respective branch in the TapTree.
+    // TODO: We might require other compile errors for Taproot.
     #[cfg(feature = "compiler")]
     pub fn compile_tr(&self, unspendable_key: Option<Pk>) -> Result<Descriptor<Pk>, Error> {
         let (internal_key, policy) = self.clone().extract_key(unspendable_key)?;
-        let tree = Descriptor::new_tr(internal_key, Some(policy.compile_leaf_taptree()?))?;
+        let tree = Descriptor::new_tr(
+            internal_key,
+            match policy {
+                Policy::Trivial => None,
+                policy => Some(policy.compile_tr_policy()?),
+            },
+        )?;
         Ok(tree)
     }
 
@@ -771,4 +796,35 @@ where
     fn from_tree(top: &expression::Tree) -> Result<Policy<Pk>, Error> {
         Policy::from_tree_prob(top, false).map(|(_, result)| result)
     }
+}
+
+/// Create a Huffman Tree from compiled [Miniscript] nodes
+#[cfg(feature = "compiler")]
+fn with_huffman_tree<Pk: MiniscriptKey>(
+    ms: Vec<(OrdF64, Miniscript<Pk, Tap>)>,
+) -> Result<TapTree<Pk>, Error> {
+    let mut node_weights = BinaryHeap::<(Reverse<OrdF64>, TapTree<Pk>)>::new();
+    for (prob, script) in ms {
+        node_weights.push((Reverse(prob), TapTree::Leaf(Arc::new(script))));
+    }
+    if node_weights.is_empty() {
+        return Err(errstr("Empty Miniscript compilation"));
+    }
+    while node_weights.len() > 1 {
+        let (p1, s1) = node_weights.pop().expect("len must atleast be two");
+        let (p2, s2) = node_weights.pop().expect("len must atleast be two");
+
+        let p = (p1.0).0 + (p2.0).0;
+        node_weights.push((
+            Reverse(OrdF64(p)),
+            TapTree::Tree(Arc::from(s1), Arc::from(s2)),
+        ));
+    }
+
+    debug_assert!(node_weights.len() == 1);
+    let node = node_weights
+        .pop()
+        .expect("huffman tree algorithm is broken")
+        .1;
+    Ok(node)
 }
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -22,8 +22,14 @@ use bitcoin::hashes::hex::FromHex;
 use bitcoin::hashes::{hash160, ripemd160, sha256, sha256d};
 #[cfg(feature = "compiler")]
 use {
-    crate::descriptor::TapTree, crate::miniscript::ScriptContext, crate::policy::compiler,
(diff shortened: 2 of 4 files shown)


src/policy/concrete.rs
```diff
@@ -137,30 +143,109 @@ impl fmt::Display for PolicyError {
...
-            Some(key) => Ok((key, policy)),
-            None => Err(errstr("No internal key found")),
+    fn extract_key(policy: &Self, unspendable_key: Option<Pk>) -> Result<(Pk, Policy<Pk>), Error> {
+        let semantic_policy = policy.lift()?;
+        let concrete_keys = policy.keys().into_iter().collect::<HashSet<_>>();
+        let mut internal_key: Option<Pk> = None;
+
+        for key in concrete_keys.iter() {
+            if semantic_policy
+                .clone()
```
sanket1729, line 226: This algorithm is poor in memory performance because it calls clone for every single iteration. We can do a new tree traversal for every key we see. I think we will eventually need a better algorithm that avoids all these allocations. 

We can first focus on the correctness/code hygine of the PR, then address these efficiency issues.

src/policy/concrete.rs
```diff
@@ -137,30 +143,109 @@ impl fmt::Display for PolicyError {
...
     #[cfg(feature = "compiler")]
     pub fn compile_tr(&self, unspendable_key: Option<Pk>) -> Result<Descriptor<Pk>, Error> {
         let (internal_key, policy) = Self::extract_key(self, unspendable_key).unwrap();
-        let tree = Descriptor::new_tr(
-            internal_key,
-            Some(Self::compile_huffman_taptree(policy).unwrap()),
-        )
-        .unwrap();
+        let tree =
+            Descriptor::new_tr(internal_key, Some(policy.compile_tr_policy().unwrap())).unwrap();
```
sanket1729, line 248: This should either propagate the error or explain why the unwrap is safe

src/policy/concrete.rs
```diff
@@ -264,6 +349,30 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
         }
     }
 
+    /// Translate `Semantic::Key(key)` to `Semantic::Unsatisfiable` when extracting TapKey
+    pub fn translate_unsatisfiable_pk(&self, key: &Pk) -> Policy<Pk> {
```
sanket1729, line 353: This is can consume Self.

sanket1729 (commented):
Did an initial review. I think we would need a few more iterations before we get this.

src/policy/semantic.rs
```diff
@@ -679,10 +679,7 @@ mod tests {
         let policy = StringPolicy::from_str("or(pkh(),older(1000))").unwrap();
         assert_eq!(
             policy,
-            Policy::Threshold(
-                1,
-                vec![Policy::KeyHash("".to_owned()), Policy::Older(1000),]
-            )
+            Policy::Threshold(1, vec![Policy::KeyHash("".to_owned()), Policy::Older(1000)])
```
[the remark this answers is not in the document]
SarcasticNastik: Yes. This code snippet was automatically formatted by `cargo fmt` (version: $1.58.0$).

image

It was necessary to keep this change to pass the linter.

src/policy/concrete.rs
```diff
@@ -128,6 +137,33 @@ impl fmt::Display for PolicyError {
 }
 
 impl<Pk: MiniscriptKey> Policy<Pk> {
+    /// Single-Node compilation
+    #[cfg(feature = "compiler")]
+    fn compile_huffman_taptree(policy: &Self) -> Result<TapTree<Pk>, Error> {
```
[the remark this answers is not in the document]
SarcasticNastik: Right. The naming convention for the function was kept keeping future in mind. I'll rename it to `compile_leaf_taptree` for proper naming convention for this commit.

src/policy/concrete.rs
```diff
@@ -137,30 +143,109 @@ impl fmt::Display for PolicyError {
...
+            let p = (p1.0).0 + (p2.0).0;
+            node_weights.push((Reverse(OrdF64(p)), Arc::from(TapTree::Tree(s1, s2))));
+        }
+
+        debug_assert!(node_weights.len() == 1);
+        let node = node_weights
+            .pop()
+            .expect("huffman tree algorithm is broken")
+            .1;
+        Ok((*node).clone())
```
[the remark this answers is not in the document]
SarcasticNastik: `node` is of type `Arc<_>` while our return type is `TapTree<Pk>`.

image

As I understand, we can't move the value out of `Arc` and the only solution is to call `clone` on the reference for node, either directly in the function or at a later stage (since we require owned value of the node as `Option<TapTree<Pk>>` for `Descriptor::new_tr()` rather than a reference to it.)

image

src/policy/concrete.rs
```diff
@@ -128,6 +137,33 @@ impl fmt::Display for PolicyError {
...
 impl<Pk: MiniscriptKey> Policy<Pk> {
+    /// Single-Node compilation
+    #[cfg(feature = "compiler")]
+    fn compile_huffman_taptree(policy: &Self) -> Result<TapTree<Pk>, Error> {
+        let compilation = policy.compile::<Tap>().unwrap();
+        Ok(TapTree::Leaf(Arc::new(compilation)))
+    }
+    /// Extract the Taproot internal_key from policy tree.
+    #[cfg(feature = "compiler")]
+    fn extract_key(policy: &Self, unspendable_key: Option<Pk>) -> Result<(Pk, &Policy<Pk>), Error> {
```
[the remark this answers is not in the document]
SarcasticNastik: This would require us to make the `compile_tr` function consume `self` which doesn't align with the `compile<Pk, Ctx>(&self)` function. I will update this function as suggested and use a cloned copy for the same until we discuss further.

src/policy/concrete.rs
SarcasticNastik: Done in [e50766](https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/e507066a5288c09c098d5eac9980066555e287ce).

src/policy/concrete.rs
SarcasticNastik: Refactored code to propagate errors in [b36c116](https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/b36c116babe70817942f2bcd95b97a6d00289ff7)

src/policy/concrete.rs
```diff
@@ -137,30 +143,109 @@ impl fmt::Display for PolicyError {
...
+            .expect("huffman tree algorithm is broken")
+            .1;
+        Ok((*node).clone())
+    }
+
+    /// Flatten the [`Policy`] tree structure into a Vector with corresponding leaf probability
+    // TODO: 1. Can try to push the maximum of scaling factors and accordingly update later for
+    // TODO: 1. integer metric. (Accordingly change metrics everywhere)
+    #[cfg(feature = "compiler")]
+    fn flatten_policy(&self, prob: f64) -> Vec<(f64, Policy<Pk>)> {
```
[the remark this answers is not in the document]
SarcasticNastik: Refactored in [e507066](https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/e507066a5288c09c098d5eac9980066555e287ce)

src/policy/concrete.rs
```diff
@@ -137,30 +143,109 @@ impl fmt::Display for PolicyError {
...
 impl<Pk: MiniscriptKey> Policy<Pk> {
-    /// Single-Node compilation
+    /// Create a Huffman Tree from compiled [Miniscript] nodes
+    #[cfg(feature = "compiler")]
+    fn with_huffman_tree<T>(
+        ms: Vec<(OrdF64, Miniscript<Pk, Tap>)>,
+        f: T,
+    ) -> Result<TapTree<Pk>, Error>
+    where
+        T: Fn(OrdF64) -> OrdF64,
```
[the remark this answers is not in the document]
SarcasticNastik: The function `f` refers to the metric we want to optimize the huffman tree over (for example: *leaf probabilities*, *script size* etc. and any combination of those).
For now this function (/signature) is just a dummy (with just *leaf probabilities*) until we definitively decide over the metrics.

src/policy/concrete.rs
```diff
@@ -32,9 +32,15 @@ use miniscript::types::extra_props::TimeLockInfo;
...
+use policy::compiler::{CompilerError, OrdF64};
+#[cfg(feature = "compiler")]
+use policy::{compiler, Semantic};
+#[cfg(feature = "compiler")]
+use std::cmp::Reverse;
 #[cfg(feature = "compiler")]
-use policy::compiler::CompilerError;
+use std::collections::BinaryHeap;
+#[cfg(feature = "compiler")]
+use std::sync::Arc;
```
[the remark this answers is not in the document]
SarcasticNastik: Refactored in [b36c11](https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/b36c116babe70817942f2bcd95b97a6d00289ff7)

src/policy/compiler.rs
```diff
@@ -38,7 +38,7 @@ type PolicyCache<Pk, Ctx> =
 
 ///Ordered f64 for comparison
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
-struct OrdF64(f64);
+pub struct OrdF64(pub f64);
```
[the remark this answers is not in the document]
SarcasticNastik: Updated in [e507066](https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/e507066a5288c09c098d5eac9980066555e287ce)

src/policy/concrete.rs
```diff
@@ -32,8 +36,13 @@ use policy::compiler;
 #[cfg(feature = "compiler")]
 use policy::compiler::CompilerError;
 #[cfg(feature = "compiler")]
+use Descriptor;
+#[cfg(feature = "compiler")]
 use Miniscript;
+#[cfg(feature = "compiler")]
+use Tap;
```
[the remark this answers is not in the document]
SarcasticNastik: Refactored in https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/b36c116babe70817942f2bcd95b97a6d00289ff7

src/policy/compiler.rs
```diff
@@ -994,7 +995,9 @@ where
                 })
                 .collect();
 
-            if key_vec.len() == subs.len() && subs.len() <= MAX_PUBKEYS_PER_MULTISIG {
+            if Ctx::sig_type() == SigType::Schnorr {
```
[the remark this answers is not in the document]
SarcasticNastik: Refactored in https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/b36c116babe70817942f2bcd95b97a6d00289ff7

src/policy/mod.rs
```diff
@@ -233,8 +233,14 @@ mod tests {
...
     };
     use bitcoin;
+    #[cfg(feature = "compiler")]
+    use descriptor::TapTree;
     use std::str::FromStr;
+    #[cfg(feature = "compiler")]
+    use std::sync::Arc;
     use DummyKey;
+    #[cfg(feature = "compiler")]
+    use {Descriptor, Tap};
```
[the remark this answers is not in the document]
SarcasticNastik: Refactored in https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/b36c116babe70817942f2bcd95b97a6d00289ff7

src/policy/concrete.rs
```diff
@@ -128,6 +137,33 @@ impl fmt::Display for PolicyError {
...
+        }
+    }
+
+    /// Compile the [`Tr`] descriptor into optimized [`TapTree`] implementation
+    #[cfg(feature = "compiler")]
+    pub fn compile_tr(&self, unspendable_key: Option<Pk>) -> Result<Descriptor<Pk>, Error> {
+        let (internal_key, policy) = Self::extract_key(self, unspendable_key).unwrap();
+        let tree = Descriptor::new_tr(
+            internal_key,
+            Some(Self::compile_huffman_taptree(policy).unwrap()),
```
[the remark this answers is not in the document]
SarcasticNastik: Updated code as suggested in https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/e507066a5288c09c098d5eac9980066555e287ce

src/policy/compiler.rs
```diff
@@ -994,18 +995,25 @@ where
...
-            // resulting threshold check (N EQUAL) in any case.
-            else if k == subs.len() {
-                let mut policy = subs.first().expect("No sub policy in thresh() ?").clone();
-                for sub in &subs[1..] {
-                    policy = Concrete::And(vec![sub.clone(), policy]);
+            match Ctx::sig_type() {
+                SigType::Schnorr => {
+                    insert_wrap!(AstElemExt::terminal(Terminal::MultiA(k, key_vec)))
+                }
+                _ => {
```
sanket1729, line 1002: match the `SigType::Ecdsa` to ensure that we don't run any other unnecessary code with SigType is added with another enum variant.

src/policy/compiler.rs
```diff
@@ -994,18 +995,25 @@ where
...
+                    if key_vec.len() == subs.len() && subs.len() <= MAX_PUBKEYS_PER_MULTISIG {
+                        insert_wrap!(AstElemExt::terminal(Terminal::Multi(k, key_vec)));
+                    }
+                    // Not a threshold, it's always more optimal to translate it to and()s as we save the
+                    // resulting threshold check (N EQUAL) in any case.
+                    else if k == subs.len() {
```
sanket1729, line 1008: We could also add this optimization to multi_a. If `k == n`, compiling as `ands() ` is more efficient.

src/policy/concrete.rs
```diff
@@ -126,27 +132,115 @@ impl fmt::Display for PolicyError {
...
+        }
+        if node_weights.is_empty() {
+            return Err(errstr("Empty Miniscript compilation"));
+        }
+        while node_weights.len() > 1 {
+            let (p1, s1) = node_weights.pop().expect("len must atleast be two");
+            let (p2, s2) = node_weights.pop().expect("len must atleast be two");
+
+            let p = (p1.0).0 + (p2.0).0;
+            node_weights.push((Reverse(OrdF64(p)), Arc::from(TapTree::Tree(s1, s2))));
```
sanket1729, line 157: This should not create an `Arc`. The vec should contain all the owned TapTree. You can then do the following:

```diff
- node_weights.push((Reverse(OrdF64(p)), Arc::from(TapTree::Tree(s1, s2))));
+ node_weights.push((Reverse(OrdF64(p)), TapTree::Tree(Arc::new(s1), Arc::new(s2))));

File: src/policy/concrete.rs (hunk 6 above)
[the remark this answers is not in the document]
Comment (sanket1729) [reply src/policy/concrete.rs:165]: suggested how to fix it above.

File: src/policy/semantic.rs (hunk 4 above)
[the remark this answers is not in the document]
Comment (sanket1729) [reply src/policy/semantic.rs:682]: I see this is because you removed the "," at the end of all lines. This change is okay, but it is not a part of this commit. But I can live with this

**Review by sanket1729 — COMMENTED** [review]
Very close. Last few comments.

File: src/policy/concrete.rs (hunk 18 above)
Comment (SarcasticNastik) [reply src/policy/concrete.rs]: Updated in https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/f7c502629c7a9972c762590b5027ff74513f6c5a

Comment (SarcasticNastik) [reply src/policy/concrete.rs]: Update in https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/1a1d19262b859c506d8a60baa3a5ac3643b0fe3f

File: src/policy/compiler.rs (hunk 17 above)
Comment (SarcasticNastik) [reply src/policy/compiler.rs:1017]: Updated in https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/1a1d19262b859c506d8a60baa3a5ac3643b0fe3f

File: src/policy/compiler.rs (hunk 16 above)
Comment (SarcasticNastik) [reply src/policy/compiler.rs]: Updated in https://github.com/rust-bitcoin/rust-miniscript/pull/291/commits/1a1d19262b859c506d8a60baa3a5ac3643b0fe3f

**Comment by SarcasticNastik** [comment]
The Lint+Fuzz test fails here because of [this](https://github.com/rust-fuzz/honggfuzz-rs/issues/61) issue. Locally running the scripts work just fine.

File: src/policy/compiler.rs [hunk 19]
```diff
@@ -994,18 +995,24 @@ where
...
+                SigType::Schnorr if key_vec.len() == subs.len() => {
+                    insert_wrap!(AstElemExt::terminal(Terminal::MultiA(k, key_vec)))
+                }
+                SigType::Ecdsa
+                    if key_vec.len() == subs.len() && subs.len() <= MAX_PUBKEYS_PER_MULTISIG =>
+                {
+                    insert_wrap!(AstElemExt::terminal(Terminal::Multi(k, key_vec)))
                 }
+                _ if k == subs.len() => {
+                    let mut policy = subs.first().expect("No sub policy in thresh() ?").clone();
```
JeremyRubin, line 1008: you can use "reduce" here, if the MSRV is 1.51 or higher.

otherwise something like
```
let mut it = policy.iter();
let first = self.next()?;
Some(self.fold(first, |(x,y)| Concrete::And(vec![x,y])))
```

src/policy/semantic.rs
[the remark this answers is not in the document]
JeremyRubin: i would recommend to bump 1.58.0 and then cargo fmt as one commit maybe?

JeremyRubin:
shouldn't this try to compile the common: And(Or(a,b), Or(c,d)) to: ac ad bc bd?

how do we judge if the increased number of nodes vs script length cutoff makes sense?

It seems like there is a script length beyond which it always makes sense, maybe like 32 byte script?

src/policy/compiler.rs
SarcasticNastik: Right! This does seem more idiomatic.

SarcasticNastik:
@JeremyRubin You're right in the intuitive analysis of the method taproot compiler should ideally adopt. I plan on introducing this type to reduction/ enumeration of policy nodes in a follow-up PR. The reason to keep that on hold was to understand if doing so would turn out to be intractable. For example:- Consider a generic `thresh(k, ...)` k-of-n keys scenario. Enumerating this as separate nodes in disjunction results in nCk (exponential in k). 
In addition, this PR just serves to provide one of the many compilation methods we plan to try/test upon. Another version of this compiler is underway which uses a heuristic to re-arrange the policy nodes and only compile the nodes (`A` and `B`) separately if resulting expected total cost (script_size + average satisfaction cost) of compilation into a taptree is lesser than compilation of a combination of the two (`Or(A, B)`).

dr-orlovsky:
@sanket1729 is there plans of getting this into the release?

sanket1729:
@dr-orlovsky, no this won't go into the release. There is still some work to be done here, and the compiler implemented here is feature complete.  We still have a few things we want before we support this. 

Maybe we will have another release with support for partial descriptors, unspendable key support, musig descriptors etc... Possibly 1 month from now

src/policy/concrete.rs
```diff
@@ -126,27 +133,137 @@ impl fmt::Display for PolicyError {
...
+            .to_tapleaf_prob_vec(1.0)
+            .into_iter()
+            .filter(|x| x.1 != Policy::Unsatisfiable)
+            .map(|(prob, ref policy)| (OrdF64(prob), compiler::best_compilation(policy).unwrap()))
+            .collect();
+        let taptree = Self::with_huffman_tree(leaf_compilations, |x| x).unwrap();
+        Ok(taptree)
+    }
+
+    /// Extract the internal_key from policy tree.
```
SarcasticNastik, line 211: The core logic for *internal-key extraction* has been changed from selecting first encountered key (satisfying the policy) to the **most-probable key** (satisfying the policy).

src/policy/concrete.rs
```diff
@@ -126,27 +133,137 @@ impl fmt::Display for PolicyError {
...
+    #[cfg(feature = "compiler")]
+    fn extract_key(self, unspendable_key: Option<Pk>) -> Result<(Pk, Policy<Pk>), Error> {
+        // Making sure the borrow ends before you move the value.
+        let mut internal_key: Option<Pk> = None;
+        {
+            let mut prob = 0.;
+            let semantic_policy = self.lift()?;
+            let concrete_keys = self.keys();
+            let key_prob_map: HashMap<_, _> = self
```
SarcasticNastik, line 220: The logic of *internal-key extraction* has been changed from selecting **first-encountered key** satisfying the policy to **most-probable key** satisfying the policy.

sanket1729:
Can you rebase on top of master.? And remove the last CI commit.

src/policy/concrete.rs
```diff
@@ -126,27 +133,137 @@ impl fmt::Display for PolicyError {
 }
 
 impl<Pk: MiniscriptKey> Policy<Pk> {
-    /// Single-Node compilation
+    /// Create a Huffman Tree from compiled [Miniscript] nodes
     #[cfg(feature = "compiler")]
-    fn compile_leaf_taptree(&self) -> Result<TapTree<Pk>, Error> {
-        let compilation = self.compile::<Tap>().unwrap();
-        Ok(TapTree::Leaf(Arc::new(compilation)))
+    fn with_huffman_tree<T>(
```
sanket1729, line 138: why is this function in `impl Policy`?

src/policy/concrete.rs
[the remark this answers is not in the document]
sanket1729: Let's not have this confusing argument for now. And hard-code whatever dummy metric we have

src/policy/concrete.rs
```diff
@@ -126,27 +133,137 @@ impl fmt::Display for PolicyError {
...
+        }
+        match (internal_key, unspendable_key) {
+            (Some(ref key), _) => Ok((key.clone(), self.translate_unsatisfiable_pk(&key))),
+            (_, Some(key)) => Ok((key, self)),
+            _ => Err(errstr("No viable internal key found.")),
         }
     }
 
     /// Compile the [`Tr`] descriptor into optimized [`TapTree`] implementation
+    // TODO: We might require other compile errors for Taproot. Will discuss and update.
```
sanket1729, line 256: Elaborate on the word optimized here.

src/policy/concrete.rs
SarcasticNastik: Right, I forgot to move the function out of the `impl` block. Will update this.

src/policy/concrete.rs
SarcasticNastik: Ig the comment here isn't apt here. Will change it altogether!

SarcasticNastik:
Changes in the force update:

1. Updated documentation for some functions.
2. Moved `with_huffman_tree` out of the `impl` block.
3. Seperate `internal_key_extraction` and taptree compilation into different commits.

sanket1729 (commented):
Looks mostly good

src/policy/concrete.rs
```diff
@@ -127,7 +133,30 @@ impl fmt::Display for PolicyError {
 }
 
 impl<Pk: MiniscriptKey> Policy<Pk> {
-    /// Single-Node compilation
+    /// Flatten the [`Policy`] tree structure into a Vector with corresponding leaf probability
```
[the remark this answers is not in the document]
SarcasticNastik: Done in 3e9214c. Kindly review it and let me know.

src/policy/concrete.rs
```diff
@@ -128,6 +127,30 @@ impl fmt::Display for PolicyError {
...
+    /// Single-Node compilation
+    #[cfg(feature = "compiler")]
+    fn compile_leaf_taptree(&self) -> Result<TapTree<Pk>, Error> {
+        let compilation = self.compile::<Tap>().unwrap();
+        Ok(TapTree::Leaf(Arc::new(compilation)))
+    }
+
+    /// Extract the Taproot internal_key from policy tree.
+    #[cfg(feature = "compiler")]
+    fn extract_key(&self, unspendable_key: Option<Pk>) -> Result<(Pk, &Policy<Pk>), Error> {
```
sanket1729: In the first commit 6fb1b29a46ad10231f50f490300ee22a0a66fc07: Is extract key not completely implemented?

src/policy/concrete.rs
sanket1729: I see the ertract key is updated in next commit. No need to fix it here.

sanket1729:
Sorry, @SarcasticNastik this would require a rebase again :( . 

@apoelstra requesting your review. This is a major contribution from a new contributor and I don't want to go ahead without your review. The commit messages and PR description are helpful in understanding how this is structured.

SarcasticNastik:
@sanket1729 I've rebased the PR onto current master. @apoelstra A kind request for your review.

src/policy/concrete.rs
SarcasticNastik: Right, this might seem confusing. Will remove it altogether in the first commit.

SarcasticNastik: Resolved in 2b136940

sanket1729 (approved):
ACK 2b1369409adb49c838253572a59d05b00060df23 . Let's push forward on this. In the interest of ACKs and multiple rebases already done here, I am merging this. We recently shifted to rust 2018. And this code has a bunch of warnings because of it.

PR #293, opened 2022-02-07
miniscript: allow up to 2^32 timelocks for after()

darosior:
See https://github.com/sipa/miniscript/pull/97 and the discussion at https://gnusha.org/miniscript/2022-02-06.log.

I still need to:
- [ ] Check the size computation with 5-bytes pushes
- [ ] Add sanity checks for the maximum values inside `after()` and `older()`

### DIFF
--- a/src/miniscript/types/extra_props.rs
+++ b/src/miniscript/types/extra_props.rs
@@ -1033,10 +1033,7 @@ impl Property for ExtData {
                 }
             }
             Terminal::After(t) => {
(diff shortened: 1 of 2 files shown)


sanket1729:
I think we decided this is not possible, right?

darosior:
[Yes](https://github.com/sipa/miniscript/pull/97#issuecomment-1031262186), let's close this.

PR #296, opened 2022-02-09
Improve Taproot descriptor tree parsing

sanket1729:


src/expression.rs
```diff
@@ -173,34 +132,39 @@ impl<'a> Tree<'a> {
...
                     }
 
                     sl = &new_sl[1..];
                     match new_sl.as_bytes()[0] {
                         b',' => {}
-                        b'}' => break,
-                        _ => return Err(Error::ExpectedChar(',')),
+                        last_byte => {
+                            if last_byte == closing_delim(delim) as u8 {
+                                break;
```
sanket1729, line 163: doing an `as u8` on } or ) correct as we have checked things to be ascii

apoelstra (approved):
ACK efe016d1c7c32a4d8d73e2dc076679d2872b8c18

sanket1729:
@apoelstra the ACK is not the tip of the branch.

PR #297, opened 2022-02-10
Add pretaproot descriptor and addressabledescriptor

sanket1729:
Still not completely sure if this is what you wanted. I think we might also want new descriptor enums for `PreTaprootDescriptor` and `AddressableDescriptor`

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -329,6 +329,43 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
             Descriptor::Tr(ref _tr) => DescriptorType::Tr,
         }
     }
+
+    /// .
+    /// Convert a Descriptor into [`pretaproot::PreTaprootDescriptor`]
+    /// # Examples
+    ///
+    /// ```
+    /// use std::str::FromStr;
+    /// use miniscript::descriptor::Descriptor;
+    /// use miniscript::{PreTaprootDescriptor, PreTaprootDescriptorTrait};
+    /// use miniscript::bitcoin;
+    ///
+    /// // A descriptor with a string generic
+    /// let desc = Descriptor::<bitcoin::PublicKey>::from_str("wpkh(02e18f242c8b0b589bfffeac30e1baa80a60933a649c7fb0f1103e78fbf58aa0ed)")
+    ///     .expect("Valid segwitv0 descriptor");
+    /// let pre_tap_desc = desc.into_pre_taproot_desc().expect("Wsh is pre taproot");
+    ///
+    /// // Now the script code and explicit script no longer fail on longer fail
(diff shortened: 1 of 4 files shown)


sanket1729:
Converted to draft because there are more changes to be done here as discussed on IRC.

src/descriptor/mod.rs
```diff
@@ -619,69 +447,161 @@ impl Descriptor<DescriptorPublicKey> {
...
+    use super::{Pkh, PreTaprootDescriptor, Sh, Wpkh, Wsh};
+    use {DescriptorTrait, Error, MiniscriptKey, Satisfier, ToPublicKey};
+
+    /// A general trait for Pre taproot bitcoin descriptor.
+    /// Similar to [`DescriptorTrait`], but `explicit_script` and `script_code` methods cannot fail
+    pub trait PreTaprootDescriptorTrait<Pk: MiniscriptKey>: DescriptorTrait<Pk> {
+        /// Same as [`DescriptorTrait::sanity_check`]
+        fn sanity_check(&self) -> Result<(), Error> {
+            <Self as DescriptorTrait<Pk>>::sanity_check(&self)
         }
```
apoelstra, line 509: I think these pass-through implementaitons aren't needed. Because `PreTaprootDescriptorTrait` implies `DescriptorTrait`, anything that implements the former is guaranteed to implement the latter.

I think that users might be forced to import both traits in order to use their methods, but I think this is a reasonable cost to charge them, in exchange for greatly reducing the amount of duplicated code in this crate.

src/macros.rs
```diff
@@ -96,3 +96,232 @@ macro_rules! serde_string_impl_pk {
...
+            fn sanity_check(&self) -> Result<(), Error> {
+                ImplDesc! {
+                    @as_expr
+                    match self {
+                        $(
+                            $struct_name::$var_name(ref v) => <$var_ty::<Pk> as $trait_name<Pk>>::sanity_check(&v),
+                        )*
+                    }
+                }
+            }
```
apoelstra, line 121: It's hard to balance going too macro-crazy, but couldn't we make the user give every method name (and its signature) to the macro, and then generate all these from the same template?

This would make the macro smaller and also more general

apoelstra:
concept ACK, I think this is the direction we want to go in, but I think we could tighten it up.

src/descriptor/mod.rs
sanket1729: I see. This is much better

src/macros.rs
sanket1729: Yup, good idea. I think this should be possible

apoelstra (approved):
ACK 121e0b444f4e2030c5482d7d3fe0454c0b5a7143

PR #299, opened 2022-02-24
Allow constructing Sh descriptors wrapping an existing Wsh/Wpkh

shesek:
Useful if you already have an `W{s,pk}h` descriptor and wish to wrap it with P2SH.

This used to be [possible](https://github.com/shesek/minsc/blob/33433462e84e3903a70a9aac6f47e591df094127/src/stdlib/miniscript.rs#L149-L153) before the new inner types for the Descriptor enum variants, which has private fields that cannot be initialized externally.

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -277,6 +277,18 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
         Ok(Descriptor::Bare(Bare::new(ms)?))
     }
 
+    // Wrap with sh
+
+    /// Create a new sh wrapper for the given wpkh descriptor
+    pub fn new_sh_with_wpkh(wpkh: Wpkh<Pk>) -> Self {
+        Descriptor::Sh(Sh::new_with_wpkh(wpkh))
+    }
+
+    /// Create a new sh wrapper for the given wsh descriptor
+    pub fn new_sh_with_wsh(wsh: Wsh<Pk>) -> Self {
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ACK bd7c642416c6176d71eca7e2a7fb374e79755da7

Looks good! I'll let @sanket1729 also check this out, in case he has opinions about the method naming.

sanket1729 (approved):
ACK https://github.com/rust-bitcoin/rust-miniscript/commit/bd7c642416c6176d71eca7e2a7fb374e79755da7

sanket1729:
Actually, re-reading it. I have some complaints about the names :(. 
This is not consistent with the other names in the API. We have names like `Descriptor::new_wsh()` instead of `Descriptor::new_with_wsh`. I would prefer to change to `new_<name>` from `new_with_<name>` without the `with` for consistency with other APIs.

We want to be more rust-idiomatic with names like `Descriptor::with_wsh` (instead of `new_with_wsh` (as suggested in this PR), or `new_wsh` as implemented in master). But that is another issue.

shesek:
@sanket1729 `Descriptor::new_sh_wsh()`, `Descriptor::new_sh_wpkh()`, `Sh::new_wsh()` and `Sh::new_wpkh()` already exists though, with a different meaning (they accept the underlying Miniscript and wrap it with a `W{s,pk}h `descriptor, rather than accept the descriptor directly).

PR #301, opened 2022-02-25
Taproot interpreter support

sanket1729:
Add taproot interpreter support. This is the second to last PR for taproot miniscript support. The last PR will add some more high-level APIs for taproot psbt and integration tests.

### DIFF
--- a/src/interpreter/inner.rs
+++ b/src/interpreter/inner.rs
@@ -376,7 +376,6 @@ pub(super) fn pre_taproot_to_no_checks<Ctx: ScriptContext>(
 }
 
 // Convert a miniscript from a specified context into an insane miniscript
-#[allow(dead_code)]
 pub(super) fn taproot_to_no_checks<Ctx: ScriptContext>(
     ms: &Miniscript<bitcoin::XOnlyPublicKey, Ctx>,
 ) -> Miniscript<BitcoinKey, NoChecks> {
--- a/src/interpreter/inner.rs
+++ b/src/interpreter/inner.rs
@@ -154,7 +154,7 @@ pub fn from_txdata<'txin>(
             match wit_stack.pop() {
                 Some(elem) => {
                     let pk = pk_from_stackelem(&elem, true)?;
-                    if *spk == bitcoin::Script::new_v0_wpkh(&pk.to_pubkeyhash().into()) {
+                    if *spk == bitcoin::Script::new_v0_p2wpkh(&pk.to_pubkeyhash().into()) {
                         Ok((
                             Inner::PublicKey(pk, PubkeyType::Wpkh),
                             wit_stack,
@@ -177,7 +177,7 @@ pub fn from_txdata<'txin>(
                     let miniscript = script_from_stackelem(&elem)?;
                     let script = miniscript.encode();
                     let scripthash = sha256::Hash::hash(&script[..]);
-                    if *spk == bitcoin::Script::new_v0_wsh(&scripthash.into()) {
+                    if *spk == bitcoin::Script::new_v0_p2wsh(&scripthash.into()) {
                         Ok((
                             Inner::Script(miniscript, ScriptType::Wsh),
                             wit_stack,
@@ -208,8 +208,9 @@ pub fn from_txdata<'txin>(
                                 } else {
                                     let pk = pk_from_stackelem(&elem, true)?;
                                     if slice
-                                        == &bitcoin::Script::new_v0_wpkh(&pk.to_pubkeyhash().into())
-                                            [..]
+                                        == &bitcoin::Script::new_v0_p2wpkh(
+                                            &pk.to_pubkeyhash().into(),
+                                        )[..]
                                     {
                                         Ok((
                                             Inner::PublicKey(pk, PubkeyType::ShWpkh),
@@ -233,7 +234,8 @@ pub fn from_txdata<'txin>(
                                     let miniscript = script_from_stackelem(&elem)?;
                                     let script = miniscript.encode();
                                     let scripthash = sha256::Hash::hash(&script[..]);
-                                    if slice == &bitcoin::Script::new_v0_wsh(&scripthash.into())[..]
+                                    if slice
+                                        == &bitcoin::Script::new_v0_p2wsh(&scripthash.into())[..]
                                     {
                                         Ok((
                                             Inner::Script(miniscript, ScriptType::ShWsh),
@@ -322,7 +324,7 @@ mod tests {
 
             let pkhash = key.to_pubkeyhash().into();
             let wpkhash = key.to_pubkeyhash().into();
-            let wpkh_spk = bitcoin::Script::new_v0_wpkh(&wpkhash);
+            let wpkh_spk = bitcoin::Script::new_v0_p2wpkh(&wpkhash);
             let wpkh_scripthash = hash160::Hash::hash(&wpkh_spk[..]).into();
 
             KeyTestData {
@@ -670,7 +672,7 @@ mod tests {
         let wit_hash = sha256::Hash::hash(&witness_script[..]).into();
         let wit_stack = Witness::from_vec(vec![witness_script.to_bytes()]);
 
-        let spk = Script::new_v0_wsh(&wit_hash);
+        let spk = Script::new_v0_p2wsh(&wit_hash);
         let blank_script = bitcoin::Script::new();
 
         // wsh without witness
@@ -708,7 +710,7 @@ mod tests {
         let wit_hash = sha256::Hash::hash(&witness_script[..]).into();
         let wit_stack = Witness::from_vec(vec![witness_script.to_bytes()]);
 
-        let redeem_script = Script::new_v0_wsh(&wit_hash);
+        let redeem_script = Script::new_v0_p2wsh(&wit_hash);
         let script_sig = script::Builder::new()
             .push_slice(&redeem_script[..])
             .into_script();
--- a/src/interpreter/inner.rs
+++ b/src/interpreter/inner.rs
@@ -83,7 +83,6 @@ pub enum ScriptType {
     Sh,
     Wsh,
     ShWsh,
-    #[allow(dead_code)]
     Tr, // Script Spend
 }
 
@@ -234,17 +233,23 @@ pub(super) fn from_txdata<'txin>(
                     let ms = taproot_to_no_checks(&tap_script);
                     // Creating new contexts is cheap
                     let secp = bitcoin::secp256k1::Secp256k1::verification_only();
+                    let tap_script = tap_script.encode();
                     // Should not really need to call dangerous assumed tweaked here.
                     // Should be fixed after RC
                     if ctrl_blk.verify_taproot_commitment(
                         &secp,
                         &output_key.dangerous_assume_tweaked(),
-                        &tap_script.encode(),
+                        &tap_script,
                     ) {
                         Ok((
                             Inner::Script(ms, ScriptType::Tr),
                             wit_stack,
-                            None, // Tr script code None
+                            // Tr script spend code is stored in script spend. This is an internal hack to store the
+                            // encoded script instead yet another tagged enum for saving tap_script
+                            // We can recompute the script again by translating from nochecks to tap and re-encoding it
+                            // Having this hack seems simple because this is not exposed publicly and the internal code is
+                            // easy to audit
+                            Some(tap_script),
                         ))
                     } else {
                         return Err(Error::ControlBlockVerificationError);
--- a/src/interpreter/inner.rs
+++ b/src/interpreter/inner.rs
@@ -16,8 +16,10 @@ use bitcoin;
 use bitcoin::blockdata::witness::Witness;
 use bitcoin::hashes::{hash160, sha256, Hash};
 
-use super::{stack, Error, Stack};
-use miniscript::context::NoChecksEcdsa;
+use {BareCtx, Legacy, Segwitv0};
+
+use super::{stack, BitcoinKey, Error, Stack, TaggedHash160};
+use miniscript::context::{NoChecksEcdsa, ScriptContext};
 use {Miniscript, MiniscriptKey};
 
 /// Attempts to parse a slice as a Bitcoin public key, checking compressedness
@@ -46,13 +48,14 @@ fn pk_from_stackelem<'a>(
     pk_from_slice(slice, require_compressed)
 }
 
-fn script_from_stackelem<'a>(
+// Parse the script with appropriate context to check for context errors like
+// correct usage of x-only keys or multi_a
+fn script_from_stackelem<'a, Ctx: ScriptContext>(
     elem: &stack::Element<'a>,
-) -> Result<Miniscript<bitcoin::PublicKey, NoChecksEcdsa>, Error> {
+) -> Result<Miniscript<Ctx::Key, Ctx>, Error> {
     match *elem {
         stack::Element::Push(sl) => {
-            Miniscript::<bitcoin::PublicKey, _>::parse_insane(&bitcoin::Script::from(sl.to_owned()))
-                .map_err(Error::from)
+            Miniscript::parse_insane(&bitcoin::Script::from(sl.to_owned())).map_err(Error::from)
         }
         stack::Element::Satisfied => Miniscript::from_ast(::Terminal::True).map_err(Error::from),
         stack::Element::Dissatisfied => {
@@ -68,6 +71,7 @@ pub enum PubkeyType {
     Pkh,
     Wpkh,
     ShWpkh,
+    Tr, // Key Spend
 }
 
 /// Helper type to indicate the origin of the bare miniscript that the interpereter uses
@@ -77,16 +81,20 @@ pub enum ScriptType {
     Sh,
     Wsh,
     ShWsh,
+    #[allow(dead_code)]
+    Tr, // Script Spend
 }
 
 /// Structure representing a script under evaluation as a Miniscript
 #[derive(Clone, PartialEq, Eq, Debug)]
-pub enum Inner {
+pub(super) enum Inner {
     /// The script being evaluated is a simple public key check (pay-to-pk,
     /// pay-to-pkhash or pay-to-witness-pkhash)
-    PublicKey(bitcoin::PublicKey, PubkeyType),
+    // Technically, this allows representing a (XonlyKey, Sh) output but we make sure
+    // that only appropriate outputs are created
+    PublicKey(super::BitcoinKey, PubkeyType),
     /// The script being evaluated is an actual script
-    Script(Miniscript<bitcoin::PublicKey, NoChecksEcdsa>, ScriptType),
+    Script(Miniscript<super::BitcoinKey, NoChecksEcdsa>, ScriptType),
 }
 
 // The `Script` returned by this method is always generated/cloned ... when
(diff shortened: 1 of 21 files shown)


src/interpreter/mod.rs
```diff
@@ -33,18 +36,109 @@ mod error;
...
+    }
+}
+
+impl From<bitcoin::XOnlyPublicKey> for BitcoinKey {
+    fn from(xpk: bitcoin::XOnlyPublicKey) -> Self {
+        BitcoinKey::XOnlyPublicKey(xpk)
+    }
+}
+
+// While parsing we need to remember how to the hash was parsed so that we can
```
apoelstra, line 126: In 374ab9a2a389da47779f0f4fa33dc7e94f491455:

`so that we can` what?

src/interpreter/inner.rs
```diff
@@ -211,7 +211,7 @@ pub(super) fn from_txdata<'txin>(
                 // Key spend
                 // let sig =
                 // Tr inference to be done in future commit
-                todo!();
+                panic!("TODO");
```
apoelstra: In 277c78392be39bb2755cd00fe6168fb2de7e685c:

lol does this change anything?

sanket1729: Nope. Just allows compiling on MSRV :P . I was testing whether that particular commit compiles on rust 1.29

apoelstra (approved):
ACK 4e6352aec8b00bc6ac0dd226c6ec0031db6d16b2

I have a single fixup commit in https://github.com/apoelstra/rust-miniscript/tree/2022-03--301-fixup which you can take or leave; it just cleans up one comment that I didn't really understand.

sanket1729:
Cherry-picked your commit

src/interpreter/inner.rs
```diff
@@ -77,16 +81,20 @@ pub enum ScriptType {
...
+pub(super) enum Inner {
     /// The script being evaluated is a simple public key check (pay-to-pk,
     /// pay-to-pkhash or pay-to-witness-pkhash)
-    PublicKey(bitcoin::PublicKey, PubkeyType),
+    // Technically, this allows representing a (XonlyKey, Sh) output but we make sure
+    // that only appropriate outputs are created
+    PublicKey(super::BitcoinKey, PubkeyType),
     /// The script being evaluated is an actual script
-    Script(Miniscript<bitcoin::PublicKey, NoChecksEcdsa>, ScriptType),
+    Script(Miniscript<super::BitcoinKey, NoChecksEcdsa>, ScriptType),
```
JeremyRubin: When would we want a TR with an ECDSA here?

src/interpreter/inner.rs
```diff
@@ -95,11 +103,12 @@ pub enum Inner {
...
 /// Parses an `Inner` and appropriate `Stack` from completed transaction data,
 /// as well as the script that should be used as a scriptCode in a sighash
-pub fn from_txdata<'txin>(
+/// Tr outputs don't have script code and return None.
+pub(super) fn from_txdata<'txin>(
     spk: &bitcoin::Script,
     script_sig: &'txin bitcoin::Script,
     witness: &'txin Witness,
-) -> Result<(Inner, Stack<'txin>, bitcoin::Script), Error> {
+) -> Result<(Inner, Stack<'txin>, Option<bitcoin::Script>), Error> {
```
JeremyRubin, line 112: should the script_sig also be optional here then?

src/interpreter/inner.rs
```diff
@@ -190,6 +200,25 @@ pub fn from_txdata<'txin>(
...
+        if !ssig_stack.is_empty() {
+            Err(Error::NonEmptyScriptSig)
+        } else {
+            let output_key = bitcoin::XOnlyPublicKey::from_slice(&spk[2..])
+                .map_err(|_| Error::XOnlyPublicKeyParseError)?;
+            if wit_stack.len() == 1 {
+                // Key spend
+                // let sig =
+                // Tr inference to be done in future commit
+                todo!();
```
JeremyRubin: n.b. might be better not to commit a todo! here for better bisecting sometime down the line?

src/interpreter/inner.rs
```diff
@@ -174,14 +183,15 @@ pub fn from_txdata<'txin>(
         } else {
             match wit_stack.pop() {
                 Some(elem) => {
-                    let miniscript = script_from_stackelem(&elem)?;
+                    let miniscript = script_from_stackelem::<Segwitv0>(&elem)?;
                     let script = miniscript.encode();
+                    let miniscript = pre_taproot_to_no_checks(&miniscript);
```
JeremyRubin, line 189: can you add a comment here?

src/interpreter/inner.rs
```diff
@@ -231,16 +262,18 @@ pub fn from_txdata<'txin>(
                                 if !ssig_stack.is_empty() {
                                     Err(Error::NonEmptyScriptSig)
                                 } else {
-                                    let miniscript = script_from_stackelem(&elem)?;
+                                    // parse wsh with Segwitv0 context
+                                    let miniscript = script_from_stackelem::<Segwitv0>(&elem)?;
                                     let script = miniscript.encode();
+                                    let miniscript = pre_taproot_to_no_checks(&miniscript);
```
JeremyRubin, line 309: can you add a comment on this line? why

src/interpreter/inner.rs
```diff
@@ -274,18 +308,46 @@ pub fn from_txdata<'txin>(
...
         } else {
             Err(Error::NonEmptyWitness)
         }
     }
 }
 
+// Convert a miniscript from a specified context into an insane miniscript
+// We need to remember how the hash160 was translated because while doing a checksig
+// we need to know whether to parse the public key provided in witness as x-only or full
+pub(super) fn pre_taproot_to_no_checks<Ctx: ScriptContext>(
```
JeremyRubin, line 369: this function / its name is kinda confusing; maybe you can clarify the comment to discuss what the goal is (parse the public key provided in witness as x-only or full) and then what we do & why it accomplishes that?

Maybe you can also add a crate local trait IsPreTaproot to restrict what Ctx it can be called with.

src/interpreter/inner.rs
```diff
@@ -274,18 +308,46 @@ pub fn from_txdata<'txin>(
...
+    // specify the () error type as this cannot error
+    ms.real_translate_pk::<_, _, _, (), _>(&mut |pk| Ok(BitcoinKey::Fullkey(*pk)), &mut |pkh| {
+        Ok(TaggedHash160::FullKey(*pkh))
+    })
+    .expect("Translation should succeed")
+}
+
+// Convert a miniscript from a specified context into an insane miniscript
+#[allow(dead_code)]
+pub(super) fn taproot_to_no_checks<Ctx: ScriptContext>(
```
JeremyRubin, line 380: why is this conversion insane? why do we do this?

src/interpreter/inner.rs
JeremyRubin: or a trait crate local ToNoChecks and you can impl it default, and change just Tap CTX's impl?

src/interpreter/mod.rs
```diff
@@ -33,18 +36,109 @@ mod error;
...
+
+// Internally used enum for different types of bitcoin keys
+// Even though we implement MiniscriptKey for BitcoinKey, we make sure that there
+// are little mis-use
+// - The only constructors for this are only called in from_txdata that take care
+//   using the correct enum variant
+// - This does not implement ToPublicKey to avoid context dependant encoding/decoding of 33/32
+//   byte keys. This allows us to keep a single NoChecks context instead of a context for
+//   for NoChecksSchnorr/NoChecksEcdsa.
+// Long term TODO: There really should be any need for Miniscript<Pk: MiniscriptKey> struct
```
JeremyRubin, line 94: nit; typo missing "not be any need"

src/interpreter/mod.rs
```diff
@@ -33,18 +36,109 @@ mod error;
...
+
+impl From<bitcoin::XOnlyPublicKey> for BitcoinKey {
+    fn from(xpk: bitcoin::XOnlyPublicKey) -> Self {
+        BitcoinKey::XOnlyPublicKey(xpk)
+    }
+}
+
+// While parsing we need to remember how to the hash was parsed so that we can
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+enum TaggedHash160 {
```
JeremyRubin, line 130: what does "tagged" mean in this context?

just that we track what kind of key?

May call it TypedHash because TaggedHash now means something else

src/interpreter/mod.rs
```diff
@@ -173,10 +175,12 @@ impl<'txin> Interpreter<'txin> {
...
     /// In case the script is actually dissatisfied, this may return several values
     /// before ultimately returning an error.
-    pub fn iter<'iter, F: FnMut(&KeySigPair) -> bool>(
+    /// See [Self::iter_assume_sigs] for a simpler API without information about Prevouts
+    /// but skips the signature verification
+    pub fn iter<'iter>(
         &'iter self,
-        verify_sig: F,
-    ) -> Iter<'txin, 'iter, F> {
+        verify_sig: Box<dyn FnMut(&KeySigPair) -> bool + 'iter>,
```
JeremyRubin, line 191: i think you can make F: dyn FnMut... + 'iter without forcing it to be Boxed explicitly

JeremyRubin:
looks pretty good to me!

JeremyRubin:
n.b. https://github.com/rust-bitcoin/rust-bitcoin/commit/04787d486758a9684a9b9639e6fd395228559a4e will seem to break some of the code using Prevouts, not clear if that's going into a 0.28 or not.

is this commit going into RC2?

src/interpreter/inner.rs
```diff
@@ -190,6 +201,65 @@ pub fn from_txdata<'txin>(
...
+                // Check for annex
+                let ctrl_blk = wit_stack.pop().ok_or(Error::UnexpectedStackEnd)?;
+                let ctrl_blk = ctrl_blk.as_push()?;
+                let tap_script = wit_stack.pop().ok_or(Error::UnexpectedStackEnd)?;
+                if ctrl_blk.len() > 0 && ctrl_blk[0] == TAPROOT_ANNEX_PREFIX {
+                    // Annex is non-standard, bitcoin consensus rules ignore it.
+                    // Our sighash structure and signature verification
+                    // does not support annex, return error
+                    return Err(Error::TapAnnexUnsupported);
+                } else if wit_stack.len() >= 2 {
```
JeremyRubin, line 229: this line is incorrect -- wit_stack can be 1 element (or in the case of CTV, empty!)

JeremyRubin: I think the confusion comes from the fact that the tap_script and ctrl_blk are popped off the wit_stack at this point.

src/interpreter/inner.rs
sanket1729: We never want it. But we have to be careful here to not ever create it.

src/interpreter/inner.rs
sanket1729: That should be provided as `Script::new`. Every segwti tx has an empty script sig instead of the None script sig.

src/interpreter/inner.rs
sanket1729: Fair point. Will change to the `unreachable!` in the commit it is introduced. In my local development, I only test that every commit compiles for stable and not for MSRV.

src/interpreter/inner.rs
sanket1729: Yup, the script spend must contain at least two witness elements that are being popped here. The additional witness elements can be 0

src/interpreter/inner.rs
sanket1729: Will add a comment elaborating this.

src/interpreter/inner.rs
sanket1729: Is this a comment about `todo!` not working on MSRV? Or that it might panic? If it is the latter, I think it is safe to do here because there is a new feature being added.

src/interpreter/inner.rs
sanket1729: The interpreter API operates on a NoChecks context because making it context-specific would involve lots of duplicate code for each context. The interpreter is for the use case when you already have a transaction and want to figure out what constraints were satisfied. The primary use-case for this feature was miniscript friendly block explorers that would output what constraints were satisfied for this given transaction. Hence, the insane scripts. 

As for this PR, I have added a `ToNoChecksTrait` that makes the code look cleaner.

src/interpreter/inner.rs
JeremyRubin: nope; I still think it's incorrect.

referencing: https://github.com/bitcoin/bitcoin/blob/64a4483dc6798a9a7d8327d320a17b3c4d7d4ee0/src/script/interpreter.cpp#L1919

1. if stack empty, fail
1. Check if there is 2 things and one of the things is an annex, remove it
2. If the stack is length 1 do keypath
3. If the stack is length 2 do script spend

Whereas your code says:

1. if stack empty, fail
1. If the length is 1 do keyspend
2. If the length is >= 2, pop the stack for CTRL and pop for tap script (stack length now 0 to N elemented):
3. If CTRL is actually an annex (0x50), fail.
4. Now control is actually a control, and script is a script for sure
5. check that there are >= 2 elements on the stack; if not fail
6. Execute the script

Clearly, the last step should not be happending! For example, the following witness fails:

<sig> <pk checksig> <ctrl>

because the script only needs one argument.

JeremyRubin: (the right number of arguments is zero)

src/interpreter/inner.rs
JeremyRubin: not sure about that, I just meant it's nice if I want to test the git history against a taproot fragment that I can see where a bug in the interpreter gets added (e.g., like the bug I pointed out below about stack sizes would not be catchable in a bisect here).

src/interpreter/inner.rs
sanket1729: > Clearly, the last step should not be happending! For example, the following witness fails:

Are you missing adding an example?

If I am understanding this correctly, this will fail when the key spend has annex? To be clear, my only concern here is we should fail cleanly on the annex, and not the exact emulation of bitcoin consensus.

sanket1729: Added a commit to reflect the first part

sanket1729:
> n.b. https://github.com/rust-bitcoin/rust-bitcoin/commit/04787d486758a9684a9b9639e6fd395228559a4e will seem to break some of the code using Prevouts, not clear if that's going into a 0.28 or not.

I am not sure if it will be 0.28. It is  a good to have for rust-miniscript, but I think we may have to wait 0.28.1

src/interpreter/inner.rs
JeremyRubin: lol -- example was there, but markdown was hiding it because it looked like html

src/interpreter/inner.rs
```diff
@@ -216,47 +228,36 @@ pub(super) fn from_txdata<'txin>(
...
-                // wit_stack.len() >=2
-                // Check for annex
                 let ctrl_blk = wit_stack.pop().ok_or(Error::UnexpectedStackEnd)?;
                 let ctrl_blk = ctrl_blk.as_push()?;
                 let tap_script = wit_stack.pop().ok_or(Error::UnexpectedStackEnd)?;
-                if ctrl_blk.len() > 0 && ctrl_blk[0] == TAPROOT_ANNEX_PREFIX {
-                    // Annex is non-standard, bitcoin consensus rules ignore it.
-                    // Our sighash structure and signature verification
-                    // does not support annex, return error
-                    return Err(Error::TapAnnexUnsupported);
```
JeremyRubin: you should keep this, no?

src/interpreter/inner.rs
```diff
@@ -208,6 +208,18 @@ pub(super) fn from_txdata<'txin>(
...
             let output_key = bitcoin::XOnlyPublicKey::from_slice(&spk[2..])
                 .map_err(|_| Error::XOnlyPublicKeyParseError)?;
+            if wit_stack.len() >= 2 {
+                let has_annex = wit_stack
+                    .last()
+                    .and_then(|x| x.as_push().ok())
+                    .map(|x| x.len() > 0 && x[0] == TAPROOT_ANNEX_PREFIX)
+                    .unwrap_or(false);
+                if has_annex {
+                    wit_stack.pop();
```
JeremyRubin, line 218: return error here.

src/interpreter/inner.rs
```diff
@@ -216,47 +228,36 @@ pub(super) fn from_txdata<'txin>(
                     None, // Tr script code None
                 ))
             } else {
-                // wit_stack.len() >=2
-                // Check for annex
                 let ctrl_blk = wit_stack.pop().ok_or(Error::UnexpectedStackEnd)?;
```
JeremyRubin, line 231: nit: mild preference to have an explicity match over wit_stack.len() for 0, 1, _ and return Error::UnexpectedStackEnd for 0, and then make a new error Errror::ImplementationInvariantBug or something for the pops here (so that we're not using expect or unwrap) -- for readability it makes it more clear to me.

src/interpreter/inner.rs
```diff
@@ -208,6 +208,18 @@ pub(super) fn from_txdata<'txin>(
         } else {
             let output_key = bitcoin::XOnlyPublicKey::from_slice(&spk[2..])
                 .map_err(|_| Error::XOnlyPublicKeyParseError)?;
+            if wit_stack.len() >= 2 {
+                let has_annex = wit_stack
+                    .last()
+                    .and_then(|x| x.as_push().ok())
+                    .map(|x| x.len() > 0 && x[0] == TAPROOT_ANNEX_PREFIX)
+                    .unwrap_or(false);
```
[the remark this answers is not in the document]
sanket1729: Is it rust 2021 :P ? We are still with 1.29 MSRV :(

src/interpreter/inner.rs
sanket1729: Yes, indeed. Initially, I was thinking that we can just pop and ignore it. But as the comment says here, since we don't support signature verification with Annex, it is correct to Error

src/interpreter/inner.rs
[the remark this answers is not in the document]
JeremyRubin: https://adventures.michaelfbryan.com/posts/daily/slice-patterns/

1.26 for slice patterns, 1.42 for the "everything else .." matching within them.

so i think it's only really viable once we're at 1.42

src/interpreter/inner.rs
sanket1729: Changed to match over len(). 

I don't see any expects or unwraps in this code? Where are we using those?

JeremyRubin:
Verification of TR looks much cleaner now, thanks for the fixups.

sanket1729:
@apoelstra , ready for your review again

src/interpreter/inner.rs
JeremyRubin: we're not, i was just saying that now the ok_or is guaranteed to succeed, so it might make sense to make it an assertion via unwrap or expect (since it should be unreachable) or to add a Error::ImplementationError instead of unwraps or UnexpectedStackEnd (because the user should know they didn't do submit something invalid if it ever hits).

apoelstra (approved):
ACK be83ac7c2f9c1b717ff0a4adbc6365c4a59a3892

PR #304, opened 2022-03-07
Fix the length of the taproot output to 34 bytes in script_is_v1_tr

JeremyRubin:
It's impossible to finalize a taproot PSBT with this line not fixed because the finalizer thinks the output is not TR and then tries to parse it as a legacy script.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -52,11 +52,14 @@ impl MsKeyBuilder for script::Builder {
 // This belongs in rust-bitcoin, make this upstream later
 pub(crate) fn script_is_v1_tr(s: &Script) -> bool {
     let mut iter = s.instructions_minimal();
-    if s.len() != 32
-        || iter.next() != Some(Ok(Instruction::Op(opcodes::all::OP_PUSHNUM_1)))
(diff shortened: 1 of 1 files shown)


sanket1729:
This is fixed in a cleaner way by using `rust-bitcoin` script.is_v1_p2tr()? I will keep this open as an issue while it is implemented in another PR.

JeremyRubin:
The PR is actually still wrong, updated...

JeremyRubin:
yes I agree the new API makes sense once 0.28 can be used, but I think this makes sense to merge as a hotfix so that at least people can work on PSBT stuff based on 0.27 still (hopefully, e.g., tests)

apoelstra:
The current state of the library already isn't usable with 0.27 because we depend on `rust-bitcoin` types from specific git commits. We're in no-mans-land until 0.28 is published.

JeremyRubin:
that's fair, i just mean to say that this is an obvious fix to accept so that at least more of the code is not buggy, and we have an explicit commit where we've labelled what the bug was.

it's OK if it is overwritten later.

sanket1729:
@JeremyRubin , this is now cleanly addressed in the first commit of #305. That also adds psbts APIs which would be of interest to you.

PR #307, opened 2022-03-13
Make Tr::spend_info public

dr-orlovsky:
This is needed for wallet libraries downstream to access for instance merkle root of the script.


sanket1729:
Sorry, this is done as a part of #305. Don't mind merging this here. If it requires one more revision, I will merge this instead.

sanket1729 (commented):
ACK 4c0b64ad47319ac26b558c9f89da4c003e3558b4

PR #308, opened 2022-03-14
MultiA interpreter

sanket1729:
Support for multi_a interpreter. Required for interpreter check finalizer.

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -1033,7 +1033,7 @@ mod tests {
     use super::*;
     use bitcoin;
     use bitcoin::hashes::{hash160, ripemd160, sha256, sha256d, Hash};
-    use bitcoin::secp256k1::{self, Secp256k1, VerifyOnly};
+    use bitcoin::secp256k1::{self, Secp256k1};
     use miniscript::context::NoChecks;
     use Miniscript;
     use MiniscriptKey;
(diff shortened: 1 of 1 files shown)


src/interpreter/mod.rs
```diff
@@ -844,6 +844,47 @@ where
...
+                        } else {
+                            self.stack.push(stack::Element::Dissatisfied);
+                        }
+                    } else {
+                        // evaluate each key with as a pk
+                        // note that evaluate_pk will error on non-empty incorrect sigs
+                        // push 1 on satisfied sigs and push 0 on empty sigs
+                        match self
+                            .stack
+                            .evaluate_pk(&mut self.verify_sig, &subs[node_state.n_evaluated])
```
dr-orlovsky, line 860: wouldn't `subs[node_state.n_evaluated]` panic if the number of the provided signatures is less than k?

dr-orlovsky (commented):
One question

src/interpreter/mod.rs
sanket1729: `n_evaluated` denotes the numbers of sub-expressions evaluated, not the number of sigs. The number of expressions evaluated must be less than subs.len() because we add +1 whenever we evaluate one sub-expression. 

Inside `evalute_pk`, there is a check that checks if we have one sig for a given pk. This will correctly error if given sigs is less than the number of keys.

dr-orlovsky (dismissed):
utACK d30d8c6feb2ba4991de056c4190523865e7480db

src/interpreter/mod.rs
dr-orlovsky: At some distant future a test for that will be nice to have :)

sanket1729: Will add in this one. Should be a couple mins of work

dr-orlovsky (approved):
utACK 251921ef6c5481f80b9ebea4b1c641a8e0c1551b

PR #309, opened 2022-03-14
Add integration tests

sanket1729:
This is the final PR testing of many of the new features in the taproot miniscript release. This defines a new template language for checking all the Psbt APIs (and thereby all the miniscript ones). 

```rust
    // K : Compressed key available
    // K!: Compressed key with corresponding secret key unknown
    // X: X-only key available
    // X!: X-only key with corresponding secret key unknown

    // Test 1: Simple spend with internal key
    let wit = test_desc::test_desc_satisfy(cl, testdata, "tr(X)");
    assert!(wit.len() == 1);
```

Right now, it does not test error cases, but I plan to update this framework so that we can also test those.

Based on #308 and #310


apoelstra (approved):
ACK 13c47464c1fde4e5e3af0785e45f8b01885c0dd2

Wow, this is great!!

PR #310, opened 2022-03-14
Fix finalizer check for sh(wpkh)

sanket1729:
We should check the witness script hashes to the redeem script in the
nested sh wsh case. Not the script pubkey

Caught up integration tests


apoelstra (approved):
ACK 983b889a1f2349ddfe78236dd6f6778d4634b332

The new code looks weird and probably there should be a comment there -- but I won't hold up the PR for it. `git blame` will be pretty clear about what happened here.

PR #312, opened 2022-03-16
Small optimization of Tr descriptor parsing.

JeremyRubin:


JeremyRubin:
probably not the right way to do this, but currently serializations using from_str will be broken otherwise.

sanket1729:
This has been fixed in latest master, using something similar.

JeremyRubin:
Ah I missed that one XD.

Just noticed it on an integration which was using a slightly stale revision.


I updated this PR to still be relevant, since in the version on master we end up checking for the ASCII + looking for the checksum twice. By directly calling Tr::from_str we save that extra pass.

I also made the match over 'tr(', since that captures that it is a function call v.s. any word starting with tr

sanket1729 (approved):
utACK 00e40e8fb469f0a951ba367cce75469fa3937089

PR #324, opened 2022-03-27
Bitcoin 0.28.0-rc.2

dr-orlovsky:
This is a draft API adopting new version of rust-bitcoin 0.28.0-rc.2

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "miniscript"
-version = "7.0.0-rc.1"
+version = "7.0.0-rc.2"
 authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>, Sanket Kanjalkar <sanket1729@gmail.com>"]
 repository = "https://github.com/apoelstra/miniscript"
 description = "Miniscript: a subset of Bitcoin Script designed for analysis"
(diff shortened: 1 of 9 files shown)


sanket1729:
Note that this would need to be a single commit to make sure all commits compile and test. Unfortunately, there is no clean way to have atomic commits here. 

What I typically tend to do in these cases is to have separate branches with 2 PRs, one with atomic commits and one with a single commit. And then locally check that the diff between the two is null. This is not required for this case because the diff is small and I can review it in single commit.

dr-orlovsky:
@sanket1729 working on update

sanket1729:
@dr-orlovsky: One other thing that needs to be fixed here: https://github.com/rust-bitcoin/rust-miniscript/blob/1f9affc43723a0669ba90cfcca83710753e032df/src/psbt/finalizer.rs#L119 Should return `&bitcoin::TxOut` instead of `bitcoin::Txout`. 

Also, the commit message should be 0.28.0-rc.2 instead of 0.27.0-rc.2.

dr-orlovsky:
@sanket1729 will fix both of the issues once bitcoin gets released and will do a final update to use released version

apoelstra:
Ok, rust-bitcoin rc2 is out

dr-orlovsky:
Updating this one

dr-orlovsky:
@apoelstra @sanket1729 ready for the review, CI is fixed

sanket1729 (approved):
tACK 08ed922ec9873eb363a4a95880a08a2e23658209. The commit message still says 0.27, but I can live with it

dr-orlovsky:
ouh, sorry for forgetting about that message :(

sanket1729:
Tagged and published.

PR #339, opened 2022-04-07
Add PsbtInputExt::update_with_descriptor

### LINKED ISSUE #335: PsbtEx::update_desc include derivation paths for bip32 keys
I've been looking at `7.0.0-rc.2` and was seeing if we could replace all the PSBT population code in BDK with the new shiny stuff here. Some of the fields we need that don't seem to be filled in by `PsbtExt::update_desc` are `bip32_derivation` and `tap_key_origins`. It seems like we should be able to populate these at this point. For example, for `bip32_derivation` we can iterate through each key in the descriptor and if it's an `XPub` derive it and insert into the map. ~We would need to pass in a secp context for this to work.~ see EDIT

More than willing to do a PR on this but wanted to check this approach makes sense.

As a side comment: It looks to me like the `update_desc` function should extend `Psbt::Input` rather than putting it on the `Psbt` type and then having to pass in the index you want to use (the logic completely ignores the rest of the psbt).

EDIT: ok so it looks we don't need to pass in secp because we are already cheating inside by allocating a new one each time. I have no complaints about that but I think the API could just require you pass in a pre-derived descriptor (i.e. no-wildcards `Descriptor<DescriptorPublicKey>`). At least for BDK there is no advantage to having an API that can do the searching within a particular range for you to match against a PSBT input since you should usually know what index the scriptPubKey was derived at for any UTXO you have.

LLFourn:
Fixes #335 

This populates all relevant fields in a PSBT input according to a descriptor. Includes previously not yet implemented logic for populating `bip32_derivation` and `tap_key_origins`. Also optionally checks the `witness_utxo` and `non_witness_utxo`.

I renamed `PsbtExt::update_desc` to `PsbtExt::update_inp_with_descriptor` which calls `update_with_descriptor`
internally but retrieves the vout for the input for the `non_witness_utxo` check automatically. I removed the derivation range functionality because it didn't feel useful (and if it is then it probably belongs in a separate function). Let me know if you want me to add it back where it was or to provide an additional utility function to find which index a script pubkey was derived from.

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -31,6 +31,7 @@ use std::{
 };
 
 use bitcoin::blockdata::witness::Witness;
+use bitcoin::util::address::WitnessVersion;
 use bitcoin::{self, secp256k1, Script};
 
 use self::checksum::verify_checksum;
@@ -263,12 +264,14 @@ pub enum DescriptorType {
 impl DescriptorType {
     /// Returns the segwit version implied by the descriptor type.
     ///
-    /// This will return `Some(0)` whether it is "native" segwitv0 or "wrapped" p2sh segwit.
-    pub fn segwit_version(&self) -> Option<u8> {
+    /// This will return `Some(WitnessVersion::V0)` whether it is "native" segwitv0 or "wrapped" p2sh segwit.
+    pub fn segwit_version(&self) -> Option<WitnessVersion> {
         use self::DescriptorType::*;
         match self {
-            Tr => Some(1),
-            Wpkh | ShWpkh | Wsh | ShWsh | ShWshSortedMulti | WshSortedMulti => Some(0),
+            Tr => Some(WitnessVersion::V1),
+            Wpkh | ShWpkh | Wsh | ShWsh | ShWshSortedMulti | WshSortedMulti => {
+                Some(WitnessVersion::V0)
+            }
             Bare | Sh | Pkh | ShSortedMulti => None,
         }
     }
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -771,9 +771,10 @@ impl Descriptor<DescriptorPublicKey> {
     /// Utility method for deriving the descriptor at each index in a range to find one matching
     /// `script_pubkey`.
     ///
-    /// If it finds a match then it returns the index it was derived it and the concrete descriptor
-    /// at that index. If the descriptor is non-derivable then it will simply check the script
-    /// pubkey against the descriptor (and in that case the index returned will be meaningless).
+    /// If it finds a match then it returns the index it was derived at and the concrete
+    /// descriptor at that index. If the descriptor is non-derivable then it will simply check the
+    /// script pubkey against the descriptor and return it if it matches (in this case the index
+    /// returned will be meaningless).
     pub fn find_derivation_index_for_spk<C: secp256k1::Verification>(
         &self,
         secp: &secp256k1::Secp256k1<C>,
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -260,6 +260,19 @@ pub enum DescriptorType {
     Tr,
 }
 
+impl DescriptorType {
+    /// Whether this is a segwit descriptor.
+    ///
+    /// Returns true whether it is "native" segwit or "wrapped" p2sh segwit
+    pub fn is_segwit(&self) -> bool {
+        use self::DescriptorType::*;
+        match self {
+            Wpkh | ShWpkh | Wsh | ShWsh | ShWshSortedMulti | WshSortedMulti | Tr => true,
+            Bare | Sh | Pkh | ShSortedMulti => false,
+        }
+    }
+}
+
 impl<Pk: MiniscriptKey> Descriptor<Pk> {
     // Keys
 
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -261,14 +261,15 @@ pub enum DescriptorType {
 }
 
 impl DescriptorType {
-    /// Whether this is a segwit descriptor.
+    /// Returns the segwit version implied by the descriptor type.
     ///
-    /// Returns true whether it is "native" segwit or "wrapped" p2sh segwit
-    pub fn is_segwit(&self) -> bool {
+    /// This will return `Some(0)` whether it is "native" segwitv0 or "wrapped" p2sh segwit.
+    pub fn segwit_version(&self) -> Option<u8> {
         use self::DescriptorType::*;
         match self {
-            Wpkh | ShWpkh | Wsh | ShWsh | ShWshSortedMulti | WshSortedMulti | Tr => true,
-            Bare | Sh | Pkh | ShSortedMulti => false,
+            Tr => Some(1),
+            Wpkh | ShWpkh | Wsh | ShWsh | ShWshSortedMulti | WshSortedMulti => Some(0),
+            Bare | Sh | Pkh | ShSortedMulti => None,
         }
     }
 }
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -23,6 +23,7 @@
 //! these with BIP32 paths, pay-to-contract instructions, etc.
 //!
 
+use std::ops::Range;
 use std::{collections::HashMap, sync::Arc};
 use std::{
     fmt,
@@ -748,6 +749,30 @@ impl Descriptor<DescriptorPublicKey> {
             .expect("Translation to string cannot fail");
 
         descriptor.to_string()
+    }
+
+    /// Utility method for deriving the descriptor at each index in a range to find one matching
+    /// `script_pubkey`.
+    ///
+    /// If it finds a match then it returns the index it was derived it and the concrete descriptor
+    /// at that index. If the descriptor is non-derivable then it will simply check the script
+    /// pubkey against the descriptor (and in that case the index returned will be meaningless).
+    pub fn find_derivation_index_for_spk<C: secp256k1::Verification>(
+        &self,
+        secp: &secp256k1::Secp256k1<C>,
+        script_pubkey: &Script,
+        range: Range<u32>,
+    ) -> Result<Option<(u32, Descriptor<bitcoin::PublicKey>)>, ConversionError> {
+        let range = if self.is_deriveable() { range } else { 0..1 };
+
+        for i in range {
+            let concrete = self.derived_descriptor(&secp, i)?;
+            if &concrete.script_pubkey() == script_pubkey {
+                return Ok(Some((i, concrete)));
+            }
+        }
+
+        Ok(None)
     }
 }
 
@@ -1735,5 +1760,32 @@ pk(03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8))";
         Descriptor::<DescriptorPublicKey>::from_str(&format!("wsh(pk({}))", comp_key)).unwrap();
         Descriptor::<DescriptorPublicKey>::from_str(&format!("wsh(pk({}))", x_only_key))
             .unwrap_err();
+    }
+
+    #[test]
+    fn test_find_derivation_index_for_spk() {
+        let secp = secp256k1::Secp256k1::verification_only();
+        let descriptor = Descriptor::from_str("tr([73c5da0a/86'/0'/0']xpub6BgBgsespWvERF3LHQu6CnqdvfEvtMcQjYrcRzx53QJjSxarj2afYWcLteoGVky7D3UKDP9QyrLprQ3VCECoY49yfdDEHGCtMMj92pReUsQ/0/*)").unwrap();
+        let script_at_0_1 = Script::from_str(
+            "5120a82f29944d65b86ae6b5e5cc75e294ead6c59391a1edc5e016e3498c67fc7bbb",
+        )
+        .unwrap();
+        let expected_concrete = Descriptor::from_str(
+            "tr(0283dfe85a3151d2517290da461fe2815591ef69f2b18a2ce63f01697a8b313145)",
+        )
+        .unwrap();
+
+        assert_eq!(
+            descriptor.find_derivation_index_for_spk(&secp, &script_at_0_1, 0..1),
+            Ok(None)
+        );
+        assert_eq!(
+            descriptor.find_derivation_index_for_spk(&secp, &script_at_0_1, 0..2),
+            Ok(Some((1, expected_concrete.clone())))
+        );
+        assert_eq!(
+            descriptor.find_derivation_index_for_spk(&secp, &script_at_0_1, 0..10),
+            Ok(Some((1, expected_concrete)))
+        );
     }
 }
--- a/src/psbt/mod.rs
+++ b/src/psbt/mod.rs
@@ -731,14 +731,14 @@ impl PsbtExt for Psbt {
         let expected_spk = {
             match (&input.witness_utxo, &input.non_witness_utxo) {
                 (Some(witness_utxo), None) => {
-                    if desc_type.is_segwit() {
+                    if desc_type.segwit_version().is_some() {
                         witness_utxo.script_pubkey.clone()
                     } else {
                         return Err(UtxoUpdateError::UtxoCheck);
                     }
                 }
                 (None, Some(non_witness_utxo)) => {
-                    if desc_type.is_segwit() {
+                    if desc_type.segwit_version().is_some() {
                         return Err(UtxoUpdateError::UtxoCheck);
                     }
 
--- a/src/psbt/mod.rs
+++ b/src/psbt/mod.rs
@@ -20,12 +20,12 @@
 //!
 
 use std::collections::BTreeMap;
-use std::ops::{Deref, Range};
+use std::ops::Deref;
 use std::{error, fmt};
 
 use bitcoin::hashes::{hash160, ripemd160, sha256, sha256d};
 use bitcoin::secp256k1::{self, Secp256k1};
-use bitcoin::util::psbt::PartiallySignedTransaction as Psbt;
+use bitcoin::util::psbt::{self, PartiallySignedTransaction as Psbt};
 use bitcoin::util::sighash::SighashCache;
 use bitcoin::{self, SchnorrSighashType};
 use bitcoin::{EcdsaSighashType, Script};
@@ -33,11 +33,15 @@ use bitcoin::{EcdsaSighashType, Script};
 use bitcoin::util::taproot::{self, ControlBlock, LeafVersion, TapLeafHash};
 use descriptor;
 use interpreter;
+use miniscript::iter::PkPkh;
 use miniscript::limits::SEQUENCE_LOCKTIME_DISABLE_FLAG;
 use miniscript::satisfy::{After, Older};
 use Preimage32;
 use Satisfier;
-use {Descriptor, DescriptorPublicKey, DescriptorTrait, MiniscriptKey, ToPublicKey};
+use {
+    Descriptor, DescriptorPublicKey, DescriptorTrait, MiniscriptKey, ToPublicKey, TranslatePk,
+    TranslatePk2,
+};
 
 mod finalizer;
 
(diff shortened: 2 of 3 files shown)


sanket1729:
> I removed the derivation range functionality because it didn't feel useful (and if it is then it probably belongs in a separate function)

Sorry for the delayed review. I was busy with the conference in Miami. I was trying to emulate bitcoin core's `utxoupdatepsbt` RPC. I think having the range in the descriptor is certainly useful to scan the output because users don't have to derive the output then. It might be useful in the case where the creator(the entity creating the psbt) is separate from the updater (the entity updating the psbt with utxo information). 

But I do think that you are right that this functionality belongs in a separate method. Perhaps it should be a method to `Descriptor<DescriptorPublicKey>` instead of `Psbt`.

src/psbt/mod.rs
```diff
@@ -861,6 +800,160 @@ impl PsbtExt for Psbt {
...
+                let spend_info = tr.spend_info();
+                for (_depth, ms) in tr.iter_scripts() {
+                    let leaf_script = (ms.encode(), LeafVersion::TapScript);
+                    let tapleaf_hash = TapLeafHash::from_script(&leaf_script.0, leaf_script.1);
+                    let control_block = spend_info
+                        .control_block(&leaf_script)
+                        .expect("Control block must exist in script map for every known leaf");
+                    inp.tap_scripts.insert(control_block, leaf_script);
+
+                    for pk in ms.iter_pk() {
```
sanket1729, line 886: This should either be `iter_pk_pkh` or `translate_pk2`. When we iterate `iter_pk`, we are not iterating the `pkh` fragments. 

Aside: Maybe I will PR later to rename `iter_pk` to `iter_only_pk` to highlight this

src/psbt/mod.rs
```diff
@@ -861,6 +800,160 @@ impl PsbtExt for Psbt {
...
+                None => {
+                    match derived {
+                        Descriptor::Tr(_) => { /*Tr outputs can always omit non_witness_utxo*/ }
+                        _ => return Err(UpdateInputError::NonWitnessUtxoCheck),
+                    }
+                }
+            }
+        }
+        // only set the input once all the checks have passed
+        *self = inp;
```
sanket1729, line 975: Can you please make the above into a separate function that only operates on `inp`. 

```
    fn update_with_descriptor(
        &mut self,
        ...
        )  -> ...
 {
 	let mut inp = self.clone();
 	update_with_descriptor_helper(&mut inp, ...);
 	*self = inp;
 }
```
This way we can be sure that we don't accidentally mutate `self` in the function body in the future. 
 
We should also avoid the allocation here, but I can do that might be complicated code-wise. I will perhaps address that in future as a low-priority issue.

src/psbt/mod.rs
```diff
@@ -871,54 +964,64 @@ fn script_code_wpkh(script: &Script) -> Script {
...
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
+/// Return error type of [`PsbtInputExt::update_with_descriptor`]
+pub enum UpdateInputError {
+    /// Derivation error
+    DerivationError(descriptor::ConversionError),
+    /// `witness_utxo` check failed
+    WitnessUtxoCheck,
+    /// `non_witness_utxo` check failed
+    NonWitnessUtxoCheck,
```
sanket1729, line 999: One of the `MissingUtxoError` is because it is not clear which field should be used for segwit transactions and how we should error it. 

The current implementation is finalizer is as follows:
```rust
// Get the spending utxo for this psbt input
pub(super) fn get_utxo(psbt: &Psbt, index: usize) -> Result<&bitcoin::TxOut, InputError> {
    let inp = &psbt.inputs[index];
    let utxo = if let Some(ref witness_utxo) = inp.witness_utxo {
        &witness_utxo
    } else if let Some(ref non_witness_utxo) = inp.non_witness_utxo {
        let vout = psbt.unsigned_tx.input[index].previous_output.vout;
        &non_witness_utxo.output[vout as usize]
    } else {
        return Err(InputError::MissingUtxo);
    };
    Ok(utxo)
}
```

src/psbt/mod.rs
```diff
@@ -806,12 +819,22 @@ pub trait PsbtInputExt {
...
-    /// `check_non_witness_utxo` respectively. You should set `check_non_witness_utxo` to inputs the
-    /// output index in the transaction being spent.
+    /// `check_non_witness_utxo` respectively. You should set `check_non_witness_utxo` to input's
+    /// output index in the transaction being spent. Note that `witness_utxo` or
+    /// `check_non_witness_utxo` being absent also counts as a check failure unless the field is
+    /// strictly not relevent to the descriptor. The two cases where the absence of a field will be
+    /// ignored are `witness_utxo` missing on a non-segwit descriptor and `non_witness_utxo` missing
+    /// on a taproot descriptor.
+    ///
+    /// Note that it may be suprising that `non_witness_utxo` would ever be required on a segwit
```
sanket1729, line 834: /s/suprising/surprising

sanket1729 (commented):
Did an initial review. Do you need this for the 7.0.0 release of rust-miniscript? I think this might take some time to get iterate.

I would prefer to get a release out for 7.0.0 out as soon as rust-bitcoin 0.28.0 is out. If rust-bitcoin is delayed even further maybe we can get this in 7.0.0 too.

LLFourn:
> > I removed the derivation range functionality because it didn't feel useful (and if it is then it probably belongs in a separate function)
> 
> Sorry for the delayed review. I was busy with the conference in Miami. I was trying to emulate bitcoin core's `utxoupdatepsbt` RPC. I think having the range in the descriptor is certainly useful to scan the output because users don't have to derive the output then. It might be useful in the case where the creator(the entity creating the psbt) is separate from the updater (the entity updating the psbt with utxo information).

On the contrary, that was very fast considering. Thanks.

> 
> But I do think that you are right that this functionality belongs in a separate method. Perhaps it should be a method to `Descriptor<DescriptorPublicKey>` instead of `Psbt`.

Will do. In general couldn't these update PSBT functions go on `Descriptor`? I was guessing they were on an extension trait because you wanted to strictly keep anything to do with PSBTs in one module. Perhaps we could just add additional impls on `Descriptor<DescriptorPublicKey>` in `psbt/mod.rs` (or maybe `psbt/descriptor_ext.rs`)?

> Did an initial review. Do you need this for the 7.0.0 release of rust-miniscript? I think this might take some time to get iterate.
>
> I would prefer to get a release out for 7.0.0 out as soon as rust-bitcoin 0.28.0 is out. If rust-bitcoin is delayed even further maybe we can get this in 7.0.0 too.

I think we will only be blocked on this by sometime in mid-may so I could wait for a follow up release around then. I'll try keep this PR as ready as ready as I can though in case we can sneak it in to 7.0.0 :)

Working on it!

src/descriptor/mod.rs
```diff
@@ -259,6 +259,19 @@ pub enum DescriptorType {
...
 
+impl DescriptorType {
+    /// Whether this is a segwit descriptor.
+    ///
+    /// Returns true whether it is "native" segwit or "wrapped" p2sh segwit
+    pub fn is_segwit(&self) -> bool {
+        use self::DescriptorType::*;
+        match self {
+            Wpkh | ShWpkh | Wsh | ShWsh | ShWshSortedMulti | WshSortedMulti | Tr => true,
+            _ => false,
```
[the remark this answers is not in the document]
LLFourn: Hmm good call. I wonder what the semantics should be if this is `Bare`. In the way I'm using it internally it means whether we `witness_utxo` should be there. I guess with `Bare` you just have to manually set these fields appropriately?

src/psbt/mod.rs
```diff
@@ -861,6 +800,160 @@ impl PsbtExt for Psbt {
...
+pub trait PsbtInputExt {
+    /// Given the decriptor for utxo being spent, populate the input with all the required fields to
+    /// spend it.
+    ///
+    /// If the descriptor contains wilcards or otherwise cannot be transformed into a concrete
+    /// descriptor an error will returned.
+    ///
+    /// You can make sure the descriptor is consistent with the `witness_utxo` and/or
+    /// `non_witness_utxo` fields of the PSBT by setting `check_witness_utxo` and
+    /// `check_non_witness_utxo` respectively. You should set `check_non_witness_utxo` to inputs the
```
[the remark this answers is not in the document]
LLFourn: > Is there any special reason why you want unchecked APIs? I don't think there is significant efficiency gain here.
>
> Normally, you can use the prevout information in the current transaction to find this `vout`. I think we should mandate the user to have this field instead of an `Option<u32>` one and always do the check.

The optionality of the checks here are not about efficiency but rather about being flexible with respect to the interpretation of BIP174.

Do you want to make it mandatory to have `non_witness_utxo` for segwitv0 outputs? i.e. Do we want to stop people being able to voluntarily expose themselves to the segwit bug? In the existing implementation of `update_desc` this check is optional. It is simply only turned on if `non_witness_utxo` is present and `witness_utxo` is omitted (note that this is not quite right for other reasons -- you could set `witness_utxo` for a non-segwit output and it would pass the check here). To keep this "feature" it has to be an optional check. FWIW I would be fine with dropping this misfeature. BDK currently exposes this feature too (but has the check on by default) but I think we could remove it. (cc @afilini)

I want the previous index `u32` to be a param rather than extracting it from the PSBT because I want a function that goes from `UTXO-data -> psbt::Input` without having to construct a PSBT to do that. Does that make sense? As you mention in the future this will work perfectly with `PsbtV2`. So I propose to keep the functionality on `psbt::Input` (which requires the `u32` for now) and then just drop the argument when moving to v2 (or change it to a bool depending on how the discussion relating to the previous paragraph turns out). 

So finally, what about at least making the `witness_utxo` check mandatory? I think this makes sense as long as there are no applications that simply omit `witness_utxo` and only include `non_witness_utxo`. [BIP174] does not say that `witness_utxo` is mandatory if you have already included `non_witness_utxo`. It says you can have both but doesn't say that. Do we want to be opinionated here and say you always have to include `witness_utxo` if it's segwit?

Having considered the above I think there is a bug here where you can omit `non_witness_utxo` if it's taproot without providing `witness_utxo`. I'll fix that now.

[BIP174]: https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki

src/psbt/mod.rs
LLFourn: Hmm ok this will require me to index the data by the hash of the key rather than the xonly key. Is it really possible to have a pubkey hash of an xonly key in a tapscript :weary:?

src/psbt/mod.rs
LLFourn: The lack of clarity about "which field should be used" is why I made a conscious design decision to avoid using the above logic and I move to a API where the caller is saying which fields should be checked against the descriptor within the context of the application. So the caller asserts that there should be `witness_utxo` and/or `non_witness_utxo` and the function checks that it is present and correct (at least when it is plausibly appropriate).

For example, using `get_utxo` the function would not check that `non_witness_utxo` is present even if the application requires it and even if it was present would not check that it has the correct `scriptPubkey`. Coldcard, for example, requires that `non_witness_utxo` is present correct even if `witness_utxo` is provided.

You could argue that it is not the responsibility of a function that updates the `psbt::Input` with stuff from a descriptor to enforce these policies but I think it's just way to convenient to do it here and creates a single PSBT input state transition where you can guarantee your input is correct after you call this function on it.

src/psbt/mod.rs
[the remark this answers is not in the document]
sanket1729: Good point about the segwit change calculation bug. I don't think that having this field in psbt helps prevent this bug? We also need to check the txid calculation for guard against it. Also, this is only really useful for HW wallets use-case. 

Code sample from BIP 174. 
```
    if non_witness_utxo.exists:
        assert(sha256d(non_witness_utxo) == psbt.tx.input[i].prevout.hash)
```

There are perhaps more gotchas that users will have to be aware of while signing transactions of inputs that they don't have all inputs. I think there are also other attacks where HW wallets need to be careful while signing only a partial set of inputs of a psbt that we cannot prevent. 

Note that only HW wallets might need to have this check as they cannot confirm the amounts against the blockchain. Wallets that can watch blockchain don't need this field. Secondly, this field is super big and we should avoid it if it is not necessary. How about we make two versions `update_with_descriptor` and `update_with_descriptor_unchecked`? In the former, we can mandate `non_witness_utxo` is correct. 

> I want the previous index u32 to be a param rather than extracting it from the PSBT because I want a function that goes from UTXO-data -> psbt::Input without having to construct a PSBT to do that. Does that make sense?

Indeed. I think we are on the same page.

src/psbt/mod.rs
sanket1729: The Hash is the same type as of the Key. For `DescriptorPublicKey`, `Pk == Pk::Hash`

sanket1729 (commented):
Addressed some comments

src/psbt/mod.rs
LLFourn: This is `bitcoin::Publickey` here though. We're working with the the concrete one because we need it to compute the tapleaf hash with the script pubkey. My clunky plan would be to:

1. When deriving `DecriptorPublicKey` create an index from pubkey hash to `(xonly, fingerprint,  derivation path)`
2. When iterating over pubkey hashes in each left read the index and insert what I find in the index into `tap_leaf_origins` (if it hasn't been inserted yet) and add the leaf.

Quite awkward but I'm not sure there's another good way to do.

src/psbt/mod.rs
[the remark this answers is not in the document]
LLFourn: Yes you're right we should check the prevout matches the txid otherwise we are not guaranteeing anything much. It still helps to make sure it's there so the hardware wallet can check it but perhaps this code will make it onto a hardware wallet someday :)

I think the `checked` and `unchecked` idea is good. I'll try making  the method on the PSBT itself the fully checked version since it can check the TXID against the prevout and the one directly on the `psbt::Input` the unchecked version? This can be revisited once we have psbtv2.

LLFourn:
Made it a draft while it's not ready for merge. I've changed the API so there's a checked and unchecked version of the update input method. The checked version is on the PSBT (where the full data is there so we can do the `non_witness_utxo` txid check) and the unchecked version is on the input. It think it's better :)

What remains:
1. Fix `pkh` keys not being included in tapscripts (test that fails has been done)
2. Create new separate test for the checked version of the function.
3. separate the checked version into helper as mentioned here: https://github.com/rust-bitcoin/rust-miniscript/pull/339#discussion_r846838184

As a heads up I don't think I'm going to be able to work on this again until Friday.

src/psbt/mod.rs
sanket1729: I think you can have both things together in the for loop like

```rust
for ((_, ms_derived), (_, ms_underived) in tr.iter_scripts().zip(underived_tr.iter_scripts)
```

sanket1729: Iterator is guaranteed to be consistent to traverse the leaves in the same order

LLFourn: ah ok sure. I was thinking about that but wasn't sure I wanted to go rely on them iterating in the same order.. but I guess as long as we don't use `std` hashmaps for anything it should be fine!

sanket1729: Yeah, they are guaranteed to output leaves in the dfs walk order.

src/psbt/mod.rs
LLFourn: I removed the allocation with a helper function that both the checked and unchecked version use in: 8c8f82b895e94ec0cbb4a5eeb0ab15b94be64ead

The idea is that the helper optionally checks the SPK is correct before it mutates anything.

LLFourn:
This is almost ready it just needs some tests on the checked version of the function. I've just moved logic to a helper function and removed the clone of the PSBT. I also stop requiring that `non_witness_utxo` is present for segwitv0 descriptors since this is easy to enforce outside. Here's what's changed so far for `CHANGELOG`:

1. Rename `update_desc` to `update_input_with_descriptor` which additionally populates `bip32_derivation` and `tap_key_origins` and more thoroughly checks the validity of `witness_utxo` and `non_witness_utxo`.
2. Add `PsbtInputExt::update_with_descriptor_unchecked` which operates on an `psbt::Input` the above but without the checks.
3. Add `is_segwit` to `DescriptorType`

afilini:
I was quickly skimming through the code and noticed:

> The `descriptor` **must not have any wildcards** in it otherwise an error will be returned however it can (and should) have extended keys in it.

I wonder if I should prioritize my work upstreaming the equivalent of the `DerivedDescriptorKey` type that we have in BDK, which would maybe help the users make less mistakes by enforcing rules with the type system.

Sidenote: I love all these new traits and various bits of utilities being exposed in miniscript. We currently have our own in BDK, but I'll be very happy to delete some code and start using yours!

src/descriptor/mod.rs
```diff
@@ -259,6 +260,19 @@ pub enum DescriptorType {
     Tr,
 }
 
+impl DescriptorType {
```
afilini, line 263: I would also add a `is_taproot()` method, even though it will be unused here in miniscript. I'm sure somebody will find that useful!

LLFourn: Maybe `segwit_version(&self) -> Option<u8>` would be the most general?

LLFourn:
hey @afilini. Thanks for taking a look.

> I was quickly skimming through the code and noticed:
> 
> > The `descriptor` **must not have any wildcards** in it otherwise an error will be returned however it can (and should) have extended keys in it.
> 
> I wonder if I should prioritize my work upstreaming the equivalent of the `DerivedDescriptorKey` type that we have in BDK, which would maybe help the users make less mistakes by enforcing rules with the type system.

Indeed it would be good to have some type safety here for keys that are wildcard free (but might require further derivation).

> 
> Sidenote: I love all these new traits and various bits of utilities being exposed in miniscript. We currently have our own in BDK, but I'll be very happy to delete some code and start using yours!

FWIW the context of this PR is that I was porting your work from BDK to my sketch of `bdk_core` and saw that this lib was offering some of the same functionality already but just needed to set the key origin stuff. I think it would great if it could all be maintained here.

sanket1729 (commented):
concept ACK and code review ACK. Can you please squash this into cleaner commits?

src/psbt/mod.rs
```diff
@@ -894,117 +897,137 @@ impl PsbtInputExt for psbt::Input {
...
+            bip32_derivation.borrow_mut().insert(
+                derived.to_public_key().inner,
+                (xpk.master_fingerprint(), xpk.full_derivation_path()),
+            );
+            Ok(derived)
+        })?;
+
+        if let Some(check_script) = check_script {
+            if check_script != derived.script_pubkey() {
+                return Ok((derived, false));
```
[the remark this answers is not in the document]
LLFourn: I opted to not do `Option<...>` because it would mean an `unwrap` for one of the unchecked caller which seemed like the worse option. I added a comment explaining the awkwardness.

LLFourn:
I've just added a test for the checked version of the function and cleaned up some comments/naming. Commits have been squashed. There's one on the end which adds `segwit_version` instead of `is_segwit` which hasn't been reviewed yet.

Updated CHANGELOG

```
- Rename `update_desc` to `update_input_with_descriptor` which additionally populates `bip32_derivation` and `tap_key_origins` and more thoroughly checks the validity of `witness_utxo` and `non_witness_utxo`.
- Add `PsbtInputExt::update_with_descriptor_unchecked` which operates on an `psbt::Input` and does the above but without the checks.
- Add `find_derivation_index_for_spk` which allows you to scan a derivation range to find the one that produced a script pubkey (this feature was extracted and removed from `update_desc`).
- Add `segwit_version` to `DescriptorType`
```

Ready to go from my perspective :)

LLFourn:
Rebased on master.

LLFourn:
CI failed but I don't think it's because of the PR.

sanket1729:
Another rebase should resolve the CI issue. We just merged #363

LLFourn:
Rebased again :)

src/descriptor/mod.rs
```diff
@@ -261,14 +261,15 @@ pub enum DescriptorType {
 }
 
 impl DescriptorType {
-    /// Whether this is a segwit descriptor.
+    /// Returns the segwit version implied by the descriptor type.
     ///
-    /// Returns true whether it is "native" segwit or "wrapped" p2sh segwit
-    pub fn is_segwit(&self) -> bool {
+    /// This will return `Some(0)` whether it is "native" segwitv0 or "wrapped" p2sh segwit.
+    pub fn segwit_version(&self) -> Option<u8> {
```
sanket1729, line 267: nit: There is a `[WitnessVersion](https://docs.rs/bitcoin/0.28.0/bitcoin/util/address/enum.WitnessVersion.html)` struct upstream that we can use here. 

You need not fix this in the current PR as it is good to go from my end, I can open it up as a good first issue.

sanket1729 (dismissed):
ACK e85b99c385b718d18f2710d330bc0932527d418d. Left a small comment, that need not be addressed here. 

Will wait sometime for a review from @apoelstra

src/psbt/mod.rs
```diff
@@ -871,54 +1043,51 @@ fn script_code_wpkh(script: &Script) -> Script {
...
+impl error::Error for UtxoUpdateError {}
 
 /// Return error type for [`PsbtExt::sighash_msg`]
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
 pub enum SighashError {
     /// Index out of bounds
     IndexOutOfBounds(usize, usize),
     /// Missing input utxo
-    MissingInputUxto,
+    MissingInputUtxo,
```
tcharding, line 1090: Took me so long to see the change in this line :)

tcharding (commented):
My review adds very little of value, please do not implement suggestions unless you are rebasing anyways. I have next to zero knowledge of `rust-miniscript`.

src/psbt/mod.rs
```diff
@@ -861,6 +868,171 @@ impl PsbtExt for Psbt {
...
+    /// Given the descriptor for a utxo being spent populate the PSBT input's fields so it can be signed.
+    ///
+    /// If the descriptor contains wildcards or otherwise cannot be transformed into a concrete
+    /// descriptor an error will be returned. The descriptor *can* (and should) have extended keys in
+    /// it so PSBT fields like `bip32_derivation` and `tap_key_origins` can be populated.
+    ///
+    /// Note that his method doesn't check that the `witness_utxo` or `non_witness_utxo` is
+    /// consistent with the descriptor. To do that see [`update_input_with_descriptor`].
+    ///
+    /// ## Return value
```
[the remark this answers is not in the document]
LLFourn: I can't find this convention on that page?

src/descriptor/mod.rs
```diff
@@ -749,6 +764,30 @@ impl Descriptor<DescriptorPublicKey> {
 
         descriptor.to_string()
     }
+
+    /// Utility method for deriving the descriptor at each index in a range to find one matching
+    /// `script_pubkey`.
+    ///
+    /// If it finds a match then it returns the index it was derived it and the concrete descriptor
```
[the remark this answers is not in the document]
LLFourn: changed it to `at`

src/descriptor/mod.rs
LLFourn: I added a fixup commit for this.

LLFourn:
@sanket1729 @tcharding Thanks. I added two fixup commits to address the remarks. Seemed better to avoid doing more rebasing until @apoelstra has had a look. Unfortunately it looks like github still doesn't have a autosquash and merge feature so that'll have to be done manually once it's ready.

src/psbt/mod.rs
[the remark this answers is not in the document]
tcharding: The link uses shows the `# Examples` header not the `# Returns` header, I couldn't find a link with all the typical headers (Returns, Panics, Examplse ect.).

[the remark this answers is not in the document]
tcharding: I'm not at all fussed, I haven't even started shaking my docs clean up wand at `rust-miniscript` yet :)

src/descriptor/mod.rs
```diff
@@ -771,9 +771,10 @@ impl Descriptor<DescriptorPublicKey> {
     /// Utility method for deriving the descriptor at each index in a range to find one matching
     /// `script_pubkey`.
     ///
-    /// If it finds a match then it returns the index it was derived it and the concrete descriptor
-    /// at that index. If the descriptor is non-derivable then it will simply check the script
-    /// pubkey against the descriptor (and in that case the index returned will be meaningless).
+    /// If it finds a match then it returns the index it was derived at and the concrete
+    /// descriptor at that index. If the descriptor is non-derivable then it will simply check the
+    /// script pubkey against the descriptor and return it if it matches (in this case the index
+    /// returned will be meaningless).
```
tcharding, line 777: Cool, TIL we derive _at_ an index not _from_ an index. Cheers.

tcharding:
FTR my lack of ACK should not be seen to imply anything. I just don't know this code well enough to do so.

sanket1729 (approved):
utACk f40dc83108a56ee035aa92da111a9b1696847062

apoelstra (approved):
ACK e85b99c385b718d18f2710d330bc0932527d418d

I didn't check that the PSBT logic is exactly what it should be, but ACK this API change (I think it's much cleaner than what we had before) as well as the implementation.

sanket1729:
You ACKed a different tip, but I think we are good to merge this.

PR #340, opened 2022-04-07
SigType check for cost-analysis in SegwitV1 scripts

SarcasticNastik:
Previously, the struct `CompilerExtData` provided cost analysis for a given script considering the **SegwitV0** standards. This PR aims to track the type/ (context) of the script and provide the respective cost analysis.

### DIFF
--- a/src/miniscript/types/correctness.rs
+++ b/src/miniscript/types/correctness.rs
@@ -15,6 +15,7 @@
 //! Correctness/Soundness type properties
 
 use super::{ErrorKind, Property};
+use crate::ScriptContext;
 
 /// Basic type representing where the fragment can go
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
@@ -142,7 +143,7 @@ impl Property for Correctness {
         }
     }
 
-    fn from_pk_k() -> Self {
+    fn from_pk_k<Ctx: ScriptContext>() -> Self {
         Correctness {
             base: Base::K,
             input: Input::OneNonZero,
@@ -151,7 +152,7 @@ impl Property for Correctness {
         }
     }
 
-    fn from_pk_h() -> Self {
+    fn from_pk_h<Ctx: ScriptContext>() -> Self {
         Correctness {
             base: Base::K,
             input: Input::AnyNonZero,
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -24,6 +24,7 @@ use std::marker::PhantomData;
 use std::sync::Arc;
 use std::{cmp, error, f64, fmt, hash, mem};
 
+use crate::miniscript::context::SigType;
 use crate::miniscript::limits::MAX_PUBKEYS_PER_MULTISIG;
 use crate::miniscript::types::{self, ErrorKind, ExtData, Property, Type};
 use crate::miniscript::ScriptContext;
@@ -165,19 +166,30 @@ impl Property for CompilerExtData {
         }
     }
 
-    fn from_pk_k() -> Self {
+    fn from_pk_k<Ctx: ScriptContext>() -> Self {
         CompilerExtData {
             branch_prob: None,
-            sat_cost: 73.0,
+            sat_cost: match Ctx::sig_type() {
+                SigType::Ecdsa => 73.0,
+                SigType::Schnorr => 1.0 /* <var_int> */ + 64.0 /* sig */ + 1.0, /* <sighash_type> */
+            },
             dissat_cost: Some(1.0),
         }
     }
 
-    fn from_pk_h() -> Self {
+    fn from_pk_h<Ctx: ScriptContext>() -> Self {
         CompilerExtData {
             branch_prob: None,
-            sat_cost: 73.0 + 34.0,
-            dissat_cost: Some(1.0 + 34.0),
+            sat_cost: match Ctx::sig_type() {
+                SigType::Ecdsa => 73.0 + 34.0,
+                SigType::Schnorr => 66.0 + 33.0,
+            },
+            dissat_cost: Some(
+                1.0 + match Ctx::sig_type() {
+                    SigType::Ecdsa => 34.0,
+                    SigType::Schnorr => 33.0,
+                },
+            ),
         }
     }
 
@@ -186,6 +198,14 @@ impl Property for CompilerExtData {
             branch_prob: None,
             sat_cost: 1.0 + 73.0 * k as f64,
             dissat_cost: Some(1.0 * (k + 1) as f64),
+        }
+    }
+
+    fn from_multi_a(k: usize, n: usize) -> Self {
+        CompilerExtData {
+            branch_prob: None,
+            sat_cost: 66.0 * k as f64 + (n - k) as f64,
+            dissat_cost: Some(n as f64), /* <w_n> ... <w_1> := 0x00 ... 0x00 (n times) */
         }
     }
 
@@ -1258,7 +1278,7 @@ mod tests {
         let compilation: DummyTapAstElemExt =
             best_t(&mut BTreeMap::new(), &policy, 1.0, None).unwrap();
 
-        assert_eq!(compilation.cost_1d(1.0, None), 88.0 + 74.109375);
+        assert_eq!(compilation.cost_1d(1.0, None), 87.0 + 67.0390625);
         assert_eq!(
             policy.lift().unwrap().sorted(),
             compilation.ms.lift().unwrap().sorted()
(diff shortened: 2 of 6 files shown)


src/policy/compiler.rs
```diff
@@ -164,38 +169,64 @@ impl Property for CompilerExtData {
...
 
     fn from_multi(k: usize, _n: usize) -> Self {
         CompilerExtData {
             branch_prob: None,
-            sat_cost: 1.0 + 73.0 * k as f64,
+            sat_cost: 1.0
+                + match Ctx::sig_type() {
+                    SigType::Ecdsa => 73.0 * k as f64,
+                    SigType::Schnorr => 64. * k as f64,
+                },
```
sanket1729, line 212: You would need a `from_multi_a` here. Add appropriate `unreachable!` depending on context

src/policy/compiler.rs
```diff
@@ -164,38 +169,64 @@ impl Property for CompilerExtData {
...
     }
 
     fn from_hash() -> Self {
         CompilerExtData {
             branch_prob: None,
-            sat_cost: 33.0,
-            dissat_cost: Some(33.0),
+            sat_cost: match Ctx::sig_type() {
+                SigType::Ecdsa => 33.0,
+                SigType::Schnorr => 32.,
```
sanket1729, line 223: This should be 33 because we also include the push_bytes in the cost.

src/policy/compiler.rs
```diff
@@ -164,38 +169,64 @@ impl Property for CompilerExtData {
...
     }
 
     fn from_pk_h() -> Self {
         CompilerExtData {
             branch_prob: None,
-            sat_cost: 73.0 + 34.0,
-            dissat_cost: Some(1.0 + 34.0),
+            sat_cost: match Ctx::sig_type() {
+                SigType::Ecdsa => 73.0 + 34.0,
+                SigType::Schnorr => 64. + 32.,
```
sanket1729, line 193: nit: We should be consistent in using `.` and `.0`

src/policy/compiler.rs
```diff
@@ -148,14 +149,18 @@ struct CompilerExtData {
     /// (total length of all witness pushes, plus their own length prefixes)
     /// for fragments that can be dissatisfied without failing the script.
     dissat_cost: Option<f64>,
+    /// Signature Type for deciding the number of bytes required for [`CompilerExtData::sat_cost`]
+    /// and [`CompilerExtData::dissat_cost`] for the given script.
+    sig_type: PhantomData<Ctx>,
 }
 
-impl Property for CompilerExtData {
+impl<Ctx: ScriptContext> Property for CompilerExtData<Ctx> {
```
sanket1729, line 157: Instead of introducing `Ctx` to `Property` and `CompilerExtData`, we should introduce it to `from_pk`, `from_pkh`, `from_multi`, and `from_multi_a` methods respectively. This should cleanup a lot of code here.

That is,
```diff
     /// Type property of the `PkK` fragment
-    fn from_pk_k() -> Self;
+    fn from_pk_k<Ctx: ScriptContext>() -> Self;
 
     /// Type property of the `PkH` fragment
-    fn from_pk_h() -> Self;
+    fn from_pk_h<Ctx: ScriptContext>() -> Self;
```

You would need to modify this for multi and multi_a too.

src/policy/compiler.rs
```diff
@@ -164,38 +169,64 @@ impl Property for CompilerExtData {
...
         }
     }
 
     fn from_pk_k() -> Self {
         CompilerExtData {
             branch_prob: None,
-            sat_cost: 73.0,
+            sat_cost: match Ctx::sig_type() {
+                SigType::Ecdsa => 73.,
+                SigType::Schnorr => 64.,
```
sanket1729, line 181: I think this should probably be 66 so that we include the `varint` prefix and the hash type flag. This is because the key itself is `64` bytes. It can optionally contain a sighash flag (1 byte) and there is also a pre-fix of `varint` type denoting the length of sig that follows. It will end up as  `<var_int_prefix><sig><sighash_type>` on the stack.

We should also add a comment explaining why this is 66 and we need to update this everywhere.

src/policy/compiler.rs
```diff
@@ -308,6 +351,16 @@ impl Property for CompilerExtData {
             branch_prob: None,
             sat_cost: left.sat_cost + right.sat_cost,
             dissat_cost: None,
+            sig_type: PhantomData,
+        })
+    }
+
+    fn and_n(a: Self, b: Self) -> Result<Self, types::ErrorKind> {
```
sanket1729, line 358: Note to self and reviewers: This is not new code added. This is moved.

src/policy/compiler.rs
```diff
@@ -470,8 +521,8 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> AstElemExt<Pk, Ctx> {
         r: &AstElemExt<Pk, Ctx>,
     ) -> Result<AstElemExt<Pk, Ctx>, types::Error<Pk, Ctx>> {
         let lookup_ext = |n| match n {
-            0 => Some(l.comp_ext_data),
-            1 => Some(r.comp_ext_data),
+            0 => Some(l.comp_ext_data.clone()),
```
sanket1729, line 524: Why do you need .clone() here? `CompExtData` is already `Copy`

sanket1729 (commented):
Let some comments. I think we can clean this up a lot if we move the `Ctx` as suggested. Please try that and let me know if there are any issues.

src/policy/compiler.rs
SarcasticNastik: Addressed in 2565f5e.

src/policy/compiler.rs
SarcasticNastik: Addressed in 2565f5e, went for `.0` .

src/policy/compiler.rs
SarcasticNastik: Addressed in 2565f5e

src/policy/compiler.rs
SarcasticNastik: In the previous version having `CompilerExtData<Ctx: ScriptKey>`, the closure signature changes to `FnOnce` since it moves the variable `l` out of its environment. [This SO post](https://stackoverflow.com/questions/31371027/copy-trait-and-phantomdata-should-this-really-move) answers for this issue.

2022 04 11-05 49 33 screenshot

src/policy/compiler.rs
SarcasticNastik: Addressed in 2565f5e

src/policy/compiler.rs
```diff
@@ -167,26 +168,41 @@ impl Property for CompilerExtData {
...
 
-    fn from_pk_h() -> Self {
+    fn from_pk_h<Ctx: ScriptContext>() -> Self {
         CompilerExtData {
             branch_prob: None,
-            sat_cost: 73.0 + 34.0,
-            dissat_cost: Some(1.0 + 34.0),
+            sat_cost: match Ctx::sig_type() {
+                SigType::Ecdsa => 73.0 + 34.0,
+                SigType::Schnorr => 66.0 + 32.0,
```
sanket1729, line 187: They key cost here and elsewhere would be 33 instead of 32

src/policy/compiler.rs
sanket1729: I think the implementation of `from_multi_a` is missing here for `CompilerExtData`. Note that the `multi_a` fragment operates differently from `multi` as it consumes empty signatures for dissatisfactions.

sanket1729 (commented):
Almost there

src/miniscript/types/correctness.rs
```diff
@@ -164,7 +165,7 @@ impl Property for Correctness {
         }
     }
 
-    fn from_multi(_: usize, _: usize) -> Self {
+    fn from_multi<Ctx: ScriptContext>(_: usize, _: usize) -> Self {
```
sanket1729, line 168: Why do we need `ctx` here? multi should have SigType `Ecdsa` and `multi_a` should have `Schnorr`

sanket1729: I think this would also resolve the awkward things that you had to do later with the `0` value

SarcasticNastik: Initially I introduced this to keep a check on misuse of SigType, but that is indeed taken care of in the consensus checks. Will update that in the following commit!

sanket1729 (dismissed):
code review ACK 25560a3abc4ebadacf09202dc023b02857a5c92d

SarcasticNastik:
@sanket1729 I overlooked that you merged #346 and not this PR. My apolgoies! 
I've rebased this PR onto master.

SarcasticNastik:
@sanket1729 @tcharding The Integration test fails at this run here because of curl stopping abruptly which should not have been the case. Kindly take a look into that, thanks!
P.S. Test passes on my fork.

sanket1729 (approved):
ACK 52d66477274e9afa770fdb61f53ec8b13f68d750

PR #344, opened 2022-04-13
Enforce that x-only keys cannot be used in non-tr descriptors

sanket1729:
Reported by: @afilini . 

This was a copy-paste error on my side :(


afilini:
ACK for the change, but I was wondering: why aren't you just returning `true` there?

Unless something recently changed in master, I'm looking at the docs for rc.2 and I don't see how something other than a `XOnlyPublicKey` would end up there.

sanket1729:
@afilini, fixed. Thanks for the suggestion

afilini (approved):
ACK 8212b9c8c1922a39425c0517d8c1ff8477b69f06

apoelstra (approved):
ACK 8212b9c8c1922a39425c0517d8c1ff8477b69f06

lgtm

PR #345, opened 2022-04-15
Define a new type for derived `DescriptorPublicKey`s

afilini:
After a brief chat on IRC and in PR #339, here's a quick draft for a new type specific to derived `DescriptorPublicKey`s.

This is the same design we use in BDK, where we wrap a normal `DescriptorPublicKey` and a secp context in a new struct. This has the drawback that the struct carries a lifetime and generic for the ctx, but I couldn't think of any other way to get this to work unless we drop the implementation of `ToPublicKey` for `DerivedDescriptorKey` which in my opinion can be very useful.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -2,10 +2,9 @@ use std::{error, fmt, str::FromStr};
 
 use bitcoin::{
     self,
-    hashes::Hash,
+    hashes::{hash160, Hash},
     hashes::{hex::FromHex, HashEngine},
-    secp256k1,
-    secp256k1::{Secp256k1, Signing},
+    secp256k1::{Secp256k1, Signing, Verification},
     util::bip32,
     XOnlyPublicKey, XpubIdentifier,
 };
@@ -68,6 +67,15 @@ pub enum SinglePubKey {
     FullKey(bitcoin::PublicKey),
     /// An xonly public key.
     XOnly(XOnlyPublicKey),
+}
+
+/// A derived [`DescriptorPublicKey`]
+///
+/// Derived keys are guaranteed to never contain wildcards
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
+pub struct DerivedDescriptorKey {
+    key: DescriptorPublicKey,
+    index: u32,
 }
 
 impl fmt::Display for DescriptorSecretKey {
@@ -437,11 +445,14 @@ impl DescriptorPublicKey {
     /// - If this key is an xpub but does not have a wildcard, returns `self`.
     /// - Otherwise, returns the derived xpub at `index` (removing the wildcard).
     ///
+    /// Since it's guaranteed that extended keys won't have wildcards, the key is returned as
+    /// [`DerivedDescriptorKey`].
+    ///
     /// # Panics
     ///
     /// If `index` ≥ 2^31
-    pub fn derive(self, index: u32) -> DescriptorPublicKey {
-        match self {
+    pub fn derive(self, index: u32) -> DerivedDescriptorKey {
+        let derived = match self {
             DescriptorPublicKey::Single(_) => self,
             DescriptorPublicKey::XPub(xpub) => {
                 let derivation_path = match xpub.wildcard {
@@ -460,7 +471,10 @@ impl DescriptorPublicKey {
                     wildcard: Wildcard::None,
                 })
             }
-        }
+        };
+
+        DerivedDescriptorKey::new(derived, index)
+            .expect("The key should not contain any wildcards at this point")
     }
 
     /// Computes the public key corresponding to this descriptor key.
@@ -475,7 +489,7 @@ impl DescriptorPublicKey {
     /// to avoid hardened derivation steps, start from a `DescriptorSecretKey`
     /// and call `to_public`, or call `TranslatePk2::translate_pk2` with
     /// some function which has access to secret key data.
-    pub fn derive_public_key<C: secp256k1::Verification>(
+    pub fn derive_public_key<C: Verification>(
         &self,
         secp: &Secp256k1<C>,
     ) -> Result<bitcoin::PublicKey, ConversionError> {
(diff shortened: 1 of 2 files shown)


src/descriptor/key.rs
```diff
@@ -717,6 +730,115 @@ impl MiniscriptKey for DescriptorPublicKey {
...
+
+    /// Return the derivation index of this key
+    pub fn get_index(&self) -> u32 {
+        self.index
+    }
+
+    /// Construct an instance from a descriptor key and a derivation index
+    ///
+    /// Panics if the key contains a wildcard
+    fn new(key: DescriptorPublicKey, index: u32, secp: &'secp Secp256k1<C>) -> Self {
```
sanket1729, line 752: Usually, we try to keep `secp` as the first parameter

src/descriptor/key.rs
```diff
@@ -717,6 +730,115 @@ impl MiniscriptKey for DescriptorPublicKey {
...
+        if let DescriptorPublicKey::XPub(ref xpk) = &key {
+            assert!(
```
sanket1729, line 754: I think we should return `Result` or `Option` here.

src/descriptor/key.rs
```diff
@@ -717,6 +730,115 @@ impl MiniscriptKey for DescriptorPublicKey {
...
+    }
+
+    fn to_pubkeyhash(&self) -> Self {
+        self.clone()
+    }
+}
+
+impl<'secp, C: Verification> ToPublicKey for DerivedDescriptorKey<'secp, C> {
+    fn to_public_key(&self) -> bitcoin::PublicKey {
+        self.key.derive_public_key(&self.secp).unwrap()
```
sanket1729, line 834: Let us not derive `ToPublicKey` at all. Users can also use the https://github.com/rust-bitcoin/rust-miniscript/blob/d8cc633b5b726e7ae595031ba04dcc1c1c2d6fda/src/descriptor/mod.rs#L684 to avoid the awkward `translate_pk2` APIs.

sanket1729 (commented):
concept ACK. I think we need to also revisit the APIs in `impl Descriptor<DescriptorPublicKey>` in `descriptor/mod.rs` to use this type.

sanket1729:
> ToPublicKey for DerivedDescriptorKey which in my opinion can be very useful.

Indeed this is super useful. After we have private key derivations, we can add support for this. But for now, this would have to do. 

Also, creating new verification contexts is now cheap(not exactly free still takes 50 micros seconds) as these contexts are now static in the underlying secp library. We are waiting on some cleanups from upstream secp after which creating verification contexts would be free!

src/descriptor/key.rs
```diff
@@ -58,6 +57,15 @@ pub enum DescriptorSecretKey {
     XPrv(DescriptorXKey<bip32::ExtendedPrivKey>),
 }
 
+/// A derived [`DescriptorPublicKey`]
+///
+/// Derived keys are guaranteed to never contain wildcards
+pub struct DerivedDescriptorKey<'secp, C: 'secp + Verification> {
```
LLFourn, line 63: I think a much better name here would be `DefiniteDescriptorKey`. "Derivation" is the process of going from an extended key to a child key not from going from a key with wildcards to a key at a particular index of a wildcard.

src/descriptor/key.rs
```diff
@@ -58,6 +57,15 @@ pub enum DescriptorSecretKey {
...
+    key: DescriptorPublicKey,
+    index: u32,
+    secp: &'secp Secp256k1<C>,
```
LLFourn, line 66: +1 for removing the `secp` since @sanket1729 says that it only takes 50 microseconds to create one and this will go away totally in the future (if I understand correctly).

sanket1729: Initially, I meant it as a remark and was not suggesting a change to the PR.  

This comes at a cost that the derivation function would be twice slow. Would that be an acceptable cost to bdk? Curious what @apoelstra thinks about this. I don't think twice slow is a big deal, but I am not sure how bad it is for wallet re-scanning

src/descriptor/key.rs
sanket1729: I am not thrilled with either names, but can't come up with a better name myself

src/descriptor/key.rs
afilini: I don't think it would be too bad, we actually have a few other places where we can gain performances by caching stuff and we've never bothered doing so.

I would also wait for Andrew's opinion though, he knows secp and how to use it properly :)

apoelstra: I like the idea of dropping the secp parameter. No opinion about whether this PR is the right time to do it. But we are moving upstream toward a model where verification contexts at least can be made basically for free.

src/descriptor/key.rs
LLFourn: I think [definite vs indefinite] is the correct way of distinguishing between this property linguistically. For example in English it shows up as:

1. Definite: I'm taking **my cat** to the vet
2. Indefinite: I'm taking **a cat** to the vet (we don't know which cat)

In miniscript you have:

1. Definite: A descriptor where all derivation indexes are known
2. Indefinite: A descriptor that could be instantiated at a particular index (like "a cat" could be instantiated with any particular cat but "my cat" could not).

It might be useful to apply this naming logic to other parts of the code. There is a function named `derive` that doesn't actually derive but just sets the derivation index. Then there is a function that actually does derivation called `derive_public_key`.

[definite vs indefinite]: https://en.wikipedia.org/wiki/Definiteness

src/descriptor/key.rs
LLFourn: I think it's worth removing here since it removes the lifetime from the type which can get in the way depending on how you do things.

src/descriptor/key.rs
afilini: `derived_descriptor()` is only available on `Descriptor<DescriptorPublicKey>`, so if I did a first `derive()` and now I have a `Descriptor<DerivedDescriptorKey>` then I can't use this directly and I have to go back to the "un-derived" descriptor.

Should I implement something similar for `Descriptor<DerivedDescriptorKey>`? Ideally with a different name, otherwise it can get confusing. Something that internally translates the keys

afilini:
Addressed most comments

src/descriptor/key.rs
tcharding: I was confused by the current function naming you describe recently, +1 for 'definite'.

LLFourn: I think renaming `derive` to `set_derivation_index` would probably be better than `make_definite` or something. But I do think definite and indefinite in the type names make sense.

dr-orlovsky: Maybe `Concrete`/`Abstract`?

afilini: I'll let you guys discuss this, let me know when you are settled with one and I'll update my PR.

I personally don't have any preference.

apoelstra:
Needs rebase, and does not compile currently I think because of rust-bitcoin updates. but ACK  e619be4efbca61ea7703b4ecec54f34df1bacfd8 -- this is much smaller/cleaner than I feared.

afilini:
Rebased and fixed everything!

apoelstra (dismissed):
Nice!
ACK 28615fd32a51ce77d71b9e453867102f61342e35

sanket1729 (approved):
ACK 28615fd32a51ce77d71b9e453867102f61342e35

src/descriptor/key.rs
```diff
@@ -713,6 +723,70 @@ impl MiniscriptKey for DescriptorPublicKey {
...
+    /// guaranteed to never contain one.
+    pub fn derive_public_key<C: Verification>(
+        &self,
+        secp: &Secp256k1<C>,
+    ) -> Result<bitcoin::PublicKey, ConversionError> {
+        self.key.derive_public_key(secp)
+    }
+
+    /// Return the derivation index of this key
+    pub fn get_index(&self) -> u32 {
```
sanket1729, line 741: nit: We usually don't use the `get` prefix for accessor methods.

sanket1729 (commented):
ACK 28615fd32a51ce77d71b9e453867102f61342e35. Was going to merge this, but this would need to be rebased because of conflicts. 
Might be good to address  the nit as well

afilini:
Rebased and fixed your last nit!

apoelstra (approved):
ACK 9cf065b7fafc41a0681374df8aa83baf0b9d3425

sanket1729 (approved):
ACK 9cf065b7fafc41a0681374df8aa83baf0b9d3425

PR #346, opened 2022-04-18
Rectify `from_multi_a` max satisfaction size

SarcasticNastik:
Cost incurred for a single Schnorr signature in the witness stack := `<var_int><64-byte signature><sig_type>` = 1 + 64 + 1 (as clarified by @sanket1729 in #340 ).

### DIFF
--- a/src/miniscript/types/extra_props.rs
+++ b/src/miniscript/types/extra_props.rs
@@ -249,7 +249,7 @@ impl Property for ExtData {
             ops_count_nsat: Some(n + 1),
             stack_elem_count_sat: Some(n),
             stack_elem_count_dissat: Some(n),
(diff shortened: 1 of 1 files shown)


sanket1729 (approved):
ACK 747876ed7104738e89de5d7574e66db66adc30e9

sanket1729:
This is a minor bug, that can go in a point release. Will affect fee cost estimation by 2 bytes per pubkey in multi_a.

sanket1729:
This breaks the CI tests, because it is based on 0.28.0-rc.2 . Now that we have 0.28.0 cargo automatically fetches it and causes other misc errors on the merge commit.

sanket1729:
Fix would be to re-open tihs after we get 7.0.0 release and rebase on top on that

PR #348, opened 2022-04-19
miniscript: the 'd:' wrapper must not be 'u'

darosior:
The value it leaves on the stack depends on the last element on the
stack. However, we can't make sure this element is OP_1 (which would
give us the 'u' property) without the MINIMALIF rule.
MINIMALIF is only policy for P2WSH, therefore giving 'd:' the 'u'
property breaks consensus soundness: it makes it possible (by consensus
but not policy) for instance to satisfy a thresh() without satisfying
at least k of its subs.

This bug was found and reported by Andrew Poelstra [based on a question from Aman Kumar Kashyap](https://github.com/rust-bitcoin/rust-miniscript/discussions/341).


darosior:
Related: https://github.com/bitcoin/bitcoin/pull/24906 and https://github.com/sipa/miniscript/pull/117

sanket1729:
I am working on fixing the tests and back porting fixes

darosior:
Oh thanks, i only ran the unit test locally and didn't check the CI.

apoelstra:
I believe this was superceded by #349

PR #349, opened 2022-04-19
Fix e/o bug in miniscript threshold correctness rules

sanket1729:
Multiple type system bugs:

The first two bugs are not severe: They relax rules so existing systems should not be affected. However the third commit is a fix that will be backported. 

Pasting description #348 

> The value it leaves on the stack depends on the last element on the
stack. However, we can't make sure this element is OP_1 (which would
give us the 'u' property) without the MINIMALIF rule.
MINIMALIF is only policy for P2WSH, therefore giving 'd:' the 'u'
property breaks consensus soundness: it makes it possible (by consensus
but not policy) for instance to satisfy a thresh() without satisfying
at least k of its subs.

This bug was found and reported by Andrew Poelstra https://github.com/rust-bitcoin/rust-miniscript/discussions/341.


apoelstra (approved):
ACK c74934bf0864a463c42c5a44502f803bf61eb57c

PR #351, opened 2022-04-19
Remove `u` property from `d`

sanket1729:


apoelstra (approved):
ACK 569344400708e89eac8151797c010a90efa179a9

PR #354, opened 2022-04-19
Remove `u` property from `d`

sanket1729:


sanket1729:
Need to raise this against 4.0.3

PR #355, opened 2022-04-19
Remove `u` property from `d`

sanket1729:


apoelstra (approved):
ACK 310ef06a582f5e00f51adfde3c4ba9f2e8f94755

PR #357, opened 2022-04-19
Remove `u` property from `d`

sanket1729:


apoelstra (approved):
ACK d54a74635ad5d2974600707d1678323e5f5d34ac

PR #363, opened 2022-04-20
Hacky Fix for int CI failure

sanket1729:
While we await rust-bitcoincore-rpc release, let's have this to make
sure that integrationt tests run and we get a nice green checkmark.
I really don't like seeing a red cross on all PRs :)


apoelstra (approved):
ACK fed892f3efceae08f53b9f4d055c887d5b484602

PR #366, opened 2022-04-22
Actually return the error

tcharding:
We have what appears to be an error return but do not actually return
the error. This line needs an explicit `return` statement otherwise it
is a noop.

### DIFF
--- a/src/descriptor/sortedmulti.rs
+++ b/src/descriptor/sortedmulti.rs
@@ -237,3 +237,34 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> fmt::Display for SortedMultiVec<Pk, 
         f.write_str(")")
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use bitcoin::secp256k1::PublicKey;
+    use miniscript::context::Legacy;
+
+    #[test]
+    fn too_many_pubkeys() {
+        // Arbitrary pubic key.
+        let pk = PublicKey::from_str(
+            "02e6642fd69bd211f93f7f1f36ca51a26a5290eb2dd1b0d8279a87bb0d480c8443",
+        )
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 4d0e8fb7692a49dd140fa34711ef53a9f31b357a

tcharding:
Damn this merged, I was thinking last night that this unit test shouldn't be merged. Its informational but not useful to maintain. We do not explicitly unit test every error path in the code so it seems odd to have this one. Not to worry.

apoelstra:
I would prefer if we did unit test every error path :)

src/descriptor/sortedmulti.rs
```diff
@@ -46,7 +46,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> SortedMultiVec<Pk, Ctx> {
     pub fn new(k: usize, pks: Vec<Pk>) -> Result<Self, Error> {
         // A sortedmulti() is only defined for <= 20 keys (it maps to CHECKMULTISIG)
         if pks.len() > MAX_PUBKEYS_PER_MULTISIG {
-            Error::BadDescriptor("Too many public keys".to_string());
```
sanket1729, line 49: post-merge comment: Does Clippy warn about this?

LLFourn: I think you just need to upgrade to 2018 edition. It's a standard warning now: [`#[must_use]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute)

tcharding: > Does Clippy warn about this?

ooo you got me, usually I always mention 'found by clippy' in the commit log but I forgot this time.

> I think you just need to upgrade to 2018 edition

The PR is open!

PR #374, opened 2022-04-28
Initiate configuration of rustfmt

tcharding:
We currently use `rustfmt` with the default configuration. There is some interest in using `rustfmt` in the crates lower down the stack but it is contentious and difficult to find a configuration, if one exists, that is 'good' for everyone. By configuring `rustfmt` in this crate we can reduce the scope of the discussion from 'should we even do this' down to 'this is a good formatting configuration', with the hope that if we can get a good configuration worked out in this crate it will help ease the debate in the others.

To get this started add a default config file and then set two, hopefully non-contentious, config options. 

Please note carefully; the first patch changes CI to run linting checks using the nightly toolchain.

Re: edition 2018 changes, there are some changes to imports in the `examples/` that appear in this PR that should really be removed altogether, leaving as is so as not to clutter this PR.


.github/workflows/rust.yml
```diff
@@ -24,29 +24,28 @@ jobs:
...
           profile: minimal
-          toolchain: ${{ matrix.rust }}
+          toolchain: nightly
           override: true
-      - name: Running cargo test
+      - name: Running bench
         env:
           DO_BENCH: true
         run: ./contrib/test.sh
+      - name: Running clippy
```
sanket1729, line 45: nit: It says running Clippy over here. I think we mean rustfmt?

sanket1729 (commented):
concept ACK. We can merge this right before 2018 bump

.github/workflows/rust.yml
tcharding: Ha! I started all this because the `contrib/test.sh` used `DO_LINT` to run rustfmt, which, as you point out is wrong - then proceeded a few days later to do the same mistake - that's funny.

tcharding:
Added initial patch to the PR that does the lint->fmt change.

tcharding:
Will require manual configuration of the CI pipeline in order to merge.

tcharding:
Rebased.

apoelstra (dismissed):
ACK 9c053e5a0f8b784ea923f8cf45cccd877fb400c6

Nice!

tcharding:
Rebased and re-did the second two patches i.e., for each, updated the config option, ran `cargo +nightly fmt` and used the same commit message.

apoelstra (approved):
ACK 8e02ed0078e21e1e0999e89ac77ceaf529ee9d91

sanket1729:
There are merge conflicts here. This would need a rebase again :(

tcharding:
Changes in force push:
- rebased on master, CI workflow had conflicts
- did formatting in last two patches as before

sanket1729:
>  Will require manual configuration of the CI pipeline in order to merge.

Can you elaborate? Do you mean this would need to be merged in god mode?

tcharding:
> > Will require manual configuration of the CI pipeline in order to merge.
> 
> Can you elaborate? Do you mean this would need to be merged in god mode?

I don't know what setting has to be changed, I don't have access to settings so I'm going off what was said on another PR/repo when I changed the CI workflow.  When a PR changes a workflow job the old one seems to still appear in the checks list on the web interface and the new one does not. Sorry I'm not very knowledgeable with GitHub.

sanket1729:
I think there is  a bug in current config?  See https://github.com/rust-bitcoin/rust-miniscript/actions/runs/2303431754


> I don't have access to settings so I'm going off what was said on another PR/repo when I changed the CI workflow

I can set the changes, but I cannot see the options for the fuzz-names in the CI workflow for this PR. Perhaps it is related to the above issue?

tcharding:
Well done, you were correct. Was missing a `run: ./contrib/test.sh` line. Thanks

sanket1729 (approved):
reACK 80620166e6372b0801850edcd6c103718467ebfe

sanket1729:
Verified the diff betwen 4847894 and 8062016.

PR #375, opened 2022-04-28
Use Key instead of MiniscriptKey

tcharding:
One of the main traits in this lib is `MiniscriptKey`, we can shorten this to `Key` with no loss of meaning. This makes the whole codebase more terse. Terse code, if clear, is easier to read because there is less clutter. Also terseness assists the formatter and can reduce lines of code.

This patch is the result of running

  `s/MiniscriptKey/Key/g` on all source files.

## Notes

This is an invasive change, I will not be offended if review is a simple 'no'. Outside the `miniscript` library users can still do `miniscript::Key` if the trait is too terse for their taste.

### DIFF
--- a/examples/sign_multisig.rs
+++ b/examples/sign_multisig.rs
@@ -35,7 +35,7 @@ fn main() {
     assert_eq!(descriptor.max_satisfaction_weight().unwrap(), 258);
 
     // Sometimes it is necessary to have additional information to get the
-    // `bitcoin::PublicKey` from the `MiniscriptKey` which can be supplied by
+    // `bitcoin::PublicKey` from the `Key` which can be supplied by
     // the `to_pk_ctx` parameter. For example, when calculating the script
     // pubkey of a descriptor with xpubs, the secp context and child information
     // maybe required.
--- a/src/descriptor/bare.rs
+++ b/src/descriptor/bare.rs
@@ -30,19 +30,18 @@ use crate::policy::{semantic, Liftable};
 use crate::prelude::*;
 use crate::util::{varint_len, witness_to_scriptsig};
 use crate::{
-    BareCtx, Error, ForEachKey, Miniscript, MiniscriptKey, Satisfier, ToPublicKey, TranslatePk,
-    Translator,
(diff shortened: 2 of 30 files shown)


sanket1729:
Over time, I have been liking shorter names. I see the pros and cons of this. I am okay with `MsKey`, will not oppose to `Key` if @apoelstra is okay with it.

> Outside the miniscript library users can still do miniscript::Key

Deriving `MiniscriptKey` is only for super advanced users. We don't expect regular users to it.

apoelstra:
I'm fine with `Key`, although because it needs to be imported (you have to use `use miniscript::Key` to get the trait methods, you can't use the fully-qualified name without a lot of noise) I worry that it may conflict with other key types.

On the other hand since edition 2018 you can do `use miniscript::Key as _` right?

tcharding:
Light bulb moment, what about we add this as well?
```
/// Export a type alias to make usage of the library more ergonomic since the
/// `Key` trait name is so generic that it may cause naming conflicts.
pub use Key as MiniscriptKey;
```

apoelstra:
oo, I like it! ok, ACK renaming to `Key`, along with this shim (which also helps with backward compatibility)

tcharding:
Changes in force push:
- Rebase on master
- Add the 'trait alias' as described above

tcharding:
In the same vein we could shorten `ScriptContext` -> `Context`. This would give us, e.g.,
```
impl<P: Key, C: Context> fmt::Display for Miniscript<P, C> {
```

apoelstra (approved):
ACK c3c595919e910223fe1c8ef904d26d425c8de7fe

This is gonna cause a lot of merge conflicts..

tcharding:
> This is gonna cause a lot of merge conflicts..

Yes. This PR can be re-done at anytime. It requires only a single command (a shell alias I have) to search-and-replace `MiniscriptKey` with `Key`. Please feel free to leave this queued up for a time that is suitable in relation to other open PRs. No rush whatsoever.

sanket1729:
Agreed, keeping this as the last PR to merge.

tcharding:
Making this draft because I do not know how to time it, it can be done mechanically at any stage if/when we feel its needed.

tcharding:
Too invasive.

PR #377, opened 2022-04-29
Re-name `as_public` -> `to_public`

tcharding:
As we have been doing in `rust-bitcoin` attempt to use idiomatic names for our conversion functions.

Re-name the key conversion functions `as_public` to be `to_public` because they take a borrowed type and return an owned, non-Copy type.

While we are it at do a bunch of refactoring to the individual functions. Each done as a separate patch, can drop any not wanted.


apoelstra (approved):
ACK be5e042beaa19a731f5d826a3907979988b41cd6

Note that this is an API-breaking change.

PR #378, opened 2022-04-30
Refactor opcode calculations

sanket1729:
- Instead of explicitly counting all opcodes required while satisfaction,
we explicitly track only additional opcodes while satisfaction and
dissatisfaction.

- Removes some unnecessary code around wrapper casting

Double-checked and compared with the c++ codebase


apoelstra (approved):
ACK ae0ce13eaa4a1c2f6a19ed1d5c291feaaf0c552c

PR #380, opened 2022-05-02
Use != 0 instead of == 1 for check of masked value

tcharding:
Recently [0] we added a check on a value masked against `SEQUENCE_LOCKTIME_DISABLE_FLAG`, this check uses `x == 1` but x can never be zero since `SEQUENCE_LOCKTIME_DISABLE_FLAG` is equal to `0x80000000`.

Looking at the original commit it looks like we are trying to check for a value _higher_ than the flag value, so we can mask then use `x != 0` to see if the flag bit is enabled.

Found by clippy:

 error: incompatible bit mask: `_ & 2147483648` can never be equal to `1`

CC: @darosior, have I understood this correctly man?

[0] commit: `commit a3dd9fe840c8cdae84e9d586cd9a9d6c195f0092`

### DIFF
--- a/src/miniscript/types/extra_props.rs
+++ b/src/miniscript/types/extra_props.rs
@@ -1037,7 +1037,7 @@ impl Property for ExtData {
                 // Note that for CLTV this is a limitation not of Bitcoin but Miniscript. The
                 // number on the stack would be a 5 bytes signed integer but Miniscript's B type
                 // only consumes 4 bytes from the stack.
-                if t == 0 || (t & SEQUENCE_LOCKTIME_DISABLE_FLAG) == 1 {
+                if t == 0 || (t & SEQUENCE_LOCKTIME_DISABLE_FLAG) != 0 {
                     return Err(Error {
                         fragment: fragment.clone(),
                         error: ErrorKind::InvalidTime,
@@ -1046,7 +1046,7 @@ impl Property for ExtData {
                 Ok(Self::from_after(t))
             }
             Terminal::Older(t) => {
-                if t == 0 || (t & SEQUENCE_LOCKTIME_DISABLE_FLAG) == 1 {
+                if t == 0 || (t & SEQUENCE_LOCKTIME_DISABLE_FLAG) != 0 {
                     return Err(Error {
                         fragment: fragment.clone(),
                         error: ErrorKind::InvalidTime,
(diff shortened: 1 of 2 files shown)


darosior:
Oh, right :facepalm:. Good catch. ACK b747e518f6061ee1281c0031154d5f82c3243b42.

sanket1729 (approved):
Wow. Thanks for catching this bug! ACK b747e518f6061ee1281c0031154d5f82c3243b42

sanket1729:
@apoelstra, does this make for trying out `clippy` here? I think maybe we can find some middle ground with clippy.

We should not enforce it on every CI run. But we can schedule CI runs every week Sunday night running the clippy CI on the latest tip? This way
- New contributors won't be annoyed by CI everytime. 
- Every Monday after we see the red cross on latest master tip, we can resolve and fix the issues. 

We can debate the schedule period if there is interest.

tcharding:
I am attempting to clear _all_ the clippy warnings here and everywhere in our stack, I'm doing this in an attempt to get clippy running on CI for ever PR - I see no reason not to, is there some reason I'm missing? From my understanding the problems were all related to Rust 1.29, that's why I've waited til now to do this work.

sanket1729:
> From my understanding the problems were all related to Rust 1.29, that's why I've waited until now to do this work.

FTR, I am in favor of it. But several contributors don't always like the suggestions Clippy presents. It is yet another thing to check while submitting the pull request. And I am sure that we will end up with lots of cases where the last commit is just "Fix up Clippy lints". 

In addition to every commit compiling tests, contributors need to run Clippy on every commit which people can forget and would get annoyed whenever the CI yells at low priority things are false positives. Then we need to figure out how to disable some lints that are not annoying etc. 

As I said, I am in favor, I am just stating the reasons why people might oppose it.

tcharding:
oh I forgot about how annoying it is going back over patch sets to fix clippy stuff. Thanks for reminding me, we definitely should find a solution that does not lead to that.

LLFourn:
Note one solution to this is to add a commit hook to git which fails to commit unless clippy passes. According [to this SO](https://stackoverflow.com/questions/3462955/putting-git-hooks-into-a-repository) you can commit a repo to rust-miniscript like: `.githooks` and get people to opt in to opt into the hooks with:

```
git config --local core.hooksPath .githooks/
```

tcharding:
Sweeeet, thanks man.

PR #386, opened 2022-05-04
Remove the `DescriptorTrait`

tcharding:
We never abstract over the `DescriptorTrait` so it is questionable if it should exist at all. Removing the `DescriptorTrait` allows us to also delete the `pretaproot` module because all the fallible vs infallible issues disolve.

Delete the `DescriptorTrait`, implement the trait methods directly on `Descriptor` (fallible methods). Implement infallible versions of the trait methods on each descriptor as required.

Done as individual patches to ease review. Although the work is only really meaningful as a block each patch is discreet enough to merge as is.


apoelstra:
Now that the trait methods are infallible I wonder if we should just remove the specialized ones.

tcharding:
That was one direction I thought would come out of review, I'll explore it.

tcharding:
This PR has been totally re-written, description and title have been changed to reflect the new PR content.

tcharding:
I was hoping to re-name the `DescriptorTrait`, the best I could come up with was`Satisfy` but the three methods `sanity_check`, `script_pubkey`, and `unsigned_script_sig` don't fit into that name :(

tcharding:
I was hoping to re-name the `DescriptorTrait`, the best I could come up with was`Satisfy` but the three methods `sanity_check`, `script_pubkey`, and `unsigned_script_sig` don't fit into that name :(

src/descriptor/bare.rs
```diff
@@ -359,7 +349,9 @@ impl<Pk: MiniscriptKey> DescriptorTrait<Pk> for Pkh<Pk> {
     where
         Pk: ToPublicKey,
     {
-        self.spk()
+        // TODO: This is broken, remove the hard coded network.
```
apoelstra, line 299: In a56fe857b0647318dacf6e386895ec9ca0fd6082:

I don't understand this comment. The network was hardcoded i the old code, and anyway it doesn't affect the output of this function at all.

src/descriptor/segwitv0.rs
```diff
@@ -472,7 +459,10 @@ impl<Pk: MiniscriptKey> DescriptorTrait<Pk> for Wpkh<Pk> {
     where
         Pk: ToPublicKey,
     {
-        self.spk()
+        // TODO: This is broken, remove the hard coded network.
```
apoelstra, line 405: In a56fe857b0647318dacf6e386895ec9ca0fd6082:

Same here.

apoelstra:
I don't know about this approach of pulling individual methods into infallible traits. It means that you can't write code which is generic over all descriptor types, and still be able to compute `Address`es, unless you use the specific `enum Descriptor` type that we provide. It means that other implementors of `DescriptorTrait` that don't also implement `ToAddress` and `ExplicitScript` are second-class.

It also requires author of generic code to provide several more explicit trait bounds.

On the other hand, going forward we expect users to mostly use Taproot in which case they can ignore the `ExplicitScript` trait, and we don't expect any new `ToAddress` descriptor types to pop up do we?

@sanket1729 what do you think?

tcharding:
> I don't know about this approach of pulling individual methods into infallible traits. It means that you can't write code which is generic over all descriptor types, and still be able to compute `Address`es, unless you use the specific `enum Descriptor` type that we provide. It means that other implementors of `DescriptorTrait` that don't also implement `ToAddress` and `ExplicitScript` are second-class.

hmmm, I'm not confident that I can mentally model consuming these generic APIs, I seem to be only thinking at the level of the code on the sceen in `rust-minscript`. @thomaseizinger has a pretty good brain for this sort of thing, do you have a few clock cycles to spare to weigh in on this please?

In a nutshell the [perceived] problem this PR is trying to solve is that pre-taproot and taproot descriptors are not being treated as equivalent as is apparent by the need for the `pretaproot` module. This is (in my opinion) just a code smell and the 'solution' I implemented just fell out while I was messing around.  

> It also requires author of generic code to provide several more explicit trait bounds.

I agree with you, this is definitely bad.

> On the other hand, going forward we expect users to mostly use Taproot in which case they can ignore the `ExplicitScript` trait, and we don't expect any new `ToAddress` descriptor types to pop up do we?

I'm out of my depth here.

> @sanket1729 what do you think?

Thanks in advance @sanket1729 :)

src/descriptor/bare.rs
tcharding: Oh, you are correct, I am totally wrong with this code comment. It should perhaps be:
```
        // Fine to hard code the address here because we immediately call
        // `script_pubkey` which does not use the `network` field of `Address`.
```

thomaseizinger (commented):
Thanks for tagging me @tcharding, I'll add my 2c :)

1. In general, smaller, focused traits are better because they encourage composability. i.e. I can write code that only abstracts over `ToAddress` and compose that with code that abstracts over `ToAddress + ScriptCode`. If I can only abstract over a single trait, the individual units of code tend to be bigger.
2. It is still possible to provide a general `DescriptorTrait` that forces to implement all other traits and provide a blanket impl to reduce noise in `where` clauses. I would consider this to be somewhat of an anti-pattern though.

I've skimmed the codebase and we don't seem to have a single piece of code that says:

```rust
fn generic_code<D: DescriptorTrait>(desc: D) { .. }
```

I'd like to challenge whether this trait should even exist if _we_ never abstract over it?

In my opinion:
- Trying to guess what other uses may want abstract over is a fallacy. They can create their own traits if they need to abstract over multiple descriptors.
- Aligning APIs is not enough to justify a trait. Several descriptors can have the same API without forcing the user to import a trait to use them.
- The main API seems to be the `Descriptor` enum. The functions on that one can just live within a regular `impl` block and delegate to all known variants. No trait needed for that.

thomaseizinger:
> * Aligning APIs is not enough to justify a trait. Several, concrete descriptors can have the same API without forcing the user to import a trait to use them. This should resolve the fallible vs. infallible dilemma.

On this: The problem the `pretaproot` module seems to aim to solve is that some users may (statically) know that they don't have a taproot descriptor and thus can get access to infallible APIs. There are multiple ways of solving this, but essentially what you want is known as "dependent types". Rust doesn't have those but you can emulate them to some degree.

Instead of splitting the descriptors into two enums, we can use another type parameter to encode in the type system, which APIs are available. For example (names subject to bike-shedding):

```rust
/// Marker type to encode that we only support descriptors pre-taproot.
enum PreTaprootOnly { }

/// Marker type to encode that we support all kinds of descriptors, including taproot.
enum AllKinds { }

struct Descriptor<Pk, TaprootMarker> {
	pub fn parse_pre_taproot<Pk>(desc: &str) -> Result<Descriptor<Pk, PreTaprootOnly>, Error> {}
	pub fn parse<Pk>(desc: &str) -> Result<Descriptor<Pk, AllKinds>, Error> {}
}
```

Within `parse_pre_taproot`, we would fail in case someone gives us a taproot descriptor. Later in the process, we can then write:

```rust
struct Descriptor<Pk, PreTaprootOnly> {
   // all sorts of functions in an infallible-flavor because we are guaranteed to never have the `Tr` variant
}

struct Descriptor<Pk, AllKinds> {
   // all sorts of functions in a fallible-flavor because we may also hold a `Tr` variant
}

struct Descriptor<Pk, TaprootMarker> {
   // all functions which don't care about taproot
}
```

This design relies on the fact that nothing else other than the `parse` functions above can construct an instance of `Descriptor`. We would thus have to move away from a public enum and instead make the enum `pub(crate)` or private and hold it in an inner, private field.

Coming back to the initial goal: In case a user statically knows that they will never need taproot scripts, they can use the specialised constructor and gain access to a range of infallible APIs which would otherwise not exist. If that cannot be guaranteed statically, they need to use the other `parse` API and deal with potential failure cases. We can also offer a conversion function between two.

I am unsure whether this complexity is worth it though.

sanket1729:
I think before we tackle the reform of descriptor trait, we should 
Overall descriptors overall serve three main purposes:

1) Create new addresses: Done with address API.
2) Watch funds already contained in some descriptors:  Done with `script_pubkey` API. Note that this allows watching of descriptors that don't have a address defined for them.
3) Enable some form of semantic analysis from the descriptor: Possible with the lift trait. ExplicitScript also kind of falls into this category. 

We also clubbed other minor functionality descriptor trait with signing-related utilities. For example:
1) In case the descriptor is wrapped segwit descriptor, we can figure out `unsigned_script_sig`
2) For segwit signers that need access to script code, we have the `script_code` API. 
3) `explicit_script` offers a nice way to look at the internal policy without the script wrapping. 

None of these require functions like sanity check that should be removed and not be a part of trait API. This should be enforced(I think it already is, but I will double-check) at descriptor creation time.  

I tried to do document something similar a while ago: https://gist.github.com/sanket1729/b1ed8ed776c7ac798d5e33e40f5d996d. 

> 
Trying to guess what other uses may want abstract over is a fallacy. They can create their own traits if they need to abstract over multiple descriptors.
Aligning APIs is not enough to justify a trait. Several, concrete descriptors can have the same API without forcing the user to import a trait to use them. This should resolve the fallible vs. infallible dilemma.
The main API seems to be the Descriptor enum. The functions on that one can just live within a regular impl block and delegate to all known variants. No trait needed for that.

I agree with @thomaseizinger, I think this is the way to go. AFAIR, @afilini has a downstream implementation of `DescriptorTrait`. Would like see what he thinks of the idea of decomposing `DescriptorTrait` into multiple traits. 


> 
I don't know about this approach of pulling individual methods into infallible traits. It means that you can't write code which is generic over all descriptor types, and still be able to compute Addresses, unless you use the specific enum Descriptor type that we provide. It means that other implementors of DescriptorTrait that don't also implement ToAddress and ExplicitScript are second-class.


This is the major criticism of the idea. I agree that this is bad, but I think people implementing custom descriptor traits are already advanced users of the library that I don't mind making things hard of them. This simplifies the codebase for us and the majority of people whom we expect will use the `Descriptor` enum from this library.

sanket1729 (commented):
Sorry for the delay. I was away last weekend, will catch up with notifications across the rust-bitcoin stack

src/descriptor/mod.rs
```diff
@@ -163,16 +143,28 @@ pub trait DescriptorTrait<Pk: MiniscriptKey> {
...
+}
+
+/// A trait for getting the "witness script" for a descriptor.
+pub trait ExplicitScript<Pk: MiniscriptKey + ToPublicKey> {
+    /// Computes the "witness script" of the descriptor, i.e. the underlying
+    /// script before any hashing is done. For `Bare`, `Pkh` and `Wpkh` this
+    /// is the scriptPubkey; for `ShWpkh` and `Sh` this is the redeemScript;
+    /// for the others it is the witness script.
+    fn explicit_script(&self) -> Script;
+}
```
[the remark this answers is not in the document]
tcharding: I think this is on purpose, FTR my new version does not use this short form of the comment.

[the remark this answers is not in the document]
apoelstra: Agreed, I think it's the comment that should be changed :)

src/descriptor/bare.rs
```diff
@@ -20,7 +20,7 @@
 
 use std::{fmt, str::FromStr};
 
-use bitcoin::{self, blockdata::script, Script};
+use bitcoin::{self, blockdata::script, Address, Network, Script};
```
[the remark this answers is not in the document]
tcharding: I was not sure what to do about this. I saw, as you say, this is done a bunch. I get it for `PublicKey` and `PrivateKey` but I don't get it for `Network`/`Address` etc. since this is a bitcoin library we all know what these things are. I was also unsure because of uniformity in this crate vs uniformity across the stack, I'm not exactly sure but I think it makes sense to try for uniformity across the stack?

Is there some perspective on using `bitcoin::Address` that I'm missing. Its not the first time I've debated removing paths from identifiers so I know I tend to err on the side of too terse :)

tcharding:
> Sorry for the delay. I was away last weekend

No sweat, good to have you back :)

tcharding:
Thanks for your input @thomaseizinger, I appreciate the effort you went to.

src/descriptor/mod.rs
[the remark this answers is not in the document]
tcharding: Cheers, fixed. The code comment I left in may need further improvements.

src/descriptor/bare.rs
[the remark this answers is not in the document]
apoelstra: I do this because I work a lot on rust-elements (which has its own fork of miniscript which uses the new opcodes) so it's easy to distinguish `elements::Address` from `bitcoin::Address`. For `Address` in particular it also helps me distinguish a "bitcoin address" from a "network address" or "memory address" etc.

I agree that within rust-miniscript itself, this serves little purpose. I mildly prefer things the way they are but don't have strong feelings either way.

apoelstra:
I agree with @thomaseizinger and @sanket1729. Lets just drop our own `DescriptorTrait` (splitting it into a set of smaller traits where that makes sense) and move our "descriptor-generic" code into direct impls on the `Descriptor` enum.

People who want their own "descriptor trait" are doing advanced and probably-custom things and can roll their own traits.

Thanks for your detailed analysis, and for reverse-engineering so much of our API @thomaseizinger !!

src/descriptor/bare.rs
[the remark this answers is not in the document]
sanket1729: > I'm not exactly sure but I think it makes sense to try for uniformity across the stack

Yep, I would like to be consistent throughout the stack. I think with all the new contributors and throughout the lifetime of the project, this seems difficult to enforce/guarantee. 

I agree that using `Address` makes sense (although there is also IP address in the network module). I always have had the personal convention to `external_crate::DataStructure` or have exactly one thing before the `::` in the path. I don't have a strong rationte to do this, just a habit I subconsciously formed.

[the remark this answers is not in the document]
sanket1729: > I do this because I work a lot on rust-elements (which has its own fork of miniscript which uses the new opcodes) so it's easy to distinguish elements::Address from bitcoin::Address

For the purposes of having minimal conflicts when porting to elements-miniscript, it might be better to `Addresss` instead of `bitcoin::Address`. Considering this, if you don't have any preference between the two I would rather it be `Address` than `bitcoin::Address` as it would save me some time when porting changes from here.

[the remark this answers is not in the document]
apoelstra: Ok cool, good feedback. Let's go with `Address` (and `Network`) then.

src/descriptor/bare.rs
```diff
@@ -258,30 +231,38 @@ impl<Pk: MiniscriptKey> Pkh<Pk> {
...
+
 impl<Pk: MiniscriptKey + ToPublicKey> Pkh<Pk> {
-    /// Obtain the corresponding script pubkey for this descriptor
-    /// Non failing verion of [`DescriptorTrait::script_pubkey`] for this descriptor
-    pub fn spk(&self) -> Script {
-        let addr = bitcoin::Address::p2pkh(&self.pk.to_public_key(), bitcoin::Network::Bitcoin);
+    /// Obtains the corresponding script pubkey for this descriptor.
+    pub fn script_pubkey(&self) -> Script {
+        let addr = Address::p2pkh(&self.pk.to_public_key(), Network::Bitcoin);
         addr.script_pubkey()
```
thomaseizinger, line 250: This could delegate to `address` internally.

src/descriptor/mod.rs
```diff
@@ -607,39 +479,39 @@ impl<Pk: MiniscriptKey> DescriptorTrait<Pk> for Descriptor<Pk> {
...
     }
 
+    /// Attempts to produce a non-malleable satisfying witness and scriptSig to spend an
+    /// output controlled by the given descriptor; add the data to a given
+    /// `TxIn` output.
+    pub fn satisfy<S>(&self, txin: &mut TxIn, satisfier: S) -> Result<(), Error>
+    where
+        S: Satisfier<Pk>,
+    {
+        // easy default implementation
```
thomaseizinger, line 489: I think this comment may have been here because this used to be in a trait and could have been overridden. That is no longer possible now so perhaps remove this?

src/descriptor/mod.rs
```diff
@@ -607,39 +479,39 @@ impl<Pk: MiniscriptKey> DescriptorTrait<Pk> for Descriptor<Pk> {
...
         match *self {
             Descriptor::Bare(ref bare) => bare.max_satisfaction_weight(),
-            Descriptor::Pkh(ref pkh) => pkh.max_satisfaction_weight(),
-            Descriptor::Wpkh(ref wpkh) => wpkh.max_satisfaction_weight(),
+            Descriptor::Pkh(ref pkh) => Ok(pkh.max_satisfaction_weight()),
+            Descriptor::Wpkh(ref wpkh) => Ok(wpkh.max_satisfaction_weight()),
             Descriptor::Wsh(ref wsh) => wsh.max_satisfaction_weight(),
             Descriptor::Sh(ref sh) => sh.max_satisfaction_weight(),
             Descriptor::Tr(ref tr) => tr.max_satisfaction_weight(),
         }
```
thomaseizinger, line 513: Style-nit but I prefer to write this as:

```suggestion
        let weight = match *self {
            Descriptor::Bare(ref bare) => bare.max_satisfaction_weight(),
            Descriptor::Pkh(ref pkh) => pkh.max_satisfaction_weight()?,
            Descriptor::Wpkh(ref wpkh) => wpkh.max_satisfaction_weight()?,
            Descriptor::Wsh(ref wsh) => wsh.max_satisfaction_weight(),
            Descriptor::Sh(ref sh) => sh.max_satisfaction_weight(),
            Descriptor::Tr(ref tr) => tr.max_satisfaction_weight(),
        };

	Ok(weight)
```

src/descriptor/sh.rs
```diff
@@ -205,52 +205,89 @@ impl<Pk: MiniscriptKey> Sh<Pk> {
...
-                bitcoin::Address::p2sh(&smv.encode(), network).expect("Size checked in Miniscript")
-            }
-            ShInner::Ms(ref ms) => {
-                bitcoin::Address::p2sh(&ms.encode(), network).expect("Size checked in Miniscript")
-            }
+            ShInner::Wsh(ref wsh) => Ok(Address::p2sh(&wsh.script_pubkey(), network)?),
+            ShInner::Wpkh(ref wpkh) => Ok(Address::p2sh(&wpkh.script_pubkey(), network)?),
+            ShInner::SortedMulti(ref smv) => Ok(Address::p2sh(&smv.encode(), network)?),
+            ShInner::Ms(ref ms) => Ok(Address::p2sh(&ms.encode(), network)?),
         }
```
thomaseizinger, line 277: I think this would be easier to understand as:

```rust
        let script = match self.inner {
            ShInner::Wsh(ref wsh) => wsh.script_pubkey(),
            ShInner::Wpkh(ref wpkh) => wpkh.script_pubkey(),
            ShInner::SortedMulti(ref smv) => smv.encode(),
            ShInner::Ms(ref ms) => ms.encode(),
        };
	let address = Address::p2sh(&script, network)?;

	Ok(address)
```

thomaseizinger (commented):
Skimmed the PR again and made a few comments, looks very good overall :)

src/descriptor/mod.rs
tcharding: woops yep, thanks

src/descriptor/mod.rs
tcharding: Yes that is nicer, thanks.

src/descriptor/sh.rs
tcharding: Nice, thanks.

src/descriptor/bare.rs
tcharding: Cool, done. Added the comment about using a hard coded address from the discussion above also.

tcharding:
Implemented suggestions from @thomaseizinger, squashed it all down to two patches, rebased, and re-wrote the PR description.

afilini:
> AFAIR, @afilini has a downstream implementation of DescriptorTrait. Would like see what he thinks of the idea of decomposing DescriptorTrait into multiple traits.

Yeah kinda, we have some traits that we implement on `Descriptor` that are in a way an extension to `DescriptorTrait`, to provide other useful stuff quickly.

I haven't read the code or anything, just wanted to say that we were planning to abstract over generic `DescriptorTrait` types in BDK to support custom scripts that can't be expressed as a descriptor. So the idea was to store a generic type inside the wallet, and if you want you can implement `DescriptorTrait`on your custom type and BDK would work with it automatically.

That said, it's true that we already have our own custom traits that you probably would have to implement as well in the scenario I described, so it's not the end of the world to reimplement in BDK the equivalent of `DescriptorTrait` if you remove it from here.

apoelstra (approved):
ACK dbdcc81af970c25e63443fe7942d7acf0217605e

It's a big diff and I'm not sure I fully grokked all the API changes. But I tried to read it carefully and this all looks good to me. One minor nit is that I'd prefer the term "ec-signature" in the comments to be changed to "ECDSA siganture".

thomaseizinger:
This sentence in the 2nd commit could be worded better IMO:

> This makes using the type for ergonomic also since users no longer need to import the trait to use a descriptor.

Maybe say: This makes usage of the `Descriptor` type more ergonomic because users no longer need to import a trait.

tcharding:
I converted to draft, this is a big change. Now we have some agreement on the way forward I'll put some effort into making the patches more reviewable. Probably should have been draft this whole time, my bad.

tcharding:
Split up into smaller patches, re-wrote PR description. Its still a bit to review, thanks in advance for the time!

apoelstra (approved):
ACK 32a74af28a223f3a84b8ac35c6ed50b0bd498b9a

It is much easier to read this way, thanks!

src/descriptor/mod.rs
```diff
@@ -399,113 +296,47 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
+}
+
+impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
     /// Computes the Bitcoin address of the descriptor, if one exists
-    fn address(&self, network: bitcoin::Network) -> Result<bitcoin::Address, Error>
-    where
-        Pk: ToPublicKey,
-    {
+    ///
+    /// Some descriptors like pk() don't have any address.
```
thomaseizinger, line 323: Nit:

```suggestion
    /// Some descriptors like pk() don't have an address.
```

thomaseizinger (approved):
Nice work.

ACK 32a74af28a223f3a84b8ac35c6ed50b0bd498b9a

sanket1729 (approved):
ACK 32a74af28a223f3a84b8ac35c6ed50b0bd498b9a. This is a very good improvement.

PR #387, opened 2022-05-04
Clean up the examples

tcharding:
In an effort to make the examples more digestible to newcomers do various clean ups and refactorings. Each example is done as a separate patch to assist review.

The diffs get smaller as you go down the list :)

### DIFF
--- a/examples/htlc.rs
+++ b/examples/htlc.rs
@@ -24,7 +24,7 @@ use miniscript::DescriptorTrait;
 use std::str::FromStr;
 
 fn main() {
-    //HTLC policy with 10:1 odds for happy(co-operative) case compared to uncooperative case
+    // HTLC policy with 10:1 odds for happy (co-operative) case compared to uncooperative case.
     let htlc_policy = Concrete::<bitcoin::PublicKey>::from_str(&format!("or(10@and(sha256({secret_hash}),pk({redeem_identity})),1@and(older({expiry}),pk({refund_identity})))",
                                                   secret_hash = "1111111111111111111111111111111111111111111111111111111111111111",
                                                   redeem_identity = "022222222222222222222222222222222222222222222222222222222222222222",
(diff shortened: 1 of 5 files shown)


apoelstra (approved):
ACK f0501dae5fa617900e5f9bc6ae6b60c55d6be2bb

examples/verify_tx.rs
```diff
@@ -12,18 +12,119 @@
...
+    // as having participated in the script
+
+    println!("\n\nExample two:\n");
+    let secp = Secp256k1::new();
+
+    // Sometimes it is necessary to have additional information to get the
+    // `bitcoin::PublicKey` from the `MiniscriptKey` which can be supplied by
+    // the `to_pk_ctx` parameter. For example, when calculating the script
+    // pubkey of a descriptor with xpubs, the secp context and child information
+    // maybe required.
```
sanket1729, line 87: Ouch, we no longer have to_pk_ctx in this codebase. This comment should have been removed. Not a part of this PR.

examples/htlc.rs
```diff
@@ -39,31 +39,34 @@ fn main() {
...
     .expect("Resource limits");
 
-    // Check whether the descriptor is safe
-    // This checks whether all spend paths are accessible in bitcoin network.
-    // It maybe possible that some of the spend require more than 100 elements in Wsh scripts
-    // Or they contain a combination of timelock and heightlock.
+    // Check whether the descriptor is safe. This checks whether all spend paths are accessible in
+    // the Bitcoin network. It may be possible that some of the spend paths require more than 100
+    // elements in Wsh scripts or they contain a combination of timelock and heightlock.
     assert!(htlc_descriptor.sanity_check().is_ok());
```
sanket1729, line 45: I would really like to get rid of these sanity checks. I think we lose of using rust if users have to call sanity check to double-check things. This feels like CPubKey.isVliad() type stuff

sanket1729 (approved):
ACK f0501dae5fa617900e5f9bc6ae6b60c55d6be2bb

examples/htlc.rs
apoelstra: Yeah, I think they don't belong in the example. They'd be useful for users who are manually constructing the `Descriptor` object I think.

tcharding: Should we be doing the sanity checks in all the constructors (incl. `FromStr`) instead?

PR #388, opened 2022-05-04
Use the term 'abstract' instead of 'semantic'

tcharding:
Currently we are introducing a new term 'semantic' for what
is (according to the code comment) typically referred to as 'abstract'.
The justification give is that `abstract` is a reserved Rust keyword.

Rust provides a mechanism for using keywords as identifiers, one just
has to prefix with `r#`. Also, in this case we have aliases for abstract
and concrete `Policy` already so we barely ever need the `r#` trick.

Use the term 'abstract' instead of 'semantic', involves:

- Rename module `semantic` -> `abstract`
- Rename fuzz target `roundtrip_semantic.rs` -> `roundtrip_abstract.rs`
- Rename type `Semantic` -> `Abstract`
- Use `Abstract` everywhere instead of `semantic::Policy`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,7 @@ See this (link)[https://github.com/rust-bitcoin/rust-miniscript/pull/349/commits
 - Changes to the miniscript type system to detect an invalid
   combination of heightlocks and timelocks
      - Lift miniscripts can now fail. Earlier it always succeeded and gave
-       the resulting Semantic Policy
+       the resulting Abstract Policy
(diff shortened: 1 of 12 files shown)


apoelstra:
I'm not a huge fan of this. I like being able to read `semantic::Policy` vs having `Abstract` as the name of a type (abstract what?). It would be nice if we could say `abstract::Policy` but we'd need to `r#` prefix it which is a bit ugly.

tcharding:
No sweat, I'll close.

PR #391, opened 2022-05-05
Make the integration script more quiet

tcharding:
Currently we have two initial output lines when running the integration test script. The first is confusing (to me at least) and the second is unnecessary it seems (please correct me if I am wrong).

- Patch one removes the warning: `bitcoind: no process found`
- Patch two removes the printing of `$PATH`

### DIFF
--- a/integration_test/run.sh
+++ b/integration_test/run.sh
@@ -9,8 +9,6 @@ mkdir -p ${TESTDIR}/1
 # To kill any remaining open bitcoind.
 killall -9 bitcoind 2> /dev/null
 
-echo $PATH
-
 BLOCKFILTERARG=""
 if bitcoind -version | grep -q "v0\.\(19\|2\)"; then
     BLOCKFILTERARG="-blockfilterindex=1"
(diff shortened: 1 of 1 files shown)


integration_test/run.sh
```diff
@@ -7,9 +7,7 @@ rm -rf ${TESTDIR}
 mkdir -p ${TESTDIR}/1
 
 # To kill any remaining open bitcoind.
-killall -9 bitcoind
-
-echo $PATH
+killall -9 bitcoind 2> /dev/null
```
apoelstra, line 10: lol, I had no idea this line was here. This is pretty anti-social. But it has nothing to do with this PR, so that's fine.

apoelstra (approved):
ACK aa1bac342ef714a5b035af1252eace6f1fe060ea

integration_test/run.sh
tcharding: Couldn't think of a quick thorough fix so I raised an issue: https://github.com/rust-bitcoin/rust-miniscript/issues/395

sanket1729: I may be to blame here. I was trying to make the script work without thinking of all the consequences.

sanket1729 (approved):
ACK aa1bac342ef714a5b035af1252eace6f1fe060ea

PR #392, opened 2022-05-06
Shorten translatefpk identifier

tcharding:
Recently we refactored the translate pk traits, in doing so the identifier `fpk` and `fpkh` were favoured over `translatefpk`, and
`translatefpkh` respectively. I missed a bunch of places.

Be uniform across the code base and use the identifiers `fpk` and `fpkh`.

This patch is the result of applying `s/translatefpk/fpk/g` to all source files.

Refactor only, no logic changes.


tcharding:
I'm a goose, I did this in another branch but never pushed it up.

PR #393, opened 2022-05-06
Refactor implementations of `TranslatePk`

tcharding:
We recently refactored the `TranslatePk` traits but did not do _all_ the implementations.

- Patch 1 trivial cleanup.
- Patch 2 does the remaining implementations.
- Patch 3 shortens other instances of the `translatefpk[h]` identifier so as to be uniform.

### DIFF
--- a/src/descriptor/bare.rs
+++ b/src/descriptor/bare.rs
@@ -212,24 +212,21 @@ impl<Pk: MiniscriptKey> ForEachKey<Pk> for Bare<Pk> {
     }
 }
 
-impl<P: MiniscriptKey, Q: MiniscriptKey> TranslatePk<P, Q> for Bare<P> {
+impl<P, Q> TranslatePk<P, Q> for Bare<P>
+where
+    P: MiniscriptKey,
+    Q: MiniscriptKey,
+{
     type Output = Bare<Q>;
 
-    fn translate_pk<Fpk, Fpkh, E>(
-        &self,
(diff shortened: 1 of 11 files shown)


apoelstra (approved):
ACK 13c07716c8d25dd9ccb5f41528cebfc9ee106b6f

Long diff but almost entirely formatting changes.

tcharding:
Thanks for the review, I get a bit trigger happy when I'm learning a new codebase so feel free to be heavier on the NACK's for miniscript refactorings from me in the first few months :) There is a risk of some code churn as I play with the traits and trait bounds but I think its worth it to make the code as simple as we can.

sanket1729:
Relatedly, I have a plan for changing all the awkwardly named TranslatePk* methods into a single Trait but got sidetracked. The current TranslatePk<P, Q> instead of operating on two function parameters, would operate on a single trait object. The trait object would have methods for `translate_fpk`, `translate_hash`, `translate_time` etc. 

The main idea here is to allow users to specify hash values and time values in generic form. We implement these are additional associated values for MiniscriptKey so that write something like 
```
let htlc = "or(and(pk(A),sha256(H)), and(pk(B),after(expiry))"
```

 Hopefully, I can get it out in a few days for review.

sanket1729 (approved):
ACK 13c07716c8d25dd9ccb5f41528cebfc9ee106b6f

Issue #226, opened 2021-01-14
Derive `Hash` for everything that can derive `Hash` for all structures in the lib


sanket1729:

apoelstra:
I think we should amend this to rather say:

* Implement `Hash` for all *pub* items that *also implement `Eq`*
* Audit whether some things currently don't implement `Eq` but should

PR #394, opened 2022-05-06
Derive hash

tcharding:
Derive `Hash` for all structs and enums (excl. error enums) that can.

Patch 2 is 'wip' because I'm not sure if it makes sense to derive `Hash` for the types in that patch but the issue says "do all of them"?

Draft still for feedback on patch 2. Also I did not re-check the whole codebase, the original PR was done a long time ago, well check it over before taking off draft.

Resolves: #226


tcharding:
Done in #659

PR #399, opened 2022-05-10
Use KeySource instead of tuple

tcharding:
Clippy emits:

 warning: very complex type used. Consider factoring parts into `type` definitions

The type we return is more complex than needed because we do not use the `bip32::KeySource` alias but instead write the full tuple `(byp32::Fingerprint, bip32::DerivationPath)`.

To fix the clippy warning we have at least 3 options

1. Add an alias for the complex return type
2. Add a compiler directive to quieten clippy
3. Use `bip32::KeySource` instead of tuple

(1) seems like an overkill
(2) means we loose lint coverage for the whole function

Elect to do (3), the best of 3 imperfect solutions.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -528,8 +528,7 @@ impl FromStr for DescriptorSecretKey {
 impl<K: InnerXKey> DescriptorXKey<K> {
     fn parse_xkey_origin(
         s: &str,
-    ) -> Result<(&str, Option<(bip32::Fingerprint, bip32::DerivationPath)>), DescriptorKeyParseError>
-    {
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK f8945d05d82b90456fc68812f53807b690a5ab1e

I actually think this is a great improvement.

sanket1729 (approved):
ACK f8945d05d82b90456fc68812f53807b690a5ab1e

tcharding:
> I actually think this is a great improvement.

Oh yeah? Why is that? At some other place in some other time I favoured the tuple also because I thought it was clearer than `KeySource`, I'm interested to hear your thoughts. Cheers

tcharding:
cool, cheers.

PR #401, opened 2022-05-11
Move DummyKey to submodule

tcharding:
The `DummyKey`, according to code comments is only useful for testing. Put it in a submodule and conditionally compile it in only for tests.


sanket1729:
It might still be worth it to move it to a submodule and expose it. I think it is polluting the main `lib.rs`

tcharding:
> I think it is polluting the main lib.rs

I agree, added to my todo list.

Issue #381, opened 2022-05-02
Derive `Hash` while manually implementing `PartialEq`


tcharding:
For the `Miniscript` struct
```
#[derive(Clone, Hash)]
pub struct Miniscript<Pk: MiniscriptKey, Ctx: ScriptContext> {
```

Clippy warns us that we are deriving `Hash` and using a custom implementation of `PartialEq`, this is likely wrong as explained in the lint [docs](https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq).

Here is the clippy warning
```
error: you are deriving `Hash` but have implemented `PartialEq` explicitly
  --> src/miniscript/mod.rs:60:17
   |
60 | #[derive(Clone, Hash)]
   |                 ^^^^
   |
   = note: `#[deny(clippy::derive_hash_xor_eq)]` on by default
note: `PartialEq` implemented here
  --> src/miniscript/mod.rs:93:1
   |
93 | / impl<Pk: MiniscriptKey, Ctx: ScriptContext> PartialEq for Miniscript<Pk, Ctx> {
94 | |     fn eq(&self, other: &Miniscript<Pk, Ctx>) -> bool {
95 | |         self.node.eq(&other.node)
96 | |     }
97 | | }
   | |_^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq
   = note: this error originates in the derive macro `Hash` (in Nightly builds, run with -Z macro-backtrace for more info)
```

apoelstra:
Nice, good catch.

It's probably okay here since (in theory) `self.node` implies all the other fields but if we're going to make that explicit for `PartialEq` we should do the same for `Hash` as well.

PR #402, opened 2022-05-11
Implement hash::Hash for Miniscript

tcharding:
We manually implement a bunch of traits on `Miniscript` that pass through to the `node` field (e.g. `PartialEq`). We should do the same for `hash::Hash` instead of deriving it.

Found by clippy.

Fixes: #381


apoelstra (approved):
ACK 4ac68c9b24745e5ec28ef865ed0e9ea8af97b30c

PR #403, opened 2022-05-11
Shorten path bitcoin::secp256k1 -> secp

tcharding:
Based off of an idea I had [here](https://github.com/rust-bitcoin/rust-bitcoin/discussions/984) we can reduce the number of times devs have to write 'secp256k1' since it is a tad clunky to write. Until we get an alias in `rust-bitcoin` we can test run the usefulness of such a thing here in `rust-miniscript` by using import aliases.

Shorten the use of `bitcoin::secp25k1::Foo` -> `secp::Foo` with no loss of clarity.

The import alias can be later removed if/when `rust-bitcoin` adds an alias as described in the discussion linked above.

Please note, this PR in and of itself is not that much of an improvement, the hope is that this makes coding on `rust-miniscript` more fun by being less clunky. If this is the case it may influence other crates and make everyones life better (for those of us who define goodness of their lives by terseness of the code we write).


tcharding:
I'm going to leave this one to go into `rust-bitcoin` then we can decide if we want to use it here.

PR #405, opened 2022-05-11
Improve the test script and CI pipeline

tcharding:
Make an effort to improve the test script, taking inspiration for the test script in `rust-secp256k1`. Since Bash can, at times, be a bit obscure, I made extra effort to do very simple discreet patches.

First part of the PR is fixes to the test script, second part is improvements to the CI pipeline.

### DIFF
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -25,14 +25,12 @@ fi
 # Fuzz if told to
 if [ "$DO_FUZZ" = true ]
 then
-    (
-        cd fuzz
-        cargo test --verbose
-        ./travis-fuzz.sh
-        # Exit out of the fuzzer,
-        # run stable tests in other CI vms
-        exit 0
-    )
+    cd fuzz
+    cargo test --verbose
+    ./travis-fuzz.sh
+
+    # Exit out of the fuzzer, do not run other tests.
+    exit 0
 fi
 
 # Test without any features first
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -16,10 +16,8 @@ rustc --version
 # Format if told to
 if [ "$DO_FMT" = true ]
 then
-    (
-        rustup component add rustfmt
-        cargo fmt --all -- --check
-    )
+    rustup component add rustfmt
+    cargo fmt --all -- --check
 fi
 
 # Fuzz if told to
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -31,19 +31,29 @@ then
     exit 0
 fi
 
-# Test without any features first
-cargo test --verbose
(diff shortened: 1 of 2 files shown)


tcharding:
Removed last patch, I guess that is why we are using 1.58 for fuzzing and not stable?

contrib/test.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh -ex
 
+set -e
```
sanket1729, line 3: This is a big bug! I wonder how it ever worked till now

contrib/test.sh
```diff
@@ -31,19 +31,29 @@ then
...
-# run all examples
-run-parts ./target/debug/examples
+    # Run all the examples
+    cargo build --examples
+    cargo run --example htlc --features=compiler 
+    cargo run --example parse
+    cargo run --example sign_multisig
+    cargo run --example verify_tx > /dev/null
+    cargo run --example psbt
+    cargo run --example xpub_descriptors
```
sanket1729, line 55: Why manually list examples instead of run-parts? Is it because of the extra args to each example? We need to remember to add examples here every time.

sanket1729 (approved):
ACK ed51323a0bb3257b09abfb5eb51ca21815b815af

contrib/test.sh
tcharding: Its so the script can be run locally, but yes as you say there is a maintenance cost. Is it worth it, I don't know? I configure `target-dir` to be a global build cache so there is no `target` directory when I build so the examples don't run.

contrib/test.sh
tcharding: I guess we were getting some false positives in the CI runs. `rust-bitcoin/contrib/test.sh` doesn't have this either :)

apoelstra (approved):
ACK ed51323a0bb3257b09abfb5eb51ca21815b815af

contrib/test.sh
```diff
@@ -8,41 +10,50 @@ then
...
+        cargo test --features="$feature"
+    done
 
-# Also build and run each example to catch regressions
-cargo build --examples
-# run all examples
-run-parts ./target/debug/examples
+    # Run all the examples
+    cargo build --examples
+    cargo run --example htlc --features=compiler
```
sanket1729, line 50: nit: There is a white space at the end of this line.

tcharding: ooo, means I did not read the diff before pushing, bad Tobin. Thanks man, I'll fix it.

apoelstra: I read the diff and I missed this too, oops! Thanks Sanket.

PR #406, opened 2022-05-11
Remove use of ref keyword

tcharding:
This one smells like code churn, raising as draft for comments.

We currently use `ref` _a lot_. This patch falls victim to 'if you only have a hammer everything looks like a nail' - I did a search and replace  `s/ref //g` on every source file then fixed the build errors. This almost certainly means I removed 'good' uses of `ref`.

An example of a code improvement is:

Before this patch is applied we had:

```rust
    pub fn into_pre_taproot_desc(self) -> Result<pretaproot::PreTaprootDescriptor<Pk>, Self> {
        match *self {
            Descriptor::Bare(ref bare) => Ok(pretaproot::PreTaprootDescriptor::Bare(bare)),
            Descriptor::Pkh(ref pkh) => Ok(pretaproot::PreTaprootDescriptor::Pkh(pkh)),
            Descriptor::Wpkh(ref wpkh) => Ok(pretaproot::PreTaprootDescriptor::Wpkh(wpkh)),
            Descriptor::Sh(ref sh) => Ok(pretaproot::PreTaprootDescriptor::Sh(sh)),
            Descriptor::Wsh(ref wsh) => Ok(pretaproot::PreTaprootDescriptor::Wsh(wsh)),
            Descriptor::Tr(ref tr) => Err(Descriptor::Tr(tr)),
        }
    }
```
All these `ref`s are not required if we remove the dereference from `self`.

A not so good result of this patch is having to write `*k` in a bunch of places instead of `k`.

Please note patch includes such things as:
```rust
	WshInner::Ms(_ms) => DescriptorType::Wsh,
```

The `_ms` could become `_`, will do if this work is deemed valuable.

### DIFF
--- a/integration_test/src/test_desc.rs
+++ b/integration_test/src/test_desc.rs
@@ -121,7 +121,7 @@ pub fn test_desc_satisfy(cl: &Client, testdata: &TestData, desc: &str) -> Witnes
 
     let mut sighash_cache = SighashCache::new(&psbt.unsigned_tx);
     match derived_desc {
-        Descriptor::Tr(ref tr) => {
+        Descriptor::Tr(tr) => {
             // Fixme: take a parameter
             let hash_ty = sighash::SchnorrSighashType::Default;
 
@@ -196,11 +196,11 @@ pub fn test_desc_satisfy(cl: &Client, testdata: &TestData, desc: &str) -> Witnes
                 Descriptor::Wpkh(pk) => find_sk_single_key(*pk.as_inner(), testdata),
                 Descriptor::Sh(sh) => match sh.as_inner() {
                     miniscript::descriptor::ShInner::Wsh(wsh) => match wsh.as_inner() {
-                        miniscript::descriptor::WshInner::SortedMulti(ref smv) => {
+                        miniscript::descriptor::WshInner::SortedMulti(smv) => {
                             let ms = Miniscript::from_ast(smv.sorted_node()).unwrap();
                             find_sks_ms(&ms, testdata)
                         }
-                        miniscript::descriptor::WshInner::Ms(ref ms) => find_sks_ms(&ms, testdata),
+                        miniscript::descriptor::WshInner::Ms(ms) => find_sks_ms(&ms, testdata),
                     },
                     miniscript::descriptor::ShInner::Wpkh(pk) => {
                         find_sk_single_key(*pk.as_inner(), testdata)
@@ -212,11 +212,11 @@ pub fn test_desc_satisfy(cl: &Client, testdata: &TestData, desc: &str) -> Witnes
                     miniscript::descriptor::ShInner::Ms(ms) => find_sks_ms(&ms, testdata),
                 },
                 Descriptor::Wsh(wsh) => match wsh.as_inner() {
-                    miniscript::descriptor::WshInner::SortedMulti(ref smv) => {
+                    miniscript::descriptor::WshInner::SortedMulti(smv) => {
                         let ms = Miniscript::from_ast(smv.sorted_node()).unwrap();
                         find_sks_ms(&ms, testdata)
                     }
-                    miniscript::descriptor::WshInner::Ms(ref ms) => find_sks_ms(&ms, testdata),
+                    miniscript::descriptor::WshInner::Ms(ms) => find_sks_ms(&ms, testdata),
                 },
                 Descriptor::Tr(_tr) => unreachable!("Tr checked earlier"),
             };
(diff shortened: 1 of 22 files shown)


apoelstra:
Ok, concept ACK. Auto-`ref`ing is one of the many "helpful" things the Rust compiler does that I believe actually make things more complicated, so in rebellion I have been manually accounting for levels of indirection and using the appropriate keywords.

Because there is no mach

tcharding:
As you might have understood from the initial description I was a bit hesitant about putting this one up. 

>  I believe actually make things more complicated

If you feel like it can you explain this position more please? Some of the changes remove a `ref` only to add a `&` which seems to be solely subjective. I'm interested in understanding the position for using `ref`.

I think there are multiple types of uses of `ref` changed in this PR but since I don't fully understand the argument _for_ `ref` I might be missing some subtlety.

Cheers

apoelstra:
`ref` and `&` do the opposite thing. The fact that you can interchange them is because Rust implicitly references and de-references variables to make things typecheck. This is confusing because it doesn't work 100% of the time and when it *does* work it's a subtle form of weak typing, which undermines the programmer's model of the code.

tcharding:
Non-urgent discussion, I'd like to fully understand you position.

> which undermines the programmer's model of the code

Do you mean that not knowing explicitly whether we are dealing with a concrete type or a reference, in general, undermines the programmers model of the code?

Or maybe, are you describing the fact that, coming from C where one knows exactly what a pointer is (an integer representing a memory location) to Rust where it (at least to me at the moment) means 'something the borrow checker uses to determine ownership', undermines the programmers model?

Up until now I thought that the use of `ref` vs `&` was just stylistic.

apoelstra:
>Do you mean that not knowing explicitly whether we are dealing with a concrete type or a reference, in general, undermines the programmers model of the code?

Yep, this is what I mean.

I'm slowly coming around to the Rust point of view that it doesn't really matter (especially as my model of pointers gets increasingly abstract and un-C-like, thanks to Ralf Jung's [terrifying blog posts](https://www.ralfj.de/blog/2022/04/11/provenance-exposed.html)...) except maybe-sometimes for performance. But initially it made me very uncomfortable that I could be wrong about the amount of indirection in a variable and the compiler would just hide this from me.

tcharding:
> Specifically, both pointer-integer and integer-pointer casts have to be treated as impure operations, but for different reasons. Pointer-integer casts have a side-effect as we have seen. Integer-pointer casts are non-deterministic – they can produce different results even for identical inputs.

Ouch! Thanks for the link, it opened my eyes to a new thing :)

tcharding:
Non-urgent Rustlang discussion for giggles.

I had a play around to see if, in my subjective opinion, I could notice my mental model changing by different formats. The only thing that changed was I got much more aware of exactly what is going on - so win. I can't say that one format is better/clearer though so I lean towards using the most terse one (as I always do in such instances). The functions `what_X` represent code we have a lot of and that led me to thinking about this. Whats your thoughts?
```rust
//! Example of references in pattern matching.
#![allow(dead_code)]

#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
enum Kind {
    A(V),
    B(V),
}

#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
struct V {
    x: u32,
}

impl V {
    fn to_u32(&self) -> u32 {
        self.x
    }

    fn borrow_self(&self) {
        // Do nothing.
    }

    fn consume_self(self) {
        // Do nothing.
    }
}

impl Kind {
    fn what_0(self) {
        match self {
            Kind::A(ref x) => x.borrow_self(), // Must explicitly borrow.
            Kind::B(ref x) => x.borrow_self(),
        }
        // Can borrow by matching with &self instead of self
        match &self {
            Kind::A(x) => x.borrow_self(),
            Kind::B(x) => x.borrow_self(),
        }
        println!("accessing kind again: {:?}", self);
    }

    fn what_1(&self) -> u32 {
        match *self {
            Kind::A(ref x) => x.to_u32(),
            Kind::B(ref x) => x.to_u32(),
        }
    }

    fn what_2(&self) -> u32 {
        match self {
            &Kind::A(ref x) => x.to_u32(),
            &Kind::B(ref x) => x.to_u32(),
        }
    }

    // This is the most terse version. Is it in anyway less clear?
    fn what_3(&self) -> u32 {
        match self {
            Kind::A(ref x) => x.to_u32(),
            Kind::B(ref x) => x.to_u32(),
        }
    }

    fn what_4(&self) {
        match self {
            Kind::A(x) => x.borrow_self(),
            Kind::B(x) => x.consume_self(), // Does not compile, cannot move out of &self
        }
    }
}

fn main() {
    let k = Kind::A(V { x: 42 });
    println!("what kind: {}", k.what_1());
    println!("accessing kind again: {:?}", k);
}
```

apoelstra:
The second half of `what_0` desugars into `what_2` which is the most explicit one. `what_3` is a little bit surprising because it implicitly derefs `self`, and I'd prefer `what_1`.

tcharding:
Cool, thanks for taking the time to look at this.

PR #412, opened 2022-05-18
Refactor threshold method

tcharding:
Do a few refactorings in an attempt to make the `threshold` method easier to read.

### DIFF
--- a/src/miniscript/types/extra_props.rs
+++ b/src/miniscript/types/extra_props.rs
@@ -762,17 +762,13 @@ impl Property for ExtData {
         let mut ops_count = 0;
         let mut ops_count_sat_vec = Vec::with_capacity(n);
         let mut ops_count_nsat_sum = 0;
-        let mut op_count_sat = Some(0);
         let mut timelocks = Vec::with_capacity(n);
         let mut stack_elem_count_sat_vec = Vec::with_capacity(n);
-        let mut stack_elem_count_sat = Some(0);
         let mut stack_elem_count_dissat = Some(0);
         let mut max_sat_size_vec = Vec::with_capacity(n);
-        let mut max_sat_size = Some((0, 0));
         let mut max_dissat_size = Some((0, 0));
         // the max element count is same as max sat element count when satisfying one element + 1
         let mut exec_stack_elem_count_sat_vec = Vec::with_capacity(n);
-        let mut exec_stack_elem_count_sat = Some(0);
         let mut exec_stack_elem_count_dissat = Some(0);
 
         for i in 0..n {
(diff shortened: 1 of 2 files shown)


src/miniscript/types/extra_props.rs
```diff
@@ -1021,6 +1003,46 @@ impl Property for ExtData {
     }
 }
 
+// Function to pass to sort_by. Sort by (satisfaction cost - dissatisfaction cost).
+//
+// We sort by (satisfaction cost - dissatisfaction cost) to make a worst-case (the most
+// costy satisfactions are satisfied, the most costy dissatisfactions are dissatisfied).
+//
+// Args are of form: (<count_sat>, <count_dissat>)
+fn sat_minus_dissat<'r, 's>(
```
apoelstra, line 1025: In cb4c0376154c3699155d2c546f5a1356b30ed8e6:

I think all these functions should be defined at the top of the function that they're used in.

src/miniscript/types/extra_props.rs
```diff
@@ -1052,7 +1065,12 @@ fn opt_max<T: Ord>(a: Option<T>, b: Option<T>) -> Option<T> {
...
 }
 
-// Returns Some(x+y) is both x and y are Some. Otherwise, return none
+/// Returns Some(x+y) is both x and y are Some. Otherwise, returns `None`.
 fn opt_add(a: Option<usize>, b: Option<usize>) -> Option<usize> {
     a.and_then(|x| b.map(|y| x + y))
 }
+
+/// Returns Some((x0+y0, x1+y1)) is both x and y are Some. Otherwise, returns `None`.
+fn opt_tuple_add(a: Option<(usize, usize)>, b: Option<(usize, usize)>) -> Option<(usize, usize)> {
```
apoelstra, line 1074: In 28662d23b84abaedfb351baf0ca7c7d6876afb4a:

Same here. I think this'd be clearer defined close to its use.

apoelstra (approved):
utACK 28662d23b84abaedfb351baf0ca7c7d6876afb4a

I imagine this will be a somewhat controversial change :)

src/miniscript/types/extra_props.rs
apoelstra: I would like these functions' visibility to match the only scope where they're used, since they are essentially helpers for the code in that scope.

apoelstra: This isn't about being C-like -- you can't even do what I'm suggesting in C.

tcharding:
> I imagine this will be a somewhat controversial change :)

Do you mean the combinator stuff or the 'sort_by' functions? I'm not 100% sure the sort_by functions are better than inline.

apoelstra:
I mean the combinator stuff -- I've gotten pushback in rust-bitcoin for being too functional. (But in this library maybe it's ok..)

tcharding:
More seriously, this is a valid concern. I get that functional style combinators are hard to understand if one is not familiar with them.

On the other hand, it took me a while to work this code out, when I first read it I couldn't work out why we were setting the variable continuously, then I had to check exactly how the `and_then` chained with `map` worked in relation to `None` and think through the variable mutation - when one sees `fold` its much more obvious whats happening (IMNSHO).

tcharding:
> But in this library maybe it's ok..

I'm yet to work out whats the separation between the various crates in our stack, i.e., different people, different ethos, different coding expectations, different review expectations, different coding styles? I'm defaulting to treating them all the same, is this wrong?

src/miniscript/types/extra_props.rs
sanket1729: I have no strong preferences here. 
If the name is descriptive and is being used multiple times in different methods(I think it is the case for some of the functions here). It makes sense to have it at the bottom.  

On the other hand, if it is just a helper function, it makes sense to have it a function near the body or even a sub-function inside the main function. The readers can cleanly separate these things.

sanket1729 (approved):
ACK 28662d23b84abaedfb351baf0ca7c7d6876afb4a. IMO, these are clean improvements. 

I have a very _weak_ preference towards @tcharding's suggestion of hiding the methods away particularly because the names are descriptive. 
Reading `opt_add`, `sat_minus_dissat_cost` the reader can guess what the functions might do. However, for a generic helper function or something similar. I would side with @apoelstra.

sanket1729:
We can continue this discussion but merging this based on Andrew's ACK.

tcharding:
Every now and again I play with defining functions inside of functions but I'm never that happy with the result, because of all the types Rust has on functions its never that clean. Now I write this, perhaps defining name closures would have been nice in this case. I'll have a play with that.

apoelstra:
I wouldn't recommend naming closures -- I agree, that quickly seems to go bad in surprising ways (like, call a closure once and it'll seem to work, then years later add a second call and watch the compiler blow up). But if you define a function inside a function with the `fn` syntax there's usually no problem.

Issue #347, opened 2022-04-19
Fix and actually test the sat stack elements calculation


darosior:
There are unit tests to exerce the calculation of the maximum number of elements of a satisfaction:
https://github.com/rust-bitcoin/rust-miniscript/blob/a2d6fff55a6b4a81a92fe740ee4dea86746875cc/src/miniscript/mod.rs#L586-L631

However they *are not* executed...
https://github.com/rust-bitcoin/rust-miniscript/blob/a2d6fff55a6b4a81a92fe740ee4dea86746875cc/src/miniscript/mod.rs#L570

Actually executing them makes the test fail. It looks like most vectors were taken from the C++ implem and therefore the Rust implem of the calculation is not on par with the C++ one.
```patch
diff --git a/src/miniscript/mod.rs b/src/miniscript/mod.rs
index c20ed89..2b4d0da 100644
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -567,7 +567,7 @@ mod tests {
         non_mal: bool,
         need_sig: bool,
         ops: usize,
-        _stack: usize,
+        stack: usize,
     ) {
         let ms: Result<Segwitv0Script, _> = Miniscript::from_str_insane(ms);
         match (ms, valid) {
@@ -576,6 +576,7 @@ mod tests {
                 assert_eq!(ms.ty.mall.non_malleable, non_mal);
                 assert_eq!(ms.ty.mall.safe, need_sig);
                 assert_eq!(ms.ext.ops_count_sat.unwrap(), ops);
+                assert_eq!(ms.ext.stack_elem_count_sat.unwrap(), stack);
             }
             (Err(_), false) => return,
             _ => unreachable!(),
```
A quick look at the first vector failing with the above patch suggests that this is the calculation that needs to be fixed: the vector looks right.
https://github.com/rust-bitcoin/rust-miniscript/blob/a2d6fff55a6b4a81a92fe740ee4dea86746875cc/src/miniscript/mod.rs#L610

darosior:
Looks like it was introduced in https://github.com/rust-bitcoin/rust-miniscript/pull/49.

sanket1729:
As a part of the miniscript review for c++, I have been doing a line-by-line comparison with rust-miniscript. I still have opcode and stack size limits to double. Hopefully, will report back today/tomorrow

sanket1729:
Note that the logic of calculating stack size is slightly different in both implementations, but they should result in the same output.

PR #413, opened 2022-05-18
Debug ms attributes test

tcharding:
Try to do issue: https://github.com/rust-bitcoin/rust-miniscript/issues/347

First patch is preparation, adds a macro so each of the test vectors is checked in a separate test so we can debug.

We use the test vectors from Sipa's `miniscript/bitcoin/test/miniscript_test.cpp`. Two new vectors seem to have been added, patch 2 adds those here (they both pass).

Two of the vectors we have have different values for `ops` (also one of ours uses `nd:and_v` instead of `d:and_v`, however I do not know the implications of this), patch 2 adds the original ones as well (both fail).

With this patch set applied we have:
```
failures:
    miniscript::tests::check_attributes_15a 
    miniscript::tests::check_attributes_16 
    miniscript::tests::check_attributes_16a
    miniscript::tests::check_attributes_26
    miniscript::tests::check_attributes_27
    miniscript::tests::check_attributes_7
```

I don't have any idea how to proceed. If anyone has any suggestions I can follow them. Thanks


sanket1729:
I will look into this later, but I just looked at 16a for now. This relates to calculations about malleable scripts. The c++ code outputs incorrect values for malleable scripts. 

There may be other reasons why there is a discrepancy in other cases and we should resolve this. I would suspect this again relates to the calculation when scripts are malleable.

tcharding:
Thanks for looking @sanket1729.

sanket1729:
Based on some discussion here: https://github.com/bitcoin/bitcoin/pull/24147#discussion_r877531692, we should only care for these values if the scripts are non-malleable. 

We should either 
1) Keep our separate tests
2) Or only test these values for non-malleable vectors.

sanket1729:
In the future, if you need any assistance with a PR, feel free to tag me. I typically ignore the PRs in the marked draft.

PR #417, opened 2022-05-22
Add support for writing integration tests that fail

kanishk779:
- Earlier we did not have the ability to write integration tests that fail.
- The default behaviour was to panic and stop the execution of the program. 
- But in addition to verifying that the code works in the correct cases, we must also ensure that it should generate an appropriate error in incorrect cases.
- Some functions are modified to return Result<T, E> to support this feature.
- Four failing tests have also been added (tests 10-13) to demonstrate this.

### DIFF
--- a/tests/test_desc.rs
+++ b/tests/test_desc.rs
@@ -5,6 +5,7 @@
 //!
 
 use std::collections::BTreeMap;
+use std::{error, fmt};
 
 use actual_rand as rand;
 use bitcoin::blockdata::witness::Witness;
@@ -43,7 +44,36 @@ fn get_vout(cl: &Client, txid: Txid, value: u64, spk: Script) -> (OutPoint, TxOu
     unreachable!("Only call get vout on functions which have the expected outpoint");
 }
 
-pub fn test_desc_satisfy(cl: &Client, testdata: &TestData, desc: &str) -> Witness {
+#[derive(Debug, PartialEq)]
+pub enum DescError {
+    /// PSBT was not able to finalize
+    PsbtFinalizeError,
+    /// Problem with address computation
+    AddressComputationError,
+    /// Error while parsing the descriptor
+    DescParseError,
+    /// Invalid descriptor
+    InvalidDescriptor,
+}
+
+impl fmt::Display for DescError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match *self {
+            DescError::PsbtFinalizeError => f.write_str("PSBT was not able to finalize"),
+            DescError::InvalidDescriptor => f.write_str("Invalid descriptor"),
+            DescError::AddressComputationError => f.write_str("Problem with address computation"),
+            DescError::DescParseError => f.write_str("Not able to parse the descriptor"),
+        }
+    }
+}
+
+impl error::Error for DescError {}
+
+pub fn test_desc_satisfy(
+    cl: &Client,
+    testdata: &TestData,
+    descriptor: &str,
+) -> Result<Witness, DescError> {
     let secp = secp256k1::Secp256k1::new();
     let sks = &testdata.secretdata.sks;
     let xonly_keypairs = &testdata.secretdata.x_only_keypairs;
@@ -55,20 +85,16 @@ pub fn test_desc_satisfy(cl: &Client, testdata: &TestData, desc: &str) -> Witnes
         .unwrap();
     assert_eq!(blocks.len(), 1);
 
-    let desc = test_util::parse_test_desc(&desc, &testdata.pubdata);
+    let desc = test_util::parse_test_desc(&descriptor, &testdata.pubdata);
+    let desc = desc.map_err(|_x| DescError::DescParseError)?;
+
     let derived_desc = desc.derived_descriptor(&secp, 0).unwrap();
+    let desc_address = derived_desc.address(bitcoin::Network::Regtest);
+    let desc_address = desc_address.map_err(|_x| DescError::AddressComputationError)?;
+
     // Next send some btc to each address corresponding to the miniscript
     let txid = cl
-        .send_to_address(
-            &derived_desc.address(bitcoin::Network::Regtest).unwrap(),
-            btc(1),
-            None,
-            None,
-            None,
-            None,
-            None,
-            None,
-        )
+        .send_to_address(&desc_address, btc(1), None, None, None, None, None, None)
         .unwrap();
     // Wait for the funds to mature.
     let blocks = cl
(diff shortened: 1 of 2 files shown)


src/descriptor/tr.rs
```diff
@@ -236,9 +236,13 @@ impl<Pk: MiniscriptKey> Tr<Pk> {
...
                 let script = ms.encode();
-                builder = builder
-                    .add_leaf(depth, script)
-                    .expect("Computing spend data on a valid Tree should always succeed");
+                if let Ok(taproot_builder) = builder.add_leaf(depth, script) {
+                    builder = taproot_builder;
+                } else {
+                    return Err(Error::Unexpected(String::from(
+                        "Computing spend data on a valid Tree should always succeed",
+                    )));
```
apoelstra, line 244: How do you actually hit this case? How do you obtain a `Tr` with an invalid tree? Can you add a unit test?

Relatedly, can you separate out the changes to `src/` into a separate commit from the changes to the integration test?

Thanks for your contribution!!

kanishk779: 1. This case is hit when the height of the tree is more than 128. In the Impl for `TaprootBuilder` in file taproot.rs of the rust-bitcoin project, there is a function `insert`, which returns a `TaprootBuilderError`.
2. For which function do you want me to add the unit tests?
3. Ok I will split it into two commits.

apoelstra: Can you add a unit test for `Taproot::spend_info` which causes this new error path to be taken?

I think that would answer my "how do you hit this case" question, which I'm still confused about, sorry.

And thanks for splitting up the commits!

sanket1729: Yep, I am also curious when this gets hit. This seems like a big bug if we need to change this.

integration_test/src/test_desc.rs
```diff
@@ -40,7 +43,27 @@ fn get_vout(cl: &Client, txid: Txid, value: u64, spk: Script) -> (OutPoint, TxOu
     unreachable!("Only call get vout on functions which have the expected outpoint");
 }
 
-pub fn test_desc_satisfy(cl: &Client, testdata: &TestData, desc: &str) -> Witness {
+// Currently this is not being used, since miniscript::Error is being propagated
+// But if the types of errors grow in future, we can extend this enum and use it.
+#[derive(Debug, PartialEq)]
+pub enum MalleableDescError {
```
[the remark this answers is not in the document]
sanket1729: Here is a suggestion for how to use this enum.

1) Rename this type to DescTestError
2) Change the return type from Error to DescTestError for the test function.

```rust
pub enum DesctestError {
    DescParseError(Error),
   PsbtFinalizeError(Error),
   AddressComputationError(Error),
   ...
}
```

In the test function then, you would change the lines to make sure that correct variant is retuned.

tests/test_desc.rs
```diff
@@ -18,8 +18,13 @@ use bitcoin::{
...
 use miniscript::miniscript::iter;
 use miniscript::psbt::{PsbtExt, PsbtInputExt};
-use miniscript::{Descriptor, Miniscript, MiniscriptKey, ScriptContext, ToPublicKey};
+use miniscript::{Descriptor, Miniscript, ToPublicKey};
+use miniscript::{MiniscriptKey, ScriptContext};
+use std::error;
+use std::fmt;
+use miniscript::Error;
 mod setup;
+// use crate::test_util::{self, TestData};
```
sanket1729, line 23: remove the comment

tests/test_desc.rs
```diff
@@ -55,12 +84,30 @@ pub fn test_desc_satisfy(cl: &Client, testdata: &TestData, desc: &str) -> Witnes
         .unwrap();
     assert_eq!(blocks.len(), 1);
 
-    let desc = test_util::parse_test_desc(&desc, &testdata.pubdata);
+    let mut desc;
```
sanket1729: use map_err instead

tests/test_desc.rs
```diff
@@ -55,12 +84,30 @@ pub fn test_desc_satisfy(cl: &Client, testdata: &TestData, desc: &str) -> Witnes
...
+            desc = temp_desc;
+        },
+        Err(_) => {
+            return Err(DescError::InvalidDescriptor);
+        }
+    }
     let derived_desc = desc.derived_descriptor(&secp, 0).unwrap();
     // Next send some btc to each address corresponding to the miniscript
+    let mut rest;
+    match derived_desc.address(bitcoin::Network::Regtest) {
```
sanket1729: same as above: map_err

tests/test_desc.rs
```diff
@@ -340,30 +383,30 @@ fn test_descs(cl: &Client, testdata: &TestData) {
...
     // X! -> Internal key unknown
     // Leaf 1 -> pk(X1!) with X1 unknown
     // Leaf 2-> and_v(v:pk(X2),pk(X3)) X2 and X3 known
-    let wit = test_desc_satisfy(cl, testdata, "tr(X!,{pk(X1),and_v(v:pk(X2),pk(X3))})");
+    let wit = test_desc_satisfy(cl, testdata, "tr(X!,{pk(X1),and_v(v:pk(X2),pk(X3))})").unwrap();
     assert!(wit.len() == 3); // control block, script and one signatures
 
     // Test 5: When everything is available, we should select the key spend path
-    let wit = test_desc_satisfy(cl, testdata, "tr(X,{pk(X1),and_v(v:pk(X2),pk(X3!))})");
+    let wit = test_desc_satisfy(cl, testdata, "tr(X,{pk(X1),and_v(v:pk(X2),pk(X3!))})").unwrap();
```
sanket1729, line 394: Fix other warnings requiring the same unwrap

src/descriptor/mod.rs
```diff
@@ -332,7 +332,7 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             Descriptor::Wpkh(ref wpkh) => Ok(wpkh.address(network)),
             Descriptor::Wsh(ref wsh) => Ok(wsh.address(network)),
             Descriptor::Sh(ref sh) => Ok(sh.address(network)),
-            Descriptor::Tr(ref tr) => Ok(tr.address(network)),
+            Descriptor::Tr(ref tr) => Ok(tr.address(network)?),
```
sanket1729, line 335: Remove this commit

sanket1729 (commented):
Left a few comments

tests/test_desc.rs
sanket1729: In ac639f64c712a2e2c9488efeaa8bbbecd83030df

tests/test_desc.rs
sanket1729: In ac639f64c712a2e2c9488efeaa8bbbecd83030df

tests/test_desc.rs
sanket1729: In ac639f64c712a2e2c9488efeaa8bbbecd83030df

tests/test_desc.rs
sanket1729: In ac639f64c712a2e2c9488efeaa8bbbecd83030df, you should fix all the unwraps

kanishk779:
@sanket1729 I have made the required changes.

sanket1729 (approved):
ACK ab7959382a2ae9e0dd8aaf3500fd3951455ff464

PR #418, opened 2022-05-25
Tr compiler v3 - Incremental Enumerative Compiler

SarcasticNastik:
This PR is a follow-up in the Tr-compiler series. This introduces splitting of a `thresh(k, ...n...)` policy into different Tapleaves. 

> For now, kindly refer to the code documentation and comments for the strategy implemented

### DIFF
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -93,6 +93,105 @@ where
     }
 }
 
+/// Lightweight repr of Concrete policy which corresponds directly to a
+/// Miniscript structure, and whose disjunctions are annotated with satisfaction
+/// probabilities to assist the compiler
+#[cfg(feature = "compiler")]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+enum PolicyArc<Pk: MiniscriptKey> {
+    /// Unsatisfiable
+    Unsatisfiable,
+    /// Trivially satisfiable
+    Trivial,
+    /// A public key which must sign to satisfy the descriptor
+    Key(Pk),
+    /// An absolute locktime restriction
+    After(u32),
+    /// A relative locktime restriction
+    Older(u32),
+    /// A SHA256 whose preimage must be provided to satisfy the descriptor
+    Sha256(Pk::Sha256),
+    /// A SHA256d whose preimage must be provided to satisfy the descriptor
+    Hash256(Pk::Hash256),
+    /// A RIPEMD160 whose preimage must be provided to satisfy the descriptor
+    Ripemd160(Pk::Ripemd160),
+    /// A HASH160 whose preimage must be provided to satisfy the descriptor
+    Hash160(Pk::Hash160),
+    /// A list of sub-policies' references, all of which must be satisfied
+    And(Vec<Arc<PolicyArc<Pk>>>),
+    /// A list of sub-policies's references, one of which must be satisfied,
+    /// along with relative probabilities for each one
+    Or(Vec<(usize, Arc<PolicyArc<Pk>>)>),
+    /// A set of descriptors' references, satisfactions must be provided for `k` of them
+    Threshold(usize, Vec<Arc<PolicyArc<Pk>>>),
+}
+
+#[cfg(feature = "compiler")]
+impl<Pk: MiniscriptKey> From<PolicyArc<Pk>> for Policy<Pk> {
+    fn from(p: PolicyArc<Pk>) -> Self {
+        match p {
+            PolicyArc::Unsatisfiable => Policy::Unsatisfiable,
+            PolicyArc::Trivial => Policy::Trivial,
+            PolicyArc::Key(pk) => Policy::Key(pk),
+            PolicyArc::After(t) => Policy::After(PackedLockTime::from(LockTime::from_consensus(t))),
+            PolicyArc::Older(t) => Policy::Older(Sequence::from_consensus(t)),
+            PolicyArc::Sha256(hash) => Policy::Sha256(hash),
+            PolicyArc::Hash256(hash) => Policy::Hash256(hash),
+            PolicyArc::Ripemd160(hash) => Policy::Ripemd160(hash),
+            PolicyArc::Hash160(hash) => Policy::Hash160(hash),
+            PolicyArc::And(subs) => Policy::And(
+                subs.into_iter()
+                    .map(|pol| Self::from((*pol).clone()))
+                    .collect(),
+            ),
+            PolicyArc::Or(subs) => Policy::Or(
+                subs.into_iter()
+                    .map(|(odds, sub)| (odds, Self::from((*sub).clone())))
+                    .collect(),
+            ),
+            PolicyArc::Threshold(k, subs) => Policy::Threshold(
+                k,
+                subs.into_iter()
+                    .map(|pol| Self::from((*pol).clone()))
+                    .collect(),
+            ),
+        }
+    }
+}
+
+#[cfg(feature = "compiler")]
+impl<Pk: MiniscriptKey> From<Policy<Pk>> for PolicyArc<Pk> {
+    fn from(p: Policy<Pk>) -> Self {
+        match p {
+            Policy::Unsatisfiable => PolicyArc::Unsatisfiable,
+            Policy::Trivial => PolicyArc::Trivial,
+            Policy::Key(pk) => PolicyArc::Key(pk),
+            Policy::After(PackedLockTime(t)) => PolicyArc::After(t),
+            Policy::Older(Sequence(t)) => PolicyArc::Older(t),
+            Policy::Sha256(hash) => PolicyArc::Sha256(hash),
+            Policy::Hash256(hash) => PolicyArc::Hash256(hash),
+            Policy::Ripemd160(hash) => PolicyArc::Ripemd160(hash),
+            Policy::Hash160(hash) => PolicyArc::Hash160(hash),
+            Policy::And(subs) => PolicyArc::And(
+                subs.iter()
+                    .map(|sub| Arc::new(Self::from(sub.clone())))
+                    .collect(),
+            ),
+            Policy::Or(subs) => PolicyArc::Or(
+                subs.iter()
+                    .map(|(odds, sub)| (*odds, Arc::new(Self::from(sub.clone()))))
+                    .collect(),
+            ),
+            Policy::Threshold(k, subs) => PolicyArc::Threshold(
+                k,
+                subs.iter()
+                    .map(|sub| Arc::new(Self::from(sub.clone())))
+                    .collect(),
+            ),
+        }
+    }
+}
+
 /// Detailed Error type for Policies
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub enum PolicyError {
@@ -205,9 +304,14 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
     ///                                    B         C
     ///
     /// gives the vector [(2/3, A), (1/3 * 3/4, B), (1/3 * 1/4, C)].
+    ///
+    /// ## Constraints
+    ///
+    /// Since this splitting might lead to exponential blow-up, we constraint the number of
+    /// leaf-nodes to [`MAX_COMPILATION_LEAVES`].
     #[cfg(feature = "compiler")]
     fn to_tapleaf_prob_vec(&self, prob: f64) -> Vec<(f64, Policy<Pk>)> {
-        match *self {
+        match self {
             Policy::Or(ref subs) => {
                 let total_odds: usize = subs.iter().map(|(ref k, _)| k).sum();
                 subs.iter()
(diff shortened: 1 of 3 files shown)


src/policy/concrete.rs
```diff
@@ -278,6 +278,7 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
             )),
             _ => {
                 let (internal_key, policy) = self.clone().extract_key(unspendable_key)?;
+                policy.check_tapleaf()?;
```
sanket1729, line 411: Should update the name of the function. This implies that we can checking something on tapleafs. Perhaps, `check_num_tapleaves`?
I would prefer to inline the `check_num_tapleaves` here, but up to you.

src/policy/concrete.rs
```diff
@@ -501,6 +502,26 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+            Policy::Or(subs) => subs.iter().map(|(_prob, pol)| pol.num_tap_leaves()).sum(),
+            Policy::Threshold(k, subs) if *k == 1 => combination(subs.len(), *k),
+            _ => 1,
+        }
+    }
+
+    /// Check on the number of TapLeaves
+    #[cfg(feature = "compiler")]
+     fn check_tapleaf(&self) -> Result<(), Error> {
+        if self.num_tap_leaves() > 100_000 {
```
sanket1729: Make this is local constant `MAX_COMPILATION_TAP_LEAVES`

src/policy/concrete.rs
```diff
@@ -501,6 +502,26 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+            return Err(errstr("Too many Tapleaves"));
```
sanket1729, line 702: Let's use a dedicated error variant instead of errstr

src/policy/concrete.rs
```diff
@@ -918,5 +939,12 @@ fn with_huffman_tree<Pk: MiniscriptKey>(
         .pop()
         .expect("huffman tree algorithm is broken")
         .1;
-    Ok(node)
+
+#[cfg(feature = "compiler")]
+fn combination(n: usize, k: usize) -> usize {
```
sanket1729: Whenever we call this, we need to make sure that we are having an overflow.

src/policy/concrete.rs
```diff
@@ -501,6 +502,26 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
         }
     }
 
+    /// Get the number of [TapLeaf][`TapTree::Leaf`] considering exhaustive root-level [OR][`Policy::Or`]
+    /// and [Thresh][`Policy::Threshold`] disjunctions for the TapTree.
+    #[cfg(feature = "compiler")]
+    fn num_tap_leaves(&self) -> usize {
+        match self {
+            Policy::Or(subs) => subs.iter().map(|(_prob, pol)| pol.num_tap_leaves()).sum(),
+            Policy::Threshold(k, subs) if *k == 1 => combination(subs.len(), *k),
```
sanket1729: What happens when there is overflow here? We should use all the checked_sum, `checked_mul` variants here to check for this. Return Option<usize> where we return None when there is overflow

src/miniscript/limits.rs
```diff
@@ -50,3 +50,6 @@ pub const MAX_BLOCK_WEIGHT: usize = 4000000;
 /// Maximum pubkeys as arguments to CHECKMULTISIG
 // https://github.com/bitcoin/bitcoin/blob/6acda4b00b3fc1bfac02f5de590e1a5386cbc779/src/script/script.h#L30
 pub const MAX_PUBKEYS_PER_MULTISIG: usize = 20;
+
+/// Maximum TapLeafs allowed in a compiled TapTree
+pub const MAX_TAPLEAFS_PER_TAPTREE: usize = 10_000;
```
sanket1729, line 55: Rename this constant to `MAX_COMPILATION_LEAVES`. As this constant is exposed publicly, others importing it might be confused `MAX_TAPLEAFS_PER_TAPTREE = 10000`  and assume this as maximum leaves in a taptreee. 

This is an artificial limitation of the compiler. 1) We should not expose it as `pub` and 2) Rename it as suggested above

src/policy/concrete.rs
```diff
@@ -186,9 +187,14 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+    /// ## Constraints
+    ///
+    /// Since this splitting might lead to exponential blow-up, we constraint the number of
+    /// leaf-nodes to [`MAX_TAPLEAFS_PER_TAPTREE`].
     #[cfg(feature = "compiler")]
     fn to_tapleaf_prob_vec(&self, prob: f64) -> Vec<(f64, Policy<Pk>)> {
-        match *self {
+        match self {
             Policy::Or(ref subs) => {
                 let total_odds: usize = subs.iter().map(|(ref k, _)| k).sum();
```
sanket1729, line 199: remove the ref k here.

sanket1729: As in, remove the `ref` . It should clean up * (*k) code in the next line

src/policy/concrete.rs
```diff
@@ -198,17 +204,141 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
-            Policy::Threshold(k, ref subs) if k == 1 => {
+            Policy::Threshold(k, ref subs) if *k == 1 => {
                 let total_odds = subs.len();
                 subs.iter()
                     .map(|policy| policy.to_tapleaf_prob_vec(prob / total_odds as f64))
                     .flatten()
                     .collect::<Vec<_>>()
             }
-            ref x => vec![(prob, x.clone())],
+            // Instead of the fixed-point algorithm, consider direct k-choose-n split if limits
```
sanket1729, line 214: I don't see why this comment is relevant here. This function is only splitting ors AFAIU.

src/policy/concrete.rs
```diff
@@ -198,17 +204,141 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+            // don't exceed
+            x => vec![(prob, x.clone())],
+        }
+    }
+
+    /// Given a [`Policy`], return a vector of policies whose disjunction is isomorphic to the initial one.
+    /// This function is supposed to incrementally expand i.e. represent the policy as disjunction over
+    /// sub-policies output by it. The probability calculations are similar as
+    /// [to_tapleaf_prob_vec][`Policy::to_tapleaf_prob_vec`]
```
sanket1729, line 223: Why do you have the link twice?

src/policy/concrete.rs
```diff
@@ -198,17 +204,141 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+    #[cfg(feature = "compiler")]
+    fn enumerate_pol(&self, prob: f64) -> Vec<(f64, Policy<Pk>)> {
+        match self {
+            Policy::Or(subs) => {
+                let total_odds = subs.iter().fold(0, |acc, x| acc + x.0);
+                subs.iter()
+                    .map(|(odds, pol)| (prob * *odds as f64 / total_odds as f64, pol.clone()))
```
sanket1729, line 230: Should this not be recursive?

src/policy/concrete.rs
```diff
@@ -198,17 +204,141 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+                    .collect::<Vec<_>>()
+            }
+            Policy::Threshold(k, subs) if *k == 1 => {
+                let total_odds = subs.len();
+                subs.iter()
+                    .map(|pol| (prob / total_odds as f64, pol.clone()))
```
sanket1729, line 236: Same comment here

sanket1729 (commented):
Did some partial review.

src/policy/concrete.rs
SarcasticNastik: Yes, the function name is a misnomer. Will change the name as suggested.

src/policy/concrete.rs
SarcasticNastik: This function isn't supposed to be there. Seems I messed up while squashing the commits, my bad!

src/policy/concrete.rs
```diff
@@ -944,3 +1120,30 @@ fn with_huffman_tree<Pk: MiniscriptKey>(
...
+/// Enumerate a [Thresh][`Policy::Threshold`](k, ..n..) into `n` different thresh.
+///
+/// ## Strategy
+///
+/// `thresh(k, x_1...x_n) := thresh(1, thresh(k, x_2...x_n), thresh(k, x_1x_3...x_n), ...., thresh(k, x_1...x_{n-1}))`
+/// by the simple argument that choosing `k` conditions from `n` available conditions might not contain
+/// any one of the conditions exclusively.
+#[cfg(feature = "compiler")]
+fn generate_combination<Pk: MiniscriptKey>(
+    policy_vec: &Vec<Policy<Pk>>,
```
sanket1729, line 1133: We should use Rc<Policy> instead of cloning it here multiple times. This should help in space and time complexity

sanket1729: We would need a new data structure with Arc's here to store these instead of Policy. But we do not want so many clones here. This might be why everything is slow.

You can create something similar to `Policy` but with Arcs instead of Owned sub-policies.

sanket1729 (commented):
code logic ACK, but would need some re-factors that will make it faster

src/miniscript/limits.rs
```diff
@@ -50,3 +50,6 @@ pub const MAX_BLOCK_WEIGHT: usize = 4000000;
 /// Maximum pubkeys as arguments to CHECKMULTISIG
 // https://github.com/bitcoin/bitcoin/blob/6acda4b00b3fc1bfac02f5de590e1a5386cbc779/src/script/script.h#L30
 pub const MAX_PUBKEYS_PER_MULTISIG: usize = 20;
+
+/// Maximum TapLeafs allowed in a compiled TapTree
+pub(crate) const MAX_COMPILATION_LEAVES: usize = 10_000;
```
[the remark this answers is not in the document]
SarcasticNastik: Should we move it to concrete.rs since that's where the tr compiler is?

SarcasticNastik:
@sanket1729 I've added a new structure `PolicyArc<Pk>` according to your suggestions. But I didn't find any significant improvements for the compilation time.

sanket1729 (commented):
mostly ACK. found a bug when dealing with more than MAX_COMPILATION_LEAVES part.

See the last commit here https://github.com/sanket1729/rust-miniscript/tree/pr418.

We are really close to merging this. I think you would run in --release mode to see the performance difference.

src/policy/concrete.rs
```diff
@@ -343,6 +503,134 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+                    break 'outer;
+                } else {
+                    // Either node is enumerable, or we have
+                    // Mark all non-enumerable nodes to remove
+                    to_del.push((p.0 .0, Arc::clone(pol)));
+                }
+            }
+
+            // --- Sanity Checks ---
+            if enum_len > MAX_COMPILATION_LEAVES || *curr_policy == PolicyArc::Unsatisfiable {
```
SarcasticNastik, line 597: We **do not** enumerate the sub-policy under consideration since we either exceed `MAX_COMPILATION_LEAVES` or the policy itself compiled to `Unsatisfiable`.

sanket1729: This was causing bugs because the enumeration was incorrect. I think the logic here would need to be improved. If you replace thresh(51,...) with thresh(49) you would see a crash

src/policy/concrete.rs
sanket1729: This is still not addressed.

src/policy/concrete.rs
```diff
@@ -343,6 +503,140 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+                // Root-level enumeration exceeds the limit, so push back the initial policy
+                if ret.is_empty() {
+                    ret.push((1.0, arc_self));
```
sanket1729, line 600: This is not the expected behavior. We need to split as much as we can until we hit the limit.

sanket1729 (commented):
Last few comments.

src/policy/concrete.rs
SarcasticNastik: The variable `enum_len` stores the vector length if we consider the enumeration of  *plausible* node. If this exceeds limit, we don't replace the initial *plausible* sub-policy by its enumerations (which is done after the check). I just handled the base-case where the root-level enumeration itself exceeds the limit, where we just return the policy itself.

src/policy/concrete.rs
SarcasticNastik: This was a *major* bug. I've fixed it in the 9c8a1c3.

sanket1729 (commented):
ACK f9901ab6ad7c75984b3fc51f7b1f8865bfcab570. Can you also rebase to latest master? I think we are good to go here

sanket1729:
Unless @kanishk779 found something else is review here.

sanket1729:
I will wait a couple of days for @kanishk779 to respond. Will merge this later

sanket1729:
@SarcasticNastik , can you also update the PR description?

sanket1729 (dismissed):
ACK 265b063f01fdb5983676674aa506b3e491553d67

sanket1729:
@SarcasticNastik , sorry for this. But I promise, this would be last rebase

sanket1729 (approved):
ACK 97de242709178f2c75c8bffbe44718356aae1e71

Issue #270, opened 2021-09-18
AST Cleanup: Split After / Older into more specific types?


JeremyRubin:
Suggestion:

While keeping the parsing and output intact for compatibility, we could modify our internal representations of After and Older to be:

AfterHeight(), AfterTime(), WaitBlocks(), WaitDuration() or an internal enum (e.g., After(Height | Time), Wait(Blocks | Duration)).

In Sapio I accomplish this at the [user level](https://github.com/sapio-lang/sapio/blob/master/sapio-base/src/timelocks.rs), but it would probably make sense for those branches to be distinct. This also allows (as I've done in the library) to narrow the types used in the internal representations from e.g. u32s for relative timelocks to u16s making the values correct by construction. This also rules out using non-standardized values.

Users of the Library from Rust would have the improved interface, which would make it easier to write "correct" generic code as users would be able to be unconcerned with checking inputs u32s for well formedness to a Older/After clause, while users using this as a string interface would see the old API (unless we modify the spec to have the 4 types of clause).

If the more generic API of Older/After is desired, this could be accomplished with a function Older/After (or type with Into) that takes a u32 and downcasts into the appropriate type.

sanket1729:
Concept ACK. This will clean up some code here too. I can work on this post tapscript

apoelstra:
concept ack from me too

apoelstra:
I think we should do this using an internal enum, which we can convert from a u32 and do checks at that time

JeremyRubin:
if you prefer that route, you'd be welcome to take the code from sapio (i can relicense that segment under MIT license) https://github.com/sapio-lang/sapio/blob/master/sapio-base/src/timelocks.rs, but that might be a bit extra for the goals here (but i've found it pretty ergonomic in my personal use).

apoelstra:
Thanks for the offer to relicense -- though FYI we are CC0, not MIT. I don't have time today to dive into whether we'd want to go that route or implement our own (smaller) enum.

JeremyRubin:
CC0 is fine for that as well, if desired.

tcharding:
With the use of bitcoin v0.29 we now use `LockTime` types for `After`/`Older` - I believe this issue can be closed.

apoelstra:
I agree -- @JeremyRubin feel free to re-open but I think we have finally done this (and we fixed a pile of bugs in the process :))

sanket1729:
[...]

PR #419, opened 2022-05-25
Fix bug in absolute timelock usage

tcharding:
Currently if we mix up height/time absolute timelocks when filtering policies the result is incorrect (to the best of my understanding).

Add a `timelock` module that includes a single public function for checking if absolute timelocks are the same unit. Use the new function to fix a bug in policy filtering.

- Patch 1 is a docs bug fix
- Patch 2 is the bug fix
- Patch 3 is a unit test patch that fails if its moved before patch 2. Please review the unit test carefully to make sure I'm not confused.

## Note

There is [ongoing discussion](https://github.com/rust-bitcoin/rust-bitcoin/pull/994) around trying to design a suitable timelock API. This PR is an attempt to make some forward progress by taking baby steps _and_ making objective improvements. I decided to do it here in miniscript because it will be easier to iterate on and more obvious when there are concrete usage examples along with each change.

cc dpc

### DIFF
--- a/src/timelock.rs
+++ b/src/timelock.rs
@@ -1,0 +1,21 @@
+//! Various functions for manipulating Bitcoin timelocks.
+
+use crate::miniscript::limits::LOCKTIME_THRESHOLD;
+
+/// Returns true if `a` and `b` are the same unit i.e., both are block heights or both are UNIX
+/// timestamps. `a` and `b` are nLockTime values.
+pub fn absolute_timelocks_are_same_unit(a: u32, b: u32) -> bool {
+    n_lock_time_is_block_height(a) == n_lock_time_is_block_height(b)
+}
+
+// https://github.com/bitcoin/bitcoin/blob/9ccaee1d5e2e4b79b0a7c29aadb41b97e4741332/src/script/script.h#L39
+
+/// Returns true if nLockTime `n` is to be interpreted as a block height.
+pub fn n_lock_time_is_block_height(n: u32) -> bool {
+    n < LOCKTIME_THRESHOLD
+}
+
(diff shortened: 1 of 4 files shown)


src/timelock.rs
```diff
@@ -0,0 +1,22 @@
+//! Various functions for manipulating Bitcoin timelocks.
+
+use crate::miniscript::limits::LOCKTIME_THRESHOLD;
+
+/// Returns true if `a` and `b` are the same unit i.e., both are block heights or both are UNIX
+/// timestamps. `a` and `b` are nLockTime values.
+pub fn absolute_timelocks_are_same_unit(a: u32, b: u32) -> bool {
+    n_lock_time_is_block_height(a) && n_lock_time_is_block_height(b)
```
dpc, line 8: I wonder if `n_lock_time_is_block_height(a) == n_lock_time_is_block_height(b)` is more readable to other people, or just me.

tcharding: I like it!

tcharding:
Force push is `cargo fmt` and suggestion from @dpc.

tcharding:
Oh and I made the helper functions public because clippy warns otherwise, I think there is value in having both even though one is now not used. I specifically put the `<` and `>=` there, explicitly in front of everyone that sees it, instead of negating a function call in one to the other.

sanket1729:
Thanks for this, will catch up the lock time discussions upstream this week

sanket1729:
Edited the description to remove the @ mention

sanket1729 (approved):
ACK 42e12168f4c4bcb6f521016930fa6a92289631fd. 

This is clean improvement. A nice step towards #270

PR #420, opened 2022-05-26
Add support for running integration tests via cargo test

### LINKED ISSUE #361: Move to `bitcoind` crate for integration testing
`bitcoind` crate is a more maintained wrapper around bitcoin-core-rpc. We should shift our integration testing framework to use it. It fits it nicely with cargo tests and allows locally running integration tests easily. 

I think we should mark that as optional in the CI build, as it will delay our release cycle adding yet another dependency that needs to be updated every-time there is a rust-bitcoin release.

sanket1729:
There are still some annoying warnings present for unused functions when they are used.
rust-lang/rust#46379

Later commits will remove the hacky integration test setup. 

Running `cargo test` now should also run integration tests

Our testing infrastructure now takes a long time to build because we have more than 100 dependencies (transitive). But all of these are dev dependencies, so our library isn't getting bloated. 

Fixes #361


tests/test_cpp.rs
```diff
@@ -241,3 +240,10 @@ pub fn test_from_cpp_ms(cl: &Client, testdata: &TestData) {
         assert!(num_conf > 0);
     }
 }
+
+#[test]
+fn tests_from_cpp() {
+    let cl = &setup::setup().client;
```
RCasatta, line 246: I am a little puzzled in understanding how this test could work, since once `BitcoinD` struct goes out of scope, the node is stopped

Cargo.toml
```diff
@@ -22,6 +22,11 @@ rand = ["bitcoin/rand"]
 bitcoin = "0.28.0"
 serde = { version = "1.0", optional = true}
 
+[dev-dependencies]
+bitcoind = {version = "0.26.1", features=["22_0"]}
+actual-rand = { package = "rand", version = "0.8.4"}
+bitcoin = { version = "0.28", features = ["rand"]}
```
RCasatta, line 28: since there is rand anyway, can't we use `sign_schnorr_with_aux_rand` and remove the feature rand so that we don't have both bitcoin and bitcoin-with-rand?

apoelstra (approved):
ACK 139324a641d364e6c5f5c3659d81c937da7a935c

tests/test_cpp.rs
sanket1729: setup returns BitcoinD. It never goes out of scope.

Cargo.toml
sanket1729: Adressing in a follow-up PR.

PR #424, opened 2022-06-03
Fix up mixup of age/height

tcharding:
We currently use age/height as identifiers for values related to CLTV and CSV. However we get them mixed up in the function call to `from_txdata`. Do a few improvements to fix the mixup and improve the overall situation.

- Patch 2 is a breaking change, changes a public method name. 
- Patch 1 an 4 are bug fixes.

(This description and PR are totally re-written, the comment directly below is now stale.)

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -605,7 +605,7 @@ where
                 Terminal::After(ref n) => {
                     debug_assert_eq!(node_state.n_evaluated, 0);
                     debug_assert_eq!(node_state.n_satisfied, 0);
-                    let res = self.stack.evaluate_after(n, self.age);
+                    let res = self.stack.evaluate_after(n, self.height);
                     if res.is_some() {
                         return res;
                     }
@@ -613,7 +613,7 @@ where
                 Terminal::Older(ref n) => {
                     debug_assert_eq!(node_state.n_evaluated, 0);
                     debug_assert_eq!(node_state.n_satisfied, 0);
-                    let res = self.stack.evaluate_older(n, self.height);
+                    let res = self.stack.evaluate_older(n, self.age);
                     if res.is_some() {
                         return res;
                     }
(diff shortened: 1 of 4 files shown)


apoelstra:
Are we actually using `age` and `height` to classify things? I would expect two classifications - `height` vs `time` and `relative` vs `absolute` ... where I would expect `age` to be a synonym for `relative`.

apoelstra:
Done reviewing 3d76d02ba95c1372d7126c08e7a1f47fc0743fb9.

I'm still confused by the logic in the last commit ... it seems like we are using "height" vs "age", which as I've mentioned in my first comment, seems like a wrong way to classify. But maybe this should be fixed in a later PR? This PR is definitely an improvement.

tcharding:
I'm trying to detangle things here in rust-miniscript while simultaneously creating the new 'timelocks' API in rust-bitcoin. I'm held back by the fact that I keep getting confused, for the last two weeks I thought 'age' went with CLTV, only after your original comment on this PR did I realize the mix ups we had. 

> I'm still confused by the logic in the last commit ... it seems like we are using "height" vs "age"

I thought I was in line with your original comment, the last commit fixes the fact that we pass the wrong value to `evaluate_older` and `evaluate_after`. It also removes the usage of 'age' in relation to CLTV but keeps the usage of 'age' for CSV.

I'm taking baby steps, mainly focusing on CLTV. The weakness in this approach is I'm not 100% across all the intricacies of CSV right at this minute so this might mean some churn when I move onto that part of the timelocks.

apoelstra:
`evaluate_after` should take an absolute timelock while `evaluate_older` should take a relative one. Height versus time is a different classification.

tcharding:
> evaluate_after should take an absolute timelock while evaluate_older should take a relative one. Height versus time is a different classification.

I think we are on the same page, I believe the PR is correct, and believe that patch 1 and 4 are rectifying real bugs.

tcharding:
I pushed an additional patch that removes `height` in favour of `lock_time`. I left it separate because I think its better to see each tiny change as a separate patch even though they are all interdependent. Can squash the whole thing into a single patch once you are convinced the changes are correct also.

apoelstra (approved):
ACK fec801414f542c5d378bc5a749d1cfac1c5b5276

sanket1729 (approved):
utACK fec801414f542c5d378bc5a749d1cfac1c5b5276

Issue #427, opened 2022-06-07
Think about how MiniscriptKey interacts with Musig keys


sanket1729:
Integrating musig2 keys with the existing API introduces a couple of new challenges:

1) We currently support `Terminal::Pk(Pk)` in our miniscript decoding. This means we should parse `musig(A,B,C)` as a single Pk. 
2) At the same time, we want the translation API to allow to translation of native `A`, `B` and `C`. 

Seems like we might to re-design some parts of the codebase to support it. cc @apoelstra

sanket1729:
It seems like the best solution would be to support parsing as a new struct `KeyExpr`.

```rust
struct KeyExpr<Pk: MiniscriptKey> {
      SingleKey(Pk),
      Musig(Vec<KeyExpr<Pk>>),
}
```

1) Then we change Terminal::Pk(Pk) => Terminal::Pk(KeyExpr<Pk>). 
2) For Pkh, as discussed elsewhere first, we need to change from Pkh(Pk::Hash) to Pkh(Option<Pk>, Pk::Hash). That will changed to `Pkh(Option<KeyExpr<Pk>>, Pk::Hash)`
3) Multi will change to use `KeyExpr` as above

apoelstra:
This sounds like the right direction to me. The keyhash stuff is a bit ugly but I think that's inherent to the problem domain.

portlandhodl:
@sanket1729  I assume you mean `enum` instead of `struct` here

```rust
struct KeyExpr<Pk: MiniscriptKey> {
      SingleKey(Pk),
      Musig(Vec<KeyExpr<Pk>>),
}
```

PR #431, opened 2022-06-09
Refactor `PkH` fragment to support `Pk`

SarcasticNastik:
As suggested in #427, this refactor helps to incorporate musig in the codebase.


src/policy/compiler.rs
```diff
@@ -840,7 +840,14 @@ where
             insert_wrap!(AstElemExt::terminal(Terminal::True));
         }
         Concrete::Key(ref pk) => {
-            insert_wrap!(AstElemExt::terminal(Terminal::PkH(pk.to_pubkeyhash())));
+            insert_wrap!(AstElemExt::terminal(Terminal::PkH(
```
SarcasticNastik, line 843: Ideally this `insert_wrap` should have been avoided (since we know `Concrete::Key` during compilation), but we have a change in internal representation for `Terminal::PkH` (found when one of tests fail in absence of this statement). As I understand, there can be 2 solutions:
1. Implement `PartialEq`, `Eq` for  `Terminal::PkH` to only check equality in the field `Pk::Hash`. We can still access the `Pk` entry while having proper checks.
2. Keep both the compilations. Although I think this might give rise to ambiguity in the compilation and parsing miniscript from string can be an issue (internal representation of `Terminal::PkH` can't be inferred from a string representation as of now).

sanket1729:
@apoelstra , thinking about your comment here: https://github.com/bitcoin/bitcoin/issues/24114#issuecomment-1017775233. I am thinking it would make more sense to change `Pkh(Pk::Hash)` to `Pk` and add a new Terminal::RawPkh(Pk::Hash), just for parsing. 

1) FromStr would not parse these. (until `rawpkh` is decided to supported upstream)
2) Compiler will always `Pkh(Pk)`
3) Parsing from the script would create a successful miniscript, but we will mark it insane. Meaning that it cannot be used in descriptors.

apoelstra:
I like it. We could then also change `semantic::Policy` to have real public keys rather than hashes, which would simplify Simplicity interop.

sanket1729:
@SarcasticNastik, we need to modify this as follows:

1) Change `Pkh(Pk::Hash)` to `Pkh(Pk)`
2) Add a RawPkh node `RawPkh(Pk::Hash)` as [commented](https://github.com/rust-bitcoin/rust-miniscript/pull/431#issuecomment-1151470715)

src/interpreter/mod.rs
```diff
@@ -595,7 +595,15 @@ where
                         return res;
                     }
                 }
-                Terminal::PkH(ref pkh) => {
+                Terminal::PkH(ref pk) => {
+                    debug_assert_eq!(node_state.n_evaluated, 0);
+                    debug_assert_eq!(node_state.n_satisfied, 0);
+                    let res = self.stack.evaluate_pk(&mut self.verify_sig, pk);
```
sanket1729, line 601: The evaluation would still be evaluate_pkh. We would pass pk.to_pubkeyhash() as an arg

src/miniscript/astelem.rs
```diff
@@ -86,7 +86,8 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
     {
         match *self {
             Terminal::PkK(ref p) => pred(ForEach(p)),
-            Terminal::PkH(ref p) => todo!("KeyHash should contain Pk"),
+            Terminal::PkH(ref p) => pred(ForEach(p)),
+            Terminal::RawPkH(ref p) => todo!("Should we require to iterate over the Hashes too?"),
```
sanket1729, line 90: We just ignore these. As these are not known keys.

src/miniscript/astelem.rs
```diff
@@ -255,7 +257,8 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> fmt::Debug for Terminal<Pk, Ctx> {
         } else {
             match *self {
                 Terminal::PkK(ref pk) => write!(f, "pk_k({:?})", pk),
-                Terminal::PkH(ref pkh) => write!(f, "pk_h({:?})", pkh),
+                Terminal::PkH(ref pk) => write!(f, "pk_h({{{:?}}} {:?})", pk, &pk.to_pubkeyhash()),
```
sanket1729, line 260: Should only format the `pk`

src/miniscript/astelem.rs
```diff
@@ -312,7 +315,8 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> fmt::Display for Terminal<Pk, Ctx> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
             Terminal::PkK(ref pk) => write!(f, "pk_k({})", pk),
-            Terminal::PkH(ref pkh) => write!(f, "pk_h({})", pkh),
+            Terminal::PkH(ref pk) => write!(f, "pk_h({})", &pk.to_pubkeyhash()),
```
sanket1729, line 318: Should only format the `pk`

src/miniscript/astelem.rs
```diff
@@ -372,9 +376,12 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> fmt::Display for Terminal<Pk, Ctx> {
                         if let Terminal::PkK(ref pk) = sub.node {
                             // alias: pk(K) = c:pk_k(K)
                             return write!(f, "pk({})", pk);
-                        } else if let Terminal::PkH(ref pkh) = sub.node {
+                        } else if let Terminal::RawPkH(ref pkh) = sub.node {
```
sanket1729, line 379: We can print this. But we should note in comments that `RawPkH` is currently unsupported in descriptor spec.

src/miniscript/astelem.rs
```diff
@@ -460,9 +469,9 @@ impl_from_tree!(
             ("pk_k", 1) => {
                 expression::terminal(&top.args[0], |x| Pk::from_str(x).map(Terminal::PkK))
             }
-            ("pk_h", 1) => {
-                expression::terminal(&top.args[0], |x| Pk::Hash::from_str(x).map(Terminal::PkH))
-            }
+            ("pk_h", 1) => expression::terminal(&top.args[0], |x| {
+                Pk::Hash::from_str(x).map(Terminal::RawPkH)
```
sanket1729, line 473: We should never parse Pk::Hash. We should only parse Pk and store it in PkH. `RawPkH` should never be created as it is unsupported.

src/policy/compiler.rs
```diff
@@ -1383,7 +1383,7 @@ mod tests {
             keys[7].to_pubkeyhash()
         );
 
-        assert_eq!(ms, ms_comp_res);
+        // assert_eq!(ms, ms_comp_res);
```
sanket1729, line 1386: Why is this commented?

sanket1729 (commented):
Will do another round tomorrow

src/policy/compiler.rs
SarcasticNastik: This test fails because now the `Miniscript::from_str` parses the `pk_h` as `Terminal::RawPkH` rather than `Terminal::PkH` (which is the compiled output).

src/interpreter/mod.rs
SarcasticNastik: The following error occurs in this case:
```bash
error[E0716]: temporary value dropped while borrowed
   --> src/interpreter/mod.rs:601:78
    |
557 | impl<'intp, 'txin: 'intp> Iter<'intp, 'txin>
    |      ----- lifetime `'intp` defined here
...
601 |                     let res = self.stack.evaluate_pkh(&mut self.verify_sig, &pk.to_pubkeyhash());
    |                               -----------------------------------------------^^^^^^^^^^^^^^^^^^-- temporary value is freed at the end of this statement
    |                               |                                              |
    |                               |                                              creates a temporary which is freed while still in use
    |                               argument requires that borrow lasts for `'intp`
```
Here, we'd need to either create the hash outliving the scope (since `'intp` is the liftime required for it's reference) or change some of our current APIs. If there's any solution, kindly let me know!

src/miniscript/astelem.rs
SarcasticNastik: Right, will address this shortly!

SarcasticNastik: Right, will address this shortly!

src/miniscript/mod.rs
```diff
@@ -607,14 +623,14 @@ mod tests {
...
-        ms_attributes_test("c:or_i(andor(c:pk_h(fcd35ddacad9f2d5be5e464639441c6065e6955d),pk_h(9652d86bedf43ad264362e6e6eba6eb764508127),pk_h(06afd46bcdfd22ef94ac122aa11f241244a37ecc)),pk_k(02d7924d4f7d43ea965a465ae3095ff41131e5946f3c85f79e44adbcf8e27e080e))", "6376a914fcd35ddacad9f2d5be5e464639441c6065e6955d88ac6476a91406afd46bcdfd22ef94ac122aa11f241244a37ecc886776a9149652d86bedf43ad264362e6e6eba6eb7645081278868672102d7924d4f7d43ea965a465ae3095ff41131e5946f3c85f79e44adbcf8e27e080e68ac", true, true, true, 17, 5);
+        ms_attributes_test("c:or_i(and_v(v:older(16),pk_h(03daed4f2be3a8bf278e70132fb0beb7522f570e144bf615c07e996d443dee8729)),pk_h(02352bbf4a4cdd12564f93fa332ce333301d9ad40271f8107181340aef25be59d5))", "6360b26976a91420d637c1a6404d2227f3561fdbaff5a680dba648886776a9148f9dff39a81ee4abcbad2ad8bafff090415a2be88868ac", true, true, true, 12, 3);
```
sanket1729, line 629: This is okay for this PR. As a followup, we should be using the same test vectors from https://github.com/sipa/miniscript/blob/fa7ea77b19d76738addce71430cb8ff50ee846b9/bitcoin/test/miniscript_tests.cpp#L461-L475 .

Now, we don't have to manually compute the hashes here

sanket1729 (commented):
Mostly ACK, but the history is unclean. Can you remove the previous commits?

We can do the updating the miniscript tests stuff in a follow up PR.

src/lib.rs
```diff
@@ -458,21 +458,13 @@ where
...
 
-/// Either a key or a keyhash
-pub enum ForEach<'a, Pk: MiniscriptKey> {
-    /// A key
-    Key(&'a Pk),
-    /// A keyhash
-    Hash(&'a Pk::Hash),
-}
+/// Either a key or keyhash, but both contain Pk
+pub struct ForEach<'a, Pk: MiniscriptKey>(&'a Pk);
```
sanket1729, line 462: Note to reviewers. We deliberately do not iterate over RawPkH here. In a follow-up PR, we will make `Miniscript::from_ast` fail when we reach a `RawPkh` terminal. 

We can update this whenever partial descriptors are supported upstream

sanket1729:
Follow things that we can do after this (I will do some of them):

1) Make the sanity_checks fail when Terminal::RawPkH is used. Partial descriptors are not specified yet and are only used internally in the codebase. 
	- Make sure scripts when parsed from blockchain are correctly parsed using the RawPkH terminal. The interpreter should be able to produce partial descriptors. This might need some revisiting of things that we did in the PR. We can get a better idea as we started here
	- Make sure that the user cannot create `RawPkH` descriptors directly. 1) Compiler 2) String or 3) Script. I think we are able to do this in this PR with correct usage of sanity_checks.

2) Update the translation APIs to remove the pkh altogether. The translation stuff is really useful when dealing with descriptors and we do not want users to reason about `rawpkh` as it contains information that they don't know about. 

3) Change the semantic Policy Pk type from Key::Hash to Key

4) Clean up the iterator API so that everything is covered under `pk`.

src/miniscript/decode.rs
```diff
@@ -134,7 +134,9 @@ pub enum Terminal<Pk: MiniscriptKey, Ctx: ScriptContext> {
     /// `<key>`
     PkK(Pk),
     /// `DUP HASH160 <keyhash> EQUALVERIFY`
-    PkH(Pk::Hash),
+    PkH(Pk),
+    /// Only for Parsing PkH
+    RawPkH(Pk::Hash),
```
apoelstra, line 139: In 36ff0d2bf8b9df789a5581f6a10f9f4457803331:

Nit: could say "only for parsing PkH from Script"

src/policy/mod.rs
```diff
@@ -124,8 +124,8 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Liftable<Pk> for Miniscript<Pk, Ctx>
 impl<Pk: MiniscriptKey, Ctx: ScriptContext> Liftable<Pk> for Terminal<Pk, Ctx> {
     fn lift(&self) -> Result<Semantic<Pk>, Error> {
         let ret = match *self {
-            Terminal::PkK(ref pk) => Semantic::KeyHash(pk.to_pubkeyhash()),
-            Terminal::PkH(ref pkh) => Semantic::KeyHash(pkh.clone()),
+            Terminal::PkK(ref pk) | Terminal::PkH(ref pk) => Semantic::KeyHash(pk.to_pubkeyhash()),
+            Terminal::RawPkH(ref pkh) => Semantic::KeyHash(pkh.clone()),
```
apoelstra, line 128: In 36ff0d2bf8b9df789a5581f6a10f9f4457803331:

I guess this is for a future PR, but I'd kinda like to make `semantic::Policy` have only `Pk`s, not `Pk::Hash`es, and then refuse to lift any scripts that had `RawPkH` in it. We could provide a transformer that took a `Miniscript<Pk>` to a `Miniscript<Pk::Hash>` maybe, converting `RawPkH`s to `PkH`s along the way, to get the old behavior.

apoelstra:
ack 36ff0d2bf8b9df789a5581f6a10f9f4457803331 aside from the comments i made

src/lib.rs
SarcasticNastik: I remember @sanket1729 and I discussed abouting removing this later on, but I don't see any reason why I can't do it here. I'm pushing another commit on top of 5cb2bcf to remove `ForEach`. If accepted, I can squash the commit anytime!

apoelstra (approved):
ACK 8d0cf061e4fcc3f7784116e43b2d2afb712cd066

sanket1729 (approved):
ACK 8d0cf061e4fcc3f7784116e43b2d2afb712cd066. The last commit message should be `Remove ForEach struct` instead of `Remove ForEach trait`. But won't holdoff this over that. 

Thanks for keeping with the reviews

src/policy/mod.rs
sanket1729: Yup, the future TODOs are listed https://github.com/rust-bitcoin/rust-miniscript/pull/431#issuecomment-1157153192

PR #433, opened 2022-06-12
Musig2

kanishk779:
Changes made in parsing logic to support musig2


sanket1729:
The general flow we are going to follow is shown on the latest commit here:
https://github.com/sanket1729/rust-miniscript/tree/musig2_supp

- You can chase the compiler errors to fix all the places where `Musig2` would be used in the Pk
- The current implementation would become the `FromStr` implementation for the `KeyExpr` struct.

PR #434, opened 2022-06-13
Add Hash256 associated type

sanket1729:
- Implement a new hash type that does not display/FromStr with `DISPLAY_BACKWARD`. Should we make this type upstream in bitcoin-hashes?
- Remove the PkTranslator trait as it was causing conflicting implementations downstream. 
- Replace the above with more macros :(

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -152,12 +152,17 @@ pub trait MiniscriptKey: Clone + Eq + Ord + fmt::Debug + fmt::Display + hash::Ha
         false
     }
 
-    /// The associated [`Hash`] type for this pubkey.
+    /// The associated PublicKey Hash for this [`MiniscriptKey`],
+    /// used in the pkh fragment
     type Hash: Clone + Eq + Ord + fmt::Display + fmt::Debug + hash::Hash;
 
-    /// The associated [`sha256::Hash`] type for this [`MiniscriptKey`] type.
-    /// Used in the sha256 fragment
+    /// The associated [`sha256::Hash`] for this [`MiniscriptKey`],
+    /// used in the hash256 fragment.
     type Sha256: Clone + Eq + Ord + fmt::Display + fmt::Debug + hash::Hash;
+
+    /// The associated [`hash256::Hash`] for this [`MiniscriptKey`],
+    /// used in the hash256 fragment.
+    type Hash256: Clone + Eq + Ord + fmt::Display + fmt::Debug + hash::Hash;
 
     /// Converts this key to the associated pubkey hash.
     fn to_pubkeyhash(&self) -> Self::Hash;
@@ -166,6 +171,7 @@ pub trait MiniscriptKey: Clone + Eq + Ord + fmt::Debug + fmt::Display + hash::Ha
 impl MiniscriptKey for bitcoin::secp256k1::PublicKey {
     type Hash = hash160::Hash;
     type Sha256 = sha256::Hash;
+    type Hash256 = hash256::Hash;
 
     fn to_pubkeyhash(&self) -> Self::Hash {
         hash160::Hash::hash(&self.serialize())
@@ -180,6 +186,7 @@ impl MiniscriptKey for bitcoin::PublicKey {
 
     type Hash = hash160::Hash;
     type Sha256 = sha256::Hash;
+    type Hash256 = hash256::Hash;
 
     fn to_pubkeyhash(&self) -> Self::Hash {
         hash160::Hash::hash(&self.to_bytes())
@@ -189,6 +196,7 @@ impl MiniscriptKey for bitcoin::PublicKey {
 impl MiniscriptKey for bitcoin::secp256k1::XOnlyPublicKey {
     type Hash = hash160::Hash;
     type Sha256 = sha256::Hash;
+    type Hash256 = hash256::Hash;
 
     fn to_pubkeyhash(&self) -> Self::Hash {
         hash160::Hash::hash(&self.serialize())
@@ -202,6 +210,7 @@ impl MiniscriptKey for bitcoin::secp256k1::XOnlyPublicKey {
 impl MiniscriptKey for String {
     type Hash = String;
     type Sha256 = String; // specify hashes as string
+    type Hash256 = String;
 
     fn to_pubkeyhash(&self) -> Self::Hash {
         (&self).to_string()
@@ -229,6 +238,9 @@ pub trait ToPublicKey: MiniscriptKey {
 
     /// Converts the generic associated [`MiniscriptKey::Sha256`] to [`sha256::Hash`]
     fn to_sha256(hash: &<Self as MiniscriptKey>::Sha256) -> sha256::Hash;
+
+    /// Converts the generic associated [`MiniscriptKey::Hash256`] to [`hash256::Hash`]
+    fn to_hash256(hash: &<Self as MiniscriptKey>::Hash256) -> hash256::Hash;
 }
 
 impl ToPublicKey for bitcoin::PublicKey {
@@ -243,6 +255,10 @@ impl ToPublicKey for bitcoin::PublicKey {
     fn to_sha256(hash: &sha256::Hash) -> sha256::Hash {
         *hash
     }
+
+    fn to_hash256(hash: &hash256::Hash) -> hash256::Hash {
+        *hash
+    }
 }
 
 impl ToPublicKey for bitcoin::secp256k1::PublicKey {
@@ -255,6 +271,10 @@ impl ToPublicKey for bitcoin::secp256k1::PublicKey {
     }
 
     fn to_sha256(hash: &sha256::Hash) -> sha256::Hash {
+        *hash
+    }
+
+    fn to_hash256(hash: &hash256::Hash) -> hash256::Hash {
         *hash
     }
 }
@@ -280,6 +300,10 @@ impl ToPublicKey for bitcoin::secp256k1::XOnlyPublicKey {
     fn to_sha256(hash: &sha256::Hash) -> sha256::Hash {
         *hash
     }
+
+    fn to_hash256(hash: &hash256::Hash) -> hash256::Hash {
+        *hash
+    }
 }
 
 /// Dummy key which de/serializes to the empty string; useful sometimes for testing
@@ -300,6 +324,7 @@ impl str::FromStr for DummyKey {
 impl MiniscriptKey for DummyKey {
     type Hash = DummyKeyHash;
     type Sha256 = DummySha256Hash;
+    type Hash256 = DummyHash256;
 
     fn to_pubkeyhash(&self) -> Self::Hash {
         DummyKeyHash
@@ -332,6 +357,11 @@ impl ToPublicKey for DummyKey {
 
     fn to_sha256(_hash: &DummySha256Hash) -> sha256::Hash {
         sha256::Hash::from_str("50863ad64a87ae8a2fe83c1af1a8403cb53f53e486d8511dad8a04887e5b2352")
+            .unwrap()
+    }
+
+    fn to_hash256(_hash: &DummyHash256) -> hash256::Hash {
+        hash256::Hash::from_str("50863ad64a87ae8a2fe83c1af1a8403cb53f53e486d8511dad8a04887e5b2352")
             .unwrap()
     }
 }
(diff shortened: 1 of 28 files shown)


apoelstra:
@sipa in your miniscript implementation, do you serialize hashes (e.g. in `sha256(...)` fragments) forward or backward?

sipa:
@apoelstra They're treated as byte arrays, so the order of 2-character-hex units is the same as the order of the bytes as they appear in the script.

apoelstra:
Perfect, thank you!

apoelstra:
Done reviewing 4ffabb78854337eebd39a0c1c94a765cb1e72f2d. ACK except for the one nit.

Arguably this should be two PRs -- the hash256 stuff in the first two commits looks unobjectionable (and it's not even breaking, since amusingly we previously had a shitload of manual byte-reversing logic..). The second two are potentially more controversial, since we introduce a macro-based API to simulate a trait hierarchy. Given the limitations of Rust's type system, I think this is the right approach, but it's definitely not a great user experience to need to know about these macros....  I guess we will just have to be careful to document them well.

nit: there are a typos in the last two commit messages

sanket1729:
@apoelstra split into two PRs.

> The second two are potentially more controversial, since we introduce a macro-based API to simulate a trait hierarchy. Given the limitations of Rust's type system, I think this is the right approach, but it's definitely not a great user experience to need to know about these macros. I guess we will just have to be careful to document them well.

I will document them more carefully in the upcoming PR and will add a macro exporting. My plan is to add an example using the API after we are merged all the associated fields (ripemd160/hash160/older/after).  

The users don't have to know about them, they still implement all of the methods, but it's annoying.

tcharding:
> @tcharding, it occurs to me that if `rustfmt` is breaking(two times now) in nightly. Perhaps we should switch to stable? Are the benefits of nightly worth it over stable?

The reason we moved to nightly was that many of the config options are not available on stable. With the current contention over using rustfmt in rust-bitcoin we need the nightly config options if we are to have any hope of enabling rustfmt there.

I do not think that 'breaking' is the correct term, rustfmt nighltly changes and our config file needed updating. 

I have two suggestions
1. Use a fixed version of nightly on CI and then update it manually periodically
2. Every time rustfmt changes do a PR to fix it

I like (1) because I can do it as part of my routine work and then others are not inconvenienced by rustfmt. We could include the current nightly version somewhere in the repo also so devs can use the same version locally.  

> We can give it a try for a few more weeks to test the stability. If it breaks again, we should consider moving back to the stable version.

I think I answered this above, are you open to the suggestions? Sorry that `rust-miniscript` is taking a hit here because of `rust-bitcoin`.

src/lib.rs
```diff
@@ -159,13 +159,18 @@ pub trait MiniscriptKey: Clone + Eq + Ord + fmt::Debug + fmt::Display + hash::Ha
     /// Used in the sha256 fragment
     type Sha256: Clone + Eq + Ord + fmt::Display + fmt::Debug + hash::Hash;
 
+    /// The associated [`hash256::Hash`] type for this [`MiniscriptKey`] type.
+    /// Used in the sha256 fragment
```
tcharding, line 163:
```suggestion
    /// The associated [`hash256::Hash`] for this `MiniscriptKey`, used in the hash256 fragment.
```

If this form is agreeably, then could use it for the comment on `Sha256` also.

src/test_utils.rs
```diff
@@ -43,6 +43,15 @@ impl Translator<String, bitcoin::PublicKey, ()> for StrKeyTranslator {
         .unwrap();
         Ok(hash)
     }
+
+    fn hash256(&mut self, _hash256: &String) -> Result<hash256::Hash, ()> {
+        // hard coded value
```
tcharding, line 48: This comment adds no value, we can see its a hard coded value. Perhaps comment _why_ this value is used or don't comment at all.

src/miniscript/astelem.rs
```diff
@@ -259,11 +259,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> fmt::Debug for Terminal<Pk, Ctx> {
                 Terminal::After(t) => write!(f, "after({})", t),
                 Terminal::Older(t) => write!(f, "older({})", t),
                 Terminal::Sha256(ref h) => write!(f, "sha256({})", h),
-                Terminal::Hash256(h) => {
-                    let mut x = h.into_inner();
-                    x.reverse();
-                    write!(f, "hash256({})", sha256d::Hash::from_inner(x))
-                }
+                Terminal::Hash256(ref h) => write!(f, "hash256({})", h), // reverse would handled correctly by rust-bitcoin
```
tcharding, line 262: This comment is incorrect, right? We are handling the display here in this lib now, right?

src/miniscript/astelem.rs
```diff
@@ -316,11 +312,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> fmt::Display for Terminal<Pk, Ctx> {
             Terminal::After(t) => write!(f, "after({})", t),
             Terminal::Older(t) => write!(f, "older({})", t),
             Terminal::Sha256(ref h) => write!(f, "sha256({})", h),
-            Terminal::Hash256(h) => {
-                let mut x = h.into_inner();
-                x.reverse();
-                write!(f, "hash256({})", sha256d::Hash::from_inner(x))
-            }
+            Terminal::Hash256(ref h) => write!(f, "hash256({})", h), // reverse correctly handled upstream
```
tcharding, line 315: And this one?

src/lib.rs
```diff
@@ -159,13 +159,18 @@ pub trait MiniscriptKey: Clone + Eq + Ord + fmt::Debug + fmt::Display + hash::Ha
...
+    type Hash256: Clone + Eq + Ord + fmt::Display + fmt::Debug + hash::Hash;
+
```
tcharding, line 165: re: `Hash`, `Sha256`, and `Hash256`

From the docs on the trait its not really clear, at least to me, what is the relation between these three types. Two of them are named after fragments but one is not and they are all 'hashes' but there are other hash fragments as well. Can be add some documentation to try and help devs (*cough* me) to understand this?

src/miniscript/hash256.rs
```diff
@@ -0,0 +1,94 @@
...
+//
+// You should have received a copy of the CC0 Public Domain Dedication
+// along with this software.
+// If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
+//
+
+//! Hash256 type
+//!
+//! Functionality similar to `sha256d` Hash type in bitcoin, but the FromStr
+//! `Display` are *NOT* REVERSE.
```
tcharding, line 18:
```suggestion
//! This module is _identical_ in functionality to the `bitcoin_hashes::sha256d` hash type
//! but the `FromHex` and `ToHex` implementations use `DISPLAY_BACKWARDS = false`.
```

tcharding (commented):
This can be rebased now that https://github.com/rust-bitcoin/rust-miniscript/pull/424 has merged (the first 5 patches of this PR).

I'd prefer to see the rustfmt stuff as a separate PR so other open PRs could use it, I imagine everyone is going to have red in the CI pipeline because of this but its not that big a deal.

High level comment: seems like this `DISPLAY_BACKWARDS` const makes the types in `bitcoin_hashes` less usable. Perhaps it should be a field on the type so users can control the direction of the output? I had to hack around it once recently also.

src/miniscript/hash256.rs
```diff
@@ -0,0 +1,94 @@
...
+};
+
+/// Output of the SHA256d hash function
+#[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Hash)]
+#[repr(transparent)]
+pub struct Hash(
+    #[cfg_attr(
+        feature = "schemars",
+        schemars(schema_with = "crate::util::json_hex_string::len_32")
+    )]
```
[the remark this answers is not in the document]
sanket1729: Wow, embarrassing copy-paste error here.

src/lib.rs
sanket1729: Yep, this is indeed clarification now that we have multiple hashes. I will add some clarification here.

src/miniscript/astelem.rs
sanket1729: Yep, this comment is incorrect. Thanks for catching this!

src/lib.rs
apoelstra: `Sha256` and `Hash256` are two of the five hash types supported by Script (the others are `Sha1` (not supported by miniscript), `Ripemd160` and `Hash160`). Meanwhile `Hash` is a trait from bitcoin_hashes which abstracts over all of these and allows feeding data into it.

(IIRC `Hash` is basically just a copy of `io::Write` but `io::Write` is std-only, for unclear reasons.)

tcharding:
Hey man, I'm not sure what the force push means, did you think this is ready for re-review? The schemars stuff is still in there.

src/lib.rs
sanket1729: @apoelstra , I think the `Hash` is @tcharding is talking about here is MiniscirptKey::Hash which is different from `bitcoin_hashes::Hash`. 

The `Hash` represents the associated type corresponding to PkH fragment. Perhaps, with multiple hashes in the picture and considering #431  we should rename that to `RawPkHash`? 

I think we should re-evaluate whether we even need a Pk::Hash type anymore. That can be useful for abstractions over RawPkH in the future when we have partial descriptors. 

We can consider renaming that as a part of a separate PR.

src/test_utils.rs
sanket1729: Removing the comment

sanket1729 (commented):
Sorry about the previous push. This is ready for review now @tcharding

src/miniscript/hash256.rs
sanket1729: Changed and also added a comment about `FromStr` and `Display`

src/lib.rs
sanket1729: Done

sanket1729:
@apoelstra, let's try to merge #431 before this.

apoelstra:
Yeah, renaming `MiniscriptKey::Hash` sounds like a good idea :)

sanket1729:
@apoelstra, we could also remove the `MiniscriptKey::RawPkH` completely. And make it fixed hash160::Hash. I don't see any specific reason why we would want to reason about raw things in an abstract. 

But on the other hand, we already have the code. So part of me wants to keep it in case we need it in the future after partial descriptors are officially supported.

apoelstra:
Yeah, let's keep it for now.

sanket1729:
@tcharding , @apoelstra , ready for review again.

apoelstra (approved):
ACK e49b6d8f7f748014e84aa81eff7db2429cfd7647

sanket1729:
Rebased after #417 .

sanket1729:
Another problem that I am noting with merging the imports is that there are always conflicts when merging PRs that touch the same file. 

PR1:

```rust
use crate::{A, B, SomethingFromPR1, T, U};
```

```rust
use crate::{A, AnotherThingFromPR2, B, T, U};
```

While previously it was no conflicts
```rust
use crate::SomethingFromPR1;
```

```rust
use crate::AnotherThingFromPR2;
```

tcharding:
> Another problem that I am noting with merging the imports is that there are always conflicts when merging PRs that touch the same file.

Yes you are correct. Its a trade off right, the more you merge the imports (not at all, merge last module, merge crate `use foo::bar::{{baz::A, B}, C}`) the more conflicts we get. The less we merge imports the more lines if code. There is no 'correct'. I personally like the `imports_garularity = "Module"` level, but its totally subjective.

sanket1729:
> Although my view was formed in projects that do not use the '2 explicit acks to merge' policy so perhaps the additional merge conflicts are not good in the rust-bitcoin stack?

I tend to agree. I think we should have this in the rust-bitcoin/rust-bitcoin.
1) It has more open PRs
2) Requires 2 ACKs

Let us keep the merging import feature in rust-miniscript for a few more weeks and learn a few more things.

tcharding (dismissed):
ACK 9d83a8f459077f464fafc6a852aa0892d7088e7a

apoelstra (approved):
ACK 905c5dc2728479501aa59097f902fba5ad846737

PR #442, opened 2022-07-01
Explicitly do not support 16 bit architectures

tcharding:
We cast `u32` to `usize` in a bunch of places, this will cause strange behaviour because of truncation on 16 bit architectures.

Add a compile time check to explicitly not support 16 bit architectures.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -90,6 +90,11 @@
 #![deny(unused_imports)]
 #![deny(missing_docs)]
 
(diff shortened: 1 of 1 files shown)


Kixunil (approved):

apoelstra (approved):
utACK d02c31309985cbae8af4ea0f61272c036737a639

apoelstra (approved):
ACK d02c31309985cbae8af4ea0f61272c036737a639

PR #443, opened 2022-07-01
Add initial support for KeyExpr

kanishk779:
As of now, miniscript does not have support for handling Multi-signature keys. The changes introduced in this PR would be the first steps for supporting such keys.

The idea is that we will have support for legacy keys `SingleKey(Pk)` and Musig Keys `Musig(vec<KeyExpr>)`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ edition = "2018"
 
 [features]
 default = ["std"]
-std = ["bitcoin/std", "bitcoin/secp-recovery"]
+std = ["bitcoin/std", "bitcoin/secp-recovery", "secp256k1-zkp"]
 no-std = ["hashbrown", "bitcoin/no-std"]
 compiler = []
 trace = []
@@ -24,6 +24,7 @@ rand = ["bitcoin/rand"]
 bitcoin = { version = "0.28.1", default-features = false }
 serde = { version = "1.0", optional = true }
 hashbrown = { version = "0.11", optional = true }
+secp256k1-zkp = { git = "https://github.com/sanket1729/rust-secp256k1-zkp", branch = "pr29", optional = true}
 
 [dev-dependencies]
 bitcoind = {version = "0.26.1", features=["22_0"]}
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -593,6 +593,9 @@ where
                 Terminal::PkK(ref pk) => {
                     debug_assert_eq!(node_state.n_evaluated, 0);
                     debug_assert_eq!(node_state.n_satisfied, 0);
+                    let pk = pk
+                        .single_key()
+                        .expect("Musig keys cannot be parsed from Script");
                     let res = self.stack.evaluate_pk(&mut self.verify_sig, *pk);
                     if res.is_some() {
                         return res;
@@ -868,10 +871,11 @@ where
                         // evaluate each key with as a pk
                         // note that evaluate_pk will error on non-empty incorrect sigs
                         // push 1 on satisfied sigs and push 0 on empty sigs
-                        match self
-                            .stack
-                            .evaluate_pk(&mut self.verify_sig, subs[node_state.n_evaluated])
-                        {
+                        let pkk = subs[node_state.n_evaluated]
+                            .single_key()
+                            .expect("Musig keys cannot be parsed from Script");
+                        let res = self.stack.evaluate_pk(&mut self.verify_sig, *pkk);
+                        match res {
                             Some(Ok(x)) => {
                                 self.push_evaluation_state(
                                     node_state.node,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -99,6 +99,9 @@ extern crate alloc;
 #[cfg(not(feature = "std"))]
 extern crate hashbrown;
 
+#[cfg(feature = "std")]
+extern crate secp256k1_zkp;
+
 #[cfg(any(feature = "std", test))]
 extern crate core;
 
@@ -775,6 +778,8 @@ pub enum Error {
     TrNoScriptCode,
     /// No explicit script for Tr descriptors
     TrNoExplicitScript,
+    /// Parsing error for single key
+    SingleKeyParseError,
 }
 
 // https://github.com/sipa/miniscript/pull/5 for discussion on this number
@@ -848,6 +853,7 @@ impl fmt::Display for Error {
             Error::TaprootSpendInfoUnavialable => write!(f, "Taproot Spend Info not computed."),
             Error::TrNoScriptCode => write!(f, "No script code for Tr descriptors"),
             Error::TrNoExplicitScript => write!(f, "No script code for Tr descriptors"),
+            Error::SingleKeyParseError => f.write_str("not able to parse the single key"),
         }
     }
 }
@@ -888,6 +894,7 @@ impl error::Error for Error {
             | BareDescriptorAddr
             | TaprootSpendInfoUnavialable
             | TrNoScriptCode
+            | SingleKeyParseError
             | TrNoExplicitScript => None,
             Script(e) => Some(e),
             AddrError(e) => Some(e),
--- a/src/miniscript/astelem.rs
+++ b/src/miniscript/astelem.rs
@@ -26,6 +26,7 @@ use bitcoin::blockdata::{opcodes, script};
 use sync::Arc;
 
 use crate::miniscript::context::SigType;
+use crate::miniscript::musig_key::KeyExpr;
 use crate::miniscript::types::{self, Property};
 use crate::miniscript::ScriptContext;
 use crate::prelude::*;
@@ -80,7 +81,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
         Pk::RawPkHash: 'a,
     {
         match *self {
-            Terminal::PkK(ref p) => pred(p),
+            Terminal::PkK(ref p) => p.for_each_key(pred),
             Terminal::PkH(ref p) => pred(p),
             Terminal::RawPkH(..)
             | Terminal::After(..)
@@ -112,9 +113,8 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
                     && c.real_for_each_key(pred)
             }
             Terminal::Thresh(_, ref subs) => subs.iter().all(|sub| sub.real_for_each_key(pred)),
-            Terminal::Multi(_, ref keys) | Terminal::MultiA(_, ref keys) => {
-                keys.iter().all(|key| pred(key))
-            }
+            Terminal::Multi(_, ref keys) => keys.iter().all(|key| pred(key)),
+            Terminal::MultiA(_, ref keys) => keys.iter().all(|key| key.for_each_key(&mut *pred)),
         }
     }
 
@@ -125,7 +125,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
         T: Translator<Pk, Q, E>,
     {
         let frag: Terminal<Q, CtxQ> = match *self {
-            Terminal::PkK(ref p) => Terminal::PkK(t.pk(p)?),
+            Terminal::PkK(ref p) => Terminal::PkK(p.translate_pk(t)?),
             Terminal::PkH(ref p) => Terminal::PkH(t.pk(p)?),
             Terminal::RawPkH(ref p) => Terminal::RawPkH(t.pkh(p)?),
             Terminal::After(n) => Terminal::After(n),
@@ -186,7 +186,8 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
                 Terminal::Multi(k, keys?)
             }
             Terminal::MultiA(k, ref keys) => {
-                let keys: Result<Vec<Q>, _> = keys.iter().map(|k| t.pk(k)).collect();
+                let keys: Result<Vec<KeyExpr<Q>>, _> =
+                    keys.iter().map(|k| k.translate_pk(t)).collect();
                 Terminal::MultiA(k, keys?)
             }
         };
@@ -455,7 +456,7 @@ impl_from_tree!(
         }
         let mut unwrapped = match (frag_name, top.args.len()) {
             ("pk_k", 1) => {
-                expression::terminal(&top.args[0], |x| Pk::from_str(x).map(Terminal::PkK))
+                expression::unary(top, Terminal::PkK)
             }
             ("pk_h", 1) => expression::terminal(&top.args[0], |x| Pk::from_str(x).map(Terminal::PkH)),
             ("after", 1) => expression::terminal(&top.args[0], |x| {
@@ -522,15 +523,18 @@ impl_from_tree!(
                     return Err(errstr("higher threshold than there were keys in multi"));
                 }
 
-                let pks: Result<Vec<Pk>, _> = top.args[1..]
-                    .iter()
-                    .map(|sub| expression::terminal(sub, Pk::from_str))
-                    .collect();
-
                 if frag_name == "multi" {
+                    let pks: Result<Vec<Pk>, _> = top.args[1..]
+                        .iter()
+                        .map(|sub| expression::terminal(sub, Pk::from_str))
+                        .collect();
                     pks.map(|pks| Terminal::Multi(k, pks))
                 } else {
                     // must be multi_a
+                    let pks: Result<Vec<KeyExpr<Pk>>, _> = top.args[1..]
+                        .iter()
+                        .map(|sub| KeyExpr::<Pk>::from_tree(sub))
+                        .collect();
                     pks.map(|pks| Terminal::MultiA(k, pks))
                 }
             }
@@ -734,7 +738,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
                 builder = builder.push_ms_key::<_, Ctx>(&keys[0]);
                 builder = builder.push_opcode(opcodes::all::OP_CHECKSIG);
                 for pk in keys.iter().skip(1) {
-                    builder = builder.push_ms_key::<_, Ctx>(pk);
+                    builder = builder.push_ms_key::<_, Ctx>(&pk);
                     builder = builder.push_opcode(opcodes::all::OP_CHECKSIGADD);
                 }
                 builder
@@ -753,7 +757,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
     /// will handle the segwit/non-segwit technicalities for you.
     pub fn script_size(&self) -> usize {
         match *self {
-            Terminal::PkK(ref pk) => Ctx::pk_len(pk),
+            Terminal::PkK(ref pk) => Ctx::key_expr_len(pk),
             Terminal::PkH(..) | Terminal::RawPkH(..) => 24,
             Terminal::After(n) => script_num_size(n as usize) + 1,
             Terminal::Older(n) => script_num_size(n as usize) + 1,
@@ -798,7 +802,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
             Terminal::MultiA(k, ref pks) => {
                 script_num_size(k)
                     + 1 // NUMEQUAL
-                    + pks.iter().map(|pk| Ctx::pk_len(pk)).sum::<usize>() // n keys
+                    + pks.iter().map(|pk| Ctx::key_expr_len(pk)).sum::<usize>() // n keys
                     + pks.len() // n times CHECKSIGADD
             }
         }
--- a/src/miniscript/context.rs
+++ b/src/miniscript/context.rs
@@ -26,6 +26,7 @@ use crate::miniscript::limits::{
     MAX_SCRIPT_SIZE, MAX_STACK_SIZE, MAX_STANDARD_P2WSH_SCRIPT_SIZE,
     MAX_STANDARD_P2WSH_STACK_ITEMS,
 };
+use crate::miniscript::musig_key::KeyExpr;
 use crate::miniscript::types;
 use crate::prelude::*;
 use crate::util::witness_to_scriptsig;
@@ -77,6 +78,8 @@ pub enum ScriptContextError {
     CheckMultiSigLimitExceeded,
     /// MultiA is only allowed in post tapscript
     MultiANotAllowed,
+    /// Musig is only allowed in tapscript and taproot descriptors
+    MusigNotAllowed(String),
 }
 
 #[cfg(feature = "std")]
@@ -100,6 +103,7 @@ impl error::Error for ScriptContextError {
             | TaprootMultiDisabled
             | StackSizeLimitExceeded { .. }
             | CheckMultiSigLimitExceeded
+            | MusigNotAllowed(_)
             | MultiANotAllowed => None,
         }
     }
@@ -179,6 +183,9 @@ impl fmt::Display for ScriptContextError {
             }
             ScriptContextError::MultiANotAllowed => {
                 write!(f, "Multi a(CHECKSIGADD) only allowed post tapscript")
+            }
+            ScriptContextError::MusigNotAllowed(ref err) => {
+                write!(f, "Musig is only allowed in tapscript : err {}", err)
             }
         }
     }
@@ -334,6 +341,13 @@ where
     /// 34 for Segwitv0, 33 for Tap
     fn pk_len<Pk: MiniscriptKey>(pk: &Pk) -> usize;
 
+    /// Get the len of the keyexpr
+    fn key_expr_len<Pk: MiniscriptKey>(pk: &KeyExpr<Pk>) -> usize {
+        match pk {
+            KeyExpr::SingleKey(pk) => Self::pk_len(pk),
+            KeyExpr::MuSig(_) => 33,
+        }
+    }
     /// Local helper function to display error messages with context
     fn name_str() -> &'static str;
 }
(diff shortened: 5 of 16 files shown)


src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,102 @@
...
+use std;
+use std::str::FromStr;
+use std::{error, fmt};
+
+use crate::expression::Tree;
+use crate::MiniscriptKey;
+
+#[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq, Hash)]
+/// Enum for representing keys in miniscript
+pub enum KeyExpr<Pk: MiniscriptKey> {
```
sanket1729, line 11: Can you manually implement Debug?

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,102 @@
+//! Support for multi-signature keys
+use std;
+use std::str::FromStr;
+use std::{error, fmt};
+
+use crate::expression::Tree;
+use crate::MiniscriptKey;
+
+#[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq, Hash)]
+/// Enum for representing keys in miniscript
```
sanket1729, line 10: *representing musig keys

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+impl error::Error for KeyExprError {}
+
+impl<Pk: MiniscriptKey + FromStr> FromStr for KeyExpr<Pk> {
+    type Err = KeyExprError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (key_tree, _) = Tree::from_slice(s).unwrap();
+        fn expr_from_tree<Pk: MiniscriptKey + FromStr>(
```
sanket1729, line 44: You should 

impl<Pk> FromTree for KeyExpr<Pk> instead of this local function

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,102 @@
...
+            tree: &Tree,
+        ) -> Result<KeyExpr<Pk>, KeyExprError> {
+            if tree.name == "musig" {
+                let mut key_expr_vect = vec![];
```
sanket1729, line 48: bit: use `_vec` instead of '_vect'

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,102 @@
...
+                for sub_tree in tree.args.iter() {
+                    let temp_res = expr_from_tree(sub_tree)?;
+                    key_expr_vect.push(temp_res);
+                }
+                Ok(KeyExpr::MuSig(key_expr_vect))
+            } else {
+                if tree.name != "" {
```
sanket1729, line 55: Empty is also a valid string. There is no reason to reject it.

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,102 @@
...
+                    let single_key = match Pk::from_str(tree.name) {
```
sanket1729, line 56: use map_err instead of match

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+}
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_one() {
+        let dummy_key = "musig(A,B,musig(C,musig(D,E)))";
```
sanket1729, line 97: The test case "musig(,,)" should also work. There should be no special rule to rule out empty strings

sanket1729, line 97: nit: name this `musig_key` instead of  `dummy_key`

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,102 @@
...
+        let pk = KeyExpr::<String>::from_str(dummy_key).unwrap();
+        println!("{}", pk);
+        assert_eq!(dummy_key, format!("{}", pk))
```
sanket1729, line 100: can you also more test cases here? Testing single key only. Musig with a single key.

sanket1729 (commented):
Left some initial feedback apart from the no-std failures.

src/miniscript/musig_key.rs
kanishk779: Shouldn't the output of debug be same as display? If not, what should be the format while debugging ?
For example if the key is `musig(A,musig(B,C),musig(D))` what should debug print for this ?

src/miniscript/musig_key.rs
kanishk779: Added some more test cases.

src/miniscript/musig_key.rs
kanishk779: Done.

kanishk779: Yes now "musig(,,)" also passes the test.

src/miniscript/musig_key.rs
kanishk779: Done.

src/miniscript/musig_key.rs
kanishk779: Ok, removed the if condition.

src/miniscript/musig_key.rs
kanishk779: Ok

src/miniscript/musig_key.rs
kanishk779: Implemented FromTreef for KeyExpr.

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::expression::Tree;
+use crate::MiniscriptKey;
+
+#[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq, Hash)]
+/// Enum for representing keys in miniscript
+pub enum KeyExpr<Pk: MiniscriptKey> {
+    /// Single-key (e.g pk(a), here 'a' is a single key)
+    SingleKey(Pk),
+
+    /// Collection of keys in used for multi-signature
```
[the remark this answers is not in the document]
kanishk779: Ok

src/miniscript/musig_key.rs
kanishk779: Ok

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,166 @@
...
+impl<'a, Pk> Iterator for KeyExprIter<'a, Pk>
+where
+    Pk: MiniscriptKey + 'a,
+{
+    type Item = &'a Pk;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while !self.stack.is_empty() {
+            let last = self.stack.pop().expect("Size checked above");
+            match &*last {
```
SarcasticNastik, line 58: nit: Since we don't have any overloaded `Deref` for `KeyExpr`, ig it's safe to just match it as:
```suggestion
            match last {
```

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,166 @@
...
+                KeyExpr::MuSig(key_vec) => {
+                    // push the elements in reverse order
+                    for key in key_vec.iter().rev() {
+                        self.stack.push(key)
+                    }
+                }
```
SarcasticNastik, line 64: nit: Since for loops themselves are implemented using iterators, there shouldn't much difference in performance.
```suggestion
                KeyExpr::MuSig(key_vec) => key_vec.iter().rev().for_each(|key| self.stack.push(key));
```

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,166 @@
...
+impl<Pk: MiniscriptKey + FromStr> FromStr for KeyExpr<Pk> {
+    type Err = Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (key_tree, _) = Tree::from_slice(s).unwrap();
+        FromTree::from_tree(&key_tree)
+    }
+}
+
+#[derive(Debug, Clone)]
+/// Iterator for keyexpr
```
SarcasticNastik, line 44: You can add linkable references in docstrings, ex:- for `KeyExpr` here. Check this out [https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html]

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,166 @@
...
+            let single_key = Pk::from_str(tree.name).map_err(|_| Error::SingleKeyParseError)?;
+            Ok(KeyExpr::SingleKey(single_key))
+        }
+    }
+}
+
+impl<Pk: MiniscriptKey + FromStr> FromStr for KeyExpr<Pk> {
+    type Err = Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (key_tree, _) = Tree::from_slice(s).unwrap();
```
SarcasticNastik, line 38: I think we should propagate the error rather than panicking here. From what little I understand, we only use `unwrap` in cases we're sure errors shouldn't occur. In this case, `s` might be a malformed string and should be handled by the calling function itself.

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,166 @@
...
+
+impl<Pk: MiniscriptKey + FromStr> FromTree for KeyExpr<Pk> {
+    fn from_tree(tree: &Tree) -> Result<KeyExpr<Pk>, Error> {
+        if tree.name == "musig" {
+            let mut key_expr_vec = vec![];
+            for sub_tree in tree.args.iter() {
+                let temp_res = KeyExpr::<Pk>::from_tree(sub_tree)?;
+                key_expr_vec.push(temp_res);
+            }
+            Ok(KeyExpr::MuSig(key_expr_vec))
```
SarcasticNastik, line 27: nit: Can be re-writtten as:
```suggestion
            let key_expr_vec = tree
                .args
                .iter()
                .map(|subtree| KeyExpr::<Pk>::from_tree(subtree))
                .collect::<Result<Vec<KeyExpr<Pk>>, Error>>()?;
            Ok(KeyExpr::MuSig(key_expr_vec))

```
It's not the cleanest solution so really upto you if you wanna include this :p

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+    }
+}
+
+impl<Pk: MiniscriptKey> fmt::Display for KeyExpr<Pk> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            KeyExpr::SingleKey(ref pk) => write!(f, "{}", pk),
+            KeyExpr::MuSig(ref my_vec) => {
```
SarcasticNastik, line 96: I'm curious here as to why we want to print all the musig keys in Display rather than outputting the key-aggregation of the same.

src/descriptor/bare.rs
```diff
@@ -219,7 +220,7 @@ impl<Pk: MiniscriptKey> Pkh<Pk> {
     /// sighash suffix. Includes the weight of the VarInts encoding the
     /// scriptSig and witness stack length.
     pub fn max_satisfaction_weight(&self) -> usize {
-        4 * (1 + 73 + BareCtx::pk_len(&self.pk))
+        4 * (1 + 73 + BareCtx::pk_len(&KeyExpr::SingleKey(self.pk.clone())))
```
sanket1729, line 223: Let's convert back `pk_len` to accept a `&Pk` so that we avoid the clone here

sanket1729: Let's add a test for Checking the len of 1) `pk(uncompressedkey)` in Legacy. 2) pk(`compressed`) in Segwitv0 3) pk(`XonlyKey`) in Tap and 4) pk('musig') in tap

src/interpreter/mod.rs
```diff
@@ -593,7 +593,10 @@ where
                 Terminal::PkK(ref pk) => {
                     debug_assert_eq!(node_state.n_evaluated, 0);
                     debug_assert_eq!(node_state.n_satisfied, 0);
-                    let res = self.stack.evaluate_pk(&mut self.verify_sig, *pk);
+                    let res = match pk.single_key() {
+                        Some(pk) => self.stack.evaluate_pk(&mut self.verify_sig, *pk),
+                        None => None,
```
sanket1729, line 598: Doing this is wrong. We should never encounter this because know the variant here is `single_key`. We should either use expect on `single_key` with a message stating "Musig keys cannot be parsed from Script"

src/interpreter/mod.rs
```diff
@@ -868,10 +871,11 @@ where
                         // evaluate each key with as a pk
                         // note that evaluate_pk will error on non-empty incorrect sigs
                         // push 1 on satisfied sigs and push 0 on empty sigs
-                        match self
-                            .stack
-                            .evaluate_pk(&mut self.verify_sig, subs[node_state.n_evaluated])
-                        {
+                        let temp = match subs[node_state.n_evaluated].single_key() {
```
sanket1729, line 874: nit: Use a better name than `temp`. Same note about using `expect` here

src/miniscript/astelem.rs
```diff
@@ -80,7 +81,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
         Pk::RawPkHash: 'a,
     {
         match *self {
-            Terminal::PkK(ref p) => pred(p),
+            Terminal::PkK(ref p) => p.for_each_key(pred),
```
sanket1729, line 84: Add a test `for_each_key` for a fragment containing musig.

src/miniscript/context.rs
```diff
@@ -332,7 +339,7 @@ where
     /// Note that this includes the serialization prefix. Returns
     /// 34/66 for Bare/Legacy based on key compressedness
     /// 34 for Segwitv0, 33 for Tap
-    fn pk_len<Pk: MiniscriptKey>(pk: &Pk) -> usize;
+    fn pk_len<Pk: MiniscriptKey>(pk: &KeyExpr<Pk>) -> usize;
```
sanket1729, line 342: As mentioned earlier, we should not change this.

src/miniscript/context.rs
```diff
@@ -384,12 +391,17 @@ impl ScriptContext for Legacy {
...
+            Terminal::PkK(ref key) => match key {
+                KeyExpr::<Pk>::SingleKey(pk) => {
+                    if pk.is_x_only_key() {
+                        return Err(ScriptContextError::XOnlyKeysNotAllowed(
+                            pk.to_string(),
+                            Self::name_str(),
+                        ));
+                    }
+                }
+                KeyExpr::<Pk>::MuSig(_) => return Err(ScriptContextError::MusigNotAllowed),
```
sanket1729, line 403: Add a string parameter as `MusigNotAllowed(String)` and pass Self::name_str. This way whenever user can also see the error context in which musig was not allowed.

src/miniscript/context.rs
```diff
@@ -567,8 +589,11 @@ impl ScriptContext for Segwitv0 {
         ms.ext.max_sat_size.map(|x| x.0)
     }
 
-    fn pk_len<Pk: MiniscriptKey>(_pk: &Pk) -> usize {
-        34
+    fn pk_len<Pk: MiniscriptKey>(pk: &KeyExpr<Pk>) -> usize {
+        match pk {
+            KeyExpr::<Pk>::SingleKey(_) => 34,
+            KeyExpr::<Pk>::MuSig(_) => 32,
```
sanket1729, line 595: Here and everywhere, where we `32`. It should actually be `33` as it also includes the push len prefix.

src/descriptor/bare.rs
sanket1729: I have posted what to change in the comment at the function definition in the context.rs file.

src/miniscript/context.rs
sanket1729: But, we should add another method here with default implementation as 

```rust
fn key_expr_len<Pk: MiniscriptKey>(pk: &KeyExpr) {
	match pk {
		SingleKey(pk) => Ctx::pk_len(&pk),
		Musig(expr) => 33, // not 32 as mentioned in another comment below
	}
}
```

src/miniscript/context.rs
```diff
@@ -618,13 +643,24 @@ impl ScriptContext for Tap {
         }
 
         match ms.node {
-            Terminal::PkK(ref pk) => {
-                if pk.is_uncompressed() {
+            Terminal::PkK(ref key) => {
+                if key.iter().all(|pk| !pk.is_uncompressed()) {
```
sanket1729, line 647: I know this is going to sound weird, but you would need to change to 
`key.iter().any(|pk| pk.is_uncompressed)`. Checking that some is not uncompressed is not the same as checking something is compressed.

src/miniscript/context.rs
```diff
@@ -618,13 +643,24 @@ impl ScriptContext for Tap {
...
+                } else {
                     return Err(ScriptContextError::UncompressedKeysNotAllowed);
                 }
-                Ok(())
             }
             Terminal::Multi(..) => Err(ScriptContextError::TaprootMultiDisabled),
+            Terminal::MultiA(_, ref keys) => {
+                if keys
+                    .iter()
+                    .all(|keyexpr| keyexpr.iter().all(|pk| !pk.is_uncompressed()))
```
sanket1729, line 657: Same comment about compressed keys

src/miniscript/context.rs
```diff
@@ -677,8 +713,11 @@ impl ScriptContext for Tap {
         SigType::Schnorr
     }
 
-    fn pk_len<Pk: MiniscriptKey>(_pk: &Pk) -> usize {
-        33
+    fn pk_len<Pk: MiniscriptKey>(pk: &KeyExpr<Pk>) -> usize {
+        match pk {
+            KeyExpr::<Pk>::SingleKey(_) => 33,
+            KeyExpr::<Pk>::MuSig(_) => 32,
```
sanket1729, line 719: The len should always be 33

src/miniscript/context.rs
```diff
@@ -753,7 +800,13 @@ impl ScriptContext for BareCtx {
         match &ms.node {
             Terminal::Check(ref ms) => match &ms.node {
                 Terminal::RawPkH(_pkh) => Ok(()),
-                Terminal::PkK(_pk) | Terminal::PkH(_pk) => Ok(()),
+                Terminal::PkH(_pk) => Ok(()),
+                Terminal::PkK(key) => match key {
+                    KeyExpr::<Pk>::SingleKey(_pk) => Ok(()),
+                    KeyExpr::<Pk>::MuSig(_) => {
+                        return Err(Error::NonStandardBareScript);
```
sanket1729, line 807: You are returning an incorrect variant here

sanket1729 (commented):
Partial review

src/miniscript/musig_key.rs
kanishk779: Ok, changed accordingly.
```suggestion
            match last {
```

src/miniscript/musig_key.rs
kanishk779: Ok, done.

src/miniscript/musig_key.rs
kanishk779: Oh, I didn't know that.
Fixed it now.

src/miniscript/musig_key.rs
kanishk779: Yes, this seems the right way of handling the error in this case.
Done.

src/miniscript/musig_key.rs
kanishk779: Looks better and more readable. Changed according to the suggestion.

src/interpreter/mod.rs
kanishk779: Ok. fixed as suggested.

src/interpreter/mod.rs
kanishk779: done.

src/miniscript/context.rs
kanishk779: Ok, created the new method with the default implementation.

src/miniscript/context.rs
kanishk779: Done. added a string parameter.

src/miniscript/context.rs
kanishk779: Done. changed everywhere.

src/miniscript/context.rs
kanishk779: Ok. changed as suggested.

src/miniscript/context.rs
kanishk779: done.

src/miniscript/context.rs
kanishk779: changed everywhere.

src/miniscript/context.rs
kanishk779: changed it to
`Err(Error::ContextError(ScriptContextError::MusigNotAllowed(String::from(Self::name_str()))))`

src/miniscript/musig_key.rs
kanishk779: I discussed this with @sanket1729 , and it seems that this is the right way for implementing display involving Musig. Because we want to show/print the individual keys and not the final aggregate key.

src/descriptor/bare.rs
kanishk779: 1. Revert back to &Pk.
2. Add a test `check_script_size` covering all the four suggested tests.

src/miniscript/astelem.rs
kanishk779: Added a test for this `test_for_each_key` in `miniscript/mod.rs`

src/policy/mod.rs
```diff
@@ -161,12 +174,20 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Liftable<Pk> for Terminal<Pk, Ctx> {
...
                 keys.iter()
                     .map(|k| Semantic::KeyHash(k.to_pubkeyhash()))
                     .collect(),
             ),
+            Terminal::MultiA(k, ref keys) => {
+                let mut policy_vec: Vec<Semantic<Pk>> = vec![];
+                for key in keys {
+                    policy_vec.push((Terminal::<Pk, Ctx>::PkK(key.clone())).lift()?)
+                }
+                // Semantic::Threshold(keys.len(), policy_vec)
```
sanket1729, line 188: remove this comment

src/policy/mod.rs
```diff
@@ -234,13 +255,14 @@ mod tests {
     #[cfg(feature = "compiler")]
     use sync::Arc;
 
-    use super::super::miniscript::context::Segwitv0;
+    use super::super::miniscript::context::{Segwitv0, Tap};
     use super::super::miniscript::Miniscript;
     use super::{Concrete, Liftable, Semantic};
     use crate::prelude::*;
     use crate::DummyKey;
+    // use crate::MiniscriptKey;
```
sanket1729, line 263: Unnecessary import

sanket1729 (commented):
Almost there. Some last few edits.

src/policy/mod.rs
kanishk779: removed.

src/policy/mod.rs
kanishk779: removed.

src/miniscript/musig_key.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+impl<Pk: MiniscriptKey> ForEachKey<Pk> for KeyExpr<Pk> {
+    fn for_each_key<'a, F: FnMut(&'a Pk) -> bool>(&'a self, mut pred: F) -> bool
+    where
+        Pk: 'a,
+        Pk::RawPkHash: 'a,
+    {
+        let keys_res = self.iter().all(|my_key| pred(my_key));
```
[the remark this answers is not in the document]
kanishk779: Done.

src/miniscript/iter.rs
```diff
@@ -210,20 +250,53 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
...
+                } else {
+                    None
+                }
             }
             _ => None,
         }
     }
 }
+/// Parent iter for all the below iters
+pub struct BaseIter<'a, Pk: MiniscriptKey, Ctx: ScriptContext> {
```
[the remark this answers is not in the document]
kanishk779: No, this is not being outside this module. I have changed it to private.

src/lib.rs
```diff
@@ -99,6 +99,9 @@ extern crate alloc;
 #[cfg(not(feature = "std"))]
 extern crate hashbrown;
 
+#[cfg(not(feature = "std"))]
```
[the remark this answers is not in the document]
kanishk779: I added this because the last CI check was failing. But it seems that this has nothing to do with it.
Removed it now.l

src/miniscript/iter.rs
```diff
@@ -143,9 +153,17 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
...
-                keys.iter().map(Pk::to_pubkeyhash).collect()
+            Terminal::PkH(ref key) => vec![key.to_pubkeyhash()],
+            Terminal::PkK(ref key) => key.iter().map(|pk| pk.to_pubkeyhash()).collect(),
+            Terminal::Multi(_, ref keys) => keys.iter().map(Pk::to_pubkeyhash).collect(),
+            Terminal::MultiA(_, ref keys) => {
+                let mut res: Vec<Pk::RawPkHash> = Vec::<Pk::RawPkHash>::new();
+                for key in keys {
+                    let temp: Vec<Pk::RawPkHash> =
+                        key.iter().map(|pk| pk.to_pubkeyhash()).collect();
+                    res.extend(temp)
```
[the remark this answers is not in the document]
kanishk779: Done. I have removed the collect in leapk, leapkh, leapk_pkh.

apoelstra:
Closing along with #433. Definitely appreciate the effort but things have changed significantly in the last few years and the musig work will basically have to be redone.

Salarmahmo (approved):

Cargo.toml
```diff
@@ -12,7 +12,7 @@ edition = "2018"
 
 [features]
```
Salarmahmo, line 13: hi

PR #448, opened 2022-07-14
DerivedDescriptor unfinished business

LLFourn:
This PR further develops the idea introduced in #345. It has two commits both with relatively detailed commit messages that can be reviewed separately. In summary:

1. Develop the `DerivedDescriptorKey` (a key that is guaranteed not to have wildcard in it) idea more by adding missing functionality and refining the API. In addition, I removed the error cases from `ConversionError` which seems to have been omitted from #345.
2. Since the introduction of `DerivedDescriptorKey`, the overlapping usage of the term "derive" has become quite confusing. In addition to the usual definition of "derive" we have also used it to mean "replace a wildcard with a particular derivation index". I deprecated and renamed everything that uses the latter definition.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -70,14 +70,9 @@ pub enum SinglePubKey {
     XOnly(XOnlyPublicKey),
 }
 
-/// A derived [`DescriptorPublicKey`]
-///
-/// Derived keys are guaranteed to never contain wildcards
+/// A [`DescriptorPublicKey`] without any wildcards.
 #[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
-pub struct DerivedDescriptorKey {
-    key: DescriptorPublicKey,
-    index: u32,
-}
+pub struct DerivedDescriptorKey(DescriptorPublicKey);
 
 impl fmt::Display for DescriptorSecretKey {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
@@ -357,22 +352,14 @@ impl FromStr for DescriptorPublicKey {
 /// Descriptor key conversion error
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
 pub enum ConversionError {
-    /// Attempted to convert a key with a wildcard to a bitcoin public key
-    Wildcard,
     /// Attempted to convert a key with hardened derivations to a bitcoin public key
     HardenedChild,
-    /// Attempted to convert a key with a hardened wildcard to a bitcoin public key
-    HardenedWildcard,
 }
 
 impl fmt::Display for ConversionError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.write_str(match *self {
-            ConversionError::Wildcard => "uninstantiated wildcard in bip32 path",
             ConversionError::HardenedChild => "hardened child step in bip32 path",
-            ConversionError::HardenedWildcard => {
-                "hardened and uninstantiated wildcard in bip32 path"
-            }
         })
     }
 }
@@ -383,7 +370,7 @@ impl error::Error for ConversionError {
         use self::ConversionError::*;
 
         match self {
-            Wildcard | HardenedChild | HardenedWildcard => None,
+            HardenedChild => None,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


apoelstra:
This looks great! The renaming really helps make sense of these APIs. Is "definite" a standard term or did you make it up for this PR? TBH I'm not sure what the right way to say "non-wildcard" is.

apoelstra (approved):
ACK 2ad655518a1899c402d75f34cfc51cdd13cc851d

tcharding:
> TBH I'm not sure what the right way to say "non-wildcard" is

Ha! I got confused as hell over the 'derive' usage and also came to this same conclusion, so I did a definition search just now on wildcard to try come up with some terms, could we use one of these?
- specified
- known
- concrete

src/descriptor/key.rs
```diff
@@ -441,40 +428,50 @@ impl DescriptorPublicKey {
...
+    pub fn has_wildcard(&self) -> bool {
         match *self {
             DescriptorPublicKey::Single(..) => false,
             DescriptorPublicKey::XPub(ref xpub) => xpub.wildcard != Wildcard::None,
         }
     }
 
-    /// Derives the [`DescriptorPublicKey`] at `index` if this key is an xpub and has a wildcard.
+    #[deprecated(note = "use at_derivation_index instead")]
+    /// Deprecated name of [`at_derivation_index`].
```
tcharding, line 446: I quite like this idea of removing the original comment and replacing with "Deprecated name of ..", I'm going to use that myself in future.

nit: Can you put the attribute below the comment please if you rebase for any other reason.

src/descriptor/key.rs
```diff
@@ -441,40 +428,50 @@ impl DescriptorPublicKey {
         }
     }
 
-    /// Whether or not the key has a wildcards
+    /// Whether or not the key has a wildcard
+    #[deprecated(note = "use has_wildcard instead")]
```
tcharding, line 432: @apoelstra / @sanket1729 I wonder if we can come up with a way to remind ourselves to add the 'since' field to this (and to these attributes in general) when we do the release tracking issue that bumps the version. these 'deprecated' attributes are impossible to write correctly because the next release number is not known at time of writing.

tcharding (commented):
I'm not a fully qualified miniscript reviewer yet so just shallow review. Thanks man!

LLFourn:
> This looks great! The renaming really helps make sense of these APIs. Is "definite" a standard term or did you make it up for this PR? TBH I'm not sure what the right way to say "non-wildcard" is.

Not a standard term in anything surrounding Bitcoin. I borrowed it from linguistics which I think would describe a language feature like `/*` as introducing [*definiteness*](https://en.wikipedia.org/wiki/Definiteness). `/*` allows you to talk about keys *indefinitely*. It works something like "a" does in English: I can say "I will take a car to the airport" without specifying precisely which car while still conveying useful information. It allows you to narrow things down without yet being *definite*.

To keep to this line of thinking `has_wildcard` could be `is_indefinite` but I thought it best not to go that far as when I suggested "definite" in #345 it got a mixed response.

> > TBH I'm not sure what the right way to say "non-wildcard" is
> 
> Ha! I got confused as hell over the 'derive' usage and also came to this same conclusion, so I did a definition search just now on wildcard to try come up with some terms, could we use one of these?
> 
> * specific / specified
> * known
> * concrete
> 
> (We have concrete already in policy so probably not good.)

"concrete" refers to a descriptor that has already derived keys in comments and docs already too.
Even if you are not familiar with the linguistic concept, `SpecificDescriptorKey` and `KnownDescrpitorKey` don't seem significantly better than `DefiniteDescriptorKey`. I could live with `WildcardFreeDescriptorKey` since it is precise (and rhymes!).

src/descriptor/key.rs
tcharding: But we only want it right before doing release not all the time.

apoelstra:
I can live with any of these (including `Definite` :)).

Sorry to stir up a bikeshedding discussion, especially when I have no strong preference, but I think it could be worthwhile.

tcharding:
I think you are right @LLFourn nothing else seems better than 'definite', FWIW +1 from me.

sanket1729 (approved):
ACK 2ad655518a1899c402d75f34cfc51cdd13cc851d. Thanks for the clean changes.

Issue #438, opened 2022-06-24
Change "use-serde" feature to "serde"


tcharding:
We should the package re-name trick that we used in `rust-bitcoin` to rename the "use-serde" feature to plain "serde".

From `rust-bitocin/Cargo.toml`:
```
actual-serde = { package = "serde", version = "1", default-features = false, features = [ "derive", "alloc" ], optional = true }
```

PR #449, opened 2022-07-21
Add feature "serde"

tcharding:
As we did in `rust-bitcoin`; change the "serde" feature to
"actual-serde" and use `package = "serde"`. This allows users of
miniscript to use "serde" as the feature enabling serde instead of our
custom "use-serde".

Fix: #438


apoelstra (approved):
ACK 88b1f530f79361c3d2a39b98f354642d53ec5c71

PR #450, opened 2022-07-25
Upgrade bitcoin dependency to v0.29.0 

tcharding:
This is going to need a pretty careful review by @sanket1729 and/or @apoelstra to double check all the sequence/lock_time stuff. Even with the new APIs this stuff is non-trivial. This patch changes the current behaviour around logic for these two types. I am unable to work out if current behaviour has bugs in it or if I just don't fully understand it. 

### Note

Please ignore the branch name, it stale now since this PR upgrades to the _newly released_ bitcoin v0.29.1


src/miniscript/types/mod.rs
```diff
@@ -820,16 +821,16 @@ impl Property for Type {
                 // Note that for CLTV this is a limitation not of Bitcoin but Miniscript. The
                 // number on the stack would be a 5 bytes signed integer but Miniscript's B type
                 // only consumes 4 bytes from the stack.
-                if t == 0 || (t & SEQUENCE_LOCKTIME_DISABLE_FLAG) != 0 {
+                if t == PackedLockTime::ZERO {
```
Kixunil, line 824: Did the code originally attempt to check if lock times are actually enabled or something?

Kixunil (changes requested):
I found a few mistakes. Disclaimer: didn't do super-deep review.

src/interpreter/mod.rs
```diff
@@ -491,12 +491,12 @@ pub enum SatisfiedConstraint {
...
     RelativeTimelock {
         /// The value of RelativeTimelock
-        time: u32,
+        n: Sequence,
     },
     ///Absolute Timelock for CLTV.
     AbsoluteTimelock {
         /// The value of Absolute timelock
-        time: u32,
+        n: LockTime,
```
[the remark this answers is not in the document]
tcharding: Its a block height or a block time. `n` is from `n OP_CLTV` and `nLockTime` - that's why I used `n`.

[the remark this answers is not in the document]
Kixunil: `n` in `nLockTime` literally means number. It's meaningless. I do see that `time` is confusing, so `lock_time` would be better because most people already know it can be height.

src/miniscript/hash256.rs
```diff
@@ -31,10 +33,18 @@ pub struct Hash([u8; 32]);
...
 borrow_slice_impl!(Hash);
 
+impl<I: SliceIndex<[u8]>> Index<I> for Hash {
+    type Output = I::Output;
+
+    #[inline]
+    fn index(&self, index: I) -> &Self::Output {
+        &self.0[index]
+    }
+}
```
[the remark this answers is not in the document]
tcharding: I added this because of the trait bounds on `Hash` trait (from `bitcoin_consensus`). `Deref` doesn't help with that, right? Or am I missing something?

src/miniscript/types/mod.rs
tcharding: Originally both sequnece and locktime (`After` and `Older`) where handled together. When I split them apart I kept the disable check even though it seemed wrong, finally I got the confidence to remove it :)

src/miniscript/hash256.rs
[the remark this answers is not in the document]
Kixunil: I don't see any `Hash` trait in `bitcoin_consensus`, I guess you mean `bitcoin_hashes`. You're right that `Deref` doesn't help but it's a question if that bound should be there since it's quite burdensome. I'd have to look into it deeper. The first check is will anything break if we remove it?

src/descriptor/tr.rs
```diff
@@ -257,6 +257,7 @@ impl<Pk: MiniscriptKey> Tr<Pk> {
                     TaprootBuilderError::EmptyTree => {
                         unreachable!("Taptree is a well formed tree with atleast 1 element")
                     }
+                    _ => unreachable!("non_exhaustive catchall"),
```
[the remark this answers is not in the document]
tcharding: This whole error match is strange, every error variant has an `unreachable!`. To be honest I was mainly just interested in checking that `rust-bitcoin` worked so I just hacked this up to make it build.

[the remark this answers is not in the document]
Kixunil: They have explanation why it should be OK, so I think it's fine. But it makes me question the whole API. Maybe we should've used typestate instead?

src/interpreter/error.rs
```diff
@@ -187,6 +189,7 @@ impl fmt::Display for Error {
             Error::VerifyFailed => {
                 f.write_str("Expected Satisfied Boolean at stack top for VERIFY")
             }
+            _ => f.write_str("Unknown error, non_exhaustive catch all"),
```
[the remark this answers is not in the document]
tcharding: face palm :) Cheers

src/interpreter/error.rs
```diff
@@ -234,6 +237,7 @@ impl error::Error for Error {
             Secp(e) => Some(e),
             SchnorrSig(e) => Some(e),
             SighashError(e) => Some(e),
+            _ => None, // non_exhaustive catch all.
```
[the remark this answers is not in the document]
tcharding: Fixed.

tcharding:
Thanks for the review @Kixunil!

src/descriptor/tr.rs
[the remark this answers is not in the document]
sanket1729: I will suggest some changes for the upstream API to make this clean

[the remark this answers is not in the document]
sanket1729: Opened https://github.com/rust-bitcoin/rust-bitcoin/pull/1151

src/miniscript/types/mod.rs
apoelstra: It's not just after vs older -- the original code would look at a `u32` which might contain a value that didn't encode any locktime at all.

It's not clear to me that this check can be removed.

tcharding: No worries, since this PR is not a merge candidate this bridge doesn't need crossing now. I imagine between @apoelstra and @sanket1729 this will have to be worked out when miniscript upgrades to use rust-bitcoin 0.29.0.

src/descriptor/tr.rs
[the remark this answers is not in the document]
tcharding: This is fixed now, please check my use of a single `unreachable` statement @sanket1729

src/interpreter/mod.rs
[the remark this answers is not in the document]
tcharding: Used `lock_time` for this field and similarly `sequence` for `SatisfiedConstraint::RelativeTimelock` field.

src/miniscript/hash256.rs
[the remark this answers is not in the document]
tcharding: I'll investigate this, cheers.

src/miniscript/types/mod.rs
tcharding: Flagging this thread, this will need resolving before we can consider upgrading to the newly release version of `rust-bitcoin`. Sorry to be a drag but I think this is kind of subtle and will need input by @sanket1729 and/or @apoelstra to work it out. I can code up the solution.

src/miniscript/hash256.rs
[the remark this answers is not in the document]
tcharding: Please see: https://github.com/rust-bitcoin/bitcoin_hashes/pull/168

src/policy/semantic.rs
```diff
@@ -569,15 +592,18 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
 
     /// Filter a policy by eliminating absolute timelock constraints
     /// that are not satisfied at the given `n` (`n OP_CHECKLOCKTIMEVERIFY`).
-    pub fn at_lock_time(mut self, n: u32) -> Policy<Pk> {
+    pub fn at_lock_time(mut self, n: LockTime) -> Policy<Pk> {
         self = match self {
             Policy::After(t) => {
-                if !timelock::absolute_timelocks_are_same_unit(t, n) {
+                // FIXME: If the point of the LockTime type is to make it hard to make mistakes then
+                // this code shows it fails at that since we _must_ check is_same_unit before comparing.
```
sanket1729, line 599: Indeed, sorry I was absent during the review of the initial code. I get it that there was some hesitance in implementing `PartialOrd`, but why not have a method that returns Option<cmp::Ordering>? 

We should really not need to check that they are same units here.

src/psbt/mod.rs
```diff
@@ -334,31 +332,30 @@ impl<'psbt, Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for PsbtInputSatisfie
...
-            // transaction version and sequence check
-            false
-        } else {
-            <dyn Satisfier<Pk>>::check_older(&Older(seq), n)
+        // Disable flag set => return true.
+        if !n.is_relative_lock_time() {
+            return true;
         }
+
+        if self.psbt.unsigned_tx.version < 2 || !seq.is_relative_lock_time() {
```
sanket1729, line 354: I like the other mthods. They make the code here a lot cleaner

sanket1729 (commented):
Mostly ACK. did a round of careful review. Just waiting on reply for comparison FIXME comment

src/policy/semantic.rs
apoelstra: `PartialOrd` gives us a method that returns `Option<cmp::Ordering>`. It's `Ord` that we couldn't implement, because there is no natural total ordering on timelocks, but which we wanted to implement so that you could use `Vec::sort` and derive `Ord` on structures that contain timelocks.

Kixunil: I think @sanket1729 means that `PartialOrd` still allows `<` and `>` operators which are foot guns. Maybe this should be a `clippy` lint.

tcharding:
Thanks for the review @sanket1729, I'll get back to this soonish, no rush because AFAIU there are a few other crates to upgrade first. Feel free to take this work and use it if you don't want to wait for me.

src/miniscript/satisfy.rs
```diff
@@ -109,59 +106,46 @@ pub trait Satisfier<Pk: MiniscriptKey + ToPublicKey> {
...
-
-        let mask = SEQUENCE_LOCKTIME_MASK | SEQUENCE_LOCKTIME_TYPE_FLAG;
-        let masked_n = n & mask;
-        let masked_seq = self.0 & mask;
-        if masked_n < SEQUENCE_LOCKTIME_TYPE_FLAG && masked_seq >= SEQUENCE_LOCKTIME_TYPE_FLAG {
-            false
+        if self.is_height_locked() && n.is_height_locked()
+            || self.is_time_locked() && n.is_time_locked()
+        {
+            n.to_consensus_u32() <= self.to_consensus_u32()
```
LLFourn, line 131: The previous implementation masked the sequence before doing the comparison. How can we justify not doing this here? What if one has the RBF disabled  and the other doesn't for example (disabling RBF will always be the highest value unless you mask out that bit).

PS methods on `Sequence` like `is_satisfied_by_height` and `is_satisfied_by_time` would be super useful here.

tcharding: > How can we justify not doing this here?

Thanks man, you are 100% correct, this code is not the same as the code it replaces - that's my bug.

> PS methods on Sequence like is_satisfied_by_height and is_satisfied_by_time would be super useful here.

A relative lock time type is meant to be in the works (not sure if anyone is actively working on it yet) and that type would have these methods you suggest. When `Sequence` was added relative lock time methods were explicitly out of scope.

src/psbt/mod.rs
tcharding: Sorry mate I'm not following you, which other methods?

src/miniscript/satisfy.rs
tcharding: I've reverted these changes, so we now have
```rust
        // We need a relative lock time type in rust-bitcoin to clean this up.

        /* If nSequence encodes a relative lock-time, this mask is
         * applied to extract that lock-time from the sequence field. */
        const SEQUENCE_LOCKTIME_MASK: u32 = 0x0000ffff;
        const SEQUENCE_LOCKTIME_TYPE_FLAG: u32 = 0x00400000;

        let mask = SEQUENCE_LOCKTIME_MASK | SEQUENCE_LOCKTIME_TYPE_FLAG;
        let masked_n = n.to_consensus_u32() & mask;
        let masked_seq = self.to_consensus_u32() & mask;
        if masked_n < SEQUENCE_LOCKTIME_TYPE_FLAG && masked_seq >= SEQUENCE_LOCKTIME_TYPE_FLAG {
            false
        } else {
            masked_n <= masked_seq
        }
```

src/miniscript/satisfy.rs
```diff
@@ -109,36 +106,35 @@ pub trait Satisfier<Pk: MiniscriptKey + ToPublicKey> {
     }
 
     /// Assert whether an relative locktime is satisfied
-    fn check_older(&self, _: u32) -> bool {
+    fn check_older(&self, _: Sequence) -> bool {
```
LLFourn, line 109: If you introduce a relative locktime type then it should probably be passed in here instead of `Sequence`.

src/miniscript/satisfy.rs
tcharding: Please see https://github.com/rust-bitcoin/rust-miniscript/pull/455 which implements this using a new `relative::LockTime` API.

src/miniscript/satisfy.rs
tcharding: Thanks man, I've done that now in the demo PR.

src/miniscript/astelem.rs
```diff
@@ -620,9 +621,11 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
                 .push_slice(&Pk::hash_to_hash160(hash)[..])
                 .push_opcode(opcodes::all::OP_EQUALVERIFY),
             Terminal::After(t) => builder
-                .push_int(t as i64)
+                .push_int(t.0.into())
```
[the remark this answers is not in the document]
tcharding: I think I changed them all.

src/miniscript/types/mod.rs
```diff
@@ -437,16 +438,16 @@ pub trait Property: Sized {
...
```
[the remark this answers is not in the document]
tcharding: I believe so.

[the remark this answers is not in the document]
tcharding: I'm still at a loss as to exactly what the comment means. It should be removed now though, right?

sanket1729:
I don't think we should wait for relative lock time PR in bitcoin 0.30.0. We would need a version of rust-miniscript that supports bitcoin 0.29.0

src/psbt/mod.rs
sanket1729: Oops. I meant I really like the new `added` methods upstream. They make the code here a lot cleaner. There is nothing to resolve here

src/policy/semantic.rs
```diff
@@ -569,15 +592,16 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
-    pub fn at_lock_time(mut self, n: u32) -> Policy<Pk> {
+    pub fn at_lock_time(mut self, n: LockTime) -> Policy<Pk> {
         self = match self {
             Policy::After(t) => {
-                if !timelock::absolute_timelocks_are_same_unit(t, n) {
+                let t = LockTime::from(t);
+                if !t.is_same_unit(n) {
                     Policy::Unsatisfiable
-                } else if t > n {
+                } else if t.to_consensus_u32() > n.to_consensus_u32() {
```
sanket1729, line 601: This is okay, but I would have expected an API like 
```rust
let cmp_res = t.is_greater_than(n) ; // Any other name.  

match(cmp_res) {
	Some(true),
	Some(false),
	None,
}
```

Rust miniscript/or any other downstream project should need to know that they should check !t.is_same_unit() before comparing. 

Perhaps, something worth looking at 0.30.0

src/psbt/mod.rs
tcharding: :) Will be even better when `relative::LockTime` is available, its been nice to see the API in rust-bitcoin develop huh.

src/policy/semantic.rs
tcharding: We could add a method `is_satisfied_by_lock` in the same manner as `relative::LockTime`.

tcharding: Which would give us the much nicer:
```rust
            Policy::After(t) => {
                let t = LockTime::from(t);
                } else if !t.is_satisfied_by_lock(n) {
                    Policy::Unsatisfiable
                } else {
                    Policy::After(t.into())
                }
            }
```

tcharding: https://github.com/rust-bitcoin/rust-bitcoin/pull/1258

tcharding: Face palm, we actually provide a patter for this in the rustdocs (double face palm, I even wrote the rustdocs and still forgot to use the correct pattern)

I've updated the PR to be 
```rust
            Policy::After(t) => {
                let t = LockTime::from(t);
                let is_satisfied_by = match (t, n) {
                    (Blocks(t), Blocks(n)) => t <= n,
                    (Seconds(t), Seconds(n)) => t <= n,
                    _ => false
                };
                if !is_satisfied_by {
                    Policy::Unsatisfiable
                } else {
                    Policy::After(t.into())
                }
            }
```

tcharding:
Hey @sanket1729, seems @josediegorobles is pretty keen to use the upcoming release, is there anything else that needs doing before we can consider this PR and get a release out?

sanket1729:
Hi @tcharding, there is nothing that needs to be done in particular here. I have a few cleanups in mind, but those can wait for `8.1.0` . I will look into preparing a new release this week.

josediegorobles:
Hi, @sanket1729 by suggestion of @tcharding i started a PR for the release: https://github.com/tcharding/rust-miniscript/pull/1

What version must be, 7.1.0 or 8.0.0? And what other things must be on the changelog?

sanket1729 (approved):
ACK 24a249820f403ec103ab7e3d45fa4f7a5f0dee19. Reviewed by range-diff from c51595b9c951d1f07b41dccb79d2205cc3c91160 .


I am waiting on @apoelstra to respond the comment issue he raised

src/miniscript/types/mod.rs
sanket1729: @apoelstra are you convinced that this is resolved?

apoelstra: Let me re-read this PR.

src/policy/semantic.rs
```diff
@@ -345,10 +364,10 @@ impl_from_tree!(
                 Pk::RawPkHash::from_str(pk).map(Policy::KeyHash)
             }),
             ("after", 1) => expression::terminal(&top.args[0], |x| {
-                expression::parse_num(x).map(Policy::After)
+                expression::parse_num(x).map(|x| Policy::after(x))
             }),
             ("older", 1) => expression::terminal(&top.args[0], |x| {
-                expression::parse_num(x).map(Policy::Older)
+                expression::parse_num(x).map(|x| Policy::older(x))
```
apoelstra, line 370: nit: there was no need for the `|x|` here

apoelstra (approved):
ACK 24a249820f403ec103ab7e3d45fa4f7a5f0dee19

apoelstra:
@sanket1729 in future before merging can you edit PR descriptions to remove the `@` signs before people's names? Otherwise github creates spurious notifications for them.

tcharding:
Should I just not put mentions in the PR description? I thought it was only in the commit log (on the actual patch) that mattered?

apoelstra:
@tcharding no, it matters both on the commits and in the PR description. (Though the PR description is really easy to edit, for anyone with merge permissions, while the commits are a pain.)

tcharding:
My bad, duly schooled :)

Kixunil:
@apoelstra are you sure? I was never spuriously notified from PRs. Creation doesn't count as spurious - I consider mentioning explicitly wanting to notify someone.

apoelstra:
@Kixunil yes, though I believe the reason is that our merge script copies the entire PR description into the merge commit message.

Kixunil:
Ah, good point! Could the merge script be modified to replace `([:whitespace:])@([a-zA-Z0-9_-]+[:whitespace]])` with `\1\2` in PR description?

apoelstra:
I don't know -- could you open an issue on https://github.com/bitcoin-core/bitcoin-maintainer-tools/ ?

Note that it does currently display an angry warning if there are @s in the PR description, which I always respond to by just editing the PR description and trying again ... maybe they don't want to modify PRs by default but it'd be nice if there was at least a gitconfig flag or something to do it.

PR #457, opened 2022-08-24
Example for spending taproot descriptor.

kanishk779:
The internal key is changed to `KeyExpr<Pk>`, allowing the internal key to be a `musig` now. The example elaborates the steps for spending via the internal key.


sanket1729:
@tcharding , any clue why the fuzzer is not broken here in rust-miniscript? I was expecting it to break everywhere in rust-bitcoin ecosystem.

apoelstra:
Yeah, super weird ... our CI fuzzing code is substantially the same (there are some changes to the bash notation in rust-bitcoin which might affect error reporting, but I don't think that matters, you can see from the CI output here that fuzzing is being run successfully).

Maybe we just needed to wait for something to change on the CI runner side, and re-enabling fuzzing in rust-bitcoin would Just Work now?

tcharding:
I've no clue. I was looking at CI runs here in `miniscript` when I did the "disable fuzzer" PR and could not work it out.

PR #459, opened 2022-09-12
Pin dependencies when using MSRV

tcharding:
A recent release of `form_urlencoded` breaks our test run when using 1.41.1 toolchain.
    
Pin the `url` and `form_urlencoded` dependencies when on 1.41.1 toolchain. Add a section to the readme instructing devs how to pin relevant dependencies (only when running tests).


apoelstra (approved):
ACK c6d89cd910af787c7f12f139713bba2c7662e07f

PR #461, opened 2022-09-19
Cleanly separate experimental/insane miniscripts from sane miniscripts

sanket1729:
Cleanly separates how we deal with insane and unspecified miniscripts in the library. Insane minsicripts are well-defined that could 
- It is unsafe(does not require a digital signature to spend it)
- It contains a unspendable path because of either a) Resource limitations or b) Timelock Mixing
- The script has malleable satisfaction
- It has repeated public keys

There are some unspecified features like PartialDescriptors that we use to allow parsing miniscripts from scripts or inferring descriptors. For this only includes 
- Parsing rawpkh fragments with `hash160`

All of these can be individually enabled or disabled using the `ExtParams` builder. 

As a consequence of this, we removed the Miniscript::Hash associated type and change the Semantic policy to operate only on keys.

Also fixes another issue in the usage of PkTranslator that made it impossible to implement the trait in downstream libraries


src/pub_macros.rs
```diff
@@ -0,0 +1,120 @@
...
+/// Macro for translation of associated types where the associated type is the same
+/// Handy for Derived -> concrete keys where the associated types are the same.
+///
+/// Writing the complete translator trait is tedious. This macro is handy when
+/// we are not trying the associated types for hash160, ripemd160, hash256 and
+/// sha256.
+///
+/// See also [`crate::translate_assoc_fail`]
+#[macro_export]
+macro_rules! translate_assoc_clone {
```
apoelstra, line 90: What if we instead made this `translate_assoc_into` and called `.into()` rather than `.clone()`? This would work for all the hash types since they are `Copy`, and would potentially work more generally.

Also I think both these macros should somehow be renamed to indcate that they only implement the translate functions for hashes, not for pubkeys.

sanket1729: Agreed about the renaming. I was also thinking of another refactor that removes translate_pkh altogether. Now that we have Pkh(Pk) instead of Pkh(Pk::Hash), I don't think it is needed. We should support any translation for abstract `RawPkh` until we have raw descriptors specified.

apoelstra: Yeah, agreed, let's drop `translate_pkh`.

sanket1729:
@apoelstra, this turned into a significantly bigger refactor than expected. We removed the `Pk::Hash` type altogether. It seems necessary to have abstract generic types over `RawPkH` types when they are not even specified yet. Users should almost never be dealing with `RawTypes` and lifting to raw types fails(for now). For interpreter descriptor inference purposes, we print an `expr_raw_pkh(40-byte-hex)` to display PkH fragments.

apoelstra:
Reviewed everything except 844382fee5d8fc94f0e0c3122a424b552b006351 (the big one). I have a conceptual question: are you sure we shouldn't be able to parse `rawpkh` using `from_str`? Do we gain anything by not supporting this?

What if we parse a Miniscript from a Script, it contains `rawpkh`, and we try to print that as a string? I think it should succeed, and we should be able to parse the result.

I'd be fine if we still dropped the `RawPkHash` associated type and forced pkhashes to be hash160s, so I think the diff will be substantially the same, but I think there's value in having a string representation of rawpkh, if only so that you can pass "readable" renderings of existing scripts around.

As a future improvement (not in this PR) I'd also like to extend the `Lift` trait somehow

src/miniscript/analyzable.rs
```diff
@@ -22,7 +22,115 @@ use core::fmt;
 use std::error;
 
 use crate::prelude::*;
-use crate::{Miniscript, MiniscriptKey, ScriptContext};
+use crate::{Miniscript, MiniscriptKey, ScriptContext, Terminal};
+
+/// Params for parsing miniscripts that either non-sane or non-specified(experimental) in the spec.
+/// Used as a parameter [`Miniscript::from_str_insane`] and [`Miniscript::parse_insane`].
```
apoelstra, line 28: In 89169c23aec0bac4009593e6c451bf9b0432ee6d:

This comment is out of date since you renamed these two methods :)

src/miniscript/analyzable.rs
```diff
@@ -22,7 +22,115 @@ use core::fmt;
...
+    /// Allow parsing of miniscripts with timelock mixing
+    pub timelock_mixing: bool,
+    /// Allow parsing of malleable miniscripts
+    pub malleability: bool,
+    /// Allow parsing of miniscripts with repeated public keys
+    pub repeated_pk: bool,
+    /// Allow parsing of miniscripts with raw pkh fragments without the pk.
+    /// This could be obtained when parsing miniscript from script
+    pub raw_pkh: bool,
+}
```
apoelstra, line 54: I really love this struct. It's an easy reference place for basically everything weird about Miniscript :)

src/miniscript/analyzable.rs
```diff
@@ -22,7 +22,115 @@ use core::fmt;
...
+            resource_limitations: true,
+            timelock_mixing: true,
+            malleability: true,
+            repeated_pk: true,
+            raw_pkh: false,
+        }
+    }
+
+    /// Enable all non-sane rules and experimental rules
+    pub fn all() -> ExtParams {
```
apoelstra, line 87: In 89169c23aec0bac4009593e6c451bf9b0432ee6d:

Can you rename this `allow_all` or (somewhat snarkily) `free_for_all`?

src/miniscript/mod.rs
```diff
@@ -156,6 +157,20 @@ impl<Ctx: ScriptContext> Miniscript<Ctx::Key, Ctx> {
     /// insane scripts. In general, in a multi-party setting users should only
     /// accept sane scripts.
     pub fn parse_insane(script: &script::Script) -> Result<Miniscript<Ctx::Key, Ctx>, Error> {
```
apoelstra, line 159: In 89169c23aec0bac4009593e6c451bf9b0432ee6d:

I think we should deprecate this and tell users to use `parse_with_ext` with `ExtParams::all` (or whatever we call that). What do you think?

apoelstra:
Done reviewing 89169c23aec0bac4009593e6c451bf9b0432ee6d. Only nits. This is a huge improvement!

I guess you should update the PR descriptor and maybe the title since so much of this PR involves introducing `ExtParams`.

src/miniscript/mod.rs
sanket1729: There is a small difference between `insane` and `ext` (experimental stuff). Insane things are well documented and specified in spec, but disallowed for user safety. The `raw_pkh` stuff is not specified anywhere yet.

sanket1729: I think having an insane high level API is still valuable short-hand for `parse_with_ext(ExtParams::insane())` which is different from `parse_with_ext(ExtParams::all())`

apoelstra: Ok, cool, that's fair. I didn't realize that "insane" covered well-defined things outside of this library.

sanket1729:
Force-pushed and updated the description.

apoelstra:
```
Running `target/debug/examples/htlc`
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `"or(and(pk(022222222222222222222222222222222222222222222222222222222222222222),sha256(1111111111111111111111111111111111111111111111111111111111111111)),and(pk(020202020202020202020202020202020202020202020202020202020202020202),older(4444)))"`,
 right: `"or(and(pkh(4377a5acd66dc5cb67148a24818d1e51fa183bd2),sha256(1111111111111111111111111111111111111111111111111111111111111111)),and(pkh(51814f108670aced2d77c1805ddd6634bc9d4731),older(4444)))"`', examples/htlc.rs:49:5
```
Can you do a rebase -x and check that the `htlc` example passes on every commit?

Otherwise ack b72f5b1b3fe8b56bfc95f5d7f6338810152afafd

sanket1729:
Done and force pushed. Awaiting CI

apoelstra (approved):
ACK 6a06a115978ef95e4eb53c0e9e7e12a94316b7d0

PR #463, opened 2022-09-26
Fix inconsistent naming in the dummy key hash structs

afilini:
Also derive `Default` for `DummyHash` and all of its hash types since they are all empty structs.

### DIFF
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -16,10 +16,15 @@ if cargo --version | grep "1\.41\.0"; then
     MSRV=true
 fi
 
+if cargo --version | grep "1\.47\.0"; then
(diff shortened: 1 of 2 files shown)


apoelstra:
CI failure is a broken dependency, fixed in a commit of #461

afilini:
I cherry-picked Sanket's commit and rebased on top of it

apoelstra (approved):
ACK eb2c282c48a1e85104ced7a9e9aa1c412bfdea57

sanket1729:
FTR, we plan to remove the `DummyKey*` types in the future. You should be able to test everything with `MiniscriptKey = String` type.

PR #462, opened 2022-09-19
Release 8.0.0

sanket1729:
Draft release PR. Still blocked on reviews of

1) #418 
2) #450 
3) #461

PR #465, opened 2022-09-29
Add `PsbtOutputExt::update_with_descriptor`

afilini:
This essentially mirrors `PsbtInputExt::update_with_descriptor` but it works on PSBT outputs.

To avoid code duplication, the current logic that deals with PSBT inputs has been generalized to handle both input and outputs with a new internal `PsbtFields` trait.

-----

I hope it's not too late to get this PR in the new 8.0 release (cc #462)

### DIFF
--- a/src/psbt/mod.rs
+++ b/src/psbt/mod.rs
@@ -549,10 +549,9 @@ pub trait PsbtExt {
     ///
     /// This is the checked version of [`update_with_descriptor_unchecked`]. It checks that the
     /// `witness_utxo` and `non_witness_utxo` are sane and have a `script_pubkey` that matches the
-    /// descriptor. In particular, it makes sure segwit descriptors always have `witness_utxo`
-    /// present and pre-segwit descriptors always have `non_witness_utxo` present (and the txid
-    /// matches). If both `witness_utxo` and `non_witness_utxo` are present then it also checks they
-    /// are consistent with each other.
+    /// descriptor. In particular, it makes sure pre-segwit descriptors always have `non_witness_utxo`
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK f46830c1baa0a5b2658f2b8f82aa0279f04775c3

apoelstra (approved):
ACK 9f6316512cbc593a7f893816418c764df9a24129

afilini:
Sorry if I keep pushing commits here, let me know if you'd like me to open separate PRs.

The last commit allows setting just the `non_witness_utxo` of a psbt input even if it is non-legacy. This [was reported to us](https://github.com/bitcoindevkit/bdk/issues/734#issuecomment-1250241171) recently and I think it make sense to support this (but to be fair I haven't checked what the spec says).

I may have to push another fix because I'm having a failure in a bdk test since I switched to miniscript's `update_input_with_descriptor`, I need to figure out if my previous implementation was broken or if this one is.

afilini:
Turns out nothing was really broken, but calling `update_input_with_descriptor` multiple times on the same psbt would lead to duplicated leaf hashes in `tap_key_origin` since it's just a vec, not a map.

I fixed it by sorting and the end and then running `dedup` on the vec

apoelstra (approved):
ACK 9a57e00aef86746d727e98a80372c9579f95c4d9

sanket1729 (approved):
ACK 9a57e00aef86746d727e98a80372c9579f95c4d9. Not completely happy with internal trait abstraction. I may work on a future PR to abstract that into a function that takes 1) script pubkey and 2) descriptor and outputs the relevant fields that are added to the psbt. 

We can then call this function on the input script pubkey and output script pubkey respectively.

PR #467, opened 2022-09-29
fuzz: fix `compile_descriptor` to sort lifted policies before comparing

### LINKED ISSUE #466: `compile_descriptor` fuzztest fails with `and(hash256(),pk())`
Our `compile_descriptor` fuzztest takes a policy, compiles it, lifts it, and checks the lifted policy against the original policy.

On the input `and(hash256(),pk())` the two params get swapped, causing the fuzztest to fail.

I think this swap is expected behavior from the compiler and we just need to call `.sorted()` in the fuzztest.

apoelstra:
Fixes #466


sanket1729 (approved):
utACK 67ac55ca5e3e28736cd74adc202f29a68edb2a46

PR #470, opened 2022-10-03
Multipath descriptors

### LINKED ISSUE #469: Multipath descriptors support
Multipath descriptors were proposed as a way to optimize the descriptor backup for the common case of having two derivation chains (one for receiving addresses, one for change).

A BIP was proposed a couple months ago: https://github.com/bitcoin/bips/pull/1354. There is an implementation in Bitcoin Core at https://github.com/bitcoin/bitcoin/pull/22838.

It was previously discussed at length in https://github.com/bitcoin/bitcoin/issues/17190.

I think it would be helpful for us to support them. What do you think?

darosior:
This implements [BIP389 multipath descriptors](https://github.com/bitcoin/bips/pull/1354). Fixes #469.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -19,6 +19,8 @@ pub enum DescriptorPublicKey {
     Single(SinglePub),
     /// Extended public key (xpub).
     XPub(DescriptorXKey<bip32::ExtendedPubKey>),
+    /// Multiple extended public keys.
+    MultiXPub(DescriptorMultiXKey<bip32::ExtendedPubKey>),
 }
 
 /// The descriptor secret key, either a single private key or an xprv.
@@ -28,6 +30,8 @@ pub enum DescriptorSecretKey {
     Single(SinglePriv),
     /// Extended private key (xpriv).
     XPrv(DescriptorXKey<bip32::ExtendedPrivKey>),
+    /// Multiple extended private keys.
+    MultiXPrv(DescriptorMultiXKey<bip32::ExtendedPrivKey>),
 }
 
 /// A descriptor [`SinglePubKey`] with optional origin information.
@@ -61,6 +65,39 @@ pub struct DescriptorXKey<K: InnerXKey> {
     pub wildcard: Wildcard,
 }
 
+/// The derivation paths in a multipath key expression.
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
+pub struct DerivPaths(Vec<bip32::DerivationPath>);
+
+impl DerivPaths {
+    /// Create a non empty derivation paths list.
+    pub fn new(paths: Vec<bip32::DerivationPath>) -> Option<DerivPaths> {
+        if paths.is_empty() {
+            None
+        } else {
+            Some(DerivPaths(paths))
+        }
+    }
+
+    /// Get the list of derivation paths.
+    pub fn paths(&self) -> &Vec<bip32::DerivationPath> {
+        &self.0
+    }
+}
+
+/// Instance of one or more extended keys, as specified in BIP 389.
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
+pub struct DescriptorMultiXKey<K: InnerXKey> {
+    /// Origin information
+    pub origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+    /// The extended key
+    pub xkey: K,
+    /// The derivation paths. Never empty.
+    pub derivation_paths: DerivPaths,
+    /// Whether the descriptor is wildcard
+    pub wildcard: Wildcard,
+}
+
 /// Single public key without any origin or range information.
 #[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
 pub enum SinglePubKey {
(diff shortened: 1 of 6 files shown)


sanket1729:
I briefly looked at the code and I like the overall approach for the given BIP. Overall, a strong concept and approach ACK. Just two things:

1) I think Conversion to `DefininateDescriptorKey` should fail when there is MultiPath descriptor
2) We can add `MiniscriptKey::num_derivation_paths`/`MiniscriptKey::num_paths` method to `MiniscriptKey` trait. When converting `Terminal` to `Miniscript`, we check whether all keys have the same `num_derivation_paths`/`num_pahts`, otherwise an error. For all existing types that implement `MiniscriptKey`, this can return 1. I am undecided/unsure whether we want a default impl returning `1` or if we should force all implementers to write the function.
Lastly, for taproot descriptors, it would require checking the same thing in all miniscripts and internal keys in all the initialization constructors.

darosior:
> On another note, I don't like the BIP changes in KEY expressions, but I have not formed an opinion about it. Previously, we could reason about legality of KEY expressions only based on top level contexts, but with this a KEY expression can invalidate another KEY expression in some other part of descriptor

I very much agree. This is [the feedback i gave to Achow](https://github.com/bitcoin/bips/pull/1354#discussion_r985316649) when implementing this.

I was kinda waiting for an answer from him about this, but looks like it won't come. I'll address your feedback.

darosior:
Implemented your suggestions and removed the WIP label from this PR. Having a `num_der_paths` method on `MiniscriptKey` allowed to check for inconsistent multipath keys directly in `Descriptor<Pk>::from_str`.

> I am undecided/unsure whether we want a default impl returning 1 or if we should force all implementers to write the function

Rather returning `0` by default? In any case i don't think there is a sane default for this. It really depends on the implementer. So i went for no default implementation: should we want to, i *think* we can always introduce one later on without breaking backward compatibility.

src/descriptor/key.rs
```diff
@@ -314,6 +314,44 @@ impl DescriptorSecretKey {
...
         Ok(pk)
     }
+
+    /// Whether or not this key has multiple derivation paths.
+    pub fn is_multipath(&self) -> bool {
+        match *self {
+            DescriptorSecretKey::Single(..) | DescriptorSecretKey::XPrv(..) => false,
+            DescriptorSecretKey::MultiXPrv(_) => true,
+        }
+    }
```
darosior, line 324: This seems redundant with `MiniscriptKey`'s `num_der_paths` introduced later on, but i kept it here in order to also have it for private keys.

darosior:
Rebased on master following the recent merges for the 8.0.0 release.

darosior (commented):
I've integrated this in my downstream project. I found a bug for which i'm pushing the fix and a unit test triggering it. Aside from this bug i've successfully used it and by doing so cross-checked it against the Python implementation i'm using in the functional testing framework of my software (https://github.com/darosior/python-bip380/pull/21).

darosior:
Got a spurious io issue on the CPP tests:
```
thread 'tests_from_cpp' panicked at 'called `Result::unwrap()` on an `Err` value: JsonRpc(Transport(SocketError(Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" })))', tests/test_cpp.rs:80:10
```

Force-pushing to trigger a re-run.

darosior:
I hadn't noticed the `key` submodule was made private. I just pushed a patch to re-export `DescriptorMultiXKey` like `DescriptorXKey`. H/T @edouardparis for mentioning that to me.

darosior:
Friendly ping to know when you'll be able to review this, or if you think it is blocked by something?

sanket1729:
@darosior, this is on my list for this week.

src/descriptor/mod.rs
```diff
@@ -792,6 +792,42 @@ impl Descriptor<DescriptorPublicKey> {
     }
 }
 
+impl<Pk: MiniscriptKey> Descriptor<Pk> {
+    /// Whether this descriptor is a multipath descriptor that contains any 2 multipath keys
+    /// with a different number of derivation paths.
+    /// Such a descriptor is invalid according to BIP389.
+    pub fn multipath_length_mismatch(&self) -> bool {
```
sanket1729, line 799: Just a review comment. Nothing to be addressed here. Ideally, we want to make sure that such invalid descriptors cannot be created in the first place. Adding this pub might confuse users to call it to double-check everything. 

I think this is true for all APIs that deal with Miniscript, but not for the internal `Terminal`. I have some changes in mind in future PRs to remove all the confusing sanity_check/check APIs.

sanket1729 (commented):
done reviewing a63d5a263a9006b4d29342012133a3bc919765ba. Just Left a couple of comments. The rest of it is good to go. I have some other changes in mind to come in the future that would clean up the translatePk trait so which would make it impossible to create Miniscripts with different multi-path lengths even after translating. 

Right now, it would fall upon the user to explicitly call the mismatching API to make sure that the translation is done correctly.

src/descriptor/key.rs
```diff
@@ -61,6 +65,19 @@ pub struct DescriptorXKey<K: InnerXKey> {
...
 
+/// Instance of one or more extended keys, as specified in BIP 389.
+#[derive(Debug, Eq, PartialEq, Clone, Ord, PartialOrd, Hash)]
+pub struct DescriptorMultiXKey<K: InnerXKey> {
+    /// Origin information
+    pub origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
+    /// The extended key
+    pub xkey: K,
+    /// The derivation paths. Never empty.
+    pub derivation_paths: Vec<bip32::DerivationPath>,
```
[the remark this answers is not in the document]
darosior: Heh i was on the fence about this when i wrote it. Guess that settles it. :) I added a `DerivPaths` newtype.

sanket1729 (approved):
ACK ae0f3d95fa408bfc5e535050743a0f0ed8cb2963.

sanket1729:
@apoelstra, any thoughts on merging this? This is not a merged BIP yet and could be changed. But so are many things in this repo that are not even BIP yet.

PR #473, opened 2022-10-06
Fix `max_satisfaction_weight` calculations for taproot.

evanlinjin:
* Key spend path should take into consideration the scriptSigLen, stackLen and itemLen (for the one and only stack item).
* Script spend path should consider `control_block` and `script` to be items on the stack.

I also restructured the function to improve readability.

### DIFF
--- a/src/descriptor/tr.rs
+++ b/src/descriptor/tr.rs
@@ -264,27 +264,37 @@ impl<Pk: MiniscriptKey> Tr<Pk> {
     /// # Errors
     /// When the descriptor is impossible to safisfy (ex: sh(OP_FALSE)).
     pub fn max_satisfaction_weight(&self) -> Result<usize, Error> {
-        let mut max_wieght = Some(65);
-        for (depth, ms) in self.iter_scripts() {
-            let script_size = ms.script_size();
-            let max_sat_elems = match ms.max_satisfaction_witness_elements() {
-                Ok(elem) => elem,
-                Err(..) => continue,
-            };
-            let max_sat_size = match ms.max_satisfaction_size() {
-                Ok(sz) => sz,
-                Err(..) => continue,
-            };
-            let control_block_sz = control_block_len(depth);
-            let wit_size = 4 + // scriptSig len byte
-            control_block_sz + // first element control block
-            varint_len(script_size) +
-            script_size + // second element script len with prefix
-            varint_len(max_sat_elems) +
-            max_sat_size; // witness
-            max_wieght = cmp::max(max_wieght, Some(wit_size));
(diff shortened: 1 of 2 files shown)


evanlinjin:
This work is based on my understanding of BIP-0141 and BIP-0341. I'm not sure what is the best way to test this, any ideas will be appreciated.

src/descriptor/tr.rs
```diff
@@ -264,27 +264,37 @@ impl<Pk: MiniscriptKey> Tr<Pk> {
...
-            script_size + // second element script len with prefix
-            varint_len(max_sat_elems) +
-            max_sat_size; // witness
-            max_wieght = cmp::max(max_wieght, Some(wit_size));
-        }
-        max_wieght.ok_or(Error::ImpossibleSatisfaction)
+        let tree = match self.taptree() {
+            // key spend path:
+            // scriptSigLen(4) + stackLen(1) + stack[Sig]Len(1) + stack[Sig](65)
+            None => return Ok(4 + 1 + 1 + 65),
```
LLFourn, line 270: nit: the 65 here is implicitly taking into account an optional sighash flag (because we're doing max I guess this strictly correct). Maybe worth doing a `64 + 1` to indicate this.

LLFourn (dismissed):
ACK fa2e4e2feee2626954b053b46cbdde4499142eef

src/descriptor/tr.rs
evanlinjin: @LLFourn Thank you for the review.

However, at the same time, the sig+sigHash is a single stack item? Maybe a comment would be more appropriate?

evanlinjin:
> Sorry, but could you squash these commitst so the tests pass on every commit?

@apoelstra All done :)

sanket1729 (approved):
utACK 8834b7f992979143e0e6e27dea9a5b79d2e34f03. Thanks for the significant code improvement.

LLFourn (approved):
ACK 8834b7f992979143e0e6e27dea9a5b79d2e34f03

apoelstra (approved):
ACK 8834b7f992979143e0e6e27dea9a5b79d2e34f03

apoelstra:
I'd like to cut a release today. Should we revert this for now?

evanlinjin:
> I'd like to cut a release today. Should we revert this for now?

@apoelstra I don't think this fix conflicts with the conversation we were having in regards to what would be the definition of "max satisfaction weight".

I think we should continue discussing whether we should merge #474 or go with #476 instead.

Let me know your thoughts on those PRs.

PR #475, opened 2022-10-07
Fix unreachable code from calling derive_public_key

sanket1729:
DescriptorPublicKey::derive_public_key API hit an unreachable line when
executing.

public_key.derive_public_key(&secp).unwrap_err();

This commit removes the API altogether and only exposes the method via
`DefiniteDescriptorKey`. Unfortunately, non-wildcard users would need
to pass a dummy value to .at_derivation_index to get this struct, but I
think this is okay.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -485,39 +485,6 @@ impl DescriptorPublicKey {
         DefiniteDescriptorKey::new(definite)
             .expect("The key should not contain any wildcards at this point")
     }
-
-    /// Computes the public key corresponding to this descriptor key.
-    /// When deriving from an XOnlyPublicKey, it adds the default 0x02 y-coordinate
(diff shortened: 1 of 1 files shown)


sanket1729:
Perhaps, @LLFourn may some comments here.

apoelstra:
I think we could add a `derive_no_wildcard` method or something which lets you do the conversion (and fails if there are actually any wildcards).

It could be as simple as just checking whether wildcards exist, and if not, calling `at_derivation_index(0)` internally, though maybe we want to do something more clever.

The point of #448 was that you wouldn't need to use dummy indices so I'd be sad to see that return..

sanket1729:
Removed the last commit. Now, this forces the users to convert to `DefiniteDescriptorKey`.

apoelstra (approved):
ACK c14d1ceda11329a08a4a235c03cd49ff532eeedc

PR #476, opened 2022-10-09
Another attempt at `max_satisfaction_weight` fixes

evanlinjin:
Replaces #474, refer to https://github.com/rust-bitcoin/rust-miniscript/pull/474#issuecomment-1272475306

This PR has two intentions:

1. Redefine `max_satisfaction_weight` to be the difference in `TxIn` weight between "satisfied" and "unsatisfied" states. In an "unsatisfied" state, we still need to include the `scriptSigLen` varint, as well as the `witnessStackLen` (for txs with at least one segwit spend).
2. Attempt further fixes to improve accuracy of `max_satisfaction_weight`.

Comments, tests and examples have been updated to reflect the above intentions.

### Notes for reviewers

The new definition of `max_satisfaction_weight` can be seen in this comment:
https://github.com/rust-bitcoin/rust-miniscript/blob/08cff39fa862ff957c7ff96d17a0011dd6446f87/src/descriptor/mod.rs#L320-L339

### DIFF
--- a/src/descriptor/bare.rs
+++ b/src/descriptor/bare.rs
@@ -56,6 +56,26 @@ impl<Pk: MiniscriptKey> Bare<Pk> {
         Ok(())
     }
 
+    /// Computes an upper bound on the difference between a non-satisfied
+    /// `TxIn`'s `segwit_weight` and a satisfied `TxIn`'s `segwit_weight`
+    ///
+    /// Since this method uses `segwit_weight` instead of `legacy_weight`,
+    /// if you want to include only legacy inputs in your transaction,
+    /// you should remove 1WU from each input's `max_weight_to_satisfy`
+    /// for a more accurate estimate.
+    ///
+    /// Assumes all ECDSA signatures are 73 bytes, including push opcode and
+    /// sighash suffix.
+    ///
+    /// # Errors
+    /// When the descriptor is impossible to safisfy (ex: sh(OP_FALSE)).
+    pub fn max_weight_to_satisfy(&self) -> Result<usize, Error> {
+        let scriptsig_size = self.ms.max_satisfaction_size()?;
+        // scriptSig varint difference between non-satisfied (0) and satisfied
+        let scriptsig_varint_diff = varint_len(scriptsig_size) - varint_len(0);
+        Ok(4 * (scriptsig_varint_diff + scriptsig_size))
+    }
+
     /// Computes an upper bound on the weight of a satisfying witness to the
     /// transaction.
     ///
@@ -65,6 +85,7 @@ impl<Pk: MiniscriptKey> Bare<Pk> {
     ///
     /// # Errors
     /// When the descriptor is impossible to safisfy (ex: sh(OP_FALSE)).
+    #[deprecated(note = "use max_weight_to_satisfy instead")]
     pub fn max_satisfaction_weight(&self) -> Result<usize, Error> {
         let scriptsig_len = self.ms.max_satisfaction_size()?;
         Ok(4 * (varint_len(scriptsig_len) + scriptsig_len))
@@ -200,6 +221,27 @@ impl<Pk: MiniscriptKey> Pkh<Pk> {
     /// Get the inner key
     pub fn into_inner(self) -> Pk {
         self.pk
+    }
+
+    /// Computes an upper bound on the difference between a non-satisfied
+    /// `TxIn`'s `segwit_weight` and a satisfied `TxIn`'s `segwit_weight`
+    ///
+    /// Since this method uses `segwit_weight` instead of `legacy_weight`,
+    /// if you want to include only legacy inputs in your transaction,
+    /// you should remove 1WU from each input's `max_weight_to_satisfy`
+    /// for a more accurate estimate.
+    ///
+    /// Assumes all ECDSA signatures are 73 bytes, including push opcode and
+    /// sighash suffix.
+    ///
+    /// # Errors
+    /// When the descriptor is impossible to safisfy (ex: sh(OP_FALSE)).
+    pub fn max_weight_to_satisfy(&self) -> usize {
+        // OP_72 + <sig(71)+sigHash(1)> + OP_33 + <pubkey>
+        let scriptsig_size = 73 + BareCtx::pk_len(&self.pk);
+        // scriptSig varint different between non-satisfied (0) and satisfied
+        let scriptsig_varint_diff = varint_len(scriptsig_size) - varint_len(0);
+        4 * (scriptsig_varint_diff + scriptsig_size)
     }
 
     /// Computes an upper bound on the weight of a satisfying witness to the
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -307,6 +307,56 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
         }
     }
 
+    /// Computes an upper bound on the difference between a non-satisfied
+    /// `TxIn`'s `segwit_weight` and a satisfied `TxIn`'s `segwit_weight`
+    ///
+    /// Since this method uses `segwit_weight` instead of `legacy_weight`,
+    /// if you want to include only legacy inputs in your transaction,
+    /// you should remove 1WU from each input's `max_weight_to_satisfy`
+    /// for a more accurate estimate.
+    ///
+    /// In other words, for segwit inputs or legacy inputs included in
+    /// segwit transactions, the following will hold for each input if
+    /// that input was satisfied with the largest possible witness:
+    /// ```ignore
+    /// for i in 0..transaction.input.len() {
+    ///     assert_eq!(
+    ///         descriptor_for_input[i].max_weight_to_satisfy(),
+    ///         transaction.input[i].segwit_weight() - Txin::default().segwit_weight()
+    ///     );
+    /// }
+    /// ```
+    ///
+    /// Instead, for legacy transactions, the following will hold for each input
+    /// if that input was satisfied with the largest possible witness:
+    /// ```ignore
+    /// for i in 0..transaction.input.len() {
+    ///     assert_eq!(
+    ///         descriptor_for_input[i].max_weight_to_satisfy(),
+    ///         transaction.input[i].legacy_weight() - Txin::default().legacy_weight()
+    ///     );
+    /// }
+    /// ```
+    ///
+    /// Assumes all ECDSA signatures are 73 bytes, including push opcode and
+    /// sighash suffix.
+    /// Assumes all Schnorr signatures are 66 bytes, including push opcode and
+    /// sighash suffix.
+    ///
+    /// # Errors
+    /// When the descriptor is impossible to safisfy (ex: sh(OP_FALSE)).
+    pub fn max_weight_to_satisfy(&self) -> Result<usize, Error> {
+        let weight = match *self {
+            Descriptor::Bare(ref bare) => bare.max_weight_to_satisfy()?,
+            Descriptor::Pkh(ref pkh) => pkh.max_weight_to_satisfy(),
+            Descriptor::Wpkh(ref wpkh) => wpkh.max_weight_to_satisfy(),
+            Descriptor::Wsh(ref wsh) => wsh.max_weight_to_satisfy()?,
+            Descriptor::Sh(ref sh) => sh.max_weight_to_satisfy()?,
+            Descriptor::Tr(ref tr) => tr.max_weight_to_satisfy()?,
+        };
+        Ok(weight)
+    }
+
     /// Computes an upper bound on the weight of a satisfying witness to the
     /// transaction.
     ///
@@ -316,6 +366,8 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
     ///
     /// # Errors
     /// When the descriptor is impossible to safisfy (ex: sh(OP_FALSE)).
+    #[deprecated(note = "use max_weight_to_satisfy instead")]
+    #[allow(deprecated)]
     pub fn max_satisfaction_weight(&self) -> Result<usize, Error> {
         let weight = match *self {
             Descriptor::Bare(ref bare) => bare.max_satisfaction_weight()?,
--- a/src/descriptor/segwitv0.rs
+++ b/src/descriptor/segwitv0.rs
@@ -72,6 +72,34 @@ impl<Pk: MiniscriptKey> Wsh<Pk> {
         Ok(())
     }
 
+    /// Computes an upper bound on the difference between a non-satisfied
+    /// `TxIn`'s `segwit_weight` and a satisfied `TxIn`'s `segwit_weight`
+    ///
+    /// Assumes all ECDSA signatures are 73 bytes, including push opcode and
+    /// sighash suffix.
+    ///
+    /// # Errors
+    /// When the descriptor is impossible to safisfy (ex: sh(OP_FALSE)).
+    pub fn max_weight_to_satisfy(&self) -> Result<usize, Error> {
+        let (redeem_script_size, max_sat_elems, max_sat_size) = match self.inner {
+            WshInner::SortedMulti(ref smv) => (
+                smv.script_size(),
+                smv.max_satisfaction_witness_elements(),
+                smv.max_satisfaction_size(),
+            ),
+            WshInner::Ms(ref ms) => (
+                ms.script_size(),
+                ms.max_satisfaction_witness_elements()?,
+                ms.max_satisfaction_size()?,
+            ),
+        };
+        // stack size varint difference between non-satisfied (0) and satisfied
+        // `max_sat_elems` is inclusive of the "witness script" (redeem script)
+        let stack_varint_diff = varint_len(max_sat_elems) - varint_len(0);
+
+        Ok(stack_varint_diff + varint_len(redeem_script_size) + redeem_script_size + max_sat_size)
+    }
+
     /// Computes an upper bound on the weight of a satisfying witness to the
     /// transaction.
     ///
@@ -81,6 +109,7 @@ impl<Pk: MiniscriptKey> Wsh<Pk> {
     ///
     /// # Errors
     /// When the descriptor is impossible to safisfy (ex: sh(OP_FALSE)).
+    #[deprecated(note = "use max_weight_to_satisfy instead")]
     pub fn max_satisfaction_weight(&self) -> Result<usize, Error> {
         let (script_size, max_sat_elems, max_sat_size) = match self.inner {
             WshInner::SortedMulti(ref smv) => (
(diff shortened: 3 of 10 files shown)


evanlinjin:
@LLFourn A review when you have time would be appreciated :pray:

LLFourn (commented):
Nice work. I think this is a coherent and useful definition of `max_satisfaction_weight`. I was actually beginning to wonder if this was even possible.

Not sure if it's wise to simply change it under people's noses. Everyone who was using `max_satisfaction_weight` before was probably making a mistake or two but this breaking change seems likely to lead to undershooting of fees even when this would only happen in previously when they were doing something weird like adding segwit and non-segwit coins at the same time.

Maybe this should be a new method like `max_weight_to_satisfy` or something less awkward and deprecate `max_satisfaction_weight` in favor of it and stressing the need to read the documentation.

src/lib.rs
```diff
@@ -74,7 +74,14 @@
...
 //! // Estimate the satisfaction cost
-//! assert_eq!(desc.max_satisfaction_weight().unwrap(), 293);
+//! // scriptSig:
+//! // * OP_0 OP_32 <32-byte-hash>
+//! // = (1 + 1 + 32) * 4 = 136 WU
+//! // redeemScript: varint <OP_33 <pk1> OP_CHECKSIG OP_IFDUP OP_NOTIF OP_33 <pk2> OP_CHECKSIG OP_ENDIF>
+//! // = 1 + (1 + 33 + 1 + 1 + 1 + 1 + 33 + 1 + 1) = 74 WU
+//! // stackItem[Sig]: varint <sig+sighash> = 1 + 72+1 = 74 WU (TODO: Figure out why sig is 72bytes here)
+//! // Expected satisfaction weight: 136 + 74 + 74 = 284
+//! assert_eq!(desc.max_satisfaction_weight().unwrap(), 284);
```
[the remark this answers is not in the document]
evanlinjin: I'm pretty sure the original calculations were wrong. I recalculated in the comments, but let me know if I've missed something.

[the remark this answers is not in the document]
LLFourn: But it was asserted in the doctest so they should have run in CI and reflect what the previous version did. Something else must be going on...

evanlinjin:
@apoelstra Could you please have a look at this PR when you have time? Would love to know your thoughts on these changes and proposed definition change of "satisfaction weight". Thank you in advance!

src/descriptor/bare.rs
```diff
@@ -66,18 +66,16 @@ impl<Pk: MiniscriptKey> Bare<Pk> {
...
+    /// Computes an upper bound on the difference in weight between a
+    /// non-satisfied `TxIn` (with empty `scriptSig` and `witness` fields) and a
+    /// satisfied `TxIn`.
     ///
-    /// Assumes all ec-signatures are 73 bytes, including push opcode and
-    /// sighash suffix. Includes the weight of the VarInts encoding the
-    /// scriptSig and witness stack length.
-    ///
-    /// # Errors
-    /// When the descriptor is impossible to safisfy (ex: sh(OP_FALSE)).
```
danielabrozzoni, line 77: Why is the Errors section being removed?

src/descriptor/mod.rs
```diff
@@ -317,12 +317,22 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
+    /// that have at least one witness spend, which will also record the value 0
+    /// (which is 1WU).
+    ///
+    /// Hence, "satisfaction weight" contains the additional varint weight of
+    /// `scriptSigLen` and `witnessStackLen` (which occurs when the varint value
+    /// increases over the 1byte threshold), and also the weights of `scriptSig`
+    /// and the rest of the `witnessField` (which is the stack items and the
+    /// stack-item-len varints of each item).
+    ///
+    /// This assumes a ec-signatures is always 73 bytes, and the sighash prefix
```
danielabrozzoni, line 334: nit: This assumes a ec-signatures **are** always 73 bytes

src/descriptor/mod.rs
```diff
@@ -317,12 +317,22 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
+    /// is always included (+1 byte).
```
danielabrozzoni, line 335: nit: I liked the old comment a bit better. From the old comment, it seemed that 73 bytes included the push opcode and the sighash suffix (making the actual signature 71 bytes). From this new version, it seems that ec signatues are assumed to be 73 bytes +1 byte for the sighash = 74 bytes

(But maybe it's just me)

danielabrozzoni (commented):
I have a couple of nits on wording, I just did a quick scroll, I didn't look at the code in deep.

I'm not fond of changing the definition of `max_sat_weight`, but I think it'd be great to implement LLoyds' suggestion of having a new method (`max_weight_to_satisfy`) instead.

After a chat with @evanlinjin, we decided that I'll take over this PR, as he has a lot on his plate already :). I'm going to wait for some feedback on `max_weight_to_satisfy` (or however we want to call it), and then go over the comments and update the code accordingly

src/descriptor/mod.rs
apoelstra: `a` should also be `that` :)

apoelstra:
Thanks @danielabrozzoni !

I agree with all of your nits. ACK 238014a09c1ecd84a603d72dfa9ea1b64338a26d otherwise.

Definitely agree that we should keep the old method, deprecate it, and make a new one with a new name. This is fine to go into 9.1 (since I am going to try to release 9.0 ASAP, see #488 ). I have no opinion on name. `max_weight_to_satisfy` sounds fine to me.

danielabrozzoni:
I just pushed `734acee1fb941ba283e79c0603d413a4e0f4ac0f`:
- rebased
- deprecated `max_satisfaction_weight` and introduced `max_weight_to_satisfy` instead, updated examples accordingly
- fixed nits in docs 
- fixed some bugs in this PR's code (see https://github.com/rust-bitcoin/rust-miniscript/pull/476#discussion_r1025382147, and also the same bug was in the p2sh-p2wpkh case)

apoelstra (dismissed):
ACK 734acee1fb941ba283e79c0603d413a4e0f4ac0f

src/descriptor/mod.rs
```diff
@@ -317,15 +317,50 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
+    /// non-satisfied `TxIn` and a satisfied `TxIn`.
+    ///
+    /// A non-satisfied `TxIn` contains the `scriptSigLen` varint recording the
+    /// value 0 (which is 4WU) and also the `witnessStackLen` for transactions
+    /// that have at least one witness spend, which will also record the value 0
+    /// (which is 1WU).
+    ///
+    /// Hence, "satisfaction weight" contains the additional varint weight of
+    /// `scriptSigLen` and `witnessStackLen` (which occurs when the varint value
+    /// increases over the 1byte threshold), and also the weights of `scriptSig`
```
LLFourn, line 330: So what this is wants to say about scriptSigLen and witnessStackLen is:
1. The witnessStackLen is is assumed to be 1 for segwit descriptors but 0  for legacy
2. The scriptSigLen is always by default 1 regardless.
3. This function will tell you the include how much satisfying the inputs will increase these over these assumed defaults.

I wonder if is possible to make this clearer.

I made this issue at https://github.com/rust-bitcoin/rust-bitcoin/issues/1411 where it seems I was mistaken about the definition we were using here. I thought we were doing weight over the weight of `TxIn::default()` but we are doing one less than that for segwit descriptors since we are use 1 as a base witness stack len where as it would be 0 for `TxIn::default()`. What do you think about changing the definition to simply the weight over `TxIn::default()` once it is satisfied (assuming `Txin::weight` existed) i.e. this always holds:

```rust 
for i in 0..transaction.input.len() {
    assert_eq!(descriptor_for_input[i].max_weight_to_satisfy(), transaction.input[i].weight() - Txin::default().weight()_
}
```
in the case where transaction is valid and you used the worst satisfaction path from the corresponding input in `descriptor_for_input`.
This might be a bit easier to explain (assuming `TxIn::weight` is coherent and easy to explain itself!).

apoelstra: >What do you think about changing the definition to simply the weight over `TxIn::default()` once it is satisfied (assuming `Txin::weight` existed) i.e. this always holds:

I think this is a good idea. I also tripped over the "starting from 1" logic here. It does make sense, and presumably wallet authors want us to return the smallest possible number that is strictly true (or maybe they want us to return the most predictable and reproducible number, so if this is used for validity checks we won't get inconsistencies like the one that crashed Liquid?), so I let it slide ... but IMO it's a bit surprising and confusing.

apoelstra: Maybe the right way to think about this function is that it's used for sanity/validity checking, and we want it to be a simple-to-explain value which tries not to do ad-hoc computations that feel likely to change between version of the library.

When wallets actually want to do fee estimation at spend-time, they presumably have more information and can use the planning API #481 and we can put as many accounting tricks as we want into that.

danielabrozzoni: Thanks for the feedback!I think you're right, having a coherent definition of `TxIn` and basing the `max_weight_to_satisfy` calculation on that sounds less error-prone for the end user. I'll open a PR on rust-bitcoin soon :)

evanlinjin: From my understanding, we aren't changing the implementation of `max_weight_to_satisfy`, but changing how we define things to make it clearer to the end user. We just say, "we assume `TxIn::default()`'s weight always includes 1 byte for `size(witness_len)`".

LLFourn: No I am suggesting changing the implementation. If `TxIn` were to have a `weight` function then I think it would have 0 weight for the witness tack length for `TxIn::default`.

LLFourn: That's not what the current implementation does if I am not mistaken. It assumes 0 witness stack len for non segwit descriptor and 1 for a segwit descriptor.

> But if we assume at least one input is a segwit spend (which is also the "safer" definition)

Why is this safer though. What guarantee does doing this provide? It doesn't guarantee that if you add all the `max_weight_to_satisfy` of all the inputs together that the sum is an upper bound on the weight contribution of the inputs. You have to know whether one input was segwit to know whether to add the 2WU segwit header. Since you always need to figure out whether there is a segwit input anyway then you might as well use it to add +1 WU for all non-segwit inputs while you're at it. This gives you a correct upper bound on the input weight (if you add the weight of the length byte for the number of inputs).

With assuming that the witness stack length is always 1WU for non-segwit inputs you can still get the correct upper bound so it's also a decent choice. You just have to go and subtract 1WU from all the non-segwit inputs in the case that all of them are non-segwit.

As @apoelstra points out there is no great answer here. We should just try and choose the easiest thing to explain correctly. To me the concept of `TxIn::weight` while not elegant is coherent enough to act as a focal point of this definition. Your explanation involving `Txin::max_weight` seems worse in this respect since what does the max weight of a `TxIn` even mean?

src/descriptor/segwitv0.rs
```diff
@@ -82,15 +82,45 @@ impl<Pk: MiniscriptKey> Wsh<Pk> {
...
+            ),
+            WshInner::Ms(ref ms) => (
+                ms.script_size(),
+                ms.max_satisfaction_witness_elements()?,
+                ms.max_satisfaction_size()?,
+            ),
+        };
+        // stack size varint difference between non-satisfied (0) and satisfied
+        // `max_sat_elems` is inclusive of the "witness script" (redeem script)
+        let stack_varint_diff = varint_len(max_sat_elems) - varint_len(0);
```
evanlinjin, line 109: If `max_satisfaction_witness_elements` does not include the "redeem script" item (which I assume it doesn't...

```suggestion
        let stack_varint_diff = varint_len(max_sat_elems + 1) - varint_len(0);
```

src/descriptor/mod.rs
danielabrozzoni: > That's not what the current implementation does if I am not mistaken. It assumes 0 witness stack len for non segwit descriptor and 1 for a segwit descriptor.

Yep, correct

> Since you always need to figure out whether there is a segwit input anyway then you might as well use that information to add +1 WU for all non-segwit inputs while you're at it.

(Note that this is not always possible to do: if you're creating a tx spending some P2SH utxos that aren't yours, you can't know before the other party signs if they're legacy or segwit wrapped, and `max_weight_to_satisfy` is often used before signing. But in case you don't you can just assume segwit, and you're fine)

I think we all agree that there's no way to produce the right result without the need for a "second pass" after the whole transaction is made.

Since I believe most users will forget to do that next step (and many may not care to do it at all) I think it's better if we return something that can be used "safely" straight away, even if it means potentially overshooting the fee. So always assuming segwit seems the best choice to me.

More advanced users could then check the tx and subtract -1 * n_inputs if the tx is non-segwit. We can provide an example for that and try to document it the best we can.

evanlinjin: @LLFourn I'm a little confused by what you mean, so to make sure we are on the same page I will give thoughts to your original comment...

> The witnessStackLen is assumed to be 1 for segwit descriptors but 0 for legacy

From my understanding this is not true. The definition of `max_weight_to_satisfy` is the **additional weight** introduced from satisfying a `TxIn`. For satisfying legacy descriptors, we will never add additional items to the witness stack, so the additional weight of witnessStackLen for legacy descriptors is 0.

This is independent to whether `TxIn::default()` (an unsatisfied txin) contains the `witnessStackLen` field.

LLFourn: > @LLFourn I'm a little confused by what you mean, so to make sure we are on the same page I will give thoughts to your original comment...
> 
> > The witnessStackLen is assumed to be 1 for segwit descriptors but 0 for legacy
> 
> From my understanding this is not true. The definition of `max_weight_to_satisfy` is the **additional weight** introduced from satisfying a `TxIn`. For satisfying legacy descriptors, we will never add additional items to the witness stack, so the additional weight of witnessStackLen for legacy descriptors is 0.

Correct. I don't think what you said is in contradiction to what I said.

> But if we assume at least one input is a segwit spend (which is also the "safer" definition), the current implementation can be worded as `<miniscript_thing>.max_weight_to_satisfy() - TxIn::default().max_weight()`, where `TxIn::max_weight` assumes segwit spend.

Ok now I understand what you were saying here and I think we are in the same page. I still think this definition is awkward though (I was at least confused by it!).

src/descriptor/segwitv0.rs
danielabrozzoni: Seems like `max_satisfaction_witness_elements` includes the redeem script, as the docs say "Maximum number of witness elements used to satisfy the Miniscript fragment, including the witness script itself."

danielabrozzoni:
I opened https://github.com/rust-bitcoin/rust-bitcoin/pull/1467 to introduce `TxIn::segwit_weight()` and `TxIn::legacy_weight()`, then pushed here a rewording of the definition of `max_satisfaction_weight`, which hopefully is clearer.

For segwit inputs or legacy inputs included in segwit txs, it holds:
```
for i in 0..transaction.input.len() {
    assert_eq!(descriptor_for_input[i].max_weight_to_satisfy(), transaction.input[i].segwit_weight() - Txin::default().segwit_weight())
}
```

Instead, for legacy transactions, it holds:
```
for i in 0..transaction.input.len() {
    assert_eq!(descriptor_for_input[i].max_weight_to_satisfy(), transaction.input[i].legacy_weight() - Txin::default().legacy_weight())
}
```

Which I think is pretty neat :)

Once https://github.com/rust-bitcoin/rust-bitcoin/pull/1467 is merged and miniscript is up-to-date with the latest rust-bitcoin, I can add a test to verify that the two invariants above hold.

src/descriptor/mod.rs
```diff
@@ -317,6 +317,59 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
+    ///
+    /// Since this method uses `segwit_weight` instead of `legacy_weight`,
+    /// if you want to include only legacy inputs in your transaction,
+    /// you should remove 1WU from each input's `max_weight_to_satisfy`
+    /// for a more accurate estimate.
+    ///
+    /// Since this method uses `segwit_weight` instead of `legacy_weight`,
+    /// if you want to include only legacy inputs in your transaction,
+    /// you should remove 1WU from each input's `max_weight_to_satisfy`
+    /// for a more accurate estimate.
```
LLFourn, line 331: paragraph duplicated.

src/descriptor/mod.rs
```diff
@@ -317,6 +317,59 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
+    ///
+    /// In other words, for segwit inputs or legacy inputs included in
+    /// segwit transactions, it holds:
+    /// ```ignore
+    /// for i in 0..transaction.input.len() {
+    ///     assert_eq!(
+    ///         descriptor_for_input[i].max_weight_to_satisfy(),
+    ///         transaction.input[i].segwit_weight() - Txin::default().segwit_weight()
+    ///     );
+    /// }
```
LLFourn, line 341: I think you need specify something like this:

> /// In other words, for segwit inputs or legacy inputs included in
> /// segwit transactions, the following will hold for each input if that input was satisfied with the largest possible witness.

It would be nice to have some tests against random descriptors that checks this does actually hold. Do we have a method of doing proptests like this already?

LLFourn:
New definition and docs really makes a lot of sense. I think as close to making sense as you can when trying to explain this stuff. I couple of nits but otherwise utACK

src/descriptor/mod.rs
danielabrozzoni: Thanks, updated!

> It would be nice to have some tests against random descriptors that checks this does actually hold.

I agree that it would be nice to have some tests, but unfortunately we have to wait for the newest rust-bitcoin to be released and for rust-miniscript to be updated before being able to write such tests. 

> Do we have a method of doing proptests like this already?

I'm not entirely sure, I guess I could expand `tests/test_desc.rs`

LLFourn: Ah right. Tests for this equality will have to wait I guess.

apoelstra: Re proptests, we have a foray into Kani in rust-bitcoin https://github.com/rust-bitcoin/rust-bitcoin/pull/1415 but I doubt Kani (which does model-checking using a SMT solver) could handle stuff here, unfortunately.

apoelstra:
It looks like this new version uses "ec-signatures" everywhere rather than "ECDSA`. I skimmed through the earlier discussion and didn't see, so I hope I'm not relitigating stuff, but I'm pretty sure that "ECDSA" is correct in every instance. Schnorr sigs are also EC sigs, and they are always 64/65 bytes.

Anyway I won't hold up this PR over that, it's not a huge deal.

apoelstra (approved):
ACK f3c7ad899c4b6e271b6641f8ca0612382fcc99d0

danielabrozzoni:
> It looks like this new version uses "ec-signatures" everywhere rather than "ECDSA`. I skimmed through the earlier discussion and didn't see, so I hope I'm not relitigating stuff, but I'm pretty sure that "ECDSA" is correct in every instance. Schnorr sigs are also EC sigs, and they are always 64/65 bytes.

I'm using "ECDSA signatures" *almost* everywhere (there's still one instance of "ec signatures" which I missed in `src/descriptor/sh.rs`, my bad), saying:
```
    /// Assumes all ECDSA signatures are 73 bytes, including push opcode and
    /// sighash suffix.
    /// Assumes all Schnorr signatures are 66 bytes, including push opcode and
    /// sighash suffix.
```

If you prefer to wait for tests before merging this, I can remove the "ec signatures" from `sh.rs` as well, otherwise, I can fix in the PR adding the tests, as you wish :)

apoelstra:
I'm going to wait on @sanket1729 to merge this since it's non-trivial and he's been the primary maintainer of the library lately. I don't know what his timeline is, but I think you have time to add tests and tweak the docs :). I can quickly re-ack any minor changes.

sanket1729 (dismissed):
ACK f3c7ad899c4b6e271b6641f8ca0612382fcc99d0

sanket1729:
ACK modulo requiring rebase

danielabrozzoni:
Rebased

sanket1729 (approved):
ACK 7296f8ee29c2ae8b5d582b55898743ad3ef4f8a1

sanket1729:
Let's the new year with a bang!

evanlinjin:
Thank you @danielabrozzoni for taking this forward!

PR #478, opened 2022-10-18
allow disabling the checksum with alternate `Display`

### LINKED ISSUE #477: feature request: `#` flag should turn off checksum when formatting descriptors
This would avoid the need for hacks like `to_string_no_chksum` in elements-miniscript, where we wrap descriptors in other stuff and add a new global checksum.

apoelstra:
Fixes #477

### DIFF
--- a/src/descriptor/checksum.rs
+++ b/src/descriptor/checksum.rs
@@ -3,13 +3,15 @@
 //! This module contains a re-implementation of the function used by Bitcoin Core to calculate the
 //! checksum of a descriptor
 
+#![allow(dead_code)] // will be removed in next commit
+use core::fmt;
 use core::iter::FromIterator;
 
 use crate::prelude::*;
 use crate::Error;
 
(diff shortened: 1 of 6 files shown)


src/descriptor/checksum.rs
```diff
@@ -99,6 +70,102 @@ pub(super) fn verify_checksum(s: &str) -> Result<&str, Error> {
...
+            self.fmt.write_char(ch)?;
+        }
+        Ok(())
+    }
+}
+
+impl<'f, 'a> fmt::Write for Formatter<'f, 'a> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        self.fmt.write_str(s)?;
+        if self.eng.input(s).is_ok() {
```
sanket1729, line 169: nit: This can be written as self.engine.input(s).map_err(|| fmt::Error)

sanket1729 (approved):
ACK 7f5bbdf379db2357dd6449e991805dcceb93106a. The nit is just a stylistic comment.

PR #481, opened 2022-10-25
Add the "planning" module

afilini:
This PR modifies the satisfaction logic so that instead of building a full witness it builds "witness templates", essentially containing placeholders instead of the actual signatures/preimages.

Once you have a "witness template" you can then analyze it (figure out the precise weight) and interactively (or in single shot) try to construct an actual witness by replacing the placeholders with actual values.

We introduce the `AssetProvider` trait which somewhat mirrors the `Satisfier` trait, but only returns whether something is available or not. For convenience the `AssetProvider` trait is automatically implemented for every `Satisfier` through a wrapping structure.

Finally, we also introduce the `Assets` structure which is a builder-like struct that implements the `AssetProvider` trait and can be used to easily tell miniscript which keys and preimages are available, or how long one is willing to wait for timelocks to expire. Using this, one can easily figure out, before creating a transaction, how much the witness is going to weight and what nLockTime/nSequence should be set in the transaction assuming those *assets* will be available later for signing.

The commits are co-authored by @danielabrozzoni since we've both worked together on this on and off for the past couple of weeks.

### DIFF
--- a/src/descriptor/bare.rs
+++ b/src/descriptor/bare.rs
@@ -13,8 +13,11 @@ use bitcoin::script::{self, PushBytes};
 use bitcoin::{Address, Network, ScriptBuf};
 
 use super::checksum::{self, verify_checksum};
+use crate::descriptor::DefiniteDescriptorKey;
 use crate::expression::{self, FromTree};
 use crate::miniscript::context::{ScriptContext, ScriptContextError};
+use crate::miniscript::satisfy::{Placeholder, Satisfaction, Witness};
+use crate::plan::AssetProvider;
 use crate::policy::{semantic, Liftable};
 use crate::prelude::*;
 use crate::util::{varint_len, witness_to_scriptsig};
@@ -131,6 +134,30 @@ impl<Pk: MiniscriptKey + ToPublicKey> Bare<Pk> {
         let script_sig = witness_to_scriptsig(&ms);
         let witness = vec![];
         Ok((witness, script_sig))
+    }
+}
+
+impl Bare<DefiniteDescriptorKey> {
+    /// Returns a plan if the provided assets are sufficient to produce a non-malleable satisfaction
+    pub fn plan_satisfaction<P>(
+        &self,
+        provider: &P,
+    ) -> Satisfaction<Placeholder<DefiniteDescriptorKey>>
+    where
+        P: AssetProvider<DefiniteDescriptorKey>,
+    {
+        self.ms.build_template(provider)
+    }
+
+    /// Returns a plan if the provided assets are sufficient to produce a malleable satisfaction
+    pub fn plan_satisfaction_mall<P>(
+        &self,
+        provider: &P,
+    ) -> Satisfaction<Placeholder<DefiniteDescriptorKey>>
+    where
+        P: AssetProvider<DefiniteDescriptorKey>,
+    {
+        self.ms.build_template_mall(provider)
     }
 }
 
@@ -308,6 +335,45 @@ impl<Pk: MiniscriptKey + ToPublicKey> Pkh<Pk> {
         S: Satisfier<Pk>,
     {
         self.get_satisfaction(satisfier)
+    }
+}
+
+impl Pkh<DefiniteDescriptorKey> {
+    /// Returns a plan if the provided assets are sufficient to produce a non-malleable satisfaction
+    pub fn plan_satisfaction<P>(
+        &self,
+        provider: &P,
+    ) -> Satisfaction<Placeholder<DefiniteDescriptorKey>>
+    where
+        P: AssetProvider<DefiniteDescriptorKey>,
+    {
+        let stack = if provider.provider_lookup_ecdsa_sig(&self.pk) {
+            let stack = vec![
+                Placeholder::EcdsaSigPk(self.pk.clone()),
+                Placeholder::Pubkey(self.pk.clone(), BareCtx::pk_len(&self.pk)),
+            ];
+            Witness::Stack(stack)
+        } else {
+            Witness::Unavailable
+        };
+
+        Satisfaction {
+            stack,
+            has_sig: true,
+            relative_timelock: None,
+            absolute_timelock: None,
+        }
+    }
+
+    /// Returns a plan if the provided assets are sufficient to produce a malleable satisfaction
+    pub fn plan_satisfaction_mall<P>(
+        &self,
+        provider: &P,
+    ) -> Satisfaction<Placeholder<DefiniteDescriptorKey>>
+    where
+        P: AssetProvider<DefiniteDescriptorKey>,
+    {
+        self.plan_satisfaction(provider)
     }
 }
 
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -592,6 +592,33 @@ impl DescriptorPublicKey {
         }
     }
 
+    /// Returns a vector containing the full derivation paths from the master key.
+    /// The vector will contain just one element for single keys, and multiple elements
+    /// for multipath extended keys.
+    ///
+    /// For wildcard keys this will return the path up to the wildcard, so you
+    /// can get full paths by appending one additional derivation step, according
+    /// to the wildcard type (hardened or normal).
+    pub fn full_derivation_paths(&self) -> Vec<bip32::DerivationPath> {
+        match self {
+            DescriptorPublicKey::MultiXPub(xpub) => {
+                let origin_path = if let Some((_, ref path)) = xpub.origin {
+                    path.clone()
+                } else {
+                    bip32::DerivationPath::from(vec![])
+                };
+                xpub.derivation_paths
+                    .paths()
+                    .into_iter()
+                    .map(|p| origin_path.extend(p))
+                    .collect()
+            }
+            _ => vec![self
+                .full_derivation_path()
+                .expect("Must be Some for non-multipath keys")],
+        }
+    }
+
     /// Whether or not the key has a wildcard
     #[deprecated(note = "use has_wildcard instead")]
     pub fn is_deriveable(&self) -> bool {
@@ -1075,6 +1102,12 @@ impl DefiniteDescriptorKey {
         self.0.full_derivation_path()
     }
 
+    /// Full paths from the master key. The vector will contain just one path for single
+    /// keys, and multiple ones for multipath extended keys
+    pub fn full_derivation_paths(&self) -> Vec<bip32::DerivationPath> {
+        self.0.full_derivation_paths()
+    }
+
     /// Reference to the underlying `DescriptorPublicKey`
     pub fn as_descriptor_public_key(&self) -> &DescriptorPublicKey {
         &self.0
@@ -1476,6 +1509,14 @@ mod test {
         let desc_key = DescriptorPublicKey::from_str("[abcdef00/0'/1']tpubDBrgjcxBxnXyL575sHdkpKohWu5qHKoQ7TJXKNrYznh5fVEGBv89hA8ENW7A8MFVpFUSvgLqc4Nj1WZcpePX6rrxviVtPowvMuGF5rdT2Vi/9478'/<0';1>/8h/*'").unwrap();
         assert!(desc_key.full_derivation_path().is_none());
         assert!(desc_key.is_multipath());
+        // But you can get all the derivation paths
+        assert_eq!(
+            desc_key.full_derivation_paths(),
+            vec![
+                bip32::DerivationPath::from_str("m/0'/1'/9478'/0'/8'").unwrap(),
+                bip32::DerivationPath::from_str("m/0'/1'/9478'/1/8'").unwrap(),
+            ],
+        );
         assert_eq!(desc_key.into_single_keys(), vec![DescriptorPublicKey::from_str("[abcdef00/0'/1']tpubDBrgjcxBxnXyL575sHdkpKohWu5qHKoQ7TJXKNrYznh5fVEGBv89hA8ENW7A8MFVpFUSvgLqc4Nj1WZcpePX6rrxviVtPowvMuGF5rdT2Vi/9478'/0'/8h/*'").unwrap(), DescriptorPublicKey::from_str("[abcdef00/0'/1']tpubDBrgjcxBxnXyL575sHdkpKohWu5qHKoQ7TJXKNrYznh5fVEGBv89hA8ENW7A8MFVpFUSvgLqc4Nj1WZcpePX6rrxviVtPowvMuGF5rdT2Vi/9478'/1/8h/*'").unwrap()]);
 
         // All the same but with extended private keys instead of xpubs.
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -721,6 +721,41 @@ impl DescriptorPublicKey {
             }
         }
     }
+
+    /// Whether this key is the "parent" of a [`DefiniteDescriptorKey`]
+    ///
+    /// The key is considered "parent" if it represents the non-derived version of a definite key,
+    /// meaning it contains a wildcard where the definite key has a definite derivation number.
+    ///
+    /// If `self` is a single key or doesn't contain any wildcards, the definite key will have to
+    /// be exactly the same.
+    ///
+    /// Returns the derivation path to apply to `self` to obtain the definite key.
+    pub fn is_parent(&self, definite_key: &DefiniteDescriptorKey) -> Option<bip32::DerivationPath> {
+        // If the key is `Single` or it's an `XPub` with no wildcard it will match the definite key
+        // exactly, so we try this check first
+        if self == &definite_key.0 {
+            return Some(bip32::DerivationPath::default());
+        }
+
+        match (self, &definite_key.0) {
+            (DescriptorPublicKey::XPub(self_xkey), DescriptorPublicKey::XPub(definite_xkey))
+                if definite_xkey.derivation_path.len() > 0 =>
+            {
+                let definite_path_len = definite_xkey.derivation_path.len();
+                if self_xkey.origin == definite_xkey.origin
+                    && self_xkey.xkey == definite_xkey.xkey
+                    && self_xkey.derivation_path.as_ref()
+                        == &definite_xkey.derivation_path[..(definite_path_len - 1)]
+                {
+                    Some(vec![definite_xkey.derivation_path[definite_path_len - 1]].into())
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        }
+    }
 }
 
 impl FromStr for DescriptorSecretKey {
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -22,7 +22,8 @@ use sync::Arc;
 
 use self::checksum::verify_checksum;
 use crate::miniscript::decode::Terminal;
-use crate::miniscript::{Legacy, Miniscript, Segwitv0};
+use crate::miniscript::{satisfy, Legacy, Miniscript, Segwitv0};
+use crate::plan::{AssetProvider, Plan};
 use crate::prelude::*;
 use crate::{
     expression, hash256, BareCtx, Error, ForEachKey, MiniscriptKey, Satisfier, ToPublicKey,
@@ -474,7 +475,7 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             Descriptor::Wpkh(ref wpkh) => wpkh.get_satisfaction(satisfier),
             Descriptor::Wsh(ref wsh) => wsh.get_satisfaction(satisfier),
             Descriptor::Sh(ref sh) => sh.get_satisfaction(satisfier),
-            Descriptor::Tr(ref tr) => tr.get_satisfaction(satisfier),
+            Descriptor::Tr(ref tr) => tr.get_satisfaction(&satisfier),
         }
     }
 
@@ -491,7 +492,7 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
             Descriptor::Wpkh(ref wpkh) => wpkh.get_satisfaction_mall(satisfier),
             Descriptor::Wsh(ref wsh) => wsh.get_satisfaction_mall(satisfier),
             Descriptor::Sh(ref sh) => sh.get_satisfaction_mall(satisfier),
-            Descriptor::Tr(ref tr) => tr.get_satisfaction_mall(satisfier),
+            Descriptor::Tr(ref tr) => tr.get_satisfaction_mall(&satisfier),
         }
     }
 
@@ -506,6 +507,64 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
         txin.witness = Witness::from_slice(&witness);
         txin.script_sig = script_sig;
         Ok(())
+    }
+}
+
+impl Descriptor<DefiniteDescriptorKey> {
+    /// Returns a plan if the provided assets are sufficient to produce a non-malleable satisfaction
+    ///
+    /// If the assets aren't sufficient for generating a Plan, the descriptor is returned
+    pub fn plan<P>(self, provider: &P) -> Result<Plan, Self>
+    where
+        P: AssetProvider<DefiniteDescriptorKey>,
+    {
+        let satisfaction = match self {
+            Descriptor::Bare(ref bare) => bare.plan_satisfaction(provider),
+            Descriptor::Pkh(ref pkh) => pkh.plan_satisfaction(provider),
+            Descriptor::Wpkh(ref wpkh) => wpkh.plan_satisfaction(provider),
+            Descriptor::Wsh(ref wsh) => wsh.plan_satisfaction(provider),
+            Descriptor::Sh(ref sh) => sh.plan_satisfaction(provider),
+            Descriptor::Tr(ref tr) => tr.plan_satisfaction(provider),
+        };
+
+        if let satisfy::Witness::Stack(stack) = satisfaction.stack {
+            Ok(Plan {
+                descriptor: self,
+                template: stack,
+                absolute_timelock: satisfaction.absolute_timelock.map(Into::into),
+                relative_timelock: satisfaction.relative_timelock,
+            })
+        } else {
+            Err(self)
+        }
+    }
+
+    /// Returns a plan if the provided assets are sufficient to produce a malleable satisfaction
+    ///
+    /// If the assets aren't sufficient for generating a Plan, the descriptor is returned
+    pub fn plan_mall<P>(self, provider: &P) -> Result<Plan, Self>
+    where
+        P: AssetProvider<DefiniteDescriptorKey>,
+    {
+        let satisfaction = match self {
+            Descriptor::Bare(ref bare) => bare.plan_satisfaction_mall(provider),
+            Descriptor::Pkh(ref pkh) => pkh.plan_satisfaction_mall(provider),
+            Descriptor::Wpkh(ref wpkh) => wpkh.plan_satisfaction_mall(provider),
+            Descriptor::Wsh(ref wsh) => wsh.plan_satisfaction_mall(provider),
+            Descriptor::Sh(ref sh) => sh.plan_satisfaction_mall(provider),
+            Descriptor::Tr(ref tr) => tr.plan_satisfaction_mall(provider),
+        };
+
+        if let satisfy::Witness::Stack(stack) = satisfaction.stack {
+            Ok(Plan {
+                descriptor: self,
+                template: stack,
+                absolute_timelock: satisfaction.absolute_timelock.map(Into::into),
+                relative_timelock: satisfaction.relative_timelock,
+            })
+        } else {
+            Err(self)
+        }
     }
 }
 
--- a/src/descriptor/segwitv0.rs
+++ b/src/descriptor/segwitv0.rs
@@ -11,8 +11,11 @@ use bitcoin::{Address, Network, ScriptBuf};
 
 use super::checksum::{self, verify_checksum};
 use super::SortedMultiVec;
+use crate::descriptor::DefiniteDescriptorKey;
 use crate::expression::{self, FromTree};
 use crate::miniscript::context::{ScriptContext, ScriptContextError};
+use crate::miniscript::satisfy::{Placeholder, Satisfaction, Witness};
+use crate::plan::AssetProvider;
 use crate::policy::{semantic, Liftable};
 use crate::prelude::*;
 use crate::util::varint_len;
@@ -188,6 +191,36 @@ impl<Pk: MiniscriptKey + ToPublicKey> Wsh<Pk> {
         witness.push(self.inner_script().into_bytes());
         let script_sig = ScriptBuf::new();
         Ok((witness, script_sig))
+    }
+}
+
+impl Wsh<DefiniteDescriptorKey> {
+    /// Returns a plan if the provided assets are sufficient to produce a non-malleable satisfaction
+    pub fn plan_satisfaction<P>(
+        &self,
+        provider: &P,
+    ) -> Satisfaction<Placeholder<DefiniteDescriptorKey>>
+    where
+        P: AssetProvider<DefiniteDescriptorKey>,
+    {
+        match &self.inner {
+            WshInner::SortedMulti(sm) => sm.build_template(provider),
+            WshInner::Ms(ms) => ms.build_template(provider),
+        }
+    }
+
+    /// Returns a plan if the provided assets are sufficient to produce a malleable satisfaction
+    pub fn plan_satisfaction_mall<P>(
+        &self,
+        provider: &P,
+    ) -> Satisfaction<Placeholder<DefiniteDescriptorKey>>
+    where
+        P: AssetProvider<DefiniteDescriptorKey>,
+    {
+        match &self.inner {
+            WshInner::SortedMulti(sm) => sm.build_template(provider),
+            WshInner::Ms(ms) => ms.build_template_mall(provider),
+        }
     }
 }
 
@@ -415,6 +448,45 @@ impl<Pk: MiniscriptKey + ToPublicKey> Wpkh<Pk> {
         S: Satisfier<Pk>,
     {
         self.get_satisfaction(satisfier)
+    }
+}
+
+impl Wpkh<DefiniteDescriptorKey> {
+    /// Returns a plan if the provided assets are sufficient to produce a non-malleable satisfaction
+    pub fn plan_satisfaction<P>(
+        &self,
+        provider: &P,
+    ) -> Satisfaction<Placeholder<DefiniteDescriptorKey>>
+    where
+        P: AssetProvider<DefiniteDescriptorKey>,
+    {
+        let stack = if provider.provider_lookup_ecdsa_sig(&self.pk) {
+            let stack = vec![
+                Placeholder::EcdsaSigPk(self.pk.clone()),
+                Placeholder::Pubkey(self.pk.clone(), Segwitv0::pk_len(&self.pk)),
+            ];
+            Witness::Stack(stack)
+        } else {
+            Witness::Unavailable
+        };
+
+        Satisfaction {
+            stack,
+            has_sig: true,
+            relative_timelock: None,
+            absolute_timelock: None,
+        }
+    }
+
+    /// Returns a plan if the provided assets are sufficient to produce a malleable satisfaction
+    pub fn plan_satisfaction_mall<P>(
+        &self,
+        provider: &P,
+    ) -> Satisfaction<Placeholder<DefiniteDescriptorKey>>
+    where
+        P: AssetProvider<DefiniteDescriptorKey>,
+    {
+        self.plan_satisfaction(provider)
     }
 }
 
--- a/src/miniscript/satisfy.rs
+++ b/src/miniscript/satisfy.rs
@@ -6,15 +6,16 @@
 //! scriptpubkeys.
 //!
 
-use core::{cmp, i64, mem};
+use core::{cmp, fmt, i64, mem};
 
 use bitcoin::hashes::hash160;
 use bitcoin::key::XOnlyPublicKey;
-use bitcoin::taproot::{ControlBlock, LeafVersion, TapLeafHash};
-use bitcoin::{absolute, Sequence};
+use bitcoin::taproot::{ControlBlock, LeafVersion, TapLeafHash, TapNodeHash};
+use bitcoin::{absolute, ScriptBuf, Sequence};
 use sync::Arc;
 
 use super::context::SigType;
+use crate::plan::AssetProvider;
 use crate::prelude::*;
 use crate::util::witness_size;
 use crate::{AbsLockTime, Miniscript, MiniscriptKey, ScriptContext, Terminal, ToPublicKey};
@@ -105,11 +106,19 @@ pub trait Satisfier<Pk: MiniscriptKey + ToPublicKey> {
     }
 
     /// Assert whether an relative locktime is satisfied
+    ///
+    /// NOTE: If a descriptor mixes time-based and height-based timelocks, the implementation of
+    /// this method MUST only allow timelocks of either unit, but not both. Allowing both could cause
+    /// miniscript to construct an invalid witness.
     fn check_older(&self, _: Sequence) -> bool {
         false
     }
 
     /// Assert whether a absolute locktime is satisfied
+    ///
+    /// NOTE: If a descriptor mixes time-based and height-based timelocks, the implementation of
+    /// this method MUST only allow timelocks of either unit, but not both. Allowing both could cause
+    /// miniscript to construct an invalid witness.
     fn check_after(&self, _: absolute::LockTime) -> bool {
         false
     }
@@ -436,7 +445,6 @@ macro_rules! impl_tuple_satisfier {
                 )*
                 None
             }
-
             fn lookup_raw_pkh_x_only_pk(
                 &self,
                 key_hash: &hash160::Hash,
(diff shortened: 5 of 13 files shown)


apoelstra:
Heads up that in the future we may drop `PackedLockTime` https://github.com/rust-bitcoin/rust-bitcoin/pull/1330 ... I think you should create a different newtype which implements `Ord` and calls `unreachable!()` rather than doing the wrong thing in the case that the types don't match.

src/descriptor/mod.rs
```diff
@@ -771,6 +804,46 @@ impl Descriptor<DefiniteDescriptorKey> {
...
+                            xkey: xkey.xkey,
+                            wildcard: Wildcard::Unhardened,
+                            derivation_path: xkey
+                                .derivation_path
+                                .into_iter()
+                                .take(xkey.derivation_path.len() - 1)
+                                .cloned()
+                                .collect::<Vec<_>>()
+                                .into(),
+                        };
```
apoelstra, line 830: In dc0b2804a8368bea9f5e6e4717ad31bc8933e090:

This "remove the wildcard" business should be a method on `DescriptorPublicKey`, if this is the correct logic ... though TBH I'm not certain that it is. Where does `key_map` come from and why would it have wildcard-stripped xpubs in it?

src/miniscript/satisfy.rs
```diff
@@ -500,11 +504,115 @@ impl_tuple_satisfier!(A, B, C, D, E, F);
 impl_tuple_satisfier!(A, B, C, D, E, F, G);
 impl_tuple_satisfier!(A, B, C, D, E, F, G, H);
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Placeholder for some data in a [`WitnessTemplate`]
+pub enum Placeholder<Pk: MiniscriptKey> {
+    /// Public key and its size
+    Pubkey(Pk, usize),
+    /// Public key hash and its size
+    PubkeyHash(hash160::Hash, usize),
```
apoelstra, line 513: In dc0b2804a8368bea9f5e6e4717ad31bc8933e090:

This comment should be updated to clarify that it's the size of the hashed pubkey, not the size of the hash

src/descriptor/mod.rs
afilini: Sure, I can definitely move this in a method on `DescriptorPublicKey`. Tomorrow I'll look at the logic more closely and maybe add some tests for it, but the rationale is:

- `Assets` works with `DefiniteDescriptorKey`s, because sometimes we have to refer to them given the hash (so we can't try bruteforcing derivation indexes on `DescriptorPublicKey`).
- When we encounter an "XPub" variant of a `DefiniteDescriptorKey` it means we have the derived version of a `DescriptorPublicKey`, and there are two possible cases there:
    - If the `DescriptorPublicKey` had a wildcard, then in the definite key we will find a path that has the exact same prefix plus the last element which has whatever derivation index the key have been derived to. So this is why when the derivation index is not empty we try looking up a key that doesn't have the last index in the derivation path plus the wildcard
    - If the `DescriptorPublicKey` didn't have a wildcard at all, then deriving it doesn't change it in any way, which is why we also try looking up the key directly, in case it was unchanged.

I think this, at least on paper, makes sense but I'll double check this tomorrow.

The `key_map` comes from the descriptor parsing, when you parse a descriptor with secrets you get back the `Descriptor<DescriptorPublicKey>` plus the map public -> secret keys of all the secret keys found in the descriptor. Normally the user would then derive the descriptor using `at_derivation_index()`, and then call this method on the derived descriptor, giving the original key map as an argument so that we can figure out which keys you actually know the secret for.

Ideally we would just be working on `Descriptor<DescriptorPublicKey>`, to essentially get two benefits:
1. The `Assets` structure would not be tied to the derived descriptor, but it could be created once and reused for every derivation index
2. We would not have to do this weird "unwinding" of the wildcard

Unfortunately as I was saying above, sometimes we have to answer lookups that only give us the key hash, so we can't work with keys that have wildcards because we would have to bruteforce them. For example the `lookup_raw_pkh_pk(&self, hash: &hash160::Hash)` method would be very hard to implement if the assets only contained underived keys.

afilini:
I remember you suggested on IRC that we also add some kind of API to mark entire branches as "available", instead of marking the individual items. I think this would also be useful in another case, which isn't (and can't, I guess) be covered by this "Assets" approach: setting a "minimum" locktime, i.e. asking miniscript to find a way to spend that includes a timelock of **at least** a given value. This could be useful if you know you have a branch with a timelock and one without and you want to force the use of the timelock, even if it's not the cheaper way to spend.

My idea for this feature is to figure out a way to provide a custom closure to the `satisfy_helper` function: right now it takes a "minimum" fn which finds the cheapest solution of a `thresh`. My idea is that you could swap this with your own custom function that can analyze the sub-templates and pick the ones you want.

But considering this PR is already pretty large (and I still need a bit of time to experiment with that) I figured I could submit that with a follow-up PR.

src/descriptor/mod.rs
apoelstra: My feeling is that we should make `Assets` hold `DescriptorPublicKey`s and we should implement `lookup_raw_pkh_pk` on a best-effort basis. Meaning, we store hashes for `DescriptorPublicKey`s without wildcards and don't for ones with wildcards.

I'm comfortable doing this because since #431 we've treated raw pkhs as second-class-citizens, which can only occur when you are inferring descriptors directly from `Script`. If you have a stringified descriptor then you will always have actual public keys and not need to do lookups by hash. This is also the only form of pkh that Core supports, BTW.

src/plan.rs
```diff
@@ -0,0 +1,467 @@
...
+    absolute_timelock: Option<LockTime>,
+    relative_timelock: Option<Sequence>,
+}
+
+impl AssetProvider<DefiniteDescriptorKey> for Assets {
+    fn lookup_ecdsa_sig(&self, pk: &DefiniteDescriptorKey) -> bool {
+        let pkh = pk.to_pubkeyhash(SigType::Ecdsa);
+        let we_have_pk = self.keys.contains_key(&pkh);
+        let we_have_sig = self.ecdsa_signatures.contains_key(&pkh);
+        we_have_pk || we_have_sig
```
apoelstra, line 280: Should all of these `||`s be `&&`s? I don't understand the intended logic here -- it seems like "do we have a pk" and "do we have a sig" are two independent questions.

afilini: Yeah it's an `OR` because in both cases we can satisfy that node: either we have the key (so we can produce the signature ourselves) or we have a pre-made signature that we found for example in a psbt.

apoelstra: Ah, gotcha. Maybe worth adding a comment on one of these to explain the context.

src/descriptor/key.rs
```diff
@@ -485,6 +485,30 @@ impl DescriptorPublicKey {
         DefiniteDescriptorKey::new(definite)
             .expect("The key should not contain any wildcards at this point")
     }
+
+    /// Whether this key matches a [`DefiniteDescriptorKey`]
```
apoelstra: In e16e7dd9:

This `matches` function should be private and have a short wrapper
```
pub fn matches<D: AsRef<DescriptorPublicKey>>(&self, key: D) {
    self.real_matches(key.as_ref())
}
```
And then in your `matches` method (now `real_matches`) take a `&DescriptorPublicKey` directly rather than writing `&definite_key.0`.

apoelstra: This depends on https://github.com/rust-bitcoin/rust-miniscript/pull/492 BTW

apoelstra: I also think that ths `matches` name should be different because I continue to think that this "truncate the path and check for a match" logic is confusing. Maybe it should be called `matches_prefix` or something?

apoelstra: I'd be willing to move this to a followup PR because I think this is dominating discussion here, and this PR is actually much much larger than this method :). I'm also musing about whether `DefiniteDescriptorKey` should have say, an `Option<usize>` which indicates which index it came from (if any) to assist with what you're trying to do.

apoelstra (dismissed):
ACK 632a365cc36bc55cfa142c372023eeb55533bb4d

sanket1729:
I am also interested in reviewing this. Give me a couple more days.

src/descriptor/key.rs
apoelstra: The new name is better but I'd still like to revisit this :) BTW I think the code should also have a check that the final (truncated) step in the derivation path is actually a wildcard.

apoelstra (dismissed):
ACK c3ba4d318f0c8e7eb0cf0a405faeca768a157bed

LLFourn (commented):
This is a really nice design. It fits in great with the existing satisifer API well (with one caveat about the merkle root for tap key spends).

I think the biggest improvement I'd like to see to the general idea is to make the `Plan` self-contained. I think the plan should contain everything you need to analyze and execute spending from that input i.e. you don't need to keep around the definite descriptor once you've settled on a plan. I think the only thing that is missing in terms of data is the merkle root for tap key spends. In terms of functionality the plan would need to be able to populate an `psbt::Input` with all the fields necessary to complete the witness for the plan. That way you can use the plan instead of just blindly setting all the fields with `update_input_with_descriptor` for all possible spending paths. I think this could come in a follow up PR.

I was wondering if it were possible to get rid of the two types of witness templates and unify them. Rather than trying to describe the approach I just attempted to implement it myself:

https://github.com/LLFourn/rust-miniscript/commit/7e0df6aded2c9e25ea830960fbf58cdae0995a9e

The basic strategy is to have a `Option` in each placeholder variant that can be satisfied where you can put the satisfaction data (i.e. signature). Once the signature (or whatever) is there then it will no longer show up when you ask the template what is missing. Once all these `Option`s have been set to `Some(_)` you can finally get the witness. This allows you avoid the `PartialSatisfaction` type. I think it might also allow you to get rid of the `WitnessTemplate` type entirely in the public API. It probably needs some more work to fully explore this API change. Let me know what you think and whether I should open up a PR.

src/plan.rs
```diff
@@ -0,0 +1,504 @@
...
+    }
+}
+
+/// The Assets we can use to satisfy a particular spending path
+#[derive(Debug, Default)]
+pub struct Assets {
+    keys: HashMap<hash160::Hash, DescriptorPublicKey>,
+    tap_key_spend_sig: Option<bitcoin::SchnorrSig>,
+    ecdsa_signatures: HashMap<DescriptorPublicKey, (bitcoin::EcdsaSig, usize)>,
+    schnorr_signatures: HashMap<(DescriptorPublicKey, TapLeafHash), (bitcoin::SchnorrSig, usize)>,
```
sanket1729, line 303: Why do we need the signatures here? Will post a detailed reply in main conversation.

sanket1729 (commented):
First off, thanks a lot. Asset Planning is one of the things that we are lacking. It really sucked to use the max_satisfaction APIs for change estimation. 

I did a review of the first two commits. Have some questions before I proceed to more review.

```rust
pub struct Assets {
    raw_keys: HashSet<DescriptorPublicKey>, // The keys that we cannot sign with, but used for pkh -> pk mapping
    avail_keys: HashSet<DescriptorPublicKey>, // The keys that we can sign with
    sighash_type: Option<PsbtSigHashType>, // The sighash e that we want to use while signing
    .. // All other fields for hashes stay the same.
    }
    ```
    We can also skip the raw_keys part as that is still in discussion. We can then the methods in AssetProvider to better match the API. I believe the above matches the most realistic user scenario: "Given keys, hash locks, and timelocks, construct a template before actually signing". We can then apply satisfiers on the placeholder witness to get the complete witness. 
    
    Just to make sure we are on the same page, I think the workflow using planner might look like:
    
    1) Construct Assets and obtain a plan. Retry with other `Assets` if Plan creation fails.  We could have multiple assets `hot_assets`, `countersparty_assets`, or `cold_assets`. We can change our plan based on the availability of different assets/counter parties. 
    2) Obtain the satisfaction weight for the given plan and compute the change output using the fees.
    3) Create an unsigned psbt with required nSequence/nLockTime.
    4) Update using `update_utxo_with_descriptor`. 
    5) Using the witness template obtained from the plan to get the `Template::required_signatures`. 
    6) Sign with the required keys and add the required preimages. If we face any error here, perhaps the counterparty did not provide signatures. go back to step 1. 
    7) Finalize psbt.

Here, we do not require signatures in assets, and is probably was @LLFourn is suggesting.

File: src/descriptor/segwitv0.rs [hunk 6]
```diff
@@ -393,6 +422,37 @@ impl<Pk: MiniscriptKey + ToPublicKey> Wpkh<Pk> {
     }
 }
 
+impl Wpkh<DefiniteDescriptorKey> {
+    /// Returns a plan if the provided assets are sufficient to produce a non-malleable satisfaction
+    pub fn get_plan<P>(&self, provider: &P) -> Option<Plan>
```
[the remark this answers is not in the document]
sanket1729: Agreed. I think this naming is an improvement. This PR is heavy in new data structures/changes. Can also defer nits to the future, as @afilini thinks.

src/plan.rs
afilini: The idea is that in multi-party protocols you can use signatures made by others as "assets" themselves. In that case you can't "cheat" by saying you have the key, because if you only have one signature you can either take it or it's not useful to you at all. When you say you have a key the planning code assumes you can sign anything you want with that key, which is a bit different.

Thinking about it now, I was planning to implement `IntoAssets` for a `psbt::Input` but I think I ended up forgetting about it, so I'll try to add that as well.

src/miniscript/satisfy.rs
```diff
@@ -706,6 +706,12 @@ pub struct Satisfaction {
     /// Whether or not this (dis)satisfaction has a signature somewhere
     /// in it
     pub has_sig: bool,
+    // We use PackedLockTime here as we need to compare timelocks using Ord. This is safe,
+    // as miniscript checks for us beforehand that the timelocks are of the same type.
```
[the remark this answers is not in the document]
afilini: Thanks for pointing this out, I wasn't aware of this!

src/util.rs
```diff
@@ -3,15 +3,55 @@ use bitcoin::hashes::Hash;
...
+    fn size(&self) -> usize;
+}
+
+impl<Pk: MiniscriptKey> ItemSize for Placeholder<Pk> {
+    fn size(&self) -> usize {
+        match self {
+            Placeholder::PubkeyHash(_, size) => *size,
+            Placeholder::Pubkey(_, size) => *size,
+            Placeholder::EcdsaSigPk(_) | Placeholder::EcdsaSigHash(_) => 73,
+            Placeholder::SchnorrSig(_, _) | Placeholder::SchnorrSigHash(_, _) => 66,
```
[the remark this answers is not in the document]
afilini: Yes, I think it can be done. We'll then store the sighash in the `RequiredSig` enum so that you know how to produce the signature so that it matches the expected size.

Just to clarify: we use `66` here because it counts the `OP_PUSH` byte as well, but with sighash_default it can drop to `65`.

afilini:
Thanks everyone for the review! I think I agree with most of what you pointed out, so I'll start working in that direction and see what I can come up with.

I think one of the limitations of the current `AssetProvider` trait is that it must provide essentially a "subset" of the information that the `Satisfier` does. Because we want to provide a default `AssetProvider` implementation for every type that also implements `Satisfier`.

So this complicates things like [adding the merkle root in the plan](https://github.com/rust-bitcoin/rust-miniscript/pull/481#discussion_r1027034721): as Lloyd pointed out it would require changing the `Satisfier` trait itself and I was trying to avoid that. But if @sanket1729 and @apoelstra are ok with it, I can definitely do it.

------

@sanket1729 re: your point number 4 in the flow

> Update using update_utxo_with_descriptor. We can change this function to also take in a plan argument. Or embed the descriptor in the plan. Not really sure.

Talking with @LLFourn we were thinking about changing the plan API so that the plan itself could be used to update the PSBT. This would allow you to:

1. Avoid keeping the descriptor around even after you've created a plan
2. Only add to the PSBT the relevant metadata (which is probably what you had in mind when you suggested to take the plan as an argument)

I really like this idea but I haven't tried implementing it yet, so there could be some things missing in the plan that won't allow me to (fully) update the PSBT. Worst case as you suggested we can embed the full descriptor in the plan, although it seems a bit "hacky" to me.

src/miniscript/satisfy.rs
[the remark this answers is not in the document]
apoelstra: Oof, I read this comment and totally missed that it wasn't true!

apoelstra:
I am definitely ok with changing the `Satisfier` trait to make this PR better.

src/plan.rs
sanket1729: I thought the purpose of planning was to decide which spend to use _before_ signing. In which case, why do we need signatures? I think having signatures in assets would be confusing, curious what @apoelstra thinks about this.

> In that case you can't "cheat" by saying you have the key

You would still need to assume that the other party will sign with a key because you calculated the change output based on it. So, they can still refuse to sign and cheat. 

Perhaps, we have some misunderstanding about what `Plan` is supposed to do.

src/miniscript/satisfy.rs
```diff
@@ -526,11 +530,121 @@ impl_tuple_satisfier!(A, B, C, D, E, F);
...
+    /// Public key and its size
+    Pubkey(Pk, usize),
+    /// Public key hash and the size of the key
+    PubkeyHash(hash160::Hash, usize),
+    /// ECDSA signature given the raw pubkey
+    EcdsaSigPk(Pk),
+    /// ECDSA signature given the pubkey hash
+    EcdsaSigHash(hash160::Hash),
+    /// Schnorr signature
+    SchnorrSig(Pk, Option<TapLeafHash>),
```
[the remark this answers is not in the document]
LLFourn: > The other alternative would be to store the context alongside the vec of Placeholder, in the WitnessTemplate, and then figure out a way to easily access it when you are iterating the list of required signatures.

Yeah I was thinking something more like this. A general context object which has what kind of signatures you are producing (ecdsa, tap key, tap script) and any dependencies needed to sign (merkle root, tap leaf hash). That would clean things up a bit i think.

src/util.rs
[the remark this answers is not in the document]
LLFourn: Hmm why include the varint for the size here (I assume that's what you mean by `OP_PUSH`)? the element size itself seems more useful because you can easily go from `elem_size -> varint_len + elem_size` but the other way around is harder (impossible?). Also it doesn't look like for `TapControlBlock` and `TapScript` you are including the varint_len.

src/plan.rs
LLFourn: I think @afilini is thinking of the scenario of where you get a request to sign a TX  input and a few signatures under other keys. The TX is already finalized at this point. To figure out what you're meant to be contributing for each input you construct a plan from the descriptor + existing signatures and see what's left to do. Of course, assuming you're using PBSTs you could just blindly follow what's missing from the PSBTs but this allows some analysis before doing so e.g. construct a plan with only the signatures + what you have locally and see if it's possible for you alone to finish the PSBT before actually engaging signing devices.

I think it would be a good idea to get rid of the signatures themselves from this struct though. If they've already signed then those keys can be treated as an asset. Just map those signatures to their asset keys but there's no need to actually represent them like that in the "assets" I think.

LLFourn: Oh right after writing this I realized that it's possible the signatures are only valid for one tapleaf but not the other. What if we generalized this to conditioning public keys that can only sign on particular tap leaves? E.g. maybe I have a hardware device that will only sign on this tapleaf one normally but might sign on this tap leaf (or do tap key) if some condition is met. Then existing signatures would map to keys that can only sign on a particular leaf (or as tap key).

LLFourn:
> @LLFourn , sorry I also don't follow why having a `merkle_root` is useful in `AssetProvider`. It is already a deterministic function of the descriptor. If we carry the descriptor in plan, we don't need any additional information. `Plan`
> 
> What I think of `AssetProvider` is providing information on what things could be available. These things should in principle only be information about keys, hash locks, and timelocks without any signatures/pre-images.

Agreed 100%. I was not suggesting the merkle root be in the `AssetProvider`. Including the desctiptor in the plan would be sufficient but I agree with @afilini that iit seems hacky and error prone.

> 
> I think what @afilini and you are mentioning is more along the lines of `SigningProvider`, information that you might need for signing. I think we can already use Psbts for it.
> 
> I think before @afilini puts more time into this PR, we should settle on what `Plan` should do.
> 
> > In terms of functionality the plan would need to be able to populate an psbt::Input with all the fields necessary to complete the witness for the plan.
> 
> Why put witnesses/Merkle root in `Plan` instead of adding them to psbt directly? I feel we are duplicating a lot of functionality that is already well-defined in psbts. The only psbts don't tell you is how to construct the initial psbt and which spend paths to use. IMO, this should be the only thing `Plan` does.

Whose going to add it to the PSBT though? Your proposed structure seems awkward to me. If the plan is going to tell you that you need to put the merkle root in the PSBT but not actually put it there you're going to have to re-fetch it yourself. But the plan has already traversed the descriptor and could easily have just saved the merkle root or tapleaf etc on its way down (it already saves the tapleaf but is missing merkle root).

> * Construct Assets and obtain a plan. Retry with other `Assets` if Plan creation fails.  We could have multiple assets `hot_assets`, `countersparty_assets`, or `cold_assets`. We can change our plan based on the availability of different assets/counter parties.
> 
> * Obtain the satisfaction weight for the given plan and compute the change output using the fees.
> 
> * Create an unsigned psbt with required nSequence/nLockTime.
> 
> * Update using `update_utxo_with_descriptor`. We can change this function to also take in a plan argument. Or embed the descriptor in the plan. Not really sure.
> 
> * Using the witness template obtained from the plan to get the `Template::required_signatures`.

If we are updating the PSBT with the plan then isn't the required signatures implied by whatever fields that were set in the PSBT?

> 
> * Sign with the required keys and add the required preimages. If we face any error here, perhaps the counterparty did not provide signatures. go back to step 1.
> 
> * Finalize psbt.

Overall I think we are all on the same page here with some caveats about specifics in the API (e.g. I think there should just be `Plan::update_psbt(input: &mut psbt::Input)` without needing the original descriptor). In terms of future vision I think we probably differ in that I would like the above workflow to be possible without using PSBTs one day but I don't think that's super necessary right now. 

Actually, opposite to my initial intuitions I'd say the way forward for this might be to go "all in" on PSBTs and reduce new API surface area we're adding here so:

1. Don't have `required_signatures` as they are implied from the fields that were populated in the PSBT by the plan.
2. Make `Placeholder` type private and all the fields of `Plan` private with read only accessors.
3. Merge witness template and plan so there's not need for `WitnessTemplate`:

```rust
#[derive(Debug, Clone)]
pub struct Plan {
    /// This plan's witness template
    pub template: Vec<Placeholder<DefiniteDescriptorKey>>,
    /// The absolute timelock this plan uses
    pub absolute_timelock: Option<LockTime>,
    /// The relative timelock this plan uses
    pub relative_timelock: Option<Sequence>,
    pub(crate) desc_type: DescriptorType,
}
``` 
4. Remove "interactive" satisfaction stuff. If we want the "interactive satisfaction" like API then I suggest going with the approach here: https://github.com/LLFourn/rust-miniscript/commit/7e0df6aded2c9e25ea830960fbf58cdae0995a9e but it can be added later.
5. Add `Plan::update_psbt(input: &mut psbt::Input)`.

Sorry for the "backflip" @afilini. I am somewhat torn by this recommendation since iterating over the placeholders and interactively satisfying them is one of the coolest things you can do with the plan but I think if we add PSBT input population logic we won't need it for the workflow @sanket1729 mentions above. I would love for it to be made public later once it's a little more refined and to eventually decouple it from PSBTs. Do you think this approach is viable @afilini?

src/plan.rs
apoelstra: >What if we generalized this to conditioning public keys that can only sign on particular tap leaves?

Yes, I agree that we should do this. Otherwise I don't really see the point of supporting signatures in lieu of public keys.

afilini:
Sorry for the huge delay, I've been travelling and I wanted to settle down a bit before getting back to this.

I think I agree that if we have to make a choice between having everything inside the plan or everything inside the PSBT, then the PSBT is the most reasonable choice.

I'll try to implement what you suggested @LLFourn, the only thing I'm not sure about is removing the `WitnessTemplate`, because I seem to remember it was a useful abstraction in a few places. But I'll try and see what happens.

> I would love for it to be made public later once it's a little more refined and to eventually decouple it from PSBTs. Do you think this approach is viable @afilini?

Yes, I think it should be possible. I think it's better to initially focus on just satisfying everything in one go from a PSBT, which is probably what most people are going to use anyway. The current interactive satisfaction part was already a bit awkard, so I'm happy to remove it and work on it separately with more thought.

LLFourn (commented):
Thanks @afilini. I did a more in-depth review. I think the PSBT input updating is a great addition.

src/miniscript/satisfy.rs
```diff
@@ -526,11 +527,153 @@ impl_tuple_satisfier!(A, B, C, D, E, F);
...
+    /// Script spend signature
+    ScriptSpend {
+        /// Leaf hash of the script
+        leaf_hash: TapLeafHash,
+    },
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Placeholder for some data in a [`Plan`]
+pub enum Placeholder<Pk: MiniscriptKey> {
```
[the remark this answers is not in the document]
sanket1729: Fortunately, `Plan` has a non-public field that prevents its misuse. It can only be created from valid descriptors.

src/plan.rs
```diff
@@ -294,13 +303,88 @@ impl Plan {
...
+            taproot: TaprootCanSign::default(),
+        }
+    }
+}
+
+#[derive(Debug)]
+/// Signatures which a taproot key can produce
+///
+/// Defaults to `key_spend=true`, `script_spend=Any` and `sighash_default=true`
+pub struct TaprootCanSign {
```
sanket1729, line 475: Same comment about public fields.

src/plan.rs
```diff
@@ -0,0 +1,1251 @@
...
+#[derive(Debug, Clone)]
+pub struct Plan<'d> {
+    /// This plan's witness template
+    pub template: Vec<Placeholder<DefiniteDescriptorKey>>,
+    /// The absolute timelock this plan uses
+    pub absolute_timelock: Option<LockTime>,
+    /// The relative timelock this plan uses
+    pub relative_timelock: Option<Sequence>,
+
+    pub(crate) descriptor: &'d Descriptor<DefiniteDescriptorKey>,
```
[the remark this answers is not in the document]
afilini: It seemed wasteful to clone but I think you are right, I didn't consider what would happen in multi-threaded scenarios.

Maybe I can still get away with just copying what we need, I initially started implementing that but then I had to keep updating the struct because as I worked I realized I needed more and more stuff. So I decided to just include the descriptor initially, now we can asses what we actually need (assuming we don't want to make large architectural changes) and potentially remove it.

src/plan.rs
```diff
@@ -0,0 +1,1251 @@
...
+            #[derive(Default)]
+            struct TrDescriptorData {
+                tap_script: Option<Script>,
+                control_block: Option<ControlBlock>,
+                internal_key: Option<XOnlyPublicKey>,
+                key_origins: BTreeMap<XOnlyPublicKey, (Vec<TapLeafHash>, bip32::KeySource)>,
+            }
+
+            let spend_info = tr.spend_info();
+            input.tap_merkle_root = spend_info.merkle_root();
```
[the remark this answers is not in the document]
afilini: I didn't know about this, I'll go re-read the spec to figure out what's the correct behavior

src/plan.rs
```diff
@@ -0,0 +1,1251 @@
...
+            Single(v) => v == lh,
+            Many(set) => set.contains(lh),
+        }
+    }
+}
+
+/// The Assets we can use to satisfy a particular spending path
+#[derive(Debug, Default)]
+pub struct Assets {
+    keys: HashMap<hash160::Hash, (DescriptorPublicKey, CanSign)>,
```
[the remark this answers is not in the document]
sanket1729: > Don't I usually want to provide an xpub + fingerprint + derivation path (referred to as a KeySource in rust-bitcoin) and say I am willing to sign any path under this key?

Are you saying this should be `DefiniteDescriptorKey`?

src/plan.rs
```diff
@@ -0,0 +1,1251 @@
...
+
+#[derive(Debug)]
+/// Signatures which a key can produce
+///
+/// Defaults to `ecdsa=true` and `taproot=TaprootCanSign::default()`
+pub struct CanSign {
+    /// Whether the key can produce ECDSA signatures
+    ecdsa: bool,
+    /// Whether the key can produce taproot (Schnorr) signatures
+    taproot: TaprootCanSign,
```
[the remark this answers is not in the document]
afilini: No you are right, I just forgot to make them public

src/descriptor/bare.rs
```diff
@@ -129,23 +129,25 @@ impl<Pk: MiniscriptKey + ToPublicKey> Bare<Pk> {
 
 impl Bare<DefiniteDescriptorKey> {
     /// Returns a plan if the provided assets are sufficient to produce a non-malleable satisfaction
-    pub fn get_plan<P>(&self, provider: &P) -> Option<Plan>
+    pub fn get_plan_satisfaction<P>(
```
sanket1729, line 132: Perhaps, this should be renamed as it no longer returns a plan? Any special reason why this does not return a plan?

src/plan.rs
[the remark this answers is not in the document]
sanket1729: I think signers might use it for control block verification? But I am also unsure about the spec here

sanket1729 (commented):
Did another pass-through. Overall, I really like the new design of `Plan`. Just left a few questions/comments. 

I think we also need to remove the `RequiredSig` and `RequiredPreImage` structs.

sanket1729:
Judging by responses from @LLFourn and code here by @afilini, I think we are in overall agreement about the workflow modulo API design decision about embedding descriptor. 

I will spend more time on detailed code review in this week.

src/plan.rs
```diff
@@ -0,0 +1,1251 @@
...
+impl FromIterator<DescriptorPublicKey> for Assets {
+    fn from_iter<I: IntoIterator<Item = DescriptorPublicKey>>(iter: I) -> Self {
+        Assets {
+            keys: iter
+                .into_iter()
+                .map(|pk| {
+                    (
+                        pk.clone()
+                            .at_derivation_index(0)
+                            .to_pubkeyhash(SigType::Ecdsa),
```
LLFourn, line 666: Ahh I think I am confused by public key hashes and miniscript again.

Why is this indexing by the pubkeyhash of the thing at derivation_index 0? I mean if the thing you need to sign is the pubkey hash at derivation index 7 how are you going to be able to look it up via this index.

Can we make pubkeyhashes a separate concern in a different field where you say what pubkeyhashes you can sign for and nothing else. i.e. just a `HashSet<hash160::Hash>`.

src/plan.rs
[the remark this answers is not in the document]
LLFourn: No I was saying that I think this field should be a fingerprint + derivation path + xpub
But I'm now confused about public key hashes looking closer (see my comment below). I don't understand why we are indexing the keys we can purportedly sign with by public key hash.

Thinking further why do I even need to provide the xpub -- a fingerprint + derivation path should be sufficient to say "I can sign anything under this key"? Anyway internal fields aside my expectation as a user is that I can communicate this somehow to the planning module. i.e. here's my hardware wallet fingerprint and derivation path that I can sign with.

src/plan.rs
```diff
@@ -0,0 +1,1251 @@
...
+                                        },
+                                        (pk.master_fingerprint(), pk.full_derivation_path()),
+                                    )
+                                });
+                        }
+                        _ => {}
+                    }
+
+                    data
+                });
```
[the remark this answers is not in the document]
afilini: I can't figure out a way to enforce with the type system that all the placeholders refer to the same leaf hash. I think it's guaranteed by the current implementation, but not enforced by the types.

I will update this code that updates the psbt so that it assumes all the placeholders are aligned and panics otherwise.

src/plan.rs
```diff
@@ -0,0 +1,1251 @@
...
+                    .tap_scripts
+                    .insert(control_block, (tap_script, LeafVersion::TapScript));
+            }
+
+            // Ensure there are no duplicated leaf hashes. This can happen if some of them were
+            // already present in the map when this function is called, since this only appends new
+            // data to the psbt without checking what's already present.
+            for (tapleaf_hashes, _) in input.tap_key_origins.values_mut() {
+                tapleaf_hashes.sort();
+                tapleaf_hashes.dedup();
```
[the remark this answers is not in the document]
afilini: Yeah the problem is checking if it exists, it means iterating the whole vec. In theory I think sort/dedup should be more efficient, but considering the size of the vectors we are working with it probably doesn't make a difference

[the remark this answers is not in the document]
afilini: Addressed in 503b3f8

src/plan.rs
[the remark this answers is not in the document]
afilini: Addressed in 503b3f8

src/plan.rs
afilini: If we ignore pubkey hashes entirely then the satisfaction code will not support satisfying parsed scripts anymore, will that be ok?

apoelstra: You mean, it won't support satisfying parsed scripts that have pubkey hashes in them? I think that's OK, at least for the first iteration of this API.

If you mean that we can't satisfy parsed scripts *at all* we should fix this now.

sanket1729: > If we ignore pubkey hashes entirely then the satisfaction code will not support satisfying parsed scripts anymore, will that be ok?

Parsed scripts with raw hashes are not miniscripts or miniscript descriptors. We should not be creating plans for incomplete/partial descriptors. If we have access to full descriptors, like `pkh(KEY)`, this will work as expected. 

In the future, this will be supported cleanly when we have support for partial descriptors.

afilini:
It took me a while to figure this out, but I believe the some of the tests are failing because they use `finalize_mall_mut` implemented on the psbt input directly. So when miniscript tries to finalize it, it doesn't have the full descriptor (which contained a `pkh()` fragment) and instead resorts to parsing the raw script sig, which is translated into a "expr_raw_pkh" internally. Which can't be satisfied anymore since the planning code doesn't support raw pkh.

Should I change the API of the PSBT finalizer so that it takes a reference to the descriptor? Otherwise we'll have to revert my latest changes and re-introduce raw pkh in the planner.

sanket1729:
I will take a look at it this week

sanket1729:
> Should I change the API of the PSBT finalizer so that it takes a reference to the descriptor? Otherwise we'll have to revert my latest changes and re-introduce raw pkh in the planner.

No. This is something we definitely don't want to do or avoid as much as possible.

sanket1729:
@afilini, I see the issue. The `raw_pkh` should not be removed from `Satisfier` API, but only from the `AssetProvider` API. All the plan functions have access to descriptor, so it's okay. 

But definitely don't want the finalizer dependent on having access to the descriptor. In most real-life scenarios, it probably will have access to it, but we should keep it this way and use the Script -> (partial) Minsicript translation feature here.

sanket1729:
@afilini just read your blog here https://afilini.com/blog/taking-a-break-from-bdk/. Sad to see you take a very well-deserved break :(. If you are busy, I don't mind taking over this PR to push things forward.  

Hope you will be back, rust-miniscript will be missing your contributions.

src/plan.rs
```diff
@@ -0,0 +1,1251 @@
...
+
+impl IntoAssets for hash160::Hash {
+    fn into_assets(self) -> Assets {
+        Assets {
+            hash160_preimages: vec![self].into_iter().collect(),
+            ..Default::default()
+        }
+    }
+}
+
```
danielabrozzoni, line 628: If I remember correctly we also wanted PSBTs to be `IntoAssets`

danielabrozzoni:
Hey everyone, I just pushed:
- Rebase on master
- Plan owns the descriptor instead of borrowing it

Tests still fail because of the problem @afilini highlighted:
> It took me a while to figure this out, but I believe the some of the tests are failing because they use finalize_mall_mut implemented on the psbt input directly. So when miniscript tries to finalize it, it doesn't have the full descriptor (which contained a pkh() fragment) and instead resorts to parsing the raw script sig, which is translated into a "expr_raw_pkh" internally. Which can't be satisfied anymore since the planning code doesn't support raw pkh.

However, I'm not sure if the solution that @sanket1729 proposed is doable:
> @afilini, I see the issue. The raw_pkh should not be removed from Satisfier API, but only from the AssetProvider API. All the plan functions have access to descriptor, so it's okay.

If I understand correctly, at the moment when we want to satisfy a descriptor we first build a template with the AssetProvider and then try completing it with the Satisfier:

https://github.com/rust-bitcoin/rust-miniscript/blob/7b1e66358b94f80025cb67fd9d678fafde106354/src/miniscript/satisfy.rs#L1598-L1614

I don't really see how we could fix this problem, apart from completely decoupling the Satsfier from AssetProvider (which would mean duplicating a lot of code). Do you have any suggestions?

sanket1729:
Hi @danielabrozzoni, promise this would be the next big thing we would want to get in. 

Will get back to this after #493

sanket1729:
Hi @danielabrozzoni, promise this would be the next big thing we would want to get in. 

Will get back to this after #493

sanket1729:
Hello @danielabrozzoni, what you say is indeed correct. After some thinking, I think the correct thing to do here is to update the psbt inference code such that correctly infers `pkh(full-key)` instead of `pkh(hash160)`. If psbt already contains the information to infer the full key from pkh in script, the internal psbt descriptor inference should work for it. I can work on this one in a separate PR so that this gets in.

This should make sure that the currently existing tests pass as `finalize_mut()` internally without introducing psbt to the Assets API.

Also, this would require a rebase after updating to 0.30 :( .

sanket1729:
After #557 is in, I think the test cases that you were mentioning in this PR should successfully pass.

danielabrozzoni:
I just pushed c07772a652e30fd2130e5772db81b8cdb535d540:
- Rebased on top of master
- Commits cleanup

Some tests are still failing, I'm looking into updating the code to use the method introduced by #557 :)

sanket1729:
Thanks @danielabrozzoni, I think that error probably related to bitcoind. I will spend some time investigating it today.

sanket1729:
Any particular reason why you swapped out the bitcoind binary here? I created a fix for the test  at https://github.com/rust-bitcoin/rust-miniscript/commit/a62a5c44c2cabe616986fd7a1cb5182c401c01a5 . Feel free to cherry-pick.

danielabrozzoni:
> Any particular reason why you swapped out the bitcoind binary here?

Ops, I had to modify `bitcoind` to work on my system (NixOS 😅), and I accidentally committed the modified binary.

I cherry-picked your commit and tests pass now 🎉  I still need to double-check that all the above comments have been resolved, but this should be ready for another round of review soon :)

apoelstra:
@danielabrozzoni are you aware of the new `BITCOIND_EXE` environment variable? You can set this to a bitcoind outside of the crate itself and then tests will work on NixOS.

This isn't widely advertised. It was added in https://github.com/rust-bitcoin/rust-miniscript/pull/538. Maybe we should put it in the README since NixOS seems to be *extremely* overrepresented in the rust-bitcoin word..

src/miniscript/satisfy.rs
```diff
@@ -535,11 +395,117 @@ impl_tuple_satisfier!(A, B, C, D, E, F);
...
+            ),
+        }
+    }
+}
+
+impl<Pk: MiniscriptKey + ToPublicKey> Placeholder<Pk> {
+    /// Replaces the placeholders with the information given by the satisfier
+    pub fn satisfy_self<Sat: Satisfier<Pk>>(&self, sat: &Sat) -> Option<Vec<u8>> {
+        match self {
+            Placeholder::Pubkey(pk, _) => Some(pk.to_public_key().to_bytes()),
```
Harshil-Jani, line 476: This would work in case of ECDSA Sig. But with the Schnorr sig, It is not fulfilling. 
`Placeholder::Pubkey(pk, _)`. Here the second argument is length so we can have a length instead of `_` and check if it was 32 or something else. If 32 then do `Some(pk.to_x_only_pubkey().serialize().to_vec())`

src/descriptor/mod.rs
```diff
@@ -511,6 +512,60 @@ impl<Pk: MiniscriptKey + ToPublicKey> Descriptor<Pk> {
     }
 }
 
+impl Descriptor<DefiniteDescriptorKey> {
+    /// Returns a plan if the provided assets are sufficient to produce a non-malleable satisfaction
+    pub fn get_plan<P>(self, provider: &P) -> Option<Plan>
```
tcharding, line 517: Does it matter that this consumes self and if `None` is returned we have no way to get the descriptor back?

tcharding:
What needs doing to push this forward? I'm not able to provide much value just reading the diff but I've got clock cycles available if there is anything else I can do?

src/descriptor/mod.rs
LLFourn: +1 on not consuming self here.

src/plan.rs
```diff
@@ -0,0 +1,1161 @@
+// Miniscript
+// Written in 2022 by rust-miniscript developers
+//
+// To the extent possible under law, the author(s) have dedicated all
+// copyright and related and neighboring rights to this software to
+// the public domain worldwide. This software is distributed without
+// any warranty.
+//
```
tcharding, line 8: We have SPDX identifiers now, this can all be removed and just plain old:
```
// SPDX-License-Identifier: CC0-1.0
```

src/plan.rs
```diff
@@ -0,0 +1,1161 @@
...
+// You should have received a copy of the CC0 Public Domain Dedication
+// along with this software.
+// If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
```
tcharding, line 11: We have SPDX identifiers now, this can all be removed and just plain old:
```
// SPDX-License-Identifier: CC0-1.0
```

src/descriptor/mod.rs
sanket1729: @danielabrozzoni, I think they mean `pub fn get_plan(self, provider: &P) -> Result<Plan, Descriptor>`. That is we don't lose the descriptor in case of error .

src/miniscript/satisfy.rs
danielabrozzoni: Thanks! I think it should be: if the size is 33, then I have to use to_x_only_pubkey, otherwise I can use to_public_key. This is because we push Ctx::pk_len, which is 33 for Tap. Does this make sense?
```rust
            Placeholder::Pubkey(pk, size) => {
                if *size == 33 {
                    Some(pk.to_x_only_pubkey().serialize().to_vec())
                } else {
                    Some(pk.to_public_key().to_bytes())
                }
            }
```

sanket1729:
According to my review. Only the pubkey size check is the only relevant comment unaddressed. Everything else looks good to me

danielabrozzoni:
I just pushed the fix for the pubkey length, there are still a couple of comments I need to address:
- https://github.com/rust-bitcoin/rust-miniscript/pull/481#discussion_r1046608000
- https://github.com/rust-bitcoin/rust-miniscript/pull/481#discussion_r1046625975
- https://github.com/rust-bitcoin/rust-miniscript/pull/481#discussion_r1046634178
- https://github.com/rust-bitcoin/rust-miniscript/pull/481#discussion_r1046614740
- https://github.com/rust-bitcoin/rust-miniscript/pull/481#discussion_r1047227630

Not being the PR author I can't resolve conversations, and given the length of the discussion, I might be missing something important

sanket1729:
@danielabrozzoni, would it help to split this PR and make progress on another fresh if this is annoying to load in Github? I would prefer listing out all the unresolved issues and addressing all of them in a separate PR. Github is unfriendly once it gets into hundreds of comments

src/miniscript/satisfy.rs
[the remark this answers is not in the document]
tcharding: Looks like this is resolved but the comment is still in 4849893 albeit with the `PackedLockTime` changed to `AbsLockTime` - better to remove the comment,right? @sanket1729 can you confirm please.

src/plan.rs
```diff
@@ -0,0 +1,504 @@
...
+use crate::{
+    DefiniteDescriptorKey, DescriptorPublicKey, MiniscriptKey, ScriptContext, ToPublicKey,
+};
+
+/// Trait describing a present/missing lookup table for constructing witness templates
+///
+/// This trait mirrors the [`Satisfier`] trait with the difference that instad of returning the
+/// item if it's present, it only returns a boolean to indicate its presence.
+///
+/// This trait is automatically implemented for every type that is also a satisfier, and simply
```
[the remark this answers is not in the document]
tcharding: This is resolved as of 4849893.

src/plan.rs
tcharding: As of 4849893 this is resolved.

src/descriptor/key.rs
```diff
@@ -485,6 +485,41 @@ impl DescriptorPublicKey {
         DefiniteDescriptorKey::new(definite)
             .expect("The key should not contain any wildcards at this point")
     }
+
+    /// Whether this key is the "parent" of a [`DefiniteDescriptorKey`]
+    ///
+    /// The key is considered "parent" if it represents the non-derived version of a definite key,
```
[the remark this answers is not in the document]
tcharding: As of 4849893 this is still TODO.

src/plan.rs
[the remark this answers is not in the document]
tcharding: As of 4849893 this is still TODO.

src/plan.rs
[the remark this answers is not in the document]
tcharding: As of 4849893 this is still TODO.

src/miniscript/satisfy.rs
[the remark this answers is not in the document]
tcharding: As of 4849893 this is public still, looks like we are happy with that. Flagging it just in case.

src/plan.rs
tcharding: It is not clear to me if this is resolved, or not. @sanket1729 can you confirm you are happy with the visibility of these types please.

src/plan.rs
tcharding: As of 4849893 this is still TODO.

src/plan.rs
```diff
@@ -0,0 +1,1161 @@
...
+        None
+    }
+
+    /// Given a public key and a associated leaf hash, look up an schnorr signature with that key
+    /// and return its size
+    fn lookup_tap_leaf_script_sig(&self, _: &Pk, _: &TapLeafHash) -> Option<usize> {
+        None
+    }
+
+    /// Given a SHA256 hash, look up its preimage
```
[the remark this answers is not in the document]
tcharding: The docs on `AssetProvider` need fixing up a bit, seems they are copied over from `Satisfier`. I had a go but got confused about what exactly is return by each function. Some return bool or size but others still return the provided thing.

src/plan.rs
```diff
@@ -0,0 +1,1235 @@
...
+//! signatures or other witness data for it.
+//!
+//! To make a plan you provide the descriptor with "assets" like which keys you are able to use, hash
+//! pre-images you have access to, absolute/relative timelock constraints etc.
+//!
+//! Once you've got a plan it can tell you its expected satisfaction weight which can be useful for
+//! doing coin selection. Furthermore it provides which subset of those keys and hash pre-images you
+//! will actually need as well as what locktime or sequence number you need to set.
+//!
+//! Once you've obstained signatures, hash pre-images etc required by the plan, it can create a
```
tcharding, line 23:
```suggestion
//! Once you've obtained signatures, hash pre-images etc required by the plan, it can create a
```

src/plan.rs
```diff
@@ -0,0 +1,1235 @@
...
+use crate::miniscript::hash256;
+use crate::miniscript::satisfy::{Placeholder, Satisfier, SchnorrSigType};
+use crate::prelude::*;
+use crate::util::witness_size;
+use crate::{DefiniteDescriptorKey, DescriptorPublicKey, Error, MiniscriptKey, ToPublicKey};
+
+/// Trait describing a present/missing lookup table for constructing witness templates
+///
+/// This trait mirrors the [`Satisfier`] trait with the difference that instead of returning the
+/// item if it's present, it only returns a boolean to indicate its presence.
```
tcharding, line 47: Not quite right anymore.

src/plan.rs
```diff
@@ -0,0 +1,1235 @@
...
+///
+/// This trait is automatically implemented for every type that is also a satisfier, and simply
+/// proxies the queries to the satisfier and returns whether an item is available or not.
+///
+/// All the methods have a default implementation that returns `false`.
```
tcharding, line 52: Not correct at the moment.

tcharding (commented):
I went through all the comments and attempted to convince myself if they were resolved or not. I added "As of 4849893 this is still TODO" so that one can easily grep and I flagged things that were not clear to me so someone else can say if they are done or not.

Hope this helps

sanket1729:
Thanks for the help @tcharding. @danielabrozzoni , can you raise a follow up PR that builds on top of this PR and addresses the issues highlighted? If there are no major issues there, we can merge this one first and then follow up again. 

I would like to have the followup PR in a mergeable state before we merge this one.

src/plan.rs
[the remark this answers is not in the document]
danielabrozzoni: > This should not set the tap_merkle_root unless it's a keypath spend I think.

I can't find any mention of this in BIP371, is there anywhere else I should look?

src/plan.rs
danielabrozzoni: Solved in 9a88807fda085e76b638cd66f4c35b7559b97a8a in #592

src/plan.rs
danielabrozzoni: Solved in 767e4f62edbd707693bbc1a118491c35e0b5f8ce in #592

src/plan.rs
danielabrozzoni: Solved in 14833b0a07b01b7289237e0f35e411b0af65342f in #592

src/plan.rs
danielabrozzoni: Solved in 14833b0a07b01b7289237e0f35e411b0af65342f in #592

src/plan.rs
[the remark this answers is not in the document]
danielabrozzoni: Solved in 14833b0a07b01b7289237e0f35e411b0af65342f in #592

(Even tho the docs are still a bit confusing, I think this is already an improvement)

sanket1729:
Merged as a part of #592

PR #482, opened 2022-10-27
Remove `unstable` feature

tcharding:
Currently it is not possible to run the test suite with `cargo test --all-features`, this is because we use a feature called `unstable` to enable the unstable `test` crate that is used for benchmarking.

There is another way to conditionally enable the test crate and guard the benchmark code. We can use a custom configuration option `bench` and then set it using `RUSTFLAGS='--cfg=bench`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ no-std = ["hashbrown", "bitcoin/no-std"]
 compiler = []
 trace = []
 
-unstable = []
 serde = ["actual-serde", "bitcoin/serde"]
 rand = ["bitcoin/rand"]
 base64 = ["bitcoin/base64"]
--- a/README.md
+++ b/README.md
@@ -53,6 +53,11 @@ architectural mismatches. If you have any questions or ideas you want to discuss
 please join us in
 [##miniscript](https://web.libera.chat/?channels=##miniscript) on Libera.
 
+## Benchmarks
+
+We use a custom Rust compiler configuration conditional to guard the bench mark code. To run the
+bench marks use: `RUSTFLAGS='--cfg=bench' cargo +nightly bench`.
+
 
 ## Release Notes
 
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -7,6 +7,12 @@ FEATURES="compiler serde rand base64"
 cargo --version
 rustc --version
 
+# Cache the toolchain we are using.
+NIGHTLY=false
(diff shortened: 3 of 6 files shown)


apoelstra:
Not sure what to make of this CI failure. It might be a bug in `base64` and might be a bug in cargo. Kicked it for now to see if it goes away.

apoelstra:
Could you also add a note to the README explaining how to run the benchmarks?

apoelstra (approved):
ACK 0c653bc013caff3ebe40e9ca829c8e64035518d1

tcharding:
> Not sure what to make of this CI failure.

My memory is super vague on this but I kind of remember now that we can't do this PR until we stop using `base64-compat` which is waiting on `rust-bitocincorerpc` update to pick up the `rust-jsonrpc` upgrade - I think.

https://github.com/rust-bitcoin/rust-bitcoincore-rpc/pull/253

apoelstra:
Sounds like it's about time we just wrote our own `base64` crate :/.

tcharding:
> Sounds like it's about time we just wrote our own base64 crate :/.

Shall we fork https://github.com/marshallpierce/rust-base64 at the version we depend on (0.13) and hack it to suit our project? Or do you think it is worth investing the time to write from scratch?

apoelstra:
Can you kick CI on this one to see if the failure has gone away?

Re forking base64, I think no, I think we should instead try to find a version that works and `crev` it. If it turns out there's no suitable version, then I guess we'll need to fork ... but let's get our hex story figured out first :)

tcharding:
The CI fail is because of `rust-bitcoincore-rpc` using an old version of `rust-jsonrpc` which depends on `bas64-compat`. That's why I tried prodding @stevenroose to release a new version :)

tcharding:
> I think we should instead try to find a version that works and crev it.

It works now, IIRC it was because of edition 2018 and Rust 1.29 that it did not work. I'll put `base64` at the top of my 'to crev list'.

tcharding:
This has uncovered a bug in feature gating in `rust-bitcoin`, will have to fix that first.

tcharding:
Rebase on master.

tcharding:
Back on ice till https://github.com/rust-bitcoin/rust-bitcoin/pull/1936 merges and is released.

apoelstra (approved):
ACK 1a33e67503e84f549bb739650118c96ead9f0a0f

PR #485, opened 2022-11-02
Derive comparison traits for all secret key types

quad:
The public key types (`DescriptorPublicKey`, `DescriptorXKey`, `SinglePubKey`, and `DefiniteDescriptorKey`) all have comparison derivations.

The underlying private key types (`bitcoin::PrivateKey`, `Fingerprint`, `DerivationPath`, `DescriptorXKey`, and `ExtendedPrivKey`) all have comparison derivations.

Let's catch 'em all.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -22,7 +22,7 @@ pub enum DescriptorPublicKey {
 }
 
 /// The descriptor secret key, either a single private key or an xprv.
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq, Clone)]
 pub enum DescriptorSecretKey {
     /// Single private key.
     Single(SinglePriv),
@@ -40,7 +40,7 @@ pub struct SinglePub {
 }
 
 /// A descriptor [`bitcoin::PrivateKey`] with optional origin information.
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq, Clone)]
 pub struct SinglePriv {
     /// Origin information (fingerprint and derivation path).
     pub origin: Option<(bip32::Fingerprint, bip32::DerivationPath)>,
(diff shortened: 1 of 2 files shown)


apoelstra:
Nice! cc @sanket1729 I'd like you to concept ACK before merging this

apoelstra (approved):
ACK 8e781ca14f14e1528aadd0875ee4dd8ab831f57b

sanket1729:
@apoelstra, I see that rust-secp implements `PartialOrd`, and `Ord` directly without any constant time guarantees. When did we do that? I thought we deliberately did not do that to avoid side channels.

sanket1729:
Don't mind merging this. Should we wait until https://github.com/rust-bitcoin/rust-secp256k1/issues/471 is resolved?

sanket1729:
Will merge after #488

apoelstra:
@sanket1729 it looks like they appeared in https://github.com/rust-bitcoin/rust-secp256k1/pull/364 where there is no discussion about them :/ maybe we forgot at that time about the pitfalls of `Ord` on secret key types. I also remember this discussion but I can't find it now and don't remember where we landed.

One thing I **do** remember saying is that any comparisons of secret keys is likely to be leaky anyway. Like, if you put secret keys into a `BTreeMap` or something then you know that the first key to be read is likely to be smaller than average, and it doesn't matter whether `Ord` is constant time. So my feeling was that if we had `Ord` at all, I wasn't so worried about constant-timeness as I would be for `Eq` (which I expect to be more frequently used, and to be usually used safely).

Another thing I **do** remember is that we were considering doing something crazy like, hashing the secret keys and then comparing by hash. But then somebody (kixunil?) pointed out that there are coherency requirements between `Eq`, `Ord` and `Hash` that would make this extremely hard to do, at least efficiently. So maybe there was a discussion about this but it never went anywhere since we couldn't agree on a good path forward?

sanket1729 (approved):
ACK 8e781ca14f14e1528aadd0875ee4dd8ab831f57b

PR #488, opened 2022-11-05
Release 9.0.0

### LINKED ISSUE #483: We should not be running pkh logic through rawpkh
See for example https://github.com/rust-bitcoin/rust-miniscript/blob/master/src/miniscript/satisfy.rs#L937 (not sure if there are any other examples). Here we have a public key, convert it to a hash, then look up the key again from that.

In #481 (the "planning module" PR) I am suggesting that we work with `DescriptorPublicKey`s rather than `DefinitePublicKey`s, which means that we may be working with keys where we don't know their hash160 and can't do this lookup. So in addition to being unnecessary, this roundtrip through a hash may become wrong.

sanket1729:
- Cleanup some pkh code in satisfaction
- Fixes #483
- Fixes another bug dealing with dissatisfaction of thresh inside pkh
- Adds example test vector from discussions
- release 9.0.0 with above fixes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 9.0.0 - November 5, 2022
+
+- Fixed a bug dealing with dissatisfying pkh inside thresh
+- Changed the signature of `Satisfier::lookup_raw_pkh_pk` API. Only custom implementations
+  of `Satisfier` need to be updated. The psbt APIs are unchanged.
+- Fixed a bug related to display of `raw_pk_h`. These descriptors are experimental
+  and only usable by opting via `ExtParams` while parsing string.
 # 8.0.0 - October 20, 2022
 
 This release contains several significant API overhauls, as well as a bump
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "miniscript"
-version = "8.0.0"
+version = "9.0.0"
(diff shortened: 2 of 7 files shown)


apoelstra:
I'm not sure that we can change the `Display` output in a patch release.

apoelstra:
Changing the signature of `Satisfier::lookup_raw_pkh_pk` we definitely *can't* do in a patch release.

apoelstra:
All changes in 7e1a7a098baf324a628cc426da4442a8599e6d92 look good to me, but they're all API-breaking and I think we need to release 9.0.0 with them.

sanket1729:
All of the changes are related to the raw APIs. I was thinking we can just
justify them as a bug fix instead of major release?

Thinking from a software sense, it makes sense to call it a breaking
change.

engenegr:
@apoelstra what is expected date for the next release if the fix goes with it?

sanket1729:
@engenegr, you can expect a new release early next week.

sanket1729:
Changed to release 9.0.0

apoelstra:
@sanket1729 you did not change the actual commit to 9.0.0!

sanket1729:
@apoelstra , lol. Good catch

apoelstra (approved):
ACK a113e35923dc094b2554209d2042ac8d7683ebb7

apoelstra:
Tagged & published

PR #489, opened 2022-11-05
replaced if else return with mapping

### LINKED ISSUE #479: Use map_err instead of if else block
Really simple change suggested as good first issue for newcomers to contribute

https://github.com/rust-bitcoin/rust-miniscript/pull/478#discussion_r998582977

Scharxi:
Fixes #479


apoelstra (approved):
ACK 480c3fd41741f7716adb1869c5208573dc8564a0

PR #492, opened 2022-11-15
DefiniteDescriptorKey: provide additional methods for converting to a DescriptorPublicKey

### LINKED ISSUE #491: `DefiniteDescriptorKey` does not expose a way to get the key out if it's internally a single key
Right now if you have a `DescriptorPublicKey` and you convert it to a `DefiniteDescriptorKey` with `at_derivation_index`, you lose any good way to determine whether you have a single key or not, and you lose access to the key.

With `DescriptorPublicKey` you can just match on it since it's a public enum and is either `Single` or `XPub`. But `DefiniteDescriptorKey` is an opaque struct.

Consider the case where you have a descriptor public key, potentially with hardened wildcards, but it's user-provided so it could be anything. To get a specific public key out of this you need to talk to a hww. With a `DescriptorPublicKey` you'd match on the key; for `Single` you directly pull the key out, and for `Xpub` you pass the derivation path over USB to your hww and use the result. But now with `DefiniteDescriptorKey` you're forced to call `derive_public_key`, which requires you have a secp context available, and only after this fails do you know that you need to switch to the hww.

I'm not sure what the best solution here is. Probably just a `single_key -> Option<SinglePub>` method on `DefiniteDescriptorKey`.

apoelstra:
Fixes #491 

I didn't provide a mutable version of this. I'm not sure if we want to support this or not; generally public keys should be immutable after creation.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -777,6 +777,16 @@ impl DefiniteDescriptorKey {
     pub fn full_derivation_path(&self) -> bip32::DerivationPath {
         self.0.full_derivation_path()
     }
+
+    /// Reference to the underlying `DescriptorPublicKey`
+    pub fn as_descriptor_public_key(&self) -> &DescriptorPublicKey {
+        &self.0
+    }
+
+    /// Converts the definite key into a generic one
+    pub fn into_descriptor_public_key(self) -> DescriptorPublicKey {
+        self.0
+    }
 }
 
 impl FromStr for DefiniteDescriptorKey {
@@ -839,6 +849,12 @@ impl ToPublicKey for DefiniteDescriptorKey {
 impl From<DefiniteDescriptorKey> for DescriptorPublicKey {
     fn from(d: DefiniteDescriptorKey) -> Self {
         d.0
+    }
+}
+
+impl Borrow<DescriptorPublicKey> for DefiniteDescriptorKey {
+    fn borrow(&self) -> &DescriptorPublicKey {
+        &self.0
     }
 }
 


LLFourn:
Concept ACK. I think `From<DefiniteDescriptorKey> for DescriptorPublicKey` should be included. The `to_` method seems superfluous given you have the `as_` there. Also the `From` implementation for references is unusual -- Maybe `AsRef` instead?

If we allowed `as_mut` this would violate the guarantees of `DefiniteDescriptor` since it would let you put a wildcard in there.

LLFourn:
> > Also the From implementation for references is unusual -- Maybe AsRef instead?
> 
> Is it unusual? I'm happy to do `AsRef` instead. TBH I have no idea what's idiomatic and what's not.

I think so. I checked out [`std::convert::From`](https://doc.rust-lang.org/std/convert/trait.From.html) and I can't see a single `&` to `&` implementation. TBH I don't really understand what's in good taste either but there are two choices that afaik are preferable to `From<&..> for &..`: `Borrow` or `AsRef`.  I'm not exactly sure when to go for which one but the docs make it clear that `AsRef` is meant for conversion things where you want to take a `T: AsRef<DescriptorPublicKey>` for your function which I guess is the goal here. As opposed to `From`, `AsRef` will let you to pass in `DescriptorPublicKey`, `&DefiniteDescriptorKey` even `&mut DescriptorPublicKey` and so on. The downside is you have to also implement `AsRef<DescriptorPublicKey> for DescriptorPublicKey`.

apoelstra:
Ok, after writing https://github.com/rust-bitcoin/rust-miniscript/pull/481/commits/e16e7dd93de3f8ceb41705d6d0e13bf3d627f2a6#r1025217984 I realize that `AsRef` is what feels more natural to a user :)

>I checked out std::convert::From and I can't see a single & to & implementation

Huh, I could've sworn I saw this in stdlib some time ago. Maybe this is a pre-1.0 memory or a fever dream :).

>you have to also implement AsRef<DescriptorPublicKey> for DescriptorPublicKey

No, there's a blanket `AsRef<T> for &T` impl that will suffice https://doc.rust-lang.org/std/convert/trait.AsRef.html

LLFourn:
> No, there's a blanket `AsRef<T> for &T` impl that will suffice https://doc.rust-lang.org/std/convert/trait.AsRef.html

No there isn't! From that page:

> Unlike AsRef, [Borrow](https://doc.rust-lang.org/std/borrow/trait.Borrow.html) has a blanket impl for any T, and can be used to accept either a reference or a value.

Fevered dreams indeed!

apoelstra:
Lol, ok. I'm starting to internalize what these trains are supposed to mean. Let's use `Borrow` then :)

apoelstra:
Pushed to use `Borrow` instead.

LLFourn:
> Lol, ok. I'm starting to internalize what these **trains** are supposed to mean.

:sweat_smile: 

nit: the `to_` is till there.

ACK 0b569a29343c8a5f40eeff69ab40819398d13ac1

apoelstra:
lol, removed the `to_`

sanket1729 (approved):
ACK 1477dbbb80523dde399a7cbaba25cbac54bf2816

PR #493, opened 2022-12-02
update TranslatePk again

sanket1729:
This PR has lot of cleanups that I had planned but did not have time for a long time. 

- All places now use constructors that do invariant checking on the struct being created. 
- Translate* APIs had a soundness bug you could create uncompressed keys in segwit descriptors etc.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -368,6 +368,57 @@ where
     fn hash160(&mut self, hash160: &P::Hash160) -> Result<Q::Hash160, E>;
 }
 
+/// An enum for representing translation errors
+pub enum TranslateErr<E> {
+    /// Error inside in the underlying key translation
+    TranslatorErr(E),
+    /// Error in the final translated structure. In some cases, the translated
+    /// structure might not be valid under the given context. For example, translating
+    /// from string keys to x-only keys in wsh descriptors.
+    OuterError(Error),
+}
+
+impl<E> TranslateErr<E> {
+    /// Enum used to capture errors from the [`Translator`] trait as well as
+    /// context errors from the translated structure.
+    /// The errors occurred in translation are captured in the [`TranslateErr::TranslatorErr`]
+    /// while the errors in the translated structure are captured in the [`TranslateErr::OuterError`]
+    ///
+    /// As of taproot upgrade: The following rules apply to the translation of descriptors:
+    /// - Legacy/Bare does not allow x_only keys
+    /// - SegwitV0 does not allow uncompressed keys and x_only keys
+    /// - Tapscript does not allow uncompressed keys
+    /// - Translating into multi-path descriptors should have same number of path
+    /// for all the keys in the descriptor
+    ///
+    /// # Errors
+    ///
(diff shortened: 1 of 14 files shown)


apoelstra:
These look great! Only nits are that I think in a couple cases you could've cleaned up some `map_err` calls, but I'm not sure and I didn't feel strongly enough to actually try my ideas to see if they compiled.

I'm not thrilled about some of the `expect`s done during translation of keys, say, but they're fine for now. I'm a little tempted to add a `translate_pk_unchecked` or something but that can wait for a future PR and more discussion.

apoelstra (approved):
ACK 2b406797daad704852eaf6863e688e83b94ae9b9

sanket1729:
> These look great! Only nits are that I think in a couple cases you could've cleaned up some map_err calls,
Perhaps, it might be time for clippy in rust-miniscript? It is great is suggesting idiomatic things when possible. Given that it is already existing in rust-bitcoin, we can add it here.

sanket1729:
Rebased. But based on #537

src/lib.rs
```diff
@@ -368,6 +368,57 @@ where
...
+    /// - Legacy/Bare does not allow x_only keys
+    /// - SegwitV0 does not allow uncompressed keys and x_only keys
+    /// - Tapscript does not allow uncompressed keys
+    /// - Translating into multi-path descriptors should have same number of path
+    /// for all the keys in the descriptor
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the Error is OutError.
+    pub fn try_into_translator_err(self) -> Result<E, Self> {
```
apoelstra, line 397: In a3019c7dfe891b7c141d779f082bcdb3b4f31fb5:

Since you exclusively use this function to assert, I think you should just call it `unwrap_translator_err` and have it panic if it gets an `OuterError`. I find the current name `try_into_translator_err` hard to understand, and each time you call it you've got a separate variant of `expect("outer errors can't happen")` which is also hard to understand.

apoelstra:
ack a3019c7dfe891b7c141d779f082bcdb3b4f31fb5 except for the one comment.

src/lib.rs
sanket1729: I think the point of try into is to explicitly have this message. There are different reasons why this might not fail. BIP32 derivations/translations we know that are not going to be fail etc.

And having a string message as a small comment is helpful rather than unwrap thing because it at least forces me to think and write why it might not fail

apoelstra: Can we call it `expect_translator_err` and take a string?

sanket1729:
Added one commit to address the issue.

apoelstra:
3ebdf4c243aa0b54ad1d33b4b328a72cad7dda67 does not compile.

But ACK 1a3a944c9bb9b91d4b7efd64a7980b0555e5e351 otherwise.

sanket1729:
Fixed and force pushed

apoelstra (approved):
ACK b90cdda3136636485e69bdea4699563f7f74401b

PR #501, opened 2022-12-18
Remove unreachable statement

tcharding:
Clippy warns about this code. It is not immediately obvious what the purpose of this code is - pattern match and call `unreachable`?

Remove the code, this changes the behaviour of the function, to maintain current behaviour we could panic or debug_assert:
```
  if witness == Policy::Threshold { panic!("some message") }
```
But I'd have no idea what to write in the panic message?


apoelstra:
This looks like it's trying to be a `debug_assert` but can't be, because we can't compare enum variants for equality like you're suggesting.

The panic message can just be "debug assert: unreachable code" if you want, though we could describe the condition we're checking more explicitly if you want.

What is Clippy complaining about?

tcharding:
warning: you seem to be trying to use `match` for destructuring a single pattern. Consider using `if let`

tcharding:
And run formatter.

apoelstra (approved):
ACK fb4c33b5d200dbf8522c4e57a2ed974c4767fc95

PR #502, opened 2022-12-18
Use allow attribute

tcharding:
Done in an effort to clear all clippy warnings.

Add two allow attributes to quieten clippy. This may not be the best solution but it at least flags the code for future improvement and allows us to enable clippy in CI (along with other clippy work).


apoelstra (approved):
ACK b481cbadfb5e81e23b08d3658d581b7d95c81c32

PR #503, opened 2022-12-21
CI cleanup

tcharding:
Do a bunch of cleanups to the CI script and GitHub Actions config file. None of these changes change the amount of test coverage but we do achieve some de-duplication of test runs. Otherwise, clean up only.


apoelstra (approved):
ACK 8c4853800712fd56b3fc258f6a0da2ca88e68d83

Issue #437, opened 2022-06-23
Remove DummyKey module


sanket1729:
We can use MinicriptKey = String for our testing purposes. Is there any use for the DummyKey struct? Perhaps, we can remove it?

sanket1729:
This would require replacing all the tests that use DummyKey to use String

Eunoia1729:
Picking this up :)

PR #500, opened 2022-12-18
Move dummy types to submodule

tcharding:
Clippy warns about the dummy types, in an effort to clear all clippy warnings do:

- Create a submodule `dummy` and move all the dummy key types there (refactor/clean up).
- Manually implement `PartialEq` and `Eq` for all the dummy types.

PR #508, opened 2022-12-21
Remove `DummyKey`

tcharding:
An alternative to #500 (because I just saw the issue #437)

As described in issue #437 we can use `String` in place of the `DummyKey` in all current test use cases.

Use `String` in tests and remove the `DummyKey` along with all the other (currently unused) `DummyFoo` types.

Fix: #437

### DIFF
--- a/fuzz/fuzz_targets/compile_descriptor.rs
+++ b/fuzz/fuzz_targets/compile_descriptor.rs
@@ -1,24 +1,24 @@
 extern crate miniscript;
 
 use miniscript::Segwitv0;
-use miniscript::{policy, DummyKey, Miniscript};
+use miniscript::{policy, Miniscript};
 use policy::Liftable;
 
 use std::str::FromStr;
 
-type DummyScript = Miniscript<DummyKey, Segwitv0>;
-type DummyPolicy = policy::Concrete<DummyKey>;
+type Script = Miniscript<String, Segwitv0>;
+type Policy = policy::Concrete<String>;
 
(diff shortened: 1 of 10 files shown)


apoelstra:
I think we should first take a look and see if it's feasible to use Kani with a "dumb" DummyKey. We definitely won't be able to do it with stringly-typed keys. So if we *can* with the existing struct that's one reason to keep it. (Though maybe not -- we could create special-purpose dummy keys just for the Kani tests.)

tcharding:
Do you mean that we should try writing Kani tests for types that use `dummy::Key` because we definitely won't be able to use `String` for them?

tcharding:
I did https://github.com/rust-bitcoin/rust-bitcoin/pull/1496 to see if I'm getting this.

sanket1729 (commented):
ACK, looks like CI is failing on rustfmt

tcharding:
Formatted and rebased, thanks. I'll get used to formatting my PRs soon ... I hope

tcharding:
But @sanket1729 please don't merge until I clarify with @apoelstra the kani testing thing, thanks.

apoelstra:
Never mind me. My thinking is that Kani can't handle strings but it can probably handle a trivial (or nearly trivial) type.

But if we want to use Kani we can reintroduce kani-specific key types.

apoelstra (approved):
ACK eeb823f7cf0a4c0da974bfca4cc3565c0ebe7f2f

Issue #364, opened 2022-04-20
Use upstream `script_code_p2wpkh` API


sanket1729:
After https://github.com/rust-bitcoin/rust-bitcoin/pull/954 and rust-bitcoin 0.29 release, we can use the new API to cleanup `script_code_p2wpkh` here to cleanup manual calculation in `psbt/mod.rs`

tcharding:
I can't find any code manually building the script code - is this issue still live?

sanket1729:
https://github.com/rust-bitcoin/rust-miniscript/blob/7664570e7d4401f457932633455d42660a00b6cb/src/psbt/mod.rs#L1258

PR #512, opened 2022-12-31
Use p2wpkh_script_code from rust-bitcoin

tcharding:
Currently we have a function to construct the script code for a P2WPKH, this code is now upstream in `rust-bitcoin` so we can use it.

Fix: #364


tcharding:
Trying to work out why we included 0x14 here in miniscript and not in bitcoin led me down an epic transaction signing rabbit hole that led to: https://github.com/rust-bitcoin/rust-bitcoin/pull/1522, turned out to be super trivial `push_slice` adds the 0x14 for us in rust-bitcoin but here we manually create the byte vector.

sanket1729 (approved):
ACK 842ae937bd434c05725e9b3376eaaa3d74929fe2

PR #522, opened 2023-01-17
Update TapTree::iter docs

### LINKED ISSUE #441: Iterator for Tr taptree which iterates scripts in DFS order
In order to construct correct PSBT tap tree we need to iterate the scripts in DFS order. Right now, the wallet has to manually recurse through `Tr::tree` nodes in order to generate PSBT for taproot descroptor. Alternatively, wallets may mistakingly use `spent_info().as_script_map()`, which will generate incorrect ordering and other co-signers may not recognize the tap tree coming from PSBT generated in that way as belonging to the wallet descriptor.

sanket1729:
Fixes #441

### DIFF
--- a/src/descriptor/tr.rs
+++ b/src/descriptor/tr.rs
@@ -120,7 +120,8 @@ impl<Pk: MiniscriptKey> TapTree<Pk> {
         }
     }
 
-    /// Iterate over all miniscripts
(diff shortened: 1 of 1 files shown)


apoelstra:
Could you add a newline after the first line?

apoelstra (approved):
ACK 3fcc84c6ecb5598fa17a56080cac86414d504db9

PR #524, opened 2023-02-07
hexify for better error message

Harshil-Jani:
This PR hexifies the Error implementation as requested in TODO. Let me know if this works and then I can remove the "TODO : Hexify this " Comment from there. 

Signed-off-by: Harshil Jani <harshiljani2002@gmail.com>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -512,7 +512,10 @@ impl fmt::Display for Error {
         match *self {
             Error::InvalidOpcode(op) => write!(f, "invalid opcode {}", op),
             Error::NonMinimalVerify(ref tok) => write!(f, "{} VERIFY", tok),
-            Error::InvalidPush(ref push) => write!(f, "invalid push {:?}", push), // TODO hexify this
+            Error::InvalidPush(ref push) => {
+                write!(f, "invalid push ")?;
+                bitcoin::hashes::hex::format_hex(push, f)
+            },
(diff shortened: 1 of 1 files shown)


apoelstra:
concept ACK for sure. But definitely nack this particular implementation :) it allocates a string for every character and then a new one.

You can use https://docs.rs/bitcoin_hashes/latest/bitcoin_hashes/hex/fn.format_hex.html to do it without allocations. `bitcoin_hashes` is available in this crate under the path `bitcoin::hashes`.

Harshil-Jani:
Got it. Thanks for helping @apoelstra

apoelstra:
You shouldn't need to explicitly add a `bitcoin_hashes` dependency. Just use `bitcoin::hashes`.

Harshil-Jani:
> Thanks for the iterations! This looks great now.
> 
> Can you squash all 3 commits into one?

Squashed.

apoelstra:
@Harshil-Jani thanks! But now that the code is all in one place I see that you're using `format_hex` wrong. It does not return a hex-formatted object. It actually does the hex-encoding.

To accomplish what you want, the code you need is

```rust
write!(f, "invalid push ")?;
bitcoin::hashes::hex::format_hex(push, f)
```

apoelstra (approved):
ACK a0b001b65c10551d3dc5f816affe04df51542678

PR #525, opened 2023-02-18

sanket1729:
While we are at it, do not use the default impl. This makes these types
of errors easy to make. I think this could have easily been prevented if
I was forced to explicitly write AnyNonZero in front of multi_a.

Reported by darosior.

### DIFF
--- a/tests/test_desc.rs
+++ b/tests/test_desc.rs
@@ -395,6 +395,10 @@ fn test_descs(cl: &Client, testdata: &TestData) {
     let result = test_desc_satisfy(cl, testdata, "tr(X!,{pk(X1!),pk(X2!)})");
     assert_eq!(result, Err(DescError::PsbtFinalizeError));
 
+    // Test 10: Test taproot desc with ZERO known keys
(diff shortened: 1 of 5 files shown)


apoelstra (approved):
ACK pr/525/head

apoelstra:
There is some github weirdness going on. ACK 7118c848

apoelstra:
I can't merge because Github is not publishing the tip of this PR under the right ref name.

tests/test_desc.rs
```diff
@@ -395,6 +395,10 @@ fn test_descs(cl: &Client, testdata: &TestData) {
     let result = test_desc_satisfy(cl, testdata, "tr(X!,{pk(X1!),pk(X2!)})");
     assert_eq!(result, Err(DescError::PsbtFinalizeError));
 
+    // Test 10: Test taproot desc with ZERO known keys
```
darosior, line 398: Comment should be updated?

darosior (approved):
utACK 7118c8488bc7e2c6219163648dd807c815d744f2

PR #526, opened 2023-02-21
Serde for DescriptorPublicKey

quad:
Save me from ever writing the following code again:

```rust
pub mod serde_descriptor_key {
    use std::str::FromStr;

    use miniscript::DescriptorPublicKey;
    use serde::{Deserialize, Deserializer, Serializer};

    pub fn serialize<S>(dpk: &DescriptorPublicKey, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        serializer.serialize_str(&dpk.to_string())
    }

    pub fn deserialize<'de, D>(deserializer: D) -> Result<DescriptorPublicKey, D::Error>
    where
        D: Deserializer<'de>,
    {
        let s = String::deserialize(deserializer)?;
        DescriptorPublicKey::from_str(&s).map_err(serde::de::Error::custom)
    }
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ actual-serde = { package = "serde", version = "1.0", optional = true }
 bitcoind = { version = "0.27.0", features=["23_0"] }
 actual-rand = { package = "rand", version = "0.8.4"}
 secp256k1 = {version = "0.24.0", features = ["rand-std"]}
+serde_test = "1.0.147"
 
 [[example]]
 name = "htlc"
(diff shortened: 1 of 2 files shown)


apoelstra:
LGTM! Will let @sanket1729 take a look before merging. I have two points of hesitation:

* One is that we have an existing `serde_string_impl_pk` macro...but it looks like it's inapplicable, and anyway this code is very short.
* The other is that we're allocating a `String` just to dereference it, which feels wasteful ... but OTOH I don't know how to get around this without a ton of boilerplate (and even then I'm doubtful). The [serde_with crate](https://crates.io/crates/serde_with/2.2.0) might be worth exploring in the future once we have cargo-crev going and once we've vetted all our existing deps. Amusingly, when I searched for how to do this, I found an old serde issue where dtolnay suggests doing your exact approach.

sanket1729 (approved):
ACK 3502477f007bdcfc48b487289b06f127b1429fc8

PR #527, opened 2023-02-22
Upgrade bitcoin to upcoming 0.30.0 release

tcharding:
**This is not a merge candidate**

To get a feel for using the code we are about to release I hacked up an upgrade to something resembling the upcoming `rust-bitcoin` 0.30.0 release.

FTR I found this upgrade painful, some pain points were:

- API methods missing or hard to use (or I'm being braindead), specifically:
  - `PushBytes::to_vec()`
  - `script::Builder::push_script()` (to handle `PushBytes`)

- New paths/re-exports don't match with current usage i.e., here in miniscript we tend to include `bitcoin::` on types, this upgrade removed a few types from the crate root so now we have long fully qualified paths instead e.g., `bitcoin::crypto::ecdsa::Signature` instead of `bitcoin::EcdsaSig`.

- The "don't implement `Ord` thing" from `absolute::LockTime` means we need some way to order absolute lock times, here we use a wrapper type.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,11 @@ serde = ["actual-serde", "bitcoin/serde"]
 rand = ["bitcoin/rand"]
 base64 = ["bitcoin/base64"]
 
+[patch.crates-io.bitcoin_hashes]
+path = "../rust-bitcoin/hashes"
+
 [dependencies]
-bitcoin = { version = "0.29.1", default-features = false }
+bitcoin = { path = "../rust-bitcoin/bitcoin", default-features = false }
 hashbrown = { version = "0.11", optional = true }
 
 # Do NOT use this as a feature! Use the `serde` feature instead.
--- a/examples/psbt_sign_finalize.rs
+++ b/examples/psbt_sign_finalize.rs
@@ -133,7 +133,7 @@ fn main() {
         .to_secp_msg();
 
     // Fixme: Take a parameter
-    let hash_ty = bitcoin::EcdsaSighashType::All;
+    let hash_ty = bitcoin::sighash::EcdsaSighashType::All;
 
     let sk1 = backup1_private.inner;
     let sk2 = backup2_private.inner;
@@ -150,7 +150,7 @@ fn main() {
 
     psbt.inputs[0].partial_sigs.insert(
         pk1,
-        bitcoin::EcdsaSig {
+        bitcoin::crypto::ecdsa::Signature {
             sig: sig1,
             hash_ty: hash_ty,
         },
--- a/examples/sign_multisig.rs
+++ b/examples/sign_multisig.rs
@@ -18,7 +18,7 @@ use std::collections::HashMap;
 use std::str::FromStr;
 
 use bitcoin::blockdata::witness::Witness;
-use bitcoin::{secp256k1, PackedLockTime, Sequence};
+use bitcoin::{secp256k1, absolute, Sequence};
 
 fn main() {
     let mut tx = spending_transaction();
@@ -63,7 +63,7 @@ fn main() {
     // Attempt to satisfy at age 0, height 0.
     let original_txin = tx.input[0].clone();
 
-    let mut sigs = HashMap::<bitcoin::PublicKey, miniscript::bitcoin::EcdsaSig>::new();
+    let mut sigs = HashMap::<bitcoin::PublicKey, miniscript::bitcoin::crypto::ecdsa::Signature>::new();
 
     // Doesn't work with no signatures.
     assert!(descriptor.satisfy(&mut tx.input[0], &sigs).is_err());
@@ -91,7 +91,7 @@ fn main() {
 fn spending_transaction() -> bitcoin::Transaction {
     bitcoin::Transaction {
         version: 2,
-        lock_time: PackedLockTime::ZERO,
+        lock_time: absolute::LockTime::ZERO,
         input: vec![bitcoin::TxIn {
             previous_output: Default::default(),
             script_sig: bitcoin::Script::new(),
@@ -128,8 +128,8 @@ fn list_of_three_arbitrary_public_keys() -> Vec<bitcoin::PublicKey> {
 
 // Returns a signature copied at random off the blockchain; this is not actually
 // a valid signature for this transaction; Miniscript does not verify the validity.
-fn random_signature_from_the_blockchain() -> bitcoin::EcdsaSig {
-    bitcoin::EcdsaSig {
+fn random_signature_from_the_blockchain() -> bitcoin::crypto::ecdsa::Signature {
+    bitcoin::crypto::ecdsa::Signature {
         sig: secp256k1::ecdsa::Signature::from_str(
             "3045\
              0221\
@@ -138,6 +138,6 @@ fn random_signature_from_the_blockchain() -> bitcoin::EcdsaSig {
              531d75c136272f127a5dc14acc0722301cbddc222262934151f140da345af177",
         )
         .unwrap(),
-        hash_ty: bitcoin::EcdsaSighashType::All,
+        hash_ty: bitcoin::sighash::EcdsaSighashType::All,
     }
 }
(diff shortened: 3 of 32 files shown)


tcharding:
FYI the `rust-bitcoincore-rpc` was trivial :)

apoelstra:
* Regarding `PushBytes` etc, definitely converting a `EcdsaSig` to `Vec` then a `PushBytes` is the wrong thing to do! In some cases maybe we should be more agressive in changing existing slice/vector types to `PushBytes` at the API boundary rather than converting. In this case I notice we have `Witness::push_bitcoin_signature` method but not a `script::Builder::push_bitcoin_signature` method, which is an oversight upstream.
* For new paths/exports we'll handle that case-by-case. I think we should import `bitcoin::crypto::ecdsa` and then use `ecdsa::Signature`. IMHO it's not as nice as using `bitcoin::` but having any path at all indicates "this is an import, look up top to find where it comes from".
* For `Ord`, we should really be distinguishing between `Height` and `Time` by having distinct `Terminal` fragments. Then we'd get `Ord` for free.

tcharding:
> For new paths/exports we'll handle that case-by-case. I think we should import bitcoin::crypto::ecdsa and then use ecdsa::Signature. IMHO it's not as nice as using bitcoin:: but having any path at all indicates "this is an import, look up top to find where it comes from".

This is the obvious solution but because we have `bitcoin/src/taproot` and `bitcoin/src/crypto/taproot` there are times when one layer of path doesn't work. I've brought this up before but I don't see the reason for having the two modules, and I definitely don't see how downstream users are supposed to tell why there are two and what lives in what (especially when `src/taproot` depends on secp stuff and `crytpo::taproot` stuff).

tcharding:
> For Ord, we should really be distinguishing between Height and Time by having distinct Terminal fragments. Then we'd get Ord for free.

Aren't terminal fragments based on "Miniscript fragments" from https://bitcoin.sipa.be/miniscript/?

tcharding:
Allright, second attempt to use `PushBytes` and I still cannot work it out, WTF am I doing wrong?
```rust

    /// Pushes a DER-encoded ECDSA signature with a signature hash type as a new element on the
    /// witness, requires an allocation.
    pub fn push_bitcoin_signature(self, signature: &ecdsa::SerializedSignature, hash_type: EcdsaSighashType) -> Builder {
        let mut sig = [0; ecdsa::MAX_SIG_LEN];
        sig[..signature.len()].copy_from_slice(signature);
        sig[signature.len()] = hash_type as u8;
        let push = PushBytes::try_from(&sig[..signature.len() + 1]).expect("73 is within limit");
        self.push_slice(&push)
    }
```
Error:
```
error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
   --> bitcoin/src/blockdata/script/builder.rs:78:20
    |
78  |         let push = PushBytes::try_from(&sig[..signature.len() + 1]).expect("73 is within limit");
    |                    ^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: within `push_bytes::primitive::PushBytes`, the trait `Sized` is not implemented for `[u8]`
note: required because it appears within the type `push_bytes::primitive::PushBytes`
   --> bitcoin/src/blockdata/script/push_bytes.rs:43:16
    |
43  |     pub struct PushBytes([u8]);
```

cc @Kixunil incase you don't watch this repo as closely.

tcharding:
> Regarding PushBytes etc, definitely converting a EcdsaSig to Vec then a PushBytes is the wrong thing to do! In some cases maybe we should be more agressive in changing existing slice/vector types to PushBytes at the API boundary rather than converting. In this case I notice we have Witness::push_bitcoin_signature method but not a script::Builder::push_bitcoin_signature method, which is an oversight upstream.

Ok I finally worked out how to use `PushBytes`, is there a bug/foot gun hiding in this code?
```rust

impl AsRef<PushBytes> for SerializedSignature {
    #[inline]
    fn as_ref(&self) -> &PushBytes {
        &<&PushBytes>::from(&self.data)[..self.len()]
    }
}
```
This implies that one can do `build.push_slice(sig)` but if we look in the `witness` module we have
```rust
    /// Pushes a DER-encoded ECDSA signature with a signature hash type as a new element on the
    /// witness, requires an allocation.
    pub fn push_bitcoin_signature(&mut self, signature: &ecdsa::SerializedSignature, hash_type: EcdsaSighashType) {
        // Note that a maximal length ECDSA signature is 72 bytes, plus the sighash type makes 73
        let mut sig = [0; 73];
        sig[..signature.len()].copy_from_slice(signature);
        sig[signature.len()] = hash_type as u8;
        self.push(&sig[..signature.len() + 1]);
    }
```
Implying that we need to manually include the sighash type.

cc @Kixunil

tcharding:
And am I still being brain dead or is this how we are expecting downstream to use `PushBytes`
```rust
                let redeem_script = wpkh.script_pubkey();
                script::Builder::new()
                    .push_slice(&<&PushBytes>::try_from(redeem_script.as_bytes()).expect("script is within limit"))
                    .into_script()
```

Excuse my french but that is F***ing horrible.

Kixunil:
Good idea, I should try to upgrade my projects to see if anything breaks.

* `PushBytes::to_vec` - do you actually need `Vec<u8>` or `PushBytesBuf`? For the latter it should be `to_owned`
* Pushing into script should be done like this: `script.push_slice(signature.serialize())`
* The code in `Witness` wasn't updated, it should be `self.push(&signature.serialize())`.

I was also thinking we should disallow pushing legacy signatures into `Witness` and non-legacy signatures into `ScriptBuf<Sig>` but I don't expect it to happen in this release.

apoelstra:
> > For Ord, we should really be distinguishing between Height and Time by having distinct Terminal fragments. Then we'd get Ord for free.
> 
> Aren't terminal fragments based on "Miniscript fragments" from https://bitcoin.sipa.be/miniscript/?

Yes, and height-based and time-based `older`s are unambiguously distinct (although the table doesn't list them separately, and we use the same string-based name `older` for both...though we spent years telling ourselves that we would change that eventually).

tcharding:
Am I understanding you correctly, are you saying `rust-miniscript` _is_ miniscript (i.e., its the reference implementation) so we can change the terminal names and such things?

apoelstra:
@tcharding no, I'm saying that `older(time)` and `older(height)` are **already** serialized as distinct nonoverlapping fragments in bitcoin script, so there is no problem separating them in our `Terminal` enum.

tcharding:
> Good idea, I should try to upgrade my projects to see if anything breaks.

Yeah I want to do bdk and/or electrs as well to see if we break them.

> * `PushBytes::to_vec` - do you actually need `Vec<u8>` or `PushBytesBuf`? For the latter it should be `to_owned`
> * Pushing into script should be done like this: `script.push_slice(signature.serialize())`

Wow, I did not realise yesterday that `SerializedSignature` was a new recently added type, added in #1190, I don't know what happened during my review of #1190 - epic fail.

> * The code in `Witness` wasn't updated, it should be `self.push(&signature.serialize())`.

I'll fix it up.

> I was also thinking we should disallow pushing legacy signatures into `Witness` and non-legacy signatures into `ScriptBuf<Sig>` but I don't expect it to happen in this release.

Cool, something to keep in mind when we add script tagging eh, feels like script tagging is going to be quite involved.

tcharding:
> @tcharding no, I'm saying that `older(time)` and `older(height)` are **already** serialized as distinct nonoverlapping fragments in bitcoin script, so there is no problem separating them in our `Terminal` enum.

Ooooh, thanks. I'll have a play when I get a second.

tcharding:
>> The code in Witness wasn't updated, it should be self.push(&signature.serialize()).
> I'll fix it up.

This is a minor can of worms, will leave it till after release.

sanket1729:
Using your changes from this branch to a make a release PR here

tcharding:
Unneeded now, closing.

PR #530, opened 2023-03-09

sanket1729:


sanket1729:
This is an old version of rust-minsicript (July 2021). Did not make any attempts to fix the fuzz CI and hacky integration testing infra here. I think we should go ahead with this for 7.0.1

apoelstra (approved):
ACK 0ad1fdfaba8f4f488ba234cabed9e5210e071163

PR #532, opened 2023-03-09
Fix CI. Move bitcoind tests to inner crate

sanket1729:
This will allow us to have separate test dependacies and test CI only on
stable. Ideally, we would want to have to a conditional dev dependancy
but that is not possible currently

Everytime `bitcoind` breaks with MSRV and we don't really need int tests with MSRV. 
Another issue is that testing this MSRV(2 years old) becomes a part of CI. When backporting fixes, we need to take care of making it pass. We need to support MSRV of old release(which is 2 + 2 = 4 years old). While our main code passes, the CI fails because we cannot build old bitcoind and we waste time figuring old versions.


RCasatta:
I am not against separating the crate, but for your info if you don't already know:

Most of the MSRV breakage was caused by deps needed for the auto download feature which can be disabled if CI provide the bitcoind executable.

@tnull is championing a effort to use an http client lib crate (minreq) that doesn't break MSRV (so that needed MSRV can be respected also with auto download feature)

sanket1729:
> Most of the MSRV breakage was caused by deps needed for the auto download feature which can be disabled if CI provide the bitcoind executable.

Oh nice, I did not know. That would have simplified development for us in past. 

However, I still think we should shift to using this as a separate crate. It would some other unseen issues in future because we only stick to stable. And it would also reduce the compilation time for tests significantly.

apoelstra:
@sanket1729 so, in the last 24 hours serde has broken our compilation. We now need to pin it at 1.0.156.

I can test and ACK this for now but we'll still have to do a followup which adds some extra lines to the README.

apoelstra (approved):
ACK 8a07904949c325e48a7411e88783e5a81b176177

PR #535, opened 2023-03-21
clippy fixes

apoelstra:
Also check minimum dependency versions.


sanket1729:
I think we might need 1.0.156 for serde to work.

sanket1729 (approved):
ACK 34f77df7d0f9ba51811f71fccec87370a3364aac

PR #537, opened 2023-03-24
update to bitcoin 10.0

sanket1729:


tcharding:
PR title has typo mate, 10 should we 30

tcharding:
Hey @sanket1729, I push a branch of the same name to my tree with additional changes to get green in CI (I think). Feel free to grab them and squash it down into this. No need for attribution.

sanket1729:
Thanks, will take a look shortly. Waiting for `bitcoind` release before making any progress here.

apoelstra:
@sanket1729 bitcoind is [blocked on rust-bitcoincore-rpc](https://github.com/RCasatta/bitcoind/pull/116) which is blocked on Steven's availability. Can't we figure out some other solution?

sanket1729:
@apoelstra, is this not already updated here? https://github.com/rust-bitcoin/rust-bitcoincore-rpc/commit/6746011400b12457d341699bc92c8fd8c4ed3d70

sanket1729:
Perhaps, you don't have publish priviledges?

apoelstra:
@sanket1729 correct, I do not have publish privileges.

romanz:
`bitcoind` 0.30 is now [released](https://crates.io/crates/bitcoind/0.30.0) :)

sanket1729:
This is also ready for review @apoelstra

apoelstra:
Looks good. I think we need to revisit the `assume_unchecked` address crap in bitcoin. I'd also like to iterate a bit on the src/util.rs stuff (which has a FIXME and some allocations that shouldn't be necessary) but we can do that in a followup since this PR is so massive and these things are a small part of it.

src/lib.rs
```diff
@@ -733,9 +734,68 @@ fn push_opcode_size(script_size: usize) -> usize {
...
 #[cfg(test)]
-fn hex_script(s: &str) -> bitcoin::Script {
+fn hex_script(s: &str) -> bitcoin::ScriptBuf {
     let v: Vec<u8> = bitcoin::hashes::hex::FromHex::from_hex(s).unwrap();
-    bitcoin::Script::from(v)
+    bitcoin::ScriptBuf::from(v)
+}
+
+/// An absolute locktime that implements `Ord`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
tcharding, line 743: You might want to use
```
#[allow(clippy::derive_ord_xor_partial_ord)]
```
Perhaps with a comment (currently missing from `rust-bitcoin`) that its ok to derive eq and implement ord because the enum variants hold the same u32 as returned by `to_u32()` so the impls match as required.

apoelstra:
There are also several clippy failures on this, but I am also willing to slip those to a followup PR.

src/util.rs
```diff
@@ -16,13 +18,15 @@ pub(crate) fn witness_size(wit: &[Vec<u8>]) -> usize {
...
+pub(crate) fn witness_to_scriptsig(witness: &[Vec<u8>]) -> ScriptBuf {
     let mut b = script::Builder::new();
     for wit in witness {
         if let Ok(n) = script::read_scriptint(wit) {
             b = b.push_int(n);
         } else {
-            b = b.push_slice(wit);
+            // FIXME: There has to be a better way than this.
+            let push = PushBytesBuf::try_from(wit.clone()).expect("FIXME: Handle error");
+            b = b.push_slice(&push)
```
tcharding, line 29: Hey @Kixunil can you take a look at this please. I cannot work out how to get something that implements `AsRef<PushBytes>` from a `&Vec<u8>` to pass to `push_slice`? Only using a reference and checking that the vec is a valid length before passing it to `push_slice`  is exactly what `PushBytes` is for right?

tcharding (commented):
A few minor suggestions, I mainly mention them because I think people will look at miniscript to see how to best use `rust-bitcoin` so we should use it in the best/easiest manner.

Thanks!

apoelstra (approved):
ACK 60943aa48afa8d2d9be3fddbef2b2584b03144ea

src/util.rs
Kixunil: Firstly, I don't understand why this function even exists - witnesses and scripts are different.

The only way is to return `Option` or `Result` from the function or document that it panics if a `witness` with too large element is passed in.

Also to avoid allocation:

```rust
let push: &PushBytes = wit.try_into()?; // or expect
b = b.push_slice(&push);
```

sanket1729: This is an internal to how rust-miniscript works. We have a function that constructs a `Vec<Vec<u8>>` type for all types of scripts. If the script is legacy p2sh or bare script pubkey, we need to convert this to `ScriptBuf` converting each `Vec<u8>` into a scriptPush. 

> The only way is to return Option or Result from the function or document that it panics if a witness with too large element is passed in.

This is one of the selling points of miniscript. Miniscripts are statically analyzed to make sure that all valid spend paths constructed at spend time are within bounds and spendable on bitcoin network. In other words, this function is only called on satisfactions generated from miniscript satisfier which guarantees the resource bounds enforced by bitcoin consensus rules.

It does context based checks that total witness length cannot be more than 3600 for segwitv0 and so on. 

> What KixUnil is saying is correct. 

The allocation should be avoided. I don't remember how I did this :( .

src/lib.rs
sanket1729: Have we added clippy in CI in rust-bitcoin? Perhaps, that gives a green light to add it here too

sanket1729:
Re-reviewed the PR again. And force pushed removing all PushBytesBuf related allocations.

src/util.rs
Kixunil: Would it make sense to pass in something like `&[WitnessElement]` where `WitnessElement` is a type that guarantees length at most 3600? You could then safely implement `AsRef<PushBytes>` for it. Or if it's not that important for miniscript to carry the precise bound just use `&[PushBytesBuf]`.

apoelstra: > Firstly, I don't understand why this function even exists - witnesses and scripts are different.

Prior to segwit they are not.

Kixunil: Prior to segwit witness doesn't exist.

apoelstra (dismissed):
ACK d5a564c6de4df32438bd48c3db6d6b0d95f3b0de

src/util.rs
```diff
@@ -41,6 +45,10 @@ pub(crate) trait MsKeyBuilder {
     where
         Pk: ToPublicKey,
         Ctx: ScriptContext;
+
+    /// All pushes in miniscript script construction will never exceed 2**32 bytes.
+    /// We only push slice of hashes outputs (32 bytes), and keys (33/65 bytes).
+    fn push_ms_slice(self, data: &[u8]) -> Self;
```
Kixunil, line 51: I don't see this being called anywhere.

sanket1729: Good catch. This is a remanent of previous design where I thought I could hide all `unwraps\expects` for miniscript library. But later on decided to have explicit unwraps at call site to avoid future errors.

src/util.rs
sanket1729: We could use `&PushBytesBuf`, since all witnesses we push are either signatures, hash preimages, or constants like 1 or 0.

tcharding (approved):
ACK d5a564c6de4df32438bd48c3db6d6b0d95f3b0de

Can re-ack after `push_ms_slice` is removed but feel free to merge without me :)

apoelstra:
I think we should remove `push_ms_slice`. I'm also happy to re-ack quickly.

apoelstra:
On further thought let's just merge this and delete the funciton in a followup.

sanket1729:
Force pushed. I wonder why it did not show unused warning?

sanket1729:
The trait is pub(crate).

Kixunil:
> I wonder why it did not show unused warning?

I think the analysis is harder for traits. Especially if it's used as `dyn` anywhere.

PR #538, opened 2023-04-04
Various fixes for Nix compatibility

apoelstra:
This brings the fuzztests and bitcoind-tests into workspaces along with the root crate. This means they are covered by `cargo test --all` and `cargo fmt --all` without any special action, and also means that `crate2nix` can test them independently without my manually messing around with directories. I run `cargo +nightly fmt` to bring the two new workspaces into proper formatting -- there were no changes that I'd consider controversial.

It also allows the `bitcoin-tests` directory to be run with a custom `BITCOIND_EXE` variable, and computes the default version more flexibly so that you can run the tests from any directory.

Finally I set the minimum honggfuzz version to 0.5.55, since the dep currently was set to 0.5.0, which definitely did not work.

With these changes, everything in this repo should work with a lockfile generated by `cargo +nightly update -Z minimal-versions`, and I am able to run both the bitcoind and fuzz tests (for five minutes per unit locally). I'll push some code to my [local-nix-ci repo](https://github.com/apoelstra/local-nix-ci) tomorrow to demonstrate whan I'm doing.

This PR does **not** do more dramatic rearrangemeent of the fuzztests ... though when https://github.com/rust-bitcoin/rust-bitcoin/pull/1732 gets merged I will open a similar PR here which does that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,3 +66,7 @@ required-features = ["compiler","std"]
 [[example]]
 name = "psbt_sign_finalize"
 required-features = ["std", "base64"]
+
+[workspace]
+members = ["bitcoind-tests", "fuzz"]
+exclude = ["embedded"]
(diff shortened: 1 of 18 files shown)


sanket1729 (approved):
ACK f50cbbb89b50212a5552a3727241bdb6f2c8ae05.

Issue #430, opened 2022-06-09
Cache Taptree height in the TapTree data structure instead of computing it everytime


sanket1729:
This would involve:
1) Renaming the existing `TapTree` as `TapTreeInner`. 
2) Adding a field `height` in the newly renamed `Taptree`.

```rust
    // TODO: Instead of computing this every time we add a new leaf, we should
    // add height as a separate field in taptree
    fn taptree_height(&self) -> usize {
        match *self {
            TapTree::Tree(ref left_tree, ref right_tree) => {
                1 + max(left_tree.taptree_height(), right_tree.taptree_height())
            }
            TapTree::Leaf(..) => 0,
        }
    }
```

Eunoia1729:
I have a follow-up query:
1. Does a TapTree's structure ever change (i.e. addition or removal of nodes) once instantiated? Looking at the code, my naive understanding is it doesn't.
2. From the TODO comment, its apparent that we want to compute during struct instantiation. But, that would require us creating a module or enforcing instantiation using a method instead of statements such as `TapTree::Tree(left_node, right_node);`. Do we intend to go in this route or did I get it wrong?

sanket1729:
1. It does not change once it is instantiated

2. Yes, we intend to go the same route as you suggested.

Eunoia1729:
@sanket1729 Thanks for clarifying!

ameya-deshmukh:
Hey, I'd like to take this up! Will submit a PR soon

apoelstra:
Thanks @ameya-deshmukh ! Heads up that we will be doing a major release this week and won't be able to take in any major changes until after that.

ameya-deshmukh:
Sure! I'll still submit a PR though, when it gets reviewed and merged is up to you 😄

startup-dreamer:
@apoelstra, @sanket1729  ok i guess the issue is pretty straight can you assign me to implement this

wthrajat:
Hi @sanket1729. I was really looking to work for this issue but seems like this one has already been solved by @tcharding in apparently [this commit](https://github.com/rust-bitcoin/rust-miniscript/commit/476e358f27d97ffa43d1d666f132fef40e9033ad). Also some other commits included the renaming of `taptree_height` -> `height` ([commit](https://github.com/rust-bitcoin/rust-miniscript/commit/ecf876ed6238d0482a5b5cc9d2151eaba44d2911)).
Safe to close the issue now I guess? :)

[...]

PR #539, opened 2023-04-10
Closes rust-bitcoin/rust-miniscript#430 re-named TapTree to TapTreeInner, added height caching

startup-dreamer:
Created TapTree struct to cache tree height, changed the current implementated name to TapTreeInner.


startup-dreamer:
comilation error in **impl_from_tree!** macro "mismatched types expected struct `tr::TapTree`, found enum `TapTreeInner`"

apoelstra:
I think I missed this at the time it was opened, but it has been superceded by #808 and related.

PR #546, opened 2023-05-08
Fix `ForEachKey` impls for policies

### LINKED ISSUE #541: Infinite type-level recursion in `ForEachKey` impl in `Policy`
In [this code](https://github.com/rust-bitcoin/rust-miniscript/blob/master/src/policy/concrete.rs#L666-L686) we have a method `for_each` which takes a `F: FnMut` and calls itself with an `&mut F`. This causes the compiler to try to instantiate the method with `F`, `&mut F`, `&mut &mut F`, etc., forever. Throughout our codebase, we've avoided this infinite recursion by having an internal `real_for_each` method which directly takes a `&mut F`, so that it can call itself with the same type that it was called with. Then our public-facing `for_each` simply sticks a `&mut` on its argument and calls `real_for_each`.

In the code in question, though, we didn't do this, and for some reason the Rust compiler has accepted it, until tonight. The latest nightly, 2023-04-21, does not compile the code.

Unfortunately this code has been present for a long time (I think since 1.0 of this crate) so to fix it, we will need to backport fixes to ever released version. Fortunately, there are a couple potential fixes, both of which are pretty simple:
* Add a `real_for_each` internal funciton like we do elsewhere
* Change `&mut pred` to `&mut |key| pred(key)` in two places (thanks GPT4 for this one)

See https://github.com/rust-lang/rust/issues/110656

apoelstra:
Our `ForEachKey` impls for the two policy types did not work. Both had an infinite type recursion which meant that if anybody had tried to use them they would not have compiled. (Though the fact that this has been happening for at least 2 years with zero bug reports suggests that nobody *has* tried to use them..)

Recent Rust compiler nightlies have started failing compilation even when the offending impls are *not* used, which means we need to prioritize fixing this. Possibly we should even backport it.

Fixes #541

### DIFF
--- a/src/descriptor/bare.rs
+++ b/src/descriptor/bare.rs
@@ -176,10 +176,7 @@ impl_from_str!(
 );
 
 impl<Pk: MiniscriptKey> ForEachKey<Pk> for Bare<Pk> {
-    fn for_each_key<'a, F: FnMut(&'a Pk) -> bool>(&'a self, pred: F) -> bool
-    where
-        Pk: 'a,
-    {
+    fn for_each_key<'a, F: FnMut(&'a Pk) -> bool>(&'a self, pred: F) -> bool {
         self.ms.for_each_key(pred)
     }
 }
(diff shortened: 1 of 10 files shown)


sanket1729 (approved):
ACK f2f95fa62777b0bd96071bba5249d62115c73fbb

PR #557, opened 2023-06-15
Add method to convert expr_raw_pkh into pkh

Harshil-Jani:
The PR adds up the method to convert the expr_raw_pkh into pkh by looking at the available keys and their hases being stored inside of the Hashmap of PSBT Input. If the key is found corresponding to given hash, It will update to pkh.

To-Do : 
- [ ] Write a  test case.
- [ ] Make it work for taproot.


apoelstra:
Nice, concept ACK.

sanket1729 (approved):
utACK 2f3e457fd21a022db2521280542fd74b624ccc9a. Thanks for the contribution

sanket1729:
As noted elsewhere, this does not work for taproot descriptors yet as they are not stored in bip32_keys psbt field. But merging this for now, as this is itself a strict improvement.

PR #560, opened 2023-07-03
cleanup examples

tcharding:
Nothing too crazy here, just a few cleanups to the `examples` directory. Note though that commit `122be5d Move bip-174 test out of examples and into tests` moves a file out of `examples` and into a new `tests` directory.

PR #561, opened 2023-07-12
Remove fuzz from the workspace

tcharding:
I think that the fuzz crate is messing with the 1.41/1.47 builds. There is no need for fuzzing to work with MSRV toolchain so instead of pinning deps from fuzzing just remove the `fuzz` crate from the workspace.

Note chat `cargo check` from the crate root does not check the `fuzz_target` source files like we hoped when adding `fuzz` to the workspace.

I am not entirely sure why `cargo check` does not work but dependencies from `fuzz` are still in the lock file.

Draft to see if this passes CI and to rebase other [failing PRs](https://github.com/rust-bitcoin/rust-miniscript/pull/560).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,5 +69,5 @@ name = "psbt_sign_finalize"
 required-features = ["std", "base64"]
 
 [workspace]
-members = ["bitcoind-tests", "fuzz"]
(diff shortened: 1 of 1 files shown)


tcharding:
Not correct :(

PR #562, opened 2023-07-12
Pin `serde_json`

tcharding:
Pin `serde_json` for 1.41 and 1.47 CI runs. Remove the pinning docs from the README and instead point devs to the CI script. (The docs are stale currently and are an unnecessary maintenance burden.)


apoelstra (approved):
ACK 6f54ed921be62b93ecb4c44fadeff60cac1d85ca

PR #564, opened 2023-07-13
Remove hashbrown dependency

tcharding:
We can use `BTreeSet` instead of `HashSet` and `BTreeMap` instead of `HashMap` to remove the `hashbrown` dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2018"
 [features]
 default = ["std"]
 std = ["bitcoin/std", "bitcoin/secp-recovery"]
-no-std = ["hashbrown", "bitcoin/no-std"]
+no-std = ["bitcoin/no-std"]
 compiler = []
 trace = []
 
@@ -23,7 +23,6 @@ base64 = ["bitcoin/base64"]
 
 [dependencies]
 bitcoin = { version = "0.30.0", default-features = false }
-hashbrown = { version = "0.11", optional = true }
 internals = { package = "bitcoin-private", version = "0.1.0", default_features = false }
 
 # Do NOT use this as a feature! Use the `serde` feature instead.
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -57,7 +57,7 @@ pub use self::key::{
 /// [`Descriptor::parse_descriptor`], since the descriptor will always only contain
 /// public keys. This map allows looking up the corresponding secret key given a
 /// public key from the descriptor.
-pub type KeyMap = HashMap<DescriptorPublicKey, DescriptorSecretKey>;
+pub type KeyMap = BTreeMap<DescriptorPublicKey, DescriptorSecretKey>;
 
 /// Script descriptor
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -656,7 +656,7 @@ impl Descriptor<DescriptorPublicKey> {
             Ok(public_key)
         }
 
-        let mut keymap_pk = KeyMapWrapper(HashMap::new(), secp);
+        let mut keymap_pk = KeyMapWrapper(BTreeMap::new(), secp);
 
         struct KeyMapWrapper<'a, C: secp256k1::Signing>(KeyMap, &'a secp256k1::Secp256k1<C>);
 
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -1501,7 +1501,7 @@ mod tests {
             witness: Witness::default(),
         };
         let satisfier = {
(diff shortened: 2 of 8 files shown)


tcharding:
Bother, will try to fix tomorrow.

tcharding:
Putting on ice until https://github.com/rust-bitcoin/rust-miniscript/pull/569 merges so I don't have to work out the pinning.

apoelstra:
Looking good. The CI failure suggests that this is simply incomplete.

apoelstra:
concept ACK. I actually think the duplicated code is short enough that we could just duplicate it rather than using a macro. Unsure.

tcharding:
So the readability thing is definitely real. On the other hand, and what I tried to assist with the comment 
```
// You can ignore the complexity of this and similar macros below, all they do is implement
// `Satisfier` on both `BTreeMap` and `HashMap` feature guarding the `HashMap` impl on "std"
// feature.
```

is that there is a trade off between the readability of the macro and the readability of the duplicate code. One thing I'm struggling with, as a newcomer to the miniscript codebase, is what code is duplicate and what code is slightly different. There is so much duplicate code that my eyes skim over things thinking they are duplicate and missing the details. I've attempted refactorings only to find that I missed some difference - perhaps its just me but the code duplication seems to make me more lazy at reading.

tcharding:
Would putting this on top of each macro help?
```
/// Implements `Satisfier<Pk> for $map<(Pk, TapLeafHash), taproot::Signature>`.
```

apoelstra:
Ok, this makes sense. And it is true that this library has a ton of duplicate code, some of which is almost-duplicate, and some of which should be duplicate but isn't.

Maybe if we made the macro contents look like `impl Satisfier<Pk> for BTreeMap<(Pk, TapLeafHash), taproot::Signature>;` so it's obvious what the code "should" be doing, that'd be fine with me.

tcharding:
Oh that is kinda nice, hows this?
```rust
macro_rules! impl_satisfier_for_map_key_hash_to_taproot_sig {
    ($(#[cfg($attr:meta)])* impl Satisfier<Pk> for $map:ident<$key:ty, $val:ty>) => {
        impl<Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for $map<$key, $val>
        {
            fn lookup_tap_leaf_script_sig(
                &self,
                key: &Pk,
                h: &TapLeafHash,
            ) -> Option<bitcoin::taproot::Signature> {
                // Unfortunately, there is no way to get a &(a, b) from &a and &b without allocating
                // If we change the signature the of lookup_tap_leaf_script_sig to accept a tuple. We would
                // face the same problem while satisfying PkK.
                // We use this signature to optimize for the psbt common use case.
                self.get(&(key.clone(), *h)).copied()
            }
        }
        
    };
}
impl_satisfier_for_map_key_hash_to_taproot_sig! {
    impl Satisfier<Pk> for BTreeMap<(Pk, TapLeafHash), bitcoin::taproot::Signature>
}
impl_satisfier_for_map_key_hash_to_taproot_sig! {
    #[cfg(feature = "std")]
    impl Satisfier<Pk> for HashMap<(Pk, TapLeafHash), bitcoin::taproot::Signature>
}
```

apoelstra:
Yeah, that looks great to me! This way when you're looking for impls you can still find it via grep, and once you arrive at the macro call, the actual invocation is nearby and almost entirely quoted code.

tcharding:
Cool, I'll hack it up.

sanket1729:
concept ACK. I like the current code, but I don't think we need the pinning commit anymore

tcharding:
I forgot I'd only done one of the macros with the new style. All done now, and squashed into a single commit although this has made the diff a bit hard to view. I can split it all back up if needed.

apoelstra:
The CI failure looks legit (and annoying :))

tcharding:
The problem was I forgot to add the passed in attribute inside the macro call.

apoelstra:
The first commit doesn't compile with `cargo test --no-default-features --features no-std` anymore (and I suspect the latter commit only compiles because `HashMap` is no longer used).

tcharding:
My bad, will investigate. Thanks

tcharding:
Ok, there offender turned out to be one little usage of `HashMap` in a unit test. The PR is now broken up into very small digestible chunks.

apoelstra:
In 986ed867ec3d07f874455e4a80b4c883c6b74b0d

```
cargo test --no-default-features --features='no-std compiler'
```
fails

tcharding:
Thanks, feels a bit like you are my CI again.

apoelstra (approved):
ACK 49fd7b9ad80064de42b2ff2ca63fe750b609d478

PR #566, opened 2023-07-13
Backport recursion fix to 9.x

apoelstra:
Backport of #546. It looks like we only backported this to 7.x, not to 9.x or 8.x.

I'm not sure who is using 8.x. We should probably backport it there too. But 9.x (well, a slight fork of it) is used by elements-miniscript 1.0.

This is needed because Rust released a stable compiler with https://github.com/rust-lang/rust/issues/110475 even though it breaks code that compiled fine for years and there was an open issue for 3 months.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.0.2 - July 13, 2023
+
+- Workaround for recursion bug in rustc https://github.com/rust-bitcoin/rust-miniscript/pull/566
+
 # 9.0.1 - March 8, 2023
 
 - Fixed a typing rule in `multi_a` for taproot miniscript descriptors. Current typing rules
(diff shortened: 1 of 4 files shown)


apoelstra:
CI is failing because of multiple serde MSRV breakages. Locally I can't really test because of path issues with `bitcoind`. But the normal unit tests compile and pass.

sanket1729:
I thought serde was fixed by a #562 :(

apoelstra:
@sanket1729 we are merging into the 9.x branch.

apoelstra:
@sanket1729 what needs to be done to move this forward?

sanket1729 (approved):
ACK f1c12724d04b7ad28d25c866383a948cd493e698. Tested locally

apoelstra:
Tagged and published.

PR #567, opened 2023-07-14
Eliminate a bunch of recursion

apoelstra:
This introduces a new `iter` module with facilities for iterating over tree structures, and uses it to eliminate a bunch of recursion.

There's a lot more "obvious" work after this, so I figured I'd PR this now for review and then do followup ones if it's accepted.

Then there's some less obvious work around getting rid of the recursion in `Expression::from_str` and tightening up the memory allocation/locality there, and some even less obvious work in improving memory allocation/locality in `Miniscript`. But even if those later stages don't work out, this is a step in the right direction because it eliminates a bunch of stack overflow vectors, and reduces the amount of code.


sanket1729 (approved):
Some commits are empty? Is that intentional. 

ACK 0576007de27971a57e0d61fc5446e0aa34a7d549.

sanket1729:
Agreed, this is a step in the right direction.

sanket1729:
I can merge these and address the nits in a follow up PR. Just waiting for your response about the empty commits.

apoelstra:
@sanket1729 no, I'll remove the empty commits. I believe those occur during rebasing when my gpg agent fails to sign a commit.

I'll do that and then you can merge this. For optimizations I think we should first introduce a wider set of benchmarks (miniscript parsing from string, decoding from script, serializing to string, encoding to script, then some borderline-trivial key conversions etc).

sanket1729 (approved):
reACK 67e91b9115031c7f773b67314370685e0c5d2f03

src/iter/mod.rs
```diff
@@ -0,0 +1,93 @@
+// Written in 2023 by Andrew Poelstra <apoelstra@wpsoftware.net>
```
tcharding, line 1: Hey @apoelstra, I saw this line and wondered if you added it as a habit or if I have been annoying you by removing them? Or was this PR just before we discussed all that? Cheers.

apoelstra: I think it predates your PR to remove these. I would've just mildlessly copied the line from other files.

I'm certainly not annoyed :).

tcharding: Sweet man, cheers.

PR #568, opened 2023-07-16
psbt: Rewrite input replacement to avoid forgetting fields

stevenroose:
The way it was written seemed very prone to forgetting fields if new ones would be added. Like this it just keeps the required fields and all the rest is set to `Default::default()`.


sanket1729 (approved):
utACK c50bdf8681b0bf188bbc7f1524e7251ac8e07525

apoelstra (approved):
ACK c50bdf8681b0bf188bbc7f1524e7251ac8e07525

PR #569, opened 2023-07-18
Update MSRV to 1.48. Fixes CI

sanket1729:


apoelstra:
Can you update the fuzzer rustc to 1.64 from 1.58, and also update `generate-files.sh` to reflect the `regex = "=1.4.0"` thing (if that is still necessary)?

sanket1729:
Nice. Found a new fuzz failure

tcharding (commented):
LGTM, glad to see you worked out the pinning, I tried and failed yesterday. (FWIW I used LGTM instead of ACK because I'm assuming you are going to squash the `fix!` commits.)

sanket1729:
After a lot of iteration on fuzzer, this is finally ready for review.

.gitconfig
```diff
@@ -0,0 +1,3 @@
+[remote "upstream"]
+        fetch = +refs/pull/*:refs/remotes/pr/*
+        fetch = +refs/merge-requests/*:refs/remotes/pr/*
```
apoelstra: In 3ecd2fbccc64bc707a9f418c93e019192e9113d9:

You should not have committed this file

apoelstra: it's deleted in the next commit, but could still cause weird things to happen during bisection.

src/policy/compiler.rs
```diff
@@ -1048,6 +1056,7 @@ where
 /// Helper function to compile different types of binary fragments.
 /// `sat_prob` and `dissat_prob` represent the sat and dissat probabilities of
 /// root or. `weights` represent the odds for taking each sub branch
+#[allow(clippy::too_many_arguments)]
```
apoelstra, line 1059: In e491308b6075d5274917750ad7258cb604b99aea:

I think this is fine, but FYI both this and the "too many arguments" lint thresholds can be set globally in clippy.toml.

apoelstra (dismissed):
ACK 13950fd16161afd9b28345c68192aff997048231

apoelstra:
Oh, actually, I'd like you to remove the spurious file before merging.

.gitconfig
sanket1729: Removed

sanket1729:
Added another commit that adds a static vector

apoelstra (approved):
ACK 60fde9e9519125d3b1ecccea302a0a6e6ce484fa

apoelstra (approved):
ACK 6b76997b14fd606b6ceba884012512c06200a4a4

PR #570, opened 2023-07-23
Expose inner `DescriptorPublicKey` functionality

quad:
Hi friends, it's me again! 👋 

In our use of miniscript, we find ourselves writing _constantly_ writing this idiom:

```rust
let xpub = match dpub {
    DescriptorPublicKey::Single(_) => unimplemented!(),
    DescriptorPublicKey::XPub(xpub) => xpub,
}
```

I'd *love* to directly use the `DescriptorXKey<bip32::ExtendedPubKey>` type; but, its functionality is anemic when compared to all the useful functions on `DescriptorPublicKey`.

The closer I looked at `DescriptorPublicKey`, the more I realised it was _mostly_ `match` blocks itself that delegated to different implementations based on the internal enum type. And, so, here's a PR that leans into that and reifies a trait (`DescriptorInnerKey`) that encapsulates all that behaviour and three types that implement it:

1. `DescriptorSinglePublicKey`
1. `DescriptorExtendedPublicKey`
1. `DescriptorMultiExtendedPublicKey`
---
This all said, I feel like this PR is still lacking. A few questions:

1. `DescriptorPublicKey` _could_ implement the trait too; but, I suspect that be a breaking change? Moreover …
1. `DescriptorInnerKey` is a weird name, especially when contrasted against the already existing `InnerXKey` trait. I'm open to a better name!


src/descriptor/key.rs
```diff
@@ -283,39 +287,51 @@ impl error::Error for DescriptorKeyParseError {
...
+        maybe_fmt_master_id(f, &self.origin)?;
+        self.xkey.fmt(f)?;
+        fmt_derivation_path(f, &self.derivation_path)?;
+        match self.wildcard {
+            Wildcard::None => {}
+            Wildcard::Unhardened => write!(f, "/*")?,
+            Wildcard::Hardened => write!(f, "/*h")?,
+        }
+        Ok(())
+    }
```
quad, line 312: Good candidate to extract to a `fmt_xkey(origin, [paths])` function.

src/descriptor/key.rs
```diff
@@ -525,42 +541,167 @@ impl error::Error for ConversionError {
...
+                    .ok()
+                    .ok_or(ConversionError::HardenedChild)?,
+            ),
+            Wildcard::Hardened => self.derivation_path.into_child(
+                bip32::ChildNumber::from_hardened_idx(index)
+                    .ok()
+                    .ok_or(ConversionError::HardenedChild)?,
+            ),
+        };
+        let definite = DescriptorPublicKey::XPub(DescriptorXKey {
```
quad, line 656: This is an awkward "upward" instantiation of the `DescriptorPublicKey` type. An idea:

1. `DefiniteDescriptorKey::new` could take a `DescriptorInnerKey`
2. `DescriptorInnerKey` extracts and implements `derive_public_key`, which is also a `match` block across the inner enum.

src/descriptor/key.rs
```diff
@@ -525,6 +525,53 @@ impl error::Error for ConversionError {
     }
 }
 
+pub trait DescriptorInnerKey {
```
apoelstra, line 581: In e4d9af484b996d05db261be7406d2af78488e527:

Do you think we should add `fmt::Display` and `std::str::FromStr` bounds on this? How about `Eq`, `Hash`, etc.? (My vote is yes to all four, and maybe `fmt::Debug` too.)

cc @sanket1729 who may have opinions here.

src/descriptor/key.rs
```diff
@@ -525,6 +525,53 @@ impl error::Error for ConversionError {
...
+    /// The fingerprint of the master key associated with this key, `0x00000000` if none.
+    fn master_fingerprint(&self) -> bip32::Fingerprint;
```
apoelstra, line 583: In e4d9af484b996d05db261be7406d2af78488e527:

As a separate issue, I think we should return an option here and then upstream we should have a `bip32::Fingerprint::from_opt` function that would do the 0x00000000 thing. Though I'm unsure. Maybe this function is only ever used in contexts where the result needs to be serialized (PSBT, encoding xpubs) in which case users always want the existng behavior.

src/descriptor/key.rs
apoelstra, line 581: In e4d9af484b996d05db261be7406d2af78488e527:

This can go in a separate PR, but like to have a type-level relationship between `DescriptorPublicKey` and `DescriptorSecretKey`. This would help me downstream in elements-miniscript where I want to define a "view key" for a confidential transaction descriptor, which has a private key in one slot and a public key in every other slot. I think it would also simplify a lot of usage of xprv descriptors, which currently require you maintain an xpub descriptor alongside a hashmap, though I haven't worked out the details.

One way we could do this is to add two associated types, like
```
type SecretKey: DescriptorInnerKey;
type PublicKey: DescriptorInnerKey;
```
where almost always one of these would be `Self`.

I would also like first-class support for unspendable keys in taproot. I am imagining an `UnspendableKey` whose `SecretKey` type would be `!`, or something.

Anyway these are just half-baked ideas that don't belong in this PR. But I want to bring them up in case they could guide this API.

src/descriptor/key.rs
apoelstra: Maybe `DescriptorInnerKey` should have an `Into<DescriptorPublicKey>` bound?

apoelstra:
Overall 14799e2dad3bf221e173684c19ce58e6170c02d8 looks good. It's a straightforward enough diff and I am confirming that all commits pass all tests (looking good so far).

> * `DescriptorPublicKey` _could_ implement the trait too; but, I suspect that be a breaking change? Moreover …

I don't think it would be breaking, but I also don't know that we need to worry about breaking changes here. Especially if we can also figure out the "what is the right way to relate private and public keys" and "what is the right way to handle unspendable keys" API questions.

But yes, it did occur to me when reading the code that it might make sense to implement this on `DescriptorPublicKey`.

> * `DescriptorInnerKey` is a weird name, especially when contrasted against the already existing `InnerXKey` trait. I'm open to a better name!

Maybe just `DescriptorKey`? We could maybe have a nice API where users could have a `Descriptor<Pk: MiniscriptKey>` but then get a ton of extra functionality if they actually had a `Descriptor<Pk: DescriptorKey>`.

> * `DescriptorPublicKey::into_single_keys` has been awkwardly left out of the extraction because it returns a vector of `DescriptorPublicKey`s. If I was going to do a breaking refactor, I'd probably move that method on to `DescriptorMultiExtendedPublicKey` itself. shrug

Good to know. Let's revisit that if we decide to do a breaking change.

> * What would going _full breaking change_ look like if `DescriptorPublicKey` as a `struct` just _disappeared_ … and instead there was only a `DescriptorPublicKey` trait?

So, with `Descriptor` itself we actually went the opposite direction. We originally had a trait because we imagined that downstream users might define their own descriptor types (in particular we had CTV and Elements in mind). But then we realized that it almost never made sense to be generic over an arbitrary descriptor, and using a trait was extra noise. See #386.

But having said that, for **keys** we have continuously moved toward more trait usage, adding more traits and associated types etc.

src/descriptor/key.rs
quad: 🧠 💡

src/descriptor/key.rs
quad: That's sure where I'm using it. 😅

src/descriptor/key.rs
quad: I added `Clone + fmt::Debug + fmt::Display + Eq + FromStr + std::hash::Hash + Ord` to match the existing derivations.

src/descriptor/key.rs
quad: Erm, that'll require ~`DescriptorInnerKey`~`DescriptorKey` to be made object-safe. That's beyond my ken.

quad:
> Maybe just `DescriptorKey`?

👍 

> But yes, it did occur to me when reading the code that it might make sense to implement this on `DescriptorPublicKey`.

I added a commit that does that. It's "breaking" insofar as consumers now need to import a trait. (e.g. see the change in `src/descriptor/mod.rs`)

src/descriptor/key.rs
apoelstra: I think it might be as simple as sticking a `: Sized` bound on the trait? What error do you get?

sanket1729:
Hello, thanks for this PR. 

Am I correct in summarizing that if we have a `type XPubOnlyDesc = Descriptor<DescriptorExtendedPublicKey>`, than it should satisfy your use-cases? Someone else also offered me the same feedback that most wallets today only use XPubs(without the multiX) and we should simplify our API to that end. 

 If this the case, we provide the same APIs for ` XPubOnlyDesc` as we do `Descriptor<DescriptorPubicKey>`. Our code is mostly generic over `MiniscriptKey`, so there are only a few APIs that are specialized for `DescriptorPublicKey`.  

We can then offer some granular types with some guidelines as 
```
// lib.rs
pub type Descriptor = Descriptor<DescriptorPublicKey>; // Full support for all types of keys
pub type DescriptorExtendedKeysOnly = Descriptor<DescriptorExtendedPublicKey>; // Without single or multi-keys.
```
Tihs obviously implies that you cannot single keys in your descriptor. Curious what @apoelstra thinks about this. 

### Comment about the trait approach

I am not sure about adding a trait, 1) It requires users to import it and 2) we are not using it any bounds here. Are you planning it downstream? If so, can you share you want it to be used as a bound?
If it is just a way to bundle "similar looking" APIs together, I am not sure about it. 

As Andrew mentioned in the #386, we found ourselves into a similar issue where we group things for `DescriptorTrait`  that were almost similar, but not exactly the same. I think is also one of the cases where we are intentionally merging things. I think this is evident by the fact that we already have weird cases around multi, bip32 master fingerprint. 

I do think the trait has value only if we go all the way so that `Descriptor<DescriptorExtendedPublicKey>` can be used directly. (Which would require implementing MiniscriptKey on `DescriptorExtendedPublicKey`). Otherwise, with respect to this PR as is, we can provide all the functionality without the trait. As in, SinglePub would have a method called `master_fingerprint()`, `into_single_descriptors` would only be for `DescriptorMulti..` and so on for other methods that make sense for each type. The top level enum would have all the types with the correct `Option`, `Result` return types for respective APIs.

cc @tcharding worked on changing the previous long names that we had `DescirptorSinglePublicKey` to `SinglePub`, would like to get his input on naming here.

quad:
> Am I correct in summarizing that if we have a `type XPubOnlyDesc = Descriptor<DescriptorExtendedPublicKey>`, than it should satisfy your use-cases?

Maybe? Honestly, the type hierarchy is too complex for me to follow.

That said, there's a lot of useful functionality on `DescriptorPublicKey`. `DescriptorPublicKey::from_str` being paramount!

> Someone else also offered me the same feedback that most wallets today only use XPubs(without the multiX) and we should simplify our API to that end.

I strongly agree with this sentiment.

quad:
> Are you planning it downstream? If so, can you share you want it to be used as a bound?

Our codebase has `type DescriptorExtendedKey = DescriptorXKey<ExtendedPubKey>` and almost entirely standardises on that type internally. We match / unwrap `DescriptorPublicKey` at the edges. I'd _love_ to use `DescriptorExtendedKey` at our edges, but its too aenemic right now to justify it.

sanket1729:
>  Maybe? Honestly, the type hierarchy is too complex for me to follow.

I guess put simply, if all functionality for `DescriptorPublicKey` is available in `DescriptorExtendedKey`, does that satisfy your use-cases?

src/descriptor/key.rs
```diff
@@ -18,15 +18,19 @@ use crate::prelude::*;
 use crate::serde::{Deserialize, Deserializer, Serialize, Serializer};
 use crate::{hash256, MiniscriptKey, ToPublicKey};
 
+type DescriptorSinglePublicKey = SinglePub;
+type DescriptorExtendedPublicKey = DescriptorXKey<bip32::ExtendedPubKey>;
+type DescriptorMultiExtendedPublicKey = DescriptorMultiXKey<bip32::ExtendedPubKey>;
```
tcharding, line 23: IMO we should not start putting `Extended` back into the identifiers of keys. The only other place that appears is in `bitcoin::bip32` which is not a module to copy :)

tcharding:
> cc @tcharding worked on changing the previous long names that we had DescirptorSinglePublicKey to SinglePub, would like to get his input on naming here.

In general I prefer shorter names if there is no ambiguity or loss of clarity. Thanks.

sanket1729:
@quad. I was suggesting something like: https://github.com/rust-bitcoin/rust-miniscript/pull/586 without all the forced mixing of public traits involved here.

PR #572, opened 2023-07-25
Reduce line count

tcharding:
This PR is formatting changes only, totally non-urgent.

Reduce line count by setting a couple of rustfmt config options to be the same as they are in `rust-bitcoin`. Final patch removes whitespace in trait impl blocks, and was the impetus for this PR.


apoelstra:
I think increasing `chain_width` is mostly a bad change.

apoelstra:
But 0c7a972bc3693f7cc0a6fed108ad11f2b8afdbb6 looks good otherwise.

tcharding:
So, assuming the comment I wrote in the rustfmt config file of rust-bitcoin is correct, we can't keep `chain_width` with a custom setting and not all the others if we want to use `use_small_heuristics = "Max"` - is the chain width annoying enough to warrant dropping this whole patch? Real question, this is your and @sanket1729's call, I'm jonny come lately to this crate.

tcharding:
I split each config option change into a separate patch so they can be viewed easily. I actually think the only massive win is `fn_single_line` and maybe `struct_lit_width`.

tcharding:
FTR docs say the values are a percentage of `max_width` but anything over 80 causes an error, so there is a bug somewhere.

`fn_call_width` cannot have a value that exceeds `max_width`. `fn_call_width` will be set to the same value as `max_width`

tcharding:
> Not really a fan of the array_width change either.

No worries, I'll drop that commit - that is exactly why I kept them all separate :)

tcharding:
Force push removes the `array_width` change (was the last patch of the series).

tcharding:
> Otherwise https://github.com/rust-bitcoin/rust-miniscript/commit/75d3fa8da07c4ae62f6194c3078f529c924f479b looks good.

This is the commit hash of the removed patch.

apoelstra (approved):
ACK 1cab0b872854a00bc7bfb2dc52a47a81350240fc

tcharding:
Rebased, no other changes.

apoelstra (dismissed):
ACK f41a70db0f403ac047b0e76f6585c05854bfaa60

tcharding:
Re-done after `planning` module merging. @TheBlueMatt this one should make you all warm and fuzzy inside :)

sanket1729:
@tcharding, what order do you prefer I merge the PRs?

apoelstra (approved):
ACK 26bb52e2037a9d603264943770e7f5c31be7bddd

tcharding:
This one first please, it will cause loads of conflicts but that is always going to be the case whenever it merges.

PR #573, opened 2023-07-25
Make common generics more terse 

tcharding:
This is #375 brought back to life, I don't have perms to re-open PRs. Includes all things from the discussion thread. The individual patches are marked wip and have foramatting changes separate, this is so reviewers can more easily see how the changes effect formatting. Also this PR is hella invasive, it can sit here until its a good time to merge then I can redo the whole thing and squash it into a single patch.

I will in no way take offence if this gets nack'ed for being too invasive :)


apoelstra:
sgtm. See my comments in the other PR, which match my current opinions.

tcharding:
Awesome, thanks. How about `ScriptContext` to `Context` which I only mention in discussion on the other PR?

apoelstra:
Yep, I think `ScriptContext` to `Context` is fine.

tcharding:
Cool, just to make sure, would either of these be better?
- `fn foo<Pk: Key, C: Context>();` 
- `fn foo<Pk: Key, C: Ctx>();`

apoelstra:
I think `Context` is better than `Ctx`.

tcharding:
Cool, I'm going to go with `<Pk: Key, Ctx: Context>`.

tcharding:
Patch one is a bit of a punish to review but it can be verified by doing the changes yourself and running the formatter
- `MiniscriptKey` becomes `Key`
- `ScriptContext` becomes `Context`
- Run `cargo +nightly fmt`

No other manual changes in patch 1

apoelstra:
I think in the second commit we should also re-export `ContextError` as `ScriptContextError`, since that's the only other breaking change.

apoelstra (dismissed):
ACK 2c8ca201b899d894ae1fb904c7d6cb7b804a1c1e

apoelstra:
Will hold off on merging in case you want to add the extra re-export.

tcharding:
> I think in the second commit we should also re-export ContextError as ScriptContextError, since that's the only other breaking change.

Oooo, to be honest I did not actually notice that change this time around when I redid this work. I'll add the re-export as suggested. Furthermore, do we want to use `context::Error` instead while we are touching it? (FTR I have been explicitly ignoring errors and docs in miniscript just recently, errors because of the stuff we are doing in secp/bitcoin and docs because I'll get to them later.)

tcharding:
I re-ordered the patches to put the alias stuff last. Added `ScriptContextError` alias. As is the patch maintains backwards compatibility - I'm not convinced the current re-exports are perfectly ergonomic.

tcharding:
The errors are a bit of a mess, changing to use `context::Error` is opening a can of worms. I'm going to leave it as `ContextError` for now.

tcharding:
Rebased, no other changes.

sanket1729:
Hey Tobin, sorry for the delayed response. I am waiting a few days to merge other PRs to see if we can #481 in first.

tcharding:
Oh definitely! don't make them rebase that PR. Thanks for paying attention.

apoelstra (approved):
ACK 24c11c09f66e213306ff3569e2b07adfe1bf3221

tcharding:
One invasive clean up at a time is plenty, leaving this util there is an opportune moment to push it.

PR #574, opened 2023-07-25
Implement From<Foo> for Descriptor<Pk> using macro

tcharding:
Add a macro and remove all the individual from impls.

Refactor only, no logic changes.

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -76,47 +76,18 @@ pub enum Descriptor<Pk: MiniscriptKey> {
     Tr(Tr<Pk>),
 }
 
-impl<Pk: MiniscriptKey> From<Bare<Pk>> for Descriptor<Pk> {
-    #[inline]
-    fn from(inner: Bare<Pk>) -> Self {
-        Descriptor::Bare(inner)
-    }
-}
+macro_rules! impl_from_for_descriptor {
+    ($($from:ident),*) => {
+        $(
+            impl<Pk: MiniscriptKey> From<$from<Pk>> for Descriptor<Pk> {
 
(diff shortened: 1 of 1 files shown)


apoelstra:
I wish there was a way that you could expand macros prior to committing them. So we could have a CI job that checked that all the code was actually produced from macros, but when reading it, you could read the expanded versions.

In C you can kinda do this by running `cpp` directly.

In Rust I guess we could also use `cpp` but that's maybe too evil :).

But basically, I like the guaranteed consistency and ease of writing macros but I really think they harm readability.

tcharding:
> but I really think they harm readability

The first Rust job I had they barred me from writing macros because of this.

Perhaps we can find a middle ground, if the macro was called in a form more like what the code looks like it would be easier to read. It may even be easier to read because it reduces the noise.

Something like this?
```rust
// Implements From<error::Bar> for GeneralError converting to `GeneralError::Bar(error::Bar)`
impl_from_for_error_type!(error::Bar, GeneralError::Bar);
```

apoelstra:
Can you use `for` instead of `is`? Or does Rust prevent this?

tcharding:
We can put anything we want in there I believe, I did not code the actual macro up though. (I also had `for` at one stage when editing the original post)

tcharding:
oh that's right, I removed `for` because its a bit weird to have for and then include the variant name.

apoelstra:
Ah, I see. So even in your hypotheticals I misread what was going on ... I think the explicit code is probably better than anything we've got here.

One thing we could experiment with is adding `#[from]` the way that `thiserror` does. We've shied away from syntax extensions in the past but possibly this one is easy to do. Unsure. It would probably depend on `quote` which we've had MSRV trouble with in the past.

tcharding:
I"ll leave this for now.

PR #575, opened 2023-07-25
Add call_through macro

tcharding:
Draft to see what you guys think of this. It could be used all over the place but its non trivial because loads of the call through functions differ in return type, mostly erroring vs infallible.

Add a macro that calls through to a function for each variant of an enum.

Add a single example usage of the new macro.

### DIFF
--- a/src/descriptor/bare.rs
+++ b/src/descriptor/bare.rs
@@ -223,6 +223,11 @@ impl<Pk: MiniscriptKey> Pkh<Pk> {
         self.pk
     }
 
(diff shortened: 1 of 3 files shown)


apoelstra:
Weak concept NACK. I think it makes the code too much less readable.

tcharding:
No stress, cheers. There are going to be a ton of "clean up" ideas coming over the next few days hopefully on my way to the stuff we talked about, please nack away merrily whenever you think they are not real improvements.

PR #576, opened 2023-07-25
Remove the `Liftable` trait

tcharding:
Draft to showcase this idea, I'm not sure how to write the function docs just yet.

This trait is never used as a trait bound, there is no obvious reason for its existence.

Remove the `Liftable` trait and implement `lift` on each type directly.


apoelstra:
This trait is useful when you are taking an arbitrary miniscript-shaped object and need to do semantic analysis on it.

Though arguably you could just directly take a `semantic::Policy`, since the conversion is always kinda expensive.

tcharding:
Cool, thanks for the explanation.

PR #583, opened 2023-07-27
Clean up `miniscript/mod.rs`

tcharding:
Make an effort to clean up the `miniscript/mod.rs` file. I could not see a reason for all the separate impl blocks, is there one? Please note this PR does not fix all the docs, just does ones that should not be too hard to review.

### DIFF
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -64,8 +64,8 @@ pub struct Miniscript<Pk: MiniscriptKey, Ctx: ScriptContext> {
 }
 
 /// `PartialOrd` of `Miniscript` must depend only on node and not the type information.
-/// The type information and extra_properties can be deterministically determined
-/// by the ast.
+///
+/// The type information and extra properties are implied by the AST.
 impl<Pk: MiniscriptKey, Ctx: ScriptContext> PartialOrd for Miniscript<Pk, Ctx> {
     fn partial_cmp(&self, other: &Miniscript<Pk, Ctx>) -> Option<cmp::Ordering> {
         Some(self.node.cmp(&other.node))
@@ -73,8 +73,8 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> PartialOrd for Miniscript<Pk, Ctx> {
 }
 
 /// `Ord` of `Miniscript` must depend only on node and not the type information.
-/// The type information and extra_properties can be deterministically determined
-/// by the ast.
+///
+/// The type information and extra properties are implied by the AST.
 impl<Pk: MiniscriptKey, Ctx: ScriptContext> Ord for Miniscript<Pk, Ctx> {
     fn cmp(&self, other: &Miniscript<Pk, Ctx>) -> cmp::Ordering {
         self.node.cmp(&other.node)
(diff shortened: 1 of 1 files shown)


src/miniscript/mod.rs
```diff
@@ -59,7 +59,7 @@ pub struct Miniscript<Pk: MiniscriptKey, Ctx: ScriptContext> {
     pub ty: types::Type,
     ///Additional information helpful for extra analysis.
     pub ext: types::extra_props::ExtData,
-    /// Context PhantomData. Only accessible inside this crate
+    /// Needed because not all terminals are paramatized by script context.
```
apoelstra, line 62: In c874074b4a6dc6c7e47be1a8945b0a6aa015e437

This is actually just used to simulate `#[non_exhaustive]` which we did not have in 1.29. We can just drop it and add `#[non_exhaustive]`.

src/miniscript/mod.rs
```diff
@@ -64,37 +64,40 @@ pub struct Miniscript<Pk: MiniscriptKey, Ctx: ScriptContext> {
 }
 
 /// `PartialOrd` of `Miniscript` must depend only on node and not the type information.
-/// The type information and extra_properties can be deterministically determined
-/// by the ast.
+///
+/// The type information and extra_properties can be deterministically determined by the AST.
```
apoelstra: In 33c04e24d8e9bbf7265d5d36e3b1ec53f2c26762:

While we're touching this, let's replace "can be deterministically determined" with "are implied" :P.

apoelstra:
5182529637632e8f78edaff812a86b23a9cfdeff looks good otherwise

src/miniscript/mod.rs
tcharding: Did you think this was an enum? I think the comment is correct IIUC, if we remove the phantom we get build error
```
error[E0392]: parameter `Ctx` is never used
  --> src/miniscript/mod.rs:55:42
   |
55 | pub struct Miniscript<Pk: MiniscriptKey, Ctx: ScriptContext> {
   |                                          ^^^ unused parameter
   |
   = help: consider removing `Ctx`, referring to it in a field, or using a marker such as `PhantomData`
```

tcharding:
Changes in force push:
- Fixed docs as suggested
- Added a couple of lines of whitespace that were missing from the original refactorings
- Fixed formatting in the last patch

tcharding:
... and fixed the rebase

tcharding:
And rebased on master to pick up the pinning fix.

src/miniscript/mod.rs
apoelstra: Oh, no, I think actually I'm just wrong. But I don't understand your comment. `Ctx` is clearly used in the `node` field, this error is just a bug in rustc (which I think has been well-known for many years). We should change the comment to clarify this.

And it *is* doing double-duty as a `#[non_exhaustive]` tag, which is fine. It prevents users from directly constructing this type.

apoelstra:
55e65f3987a348baa98fe34b94d71bdf7fdab6ff is good except I'd like to change the comment on the phantom.

src/miniscript/mod.rs
tcharding: Oh I thought it was being caused by the fact that not all variants of `Terminal` have `Ctx`. Not sure what to write there now, the original is definitely just noise though.

tcharding:
I just dropped the phantom doc change altogether.

apoelstra (approved):
ACK a60c642a54d685450aa9e117d854c624108ff7e5

PR #584, opened 2023-07-30
reduce binary bloat by removing generic param from type_check

benma:
The binary sizes are increased a lot by the use of generics. In one case, there is a lot of needless bloat by using a generic type param for the callback argument in the type_check function of the Property trait.

    cargo build --example parse --release && du -sb target/release/examples/parse

Results in a binary size of 1784152 bytes.

With this commit, the binary size is decreased by 12288 bytes, which is significant for embedded/no_std use.

The sizes above were achieved by adding this to Cargo.toml:

```
[profile.release]
strip = true
panic = "abort"
opt-level = 's'
codegen-units = 1
lto = true
```

Though the difference is roughly the same also without the optimizations.

As an additional benefit, code clarity is improved as the child arg was unused in some of the trait implementations.

### DIFF
--- a/src/miniscript/astelem.rs
+++ b/src/miniscript/astelem.rs
@@ -47,7 +47,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Terminal<Pk, Ctx> {
 impl<Pk: MiniscriptKey, Ctx: ScriptContext> fmt::Debug for Terminal<Pk, Ctx> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.write_str("[")?;
-        if let Ok(type_map) = types::Type::type_check(self, |_| None) {
+        if let Ok(type_map) = types::Type::type_check(self) {
             f.write_str(match type_map.corr.base {
                 types::Base::B => "B",
                 types::Base::K => "K",
--- a/src/miniscript/decode.rs
+++ b/src/miniscript/decode.rs
@@ -25,10 +25,6 @@ use crate::prelude::*;
 use crate::Descriptor;
 use crate::{bitcoin, hash256, AbsLockTime, Error, Miniscript, MiniscriptKey, ToPublicKey};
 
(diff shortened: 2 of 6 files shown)


sanket1729:
Thanks for the PR and bringing up the issue of binary sizes of rust-miniscript. I might take this up as some weekend project in the near future.

sanket1729:
There are a lot of things we can do in the longer term to help this. We can split out the crates into multiple separate crates like `psbt`, `interpreter`, `descriptor` and `policy`. This would reduce the compilation time and memory usage by offering flexibility. 

As pointed by @benma , this is because we use a lot of generics. In particular, almost every single function is being monomorphized for each `Ctx`. 
There are 4 types of Contexts (Bare, Legacyp2sh, Segwitv0, Taproot) that are compiled for each descriptor method. One idea might be to define `Terminal` without the Context. Guard `Miniscript` with appropriate Context while in the constructor `Miniscript::from_ast`. This should significantly reduce the code duplication and have a good impact on binary sizes.

apoelstra:
This PR looks like a good start reducing compile sizes. It also gets rid of that awful `return_none` function.

apoelstra:
>please make sure this change does not break anything

Well, I'm pretty sure it *is* a breaking API change, but that's okay, most users are not directly calling type inference functions or implementing type inference traits.

>I am not sure if the unit tests cover everything

I'm confident that the unit tests sufficiently cover both the `None` and `Some` cases of the old API, so if they continue to pass then I think we're good. I also ran all the fuzztests for 5 minutes each. The only thing I'm less confident about is whether the `unreachable` clause you added is actually unreachable (unless somebody goes and adds a new explicit call :)). But I think it is.

benma:
Reminder to merge. Let me know if anything is missing that I can provide. Thanks.

apoelstra:
@sanket1729 i'm good to merge this if you are.

sanket1729 (approved):
ACK 0c411db3fbf950cedccf113e634bfb8e1b89c314

PR #586, opened 2023-07-31
Add support for XpubOnly keys

sanket1729:
This allows users to directly operate on extended keys without always having to match on `DescriptorPublicKey` enum. This is quick hack PR, just looking for superficial feedback if this addresses issues by @quad. 

 All the APis that were supported on `DescirptorPublicKey`are now also exported via `XPubOnlyDescriptor` (I am not happy names I chose in the PR).
 
 Overall, 
  - [] There is some code duplication in the last commit that I want to address later. 
  - [] I am not happy with the names for exported descriiptors in lib.rs.

### DIFF
--- a/src/descriptor/descriptor_impl.rs
+++ b/src/descriptor/descriptor_impl.rs
@@ -2,6 +2,7 @@
 //! Contains common APIs for specific types of keys like `DescriptorPublicKey`,
 //! `DefinitePublicKey` and more.
 
+use core::convert::Infallible;
 use core::ops::Range;
 use core::str::{self, FromStr};
 
@@ -13,7 +14,7 @@ use crate::descriptor::DescriptorSecretKey;
 use crate::prelude::*;
 use crate::{
     hash256, DefiniteDescriptorKey, Descriptor, DescriptorPublicKey, Error, ForEachKey,
-    TranslatePk, Translator,
+    MsDescriptor, MsDescriptorXPubOnly, TranslatePk, Translator, XPubOnly,
 };
 
 /// Alias type for a map of public key to secret key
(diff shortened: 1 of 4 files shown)


quad:
… I guess this works? It adds yet another type. It'll mean that for now and into the future the methods need to be manually copied across. 🤷

sanket1729:
> It adds yet another type. 

My understanding is that most wallet developers will only use this type `MsDescriptorXPubOnly`. That way they get to use all the descriptor APIs (not only `DescriptorPublicKey` ones, but all `MiniscriptKey` ones) without having to deal with unwrap/err cases for Multi or SinglePub. 

> … I guess this works? It adds yet another type. It'll mean that for now and into the future the methods need to be manually copied across. shrug

I understand that replicating methods instead of implementing a unifying trait might seem contrived. We've succumbed to the allure of grouping similar-looking elements within traits. However, amalgamating two distinct entities to alter the trait signature into a Result would compel users to manage errors in methods that consistently yield Ok(). Alternatively, this approach might result in methods exclusively returning errors.

W.r.t to forgetting things, the `DescriptorPublicKey` enum contains the XPub, so we won't forget to add methods if add methods to `DescriptorPublicKey` type. 

Perhaps, @danielabrozzoni also has some comments for this one.

danielabrozzoni:
I'm not confident enough with the project structure to give suggestions about how to implement this, but I think that having a xpub only descriptor would be really useful for most users - but I'm not sure how useful this would be for bdk, as I don't think we'll ever drop support for single key descriptors.
We actually have a quite similar situation in bdk with multixpub descriptors, as at the moment we just deny them (https://github.com/bitcoindevkit/bdk/pull/1023/commits/792b39fa92bfa1218a33bd6ded257fc105aec67c), and in the future we plan to support multixpub by just translating them to a vector of xpub descriptors, as it seems easier to implement on our side, and seems to fit better the bdk architecture.

PR #587, opened 2023-08-07
 Inline descriptor inner types

tcharding:
Patch 1 is trivial cleanup

From commit log of patch 2
    
    The `ShInner` and `WshInner` enum's provide no benefit that I can
    discern. Both include only tuple variants with private tuple fields. We
    can therefore inline the enums into `Sh` and `Wsh` respectively with no
    loss of type safety. Doing so marginally simplifies the code, the
    biggest benefit is it stops devs wondering why there is an inner type.

### DIFF
--- a/bitcoind-tests/tests/test_cpp.rs
+++ b/bitcoind-tests/tests/test_cpp.rs
@@ -150,8 +150,8 @@ pub fn test_from_cpp_ms(cl: &Client, testdata: &TestData) {
     for i in 0..psbts.len() {
         let wsh_derived = desc_vec[i].derived_descriptor(&secp).unwrap();
         let ms = if let Descriptor::Wsh(wsh) = &wsh_derived {
-            match wsh.as_inner() {
-                miniscript::descriptor::WshInner::Ms(ms) => ms,
+            match wsh {
+                miniscript::descriptor::Wsh::Ms(ms) => ms,
                 _ => unreachable!(),
             }
         } else {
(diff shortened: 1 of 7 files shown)


apoelstra (dismissed):
ACK bbf5ee96bbde06a9b10687e2a41357f82ada7b0e

tcharding:
> This is a breaking change but hopefully not too disruptive of one. 

On that topic, whats your view on the status of miniscript, you said in the past that the 1.0 release was a mistake, do you consider the crate is in the same state as rust-bitcoin i.e., we can shake things up a bunch with the aim of trying to stabilise so we can stop making so many changes to the public API?

apoelstra:
@tcharding yeah. I think a "real" 1.0 should even be delayed until we get rid of all the recursion, make it impossible to construct invalid objects, and maybe improve the in-memory representation of things.

tcharding:
Cool cheers, plenty to do then.

apoelstra (dismissed):
ACK 21e3cbc38ba3c8657b8b92b5101e06bf3e388aa0

tcharding:
Rebase. Also fix a couple of match statements that used `match &self` to use `match *self` as is customary.

sanket1729 (commented):
@tcharding, this was done on purpose to prohibit invalid use of constructors. If these are directly used in enums, user can directly create `Wsh::Ms(invalid minisript)`. An invalid miniscript could be something that is not does have a root level type B among other things. 

The purpose of shifting to inners was to guard access strictly via constructors.

tcharding:
Ahhh, so my commit log shows the problem

> The `ShInner` and `WshInner` enum's provide no benefit that I can discern.

The problem is with me, there is a difference, as you state - maintaining the invariant. My bad.

Closing, thanks.

tcharding:
Ahhh I kind of feel vindicated, it is not immediately apparent that this function maintains the [undocumented] invariants.
```rust
    /// Create a new p2sh descriptor with the raw miniscript
    pub fn new(ms: Miniscript<Pk, Legacy>) -> Result<Self, Error> {
        // do the top-level checks
        Legacy::top_level_checks(&ms)?;
        Ok(Self { inner: ShInner::Ms(ms) })
    }
```

PR #588, opened 2023-08-07
Add height to tap tree

tcharding:
I stumbled across a `todo` in the code last week and thought I'd hack it up.

- Patch 1: Do identifier style fix
- Patch 2: Add a unit test to verify change in patch 3 is correct
- Patch 3: Remove recursive call to `tap_tree_height` 

From commit log of patch 3

    Convert tuple `TapTree` tuple to a struct and add `height` to it. This
    saves recusive call in `tap_tree_height` function.
    
    While we are at it rename private function `TapTree::tap_tree_height` to
    `TapTree::height` because it stutters.

### DIFF
--- a/examples/taproot.rs
+++ b/examples/taproot.rs
@@ -101,7 +101,7 @@ fn main() {
     let real_desc = desc.translate_pk(&mut t).unwrap();
 
     // Max satisfaction weight for compilation, corresponding to the script-path spend
-    // `multi_a(2,PUBKEY_1,PUBKEY_2) at taptree depth 1, having:
+    // `multi_a(2,PUBKEY_1,PUBKEY_2) at tap tree depth 1, having:
     //
     //     max_witness_size = varint(control_block_size) + control_block size +
     //                        varint(script_size) + script_size + max_satisfaction_size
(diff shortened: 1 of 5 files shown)


apoelstra:
In 37fea86c6304cdcec5d6c4e1f9b872bb8c2d3d3a you rename two methods -- we should add a deprecated version under the old name. (One gets renamed again in the last commit :) we only need to keep the original name.).

Otherwise c960798c5780ec6c8eaa0f30bce43bb3cebf0cc2 looks good, I think. I'm a little uncomfortable adding a public field which can be constructed in an incorrect way, but I think it's okay for now. In general I'd like to go through this library and make things non-constructable, but that's a much bigger change and I don't have a timeline for it.

tcharding:
Re deprecation, can do. Re renaming twice, sorry about that :) I'll put a bit more effort in and re-spin.

tcharding:
Changes in force push:
- Put the private function rename in a separate patch at the front.
- Left `taptree()` in as deprecated.
- Re-wrote commit logs using more correct wording.

apoelstra (approved):
ACK 476e358f27d97ffa43d1d666f132fef40e9033ad

sanket1729 (approved):
ACK 476e358f27d97ffa43d1d666f132fef40e9033ad

PR #589, opened 2023-08-07
Adding substitution of pkh for taproot

Harshil-Jani:
Recently , We have introduced the method `substitute_raw_pkh` to deal with the change of pkh to expr_raw_pkh. This is the implementation of that method for the taproot miniscript.
Reference PR : https://github.com/rust-bitcoin/rust-miniscript/pull/557

### DIFF
--- a/src/psbt/finalizer.rs
+++ b/src/psbt/finalizer.rs
@@ -21,7 +21,8 @@ use super::{sanity_check, Error, InputError, Psbt, PsbtInputSatisfier};
 use crate::prelude::*;
 use crate::util::witness_size;
 use crate::{
-    interpreter, BareCtx, Descriptor, ExtParams, Legacy, Miniscript, Satisfier, Segwitv0, Tap,
+    interpreter, BareCtx, Descriptor, ExtParams, Legacy, Miniscript, Satisfier, Segwitv0, SigType,
+    Tap, ToPublicKey,
 };
 
 // Satisfy the taproot descriptor. It is not possible to infer the complete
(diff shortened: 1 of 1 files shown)


apoelstra:
concept ACK. Though it took me quite a while to figure out what's going on. I think we should add a comment explaining that we're replacing `RawPkH` with `PkH`, which doesn't change the script in any way, but does attach the actual public key to the pkh fragment.

I was also confused that pkh was still supported in Taproot, but I was just confused, I don't think we need to add any comments to clarify that.

Finally, I don't know what the key origins table is that you're taking keys from -- I can't find it in BIP 174 or 370, at least not with the name "origin". So I'm unsure whether this is too magical...it may be that this causes certain workflows to work, but in such a way it'll stop working as soon as the set of PkHs and the key origin table stop being in sync, and this is such a nonobvious connection that it'd be a nightmare to debug. Unsure.

Harshil-Jani:
> concept ACK. Though it took me quite a while to figure out what's going on. I think we should add a comment explaining that we're replacing `RawPkH` with `PkH`, which doesn't change the script in any way, but does attach the actual public key to the pkh fragment.
> 
> I was also confused that pkh was still supported in Taproot, but I was just confused, I don't think we need to add any comments to clarify that.
> 

We need to satisfy or dissatisfy the key and thus we need all the given key from a descriptor . `tap_key_origin` is the only field which consist of all the keys added on a descriptor and thus we try to get keys from it. I am also not sure about why origin has all keys and not some other field.

sanket1729:
@apoelstra, in `update_psbt_input`, we use the `key_origin_feilds` (PSBT_IN_TAP_BIP32_DERIVATION) to inform the signer which keys we want it to sign for.

sanket1729 (approved):
utACK c6bb14216e7c8d3ecfe7fe09bbc98a86fec3aa80

apoelstra:
> @apoelstra, in `update_psbt_input`, we use the `key_origin_feilds` (PSBT_IN_TAP_BIP32_DERIVATION) to inform the signer which keys we want it to sign for.

Ah, that's right! Ok, with that in mind, I think it make sense to also use this set of keys to obtain public keys, since those are morally pretty similar to bip32 origins.

apoelstra (approved):
ACK c6bb14216e7c8d3ecfe7fe09bbc98a86fec3aa80

PR #594, opened 2023-09-02
Implement `TreeLike` for `policy::Concrete`

tcharding:
Make a start on removing recursion in the `policy::concrete` module.

- Patch 1: Remove`PolicyArc`
- Patch2: Implement `TreeLike` for `concrete::Policy` (does not use it)
- Patch 3: Trivial code comment deletion
- Subsequent patches are pairs
   - Add a basic unit test for the function we are about to patch
   - Remove recursion from a function (the one we just tested)
- Final patch is a refactoring 

There is still some to go but this is already a pretty heavy review burden.

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -15,7 +15,7 @@ pub use tree::{
 };
 
 use crate::sync::Arc;
-use crate::{Miniscript, MiniscriptKey, ScriptContext, Terminal};
+use crate::{policy, Miniscript, MiniscriptKey, ScriptContext, Terminal};
 
 impl<'a, Pk: MiniscriptKey, Ctx: ScriptContext> TreeLike for &'a Miniscript<Pk, Ctx> {
     fn as_node(&self) -> Tree<Self> {
@@ -65,6 +65,32 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> TreeLike for Arc<Miniscript<Pk, Ctx>
                 Tree::Nary(Arc::from([Arc::clone(a), Arc::clone(b), Arc::clone(c)]))
             }
             Thresh(_, ref subs) => Tree::Nary(subs.iter().map(Arc::clone).collect()),
+        }
+    }
+}
+
+impl<'a, Pk: MiniscriptKey> TreeLike for &'a policy::Concrete<Pk> {
+    fn as_node(&self) -> Tree<Self> {
+        use policy::Concrete::*;
+        match *self {
+            Unsatisfiable | Trivial | Key(_) | After(_) | Older(_) | Sha256(_) | Hash256(_)
+            | Ripemd160(_) | Hash160(_) => Tree::Nullary,
+            And(ref subs) => Tree::Nary(subs.iter().map(Arc::as_ref).collect()),
+            Or(ref v) => Tree::Nary(v.iter().map(|(_, p)| p.as_ref()).collect()),
+            Threshold(_, ref subs) => Tree::Nary(subs.iter().map(Arc::as_ref).collect()),
+        }
+    }
+}
+
+impl<'a, Pk: MiniscriptKey> TreeLike for Arc<policy::Concrete<Pk>> {
+    fn as_node(&self) -> Tree<Self> {
+        use policy::Concrete::*;
+        match self.as_ref() {
+            Unsatisfiable | Trivial | Key(_) | After(_) | Older(_) | Sha256(_) | Hash256(_)
+            | Ripemd160(_) | Hash160(_) => Tree::Nullary,
+            And(ref subs) => Tree::Nary(subs.iter().map(Arc::clone).collect()),
+            Or(ref v) => Tree::Nary(v.iter().map(|(_, p)| Arc::clone(p)).collect()),
+            Threshold(_, ref subs) => Tree::Nary(subs.iter().map(Arc::clone).collect()),
         }
     }
 }
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -696,42 +696,45 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
     // Checks whether the given concrete policy contains a combination of
     // timelocks and heightlocks
     fn check_timelocks_helper(&self) -> TimelockInfo {
-        // timelocks[csv_h, csv_t, cltv_h, cltv_t, combination]
-        match *self {
-            Policy::Unsatisfiable
-            | Policy::Trivial
-            | Policy::Key(_)
-            | Policy::Sha256(_)
-            | Policy::Hash256(_)
-            | Policy::Ripemd160(_)
-            | Policy::Hash160(_) => TimelockInfo::default(),
-            Policy::After(t) => TimelockInfo {
-                csv_with_height: false,
-                csv_with_time: false,
-                cltv_with_height: absolute::LockTime::from(t).is_block_height(),
-                cltv_with_time: absolute::LockTime::from(t).is_block_time(),
-                contains_combination: false,
-            },
-            Policy::Older(t) => TimelockInfo {
-                csv_with_height: t.is_height_locked(),
-                csv_with_time: t.is_time_locked(),
-                cltv_with_height: false,
-                cltv_with_time: false,
-                contains_combination: false,
-            },
-            Policy::Threshold(k, ref subs) => {
-                let iter = subs.iter().map(|sub| sub.check_timelocks_helper());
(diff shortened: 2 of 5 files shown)


tcharding:
Unless I'm missing something, if this goes in, we can remove the `concrete::PolicyArc` type (which is now a exactly the same as `concrete::Policy`)

tcharding:
@apoelstra hats off to your `iter` module, that tree is hot.

apoelstra:
Are you able to split this into two commits: one that drops `PolicyArc` in favor of tweaking `Policy` and one which implements TreeLike?

74cafab3f8bf7670af28da2a97d7c3c56b51f749 looks good to me.

tcharding:
Can do!

tcharding:
> Are you able to split this into two commits: one that drops PolicyArc in favor of tweaking Policy and one which implements TreeLike?

I did that then went a bit wild using `TreeLike` :)

src/policy/concrete.rs
```diff
@@ -668,19 +668,13 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
 
     /// Gets all keys in the policy.
     pub fn keys(&self) -> Vec<&Pk> {
-        match *self {
-            Policy::Key(ref pk) => vec![pk],
-            Policy::Threshold(_k, ref subs) => {
-                subs.iter().flat_map(|sub| sub.keys()).collect::<Vec<_>>()
+        let mut keys = vec![];
+        for data in self.post_order_iter() {
```
apoelstra, line 644: In 062e4009c36126e9cf39117170747bcb5cdc2aa5:

I think we should use the pre-order iterator here, which is more intuitively correct. (In general, the post-order iterator matches recursive logic, but I think that, since we're only counting leaf nodes, the two iterators will actually have the same effect.)

src/policy/concrete.rs
```diff
@@ -681,13 +681,21 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
+        let mut nums = vec![];
+        for data in Arc::new(self).post_order_iter() {
+            let num_for_child_n = |n| nums[data.child_indices[n]];
+
+            let num = match data.node {
+                Or(subs) => (0..subs.len()).map(num_for_child_n).sum(),
+                Threshold(k, subs) if *k == 1 => (0..subs.len()).map(num_for_child_n).sum(),
+                _ => 1,
+            };
+            nums.push(num);
```
apoelstra, line 667: In 7e73ffeb14a5e95a019c6abcb72d6976391e2969:

We don't need to maintain a vector of numbers here. We can just do a running sum and avoid some allocations and unwraps.

src/policy/concrete.rs
```diff
@@ -554,30 +554,18 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
...
-            | Policy::Hash160(..)
-            | Policy::After(..)
-            | Policy::Older(..) => true,
-            Policy::Threshold(_, ref subs) | Policy::And(ref subs) => {
-                subs.iter().all(|sub| sub.real_for_each_key(&mut *pred))
+        for data in self.post_order_iter() {
+            if let Policy::Key(ref pk) = data.node {
+                if !pred(pk) {
+                    return false;
+                }
```
apoelstra, line 561: In 37f379632c3629ab7887703001a6599a1d811dd7:

I also think this one should use pre-order iter; I also think it makes no difference here.

src/policy/concrete.rs
tcharding: Oh my bad, I had it in my head that the post order iter was the simpler of the two, I was wrong. Defaulting to pre-order iter when we don't need references to the child nodes.

src/policy/concrete.rs
tcharding: I'm not sure we can, don't we need to sum the results of nested `Or`s? I don't think we can do that without recursion or keeping the nums vec. Am I correct in thinking a tree like this should return 4
```
        or
      /     \ 
   or       or
/ \   /\  /\  /\
k  k k k  k  k  k
```

src/policy/concrete.rs
tcharding: Done!

tcharding:
Changes in force push:
- Used `pre_order_iter` where appropriate
- Got a bit more creative with combinators

src/policy/concrete.rs
apoelstra: Well, post-order is definitely the most *common* because it allows you to simulate depth-first recursive algorithms. But conceptually I think pre-order is more natural.

src/policy/concrete.rs
apoelstra: Yep, you're right, I looked at the algorithm too quickly.

apoelstra (approved):
ACK f145ac0707a38f1403ad096e15814eb94078dc7c

src/policy/concrete.rs
tcharding: No sweat, it made me double check it which was not a bad thing :)

tcharding:
Rebase and run formatter on each commit.

src/policy/concrete.rs
```diff
@@ -58,12 +59,12 @@ pub enum Policy<Pk: MiniscriptKey> {
     /// A HASH160 whose preimage must be provided to satisfy the descriptor.
     Hash160(Pk::Hash160),
     /// A list of sub-policies, all of which must be satisfied.
-    And(Vec<Policy<Pk>>),
+    And(Vec<Arc<Policy<Pk>>>),
     /// A list of sub-policies, one of which must be satisfied, along with
     /// relative probabilities for each one.
-    Or(Vec<(usize, Policy<Pk>)>),
+    Or(Vec<(usize, Arc<Policy<Pk>>)>),
```
tcharding, line 65: /me  gently banging head on table crying, "please god no, don't let this be a vector when it is only ever exactly 2 policies"

@apoelstra or @sanket1729  are `Or` and `And` policies only ever comprised of exactly 2 sub-policies?

apoelstra: I think we might have such a restriction right now (you'll find some `panic`s enforcing it) but we intend to lift this restriction eventually.

The main issue is the computational complexity of compiling n-ary `or`s.

apoelstra (approved):
ACK 1efc7a3d3a50ea7ea775a15ba486cfb7b8702623

src/policy/concrete.rs
tcharding: > but we intend to lift this restriction eventually.

By we do you mean miniscript in general or rust-miniscript? From the miniscript docs it looks like its limited to 2 also but I couldn't tell from the syntax if that was guaranteed. Is it ambiguous for the same reason?

apoelstra: Miniscript in general.

tcharding: Ah ok, so leave these as vecs then. Perhaps I'll add some comments. I literally nearly fell apart yesterday, after 3 days of mental modelling of vectors and tree iteration to find this "2 limit". I'm better today :)

apoelstra: Glad to hear it :). Sure, some comments would be appreciated.

```
// Currently these vectors are limited to two elements. Eventually we would like to extend these
// to be n-ary, but first we need to decide on a game plan for how to efficiently compile n-ary
// disjunctions
```

or something.

src/iter/mod.rs
```diff
@@ -68,3 +68,29 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> TreeLike for Arc<Miniscript<Pk, Ctx>
         }
     }
 }
+
+impl<'a, Pk: MiniscriptKey> TreeLike for &'a policy::Concrete<Pk> {
```
sanket1729, line 72: This is really not a concern for this PR, but I think policy stuff should belong in policy module. Not here. See #603 for more.

sanket1729 (approved):
ACK 1efc7a3d3a50ea7ea775a15ba486cfb7b8702623

src/policy/concrete.rs
tcharding: There are no such length restrictions on the `Policy::Threshold` so when `k == 1` we essentially have unbounded `Or` and when `k == v.len()` we have unbounded `And` - should we be restricting the length of the vector in these two cases also?

tcharding: FTR I'm playing around with creating a `Thresh<T>` type to maintain the invariants `v.len() > 0`, k != 0`, `k <= v.len()`.

PR #595, opened 2023-09-04
Remove `concrete::PolicyArc`

tcharding:
Draft because done on top of #594, just the second patch.

Remove the now redundant `concrete::PolicyArc` type.


tcharding:
I rolled this into #594

PR #597, opened 2023-09-12
CI: Pin memchr

tcharding:
Recently `memchr` broke our MSRV by updating to edition 2021 in a patch version, pin it in CI for MSRV builds.


apoelstra (approved):
ACK 3d1793af95e58031b1c9f7490e26eafcc2128282

PR #598, opened 2023-09-15
CI: Pin syn to 2.0.32

danielabrozzoni:


apoelstra:
I'm a little surprised that any `syn` 2.0.x works with 1.48.

The point of the existing serde pins is to keep `syn` at 1.x. I wonder if it would be possible to remove or relax those now?

apoelstra:
My approach to this CI failure would be to look at whatever's depending on syn 2.x and downgrade *that*.

apoelstra:
On second thought, I don't care enough about the detailed reasons for serde deps to be breaking our shit. We should drop serde_derive which should eliminate all the proc-macro nastiness and move on. Which is a bigger project and independent of this PR.

apoelstra (approved):
ACK 3d5cf3bb01c914438e3e79e7cb27672c92b82dfe

yancyribbens:
@danielabrozzoni thanks for fixing this!  I looked into fixing this but I didn't think to try syn 2.0.x.  How did you decide which version of syn to use?

danielabrozzoni:
@yancyribbens in this case I think I just looked at the compiler error and downgrade the syn dependency until it compiled. Generally, I think using `cargo tree` with the latest toolchain and using the output to figure out which of our dependencies is importing the broken dependency and pinning that is a better approach, as Andrew suggested above.

PR #601, opened 2023-09-26
Remove recursion in `is_valid`

tcharding:
Just the last 2 patches.

Draft because I can't work out why we do validity checks when parsing a policy string and then also have the `Concrete::is_valid` function that enforces, `Or` and `And` being exactly length 2. Is there some code path that allows creation of an invalid policy that we will then catch with `is_valid`?

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -15,7 +15,7 @@ pub use tree::{
 };
 
 use crate::sync::Arc;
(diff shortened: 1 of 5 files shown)


apoelstra:
Yes, you can directly construct these types because they are public enums.

tcharding:
Superseded by #606

PR #602, opened 2023-09-26
Document And/Or vectors contain 2 elements

tcharding:
It is not immediately obvious that the `Concrete::And` and `Or` vectors contain exactly 2 elements, add code comment to document it.

### DIFF
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -37,6 +37,9 @@ const MAX_COMPILATION_LEAVES: usize = 1024;
 /// Concrete policy which corresponds directly to a miniscript structure,
 /// and whose disjunctions are annotated with satisfaction probabilities
 /// to assist the compiler.
+// Currently the vectors in And/Or are limited to two elements, this is a general miniscript thing
+// not specific to rust-miniscript. Eventually we would like to extend these to be n-ary, but first
+// we need to decide on a game plan for how to efficiently compile n-ary disjunctions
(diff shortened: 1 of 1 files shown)


tcharding:
After #594 I might investigate whether it helps to remove the vectors all together. In `miniscript::decode::Terminal` we don't use vecs for ands and ors.

apoelstra:
@tcharding that's because `miniscirpt::decode::Terminal` is part of Miniscript itself, which is never going to have n-ary ands or ors. The policy types are a completely different beast, and they're far less rigid in how we can change them.

tcharding:
Do you mean policies are a thing that is still being improved but miniscript itself is more 'complete'?

apoelstra (approved):
ACK 47c87769d4a88005364b4ec7299ba6d9ef2f8252

apoelstra:
Miniscript is actual on-chain data. Policies are part of the "policy compiler" which is an entirely offchain tool used to produce efficient Miniscripts, and which has no guarantees about determinism or anything. As far as I know it isn't specced out or anything. So in that sense, yes, Miniscript is more "complete".

In another sense, that way more people care about Miniscirpt than care about this compiler thing, I expect more future work to go into Miniscript than the compiler. But that's just me :).

apoelstra:
We also care much less about memory efficiency for the compiler, for this reason, which is why we have `Vec`s in there even though there are no immediate plans to generalize the ands and ors beyond binary operators.

tcharding:
Awesome, thanks man.

PR #603, opened 2023-09-27
Add some high level notes about rust-miniscript design

sanket1729:
Some notes to get people started. Still needs more details, but serves as a good start. In the long run, I believe, these can even be separate features with default on. This gives the option to people who really just want miniscript to use miniscript core without the policy/psbt/interpreter stuff.

PR #604, opened 2023-09-27
Move and rename `crate::policy::semantic::Policy` to `crate::r#abstract::Abstract`

tcharding:
Based on discussion in #603. This requires #603 to be updated.

Done as a series of steps, I don't usually do PRs like this but after reviewing the recent `bitcoin_io` PR I think it does help for making review easier when there are a load of simple but slightly different changes involved in getting from one state to another. Can squash, merge, or re-split if required. Please comment on the patch series style if you have any suggestions on how I can make review easier.

Draft for feedback please, especially the last patch - the struct rename from `Policy` to `Abstract`. Note also this leaves `concrete::Policy` named as is.


tcharding:
There will be some grammar mistakes in docs 'a' vs 'an' because of semantic to abstract change. Will fix later.

tcharding:
TODO: Remove the serialization impl.

apoelstra:
Sorry, but I think we should keep the module name as `semantic`. If users only needed to type `r#abstract` when importing stuff that would be okay, but needing to write `r#abstract::Policy` all over the place is pretty ugly. And just `Abstract` as a type doesn't feel right. (`policy::Abstract` would be okay, but this PR moves the type out of the `policy` module.)

What if we called the module `lift` and changed the type name to `Lifted`?

Or we could keep it as `Policy` and users can write `lift::Policy`...though this may continue to cause confusion with the actual "policy language".

tcharding:
No sweat, I'll have another play with it.

tcharding:
I did #607 so we can see what its like, I will need to close the unwanted one and [re]write the PR description.

tcharding:
I don't think this is going anywhere

PR #606, opened 2023-10-01
Remove more recursion from `concrete::policy`

tcharding:
Remove the last of the recursion in functions that process policies. Note there is still recursion in the `from_tree_prob` function that creates the policies.

### DIFF
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -749,59 +749,46 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
     /// Validity condition also checks whether there is a possible satisfaction
     /// combination of timelocks and heightlocks
     pub fn is_valid(&self) -> Result<(), PolicyError> {
+        use Policy::*;
+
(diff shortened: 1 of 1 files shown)


apoelstra:
Looks great! Interesting observation that we're iterating thrice in `is_valid` (and each one potentially is allocating and deallocating vectors to deal with threshold branches).

I think this is fine, because improving this would result in much uglier code... but we should keep this in the back of our minds, and if it's a common pattern maybe we should try to refactor it out somehow.

apoelstra (approved):
ACK 0452040f6b31b981ee89a5c2480193c879336f65

tcharding:
Yeah this is where my question the other day about why we call `is_invalid` came from because not only is this function inefficient, its called often. I think improving the types helps because I _think_ we can get rid one out of three tree iterations - I agree doing it all in one is likely ugly.

sanket1729 (approved):
utACK 0452040f6b31b981ee89a5c2480193c879336f65

PR #607, opened 2023-10-03
Create the `lift` module

tcharding:
This is an alternative to #604 based on review suggestion.

Draft for feedback please.


apoelstra:
I like this. @sanket1729 what do you think?

PR #608, opened 2023-10-04
Use `bech32` to calculate descriptor checksums

tcharding:
This is just the last patch, the rest are in #609 now.

The `checksum::poly_mod` function implements BCH codes to calculate a checksum (appended to descriptors). We recently released a version of BCH codes in `bech32`. We can implement the `bech32::Checksum` trait for BIP-380 and use the `primitives::checksum` module, removing the custom `poly_mod` function.

### DIFF
--- a/src/descriptor/bare.rs
+++ b/src/descriptor/bare.rs
@@ -12,8 +12,8 @@ use core::fmt;
 use bitcoin::script::{self, PushBytes};
 use bitcoin::{Address, Network, ScriptBuf};
 
-use super::checksum::{self, verify_checksum};
-use crate::descriptor::DefiniteDescriptorKey;
+use super::checksum::verify_checksum;
+use crate::descriptor::{write_descriptor, DefiniteDescriptorKey};
 use crate::expression::{self, FromTree};
 use crate::miniscript::context::{ScriptContext, ScriptContextError};
 use crate::miniscript::satisfy::{Placeholder, Satisfaction, Witness};
@@ -156,12 +156,7 @@ impl<Pk: MiniscriptKey> fmt::Debug for Bare<Pk> {
 }
 
 impl<Pk: MiniscriptKey> fmt::Display for Bare<Pk> {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        use fmt::Write;
-        let mut wrapped_f = checksum::Formatter::new(f);
-        write!(wrapped_f, "{}", self.ms)?;
-        wrapped_f.write_checksum_if_not_alt()
-    }
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { write_descriptor!(f, "{}", self.ms) }
 }
 
 impl<Pk: MiniscriptKey> Liftable<Pk> for Bare<Pk> {
@@ -354,10 +349,7 @@ impl<Pk: MiniscriptKey> fmt::Debug for Pkh<Pk> {
 
 impl<Pk: MiniscriptKey> fmt::Display for Pkh<Pk> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        use fmt::Write;
-        let mut wrapped_f = checksum::Formatter::new(f);
-        write!(wrapped_f, "pkh({})", self.pk)?;
-        wrapped_f.write_checksum_if_not_alt()
+        write_descriptor!(f, "pkh({})", self.pk)
     }
 }
 
--- a/src/descriptor/checksum.rs
+++ b/src/descriptor/checksum.rs
@@ -7,39 +7,18 @@
 //!
 //! [BIP-380]: <https://github.com/bitcoin/bips/blob/master/bip-0380.mediawiki>
 
+use core::convert::TryFrom;
 use core::fmt;
 use core::iter::FromIterator;
+
+use bech32::primitives::checksum::PackedFe32;
+use bech32::{Checksum, Fe32};
 
 pub use crate::expression::VALID_CHARS;
 use crate::prelude::*;
 use crate::Error;
 
-const CHECKSUM_CHARSET: &[u8] = b"qpzry9x8gf2tvdw0s3jn54khce6mua7l";
-
 const CHECKSUM_LENGTH: usize = 8;
-
-fn poly_mod(mut c: u64, val: u64) -> u64 {
-    let c0 = c >> 35;
-
-    c = ((c & 0x7ffffffff) << 5) ^ val;
-    if c0 & 1 > 0 {
-        c ^= 0xf5dee51989
-    };
-    if c0 & 2 > 0 {
-        c ^= 0xa9fdca3312
-    };
-    if c0 & 4 > 0 {
-        c ^= 0x1bab10e32d
-    };
-    if c0 & 8 > 0 {
-        c ^= 0x3706b1677a
-    };
-    if c0 & 16 > 0 {
-        c ^= 0x644d626ffd
-    };
-
-    c
-}
 
 /// Compute the checksum of a descriptor.
 ///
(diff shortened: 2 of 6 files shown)


apoelstra:
Up to 5018673bf5642850cd6aad3d25697de217eb105e is great and could be PR'd separately if you want to move more quickly.

apoelstra:
As for 9ca3380f197f1bf93f92ff1c2729c7c70724b3c4:

* We should pull apart the `Engine` and replace the cls/clscount stuff with an iterator adaptor from `char`s to `Fe`s.
* `Osd` can be `DescriptorChecksum` or something. It doesn't have to be short; it shouldn't ever be used by users.
* I believe your `expect`s are all correct; you are construting a 3-digit base 3 number, which can be at most 222 (or 26 in decimal) which fits into a Fe32.
* Your "comment why we do this" TODO is because you're manually computing a checksum and manually futzing with the target checksum. Just use `Engine::input_target_residue` and then you can drop the code that inputs 0s and basically all the rest of this function. Also you can just replace the whole `checksum_chars` function with the `Checksummed` iterator adaptor from rust-bech32.

tcharding:
Sick! Thanks man.

tcharding:
I learned a new expression 'futzing with', not the first one I've learned from you @apoelstra, I love it.

src/descriptor/bare.rs
```diff
@@ -157,10 +157,17 @@ impl<Pk: MiniscriptKey> fmt::Debug for Bare<Pk> {
...
-        wrapped_f.write_checksum_if_not_alt()
+        let s = format!("{}", self.ms);
+
+        if f.alternate() {
+            write!(f, "{}", s)?;
+        } else {
+            for ch in Checksummed::new_unchecked(&s) {
+                f.write_char(ch)?;
+            }
+        }
```
apoelstra, line 168: In 22ee854c61b13d15fb7d0b11daf2a87604f45ba7:

We definitely want to keep the `wrapped_f` stuff. This new code is both longer and now allocates.

src/descriptor/checksum.rs
```diff
@@ -2,177 +2,180 @@
...
+    /// Creates a new checksummed iterator which adapts a descriptor string
+    /// iterator by appending a checksum.
+    ///
+    /// # Errors
+    ///
+    /// If descriptor contains invalid characters as specified by [BIP-380].
+    ///
+    /// [BIP-380]: <https://github.com/bitcoin/bips/blob/master/bip-0380.mediawiki>
+    #[inline]
+    pub fn new(descriptor: &'a str) -> Result<Checksummed<'a>, Error> {
```
apoelstra, line 41: In 22ee854c61b13d15fb7d0b11daf2a87604f45ba7:

Rather than taking a string this should take a `D: Display` and a `fmt::Formatter`.

src/descriptor/checksum.rs
```diff
@@ -2,177 +2,180 @@
...
-        self.fmt.write_str(s)?;
-        self.eng.input(s).map_err(|_| fmt::Error)
+    let mut parts = s.splitn(2, '#');
+    let desc_str = parts.next().unwrap();
+
+    if let Some(_) = parts.next() {
+        let checksummed = Checksummed::new(desc_str)?;
+        if !checksummed.eq(s.chars()) {
+            return Err(Error::BadDescriptor("invalid checksum".to_owned()));
+        }
```
apoelstra, line 176: In 22ee854c61b13d15fb7d0b11daf2a87604f45ba7:

It would be faster to feed the checksum into the checksum iterator and compare the residue against the target. Then in the happy case we only need to iterate once.

In the unhappy case, we should probably return a better error variant that has the expected and received checksum in it. It's fine if this is slow and allocating.

tcharding: Sure thing.

src/descriptor/bare.rs
tcharding: Yeah I balked at this before I pushed it. Will add it back in.

tcharding:
Alright, I think I got a little trigger happy before to use my new found iterator skills from the recent bech32 work. Instead this PR now just does the minimum required to remove the `poly_mod` function and use `bech32::primitives::checksum` instead.

tcharding:
I was futzing with the whole `descriptor` module before ;)

apoelstra:
Ok, ACK this. It's a good start.

tcharding:
Rebased to pick up changes in #609 and use newly release `bech32 v0.10.0-beta`, no other changes.

apoelstra (dismissed):
ACK c5c08fe43a914b0ae138f598f0c36725d161e7a4

apoelstra:
@sanket1729 I'll let you take a look at this before merging. It basically gets rid of some of the magic constants and uses rust-bech32 facilities to do the checksum. But this doesn't represent a "real" use of the new rust-bech32 API, which would pretty-much reduce the checksum.rs module to just encoding/decoding strings as u5s and then handing off to rust-bech32 APIs.

tcharding:
I was going to attack it in steps:
- Move all the `Formatter` stuff in `descriptor/*.rs` to use a macro `write_desciptor!`.
- Then in the macro work out how to convert the input string to a stream of `Fe32`s like you suggest.

The reason to do two steps is that I don't think we can have a general "input multiple strings" to the checksum engine because cls/clscount is state that has to be kept between successive input strings. But i noticed that we never do more than input a single descriptor.

apoelstra:
I don't like the idea of moving so much logic into a macro.

If we need to retain state then we can keep the existing `Engine` structure with `cls`/`clscount`. If we don't, we should change this to be an iterator adaptor.

But either way, once we have a stream of Fe32s, we should be able to just put them into a rust-bech32 `checksum::Engine` and everything should be easy.

tcharding:
> I don't like the idea of moving so much logic into a macro.

The logic is just the more or less duplicate lines 
```rust
        use fmt::Write;
        let mut wrapped_f = checksum::Formatter::new(f);
        write!(wrapped_f, "{}", self.ms)?;
        wrapped_f.write_checksum_if_not_alt()
```
Other descriptors have a match statement when calling `write` but they are essentially the same. Its totally just an abstraction thing, as it is it appears sane to call `write` multiple times, and the current implementation supports this. But its an unnecessary generalisation I believe because we only ever call `write` once, to write a single descriptor. 

> If we need to retain state then we can keep the existing `Engine` structure with `cls`/`clscount`. If we don't, we should change this to be an iterator adaptor.
>
> But either way, once we have a stream of Fe32s, we should be able to just put them into a rust-bech32 `checksum::Engine` and everything should be easy.

Another complication is we cannot use bech32 iterators to encode because of the `#` symbol. That is what led me to create the  `Checksummed` type. Anyways, I'll have another play with it.

apoelstra:
>The logic is just the more or less duplicate lines

Oh, yeah, this seems fine to me.

>Another complication is we cannot use bech32 iterators to encode because of the # symbol.

Yeah, that sucks. But I think your solution is good.

tcharding:
I had another look at this, I'm not sure we can do much better even using some sort of iterator. I'm pretty sure that our usage of `fmt::Write::write_str` is doing an allocation under the hood. Changing our usage is then clearly an optimization problem not a "use bech32" problem.

tcharding:
I think this should be considered for merge as is, I wrote up #610 to discuss potential improvements.

apoelstra:
>I'm not sure we can do much better even using some sort of iterator

Yeah, you may be right.

apoelstra (approved):
ACK c44ce81d94e80cd1c814146d9b6e91c7def3ff9a

sanket1729 (approved):
utACK c44ce81d94e80cd1c814146d9b6e91c7def3ff9a. Did not review the use of bech32 carefully. Any breakage there should have caused existing test cases to fail.

sanket1729:
We can also safely merge this despite the red crosses as CI is fixed in master.

tcharding:
I rebased it for good measure, no changes.

apoelstra (approved):
ACK fe1a6be3bba92ed7c27d5bc8e72dc590f2e4dc88

PR #611, opened 2023-10-11
Add back in helper function `semantic::Policy::_translate_pk`

tcharding:
Recently while removing recursion I mistakenly removed the `_translate_pk` function forgetting that this was a trick to prevent compilation explosion caused by the generics.

### DIFF
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -563,6 +563,14 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
         T: Translator<Pk, Q, E>,
         Q: MiniscriptKey,
     {
+        self._translate_pk(t)
+    }
+
+    fn _translate_pk<Q, E, T>(&self, t: &mut T) -> Result<Policy<Q>, E>
+    where
+        T: Translator<Pk, Q, E>,
+        Q: MiniscriptKey,
+    {
         use Policy::*;
 
         let mut translated = vec![];


tcharding:
Needs #613

apoelstra:
I don't understand this PR.

tcharding:
There was a post sometime recently by @sanket1729 about how we were getting and explosion of code once all the generics were expanded [0]. And something about the helper function not needing to be re-compiled each time ...

Can you help me out Sanket, I clearly did not fully grok it when I read it.


[0] Is "expanded" the correct term?

apoelstra:
There are a couple things that I think you might mean:

1. When we take generic parameters that can take a `T` or `&mut T` and have functions that call each other with a call tree a few layers deep, we wind up compiling functions for `T`, `&mut T`, `&mut mut T`, etc., which is wasteful. This happens with the rust-bitcoin encoding stuff but never here.
2. When we call functions *from themselves* we get the same thing but infinitely many times and compilation fails. No longer an issue here.
3. When a function takes a generic `AsRef` (for example) and all we do with the generic argument is call `.as_ref()`, then everything after that line uses concrete types, and it's wasteful to compile the whole function multiple times with different generic parameters. Better to have a helper function with concrete types, then the real function calls `as_ref()` before calling that.

I **think** you are trying to address issue (3). But the current code here is a no-op. You have two functions with identical signatures, with identical generics, and one is calling the other.

tcharding:
Thanks for the explanation, should have trusted myself more originally when I removed the helper because I couldn't see it was achieving anything. Closing.

tcharding:
I dug through the git index and the `_translate_pk` function was introduced when `translate_pk` had a different signature, when we changed it we didn't notice that the helper was no longer needed. The "mysterious" reason I vaguely thought existed must be some unrelated thing floating around in my brain.

For interest it was introduced in `commit ef1d69c81e3acc34ed2ce71b6484d719cc631ab6`

PR #613, opened 2023-10-11
fuzz: Bump rustc to version1.65

tcharding:
CI is complaining:
    
    package `regex v1.10.0` cannot be built because it requires rustc 1.65
     or newer, while the currently active rustc version is 1.64.0
    
Bump rustc to v1.65 in the `generate_files.sh` script and run it.


apoelstra (approved):
ACK ae330259fc62c8c2e4e8ea69658b2650b6a20b1f

sanket1729:
ACK

PR #617, opened 2023-10-13
Improve `TapTree` API

tcharding:
Improve the `TapTree` API by making the `height` and `combine` methods public.

Note the serendipitous finding that upgrading miniscript to use bitcoin v0.31.0-rc1 requires no changes. I found these fixes doing the `bdk` upgrade.


tcharding:
This PR does pose the question, "should we be maintaining an invariant that `height` is valid for `left` and `right`?" At the moment the fields are public.

apoelstra (approved):
ACK 0630491aee647a6cdcbe6c876a51fad08e918840

Issue #619, opened 2023-10-17
What is the point of all the hash conversion methods in `ToPublicKey`


tcharding:
Currently all implementors of `ToPublicKey` just return the hash argument, what is the point of these trait methods, what am I missing? Is this trait used differently by some other consumer of `rust-miniscript`?

apoelstra:
Yes. The point of these is to convert from an abstract hash value to a concrete one that can be embedded in a script.

Any implementer of `ToPublicKey` that is not working with literal hash values will need to do a nontrivial conversion.

apoelstra:
I'm going to close this. Seems like it's "asked and answered".

PR #620, opened 2023-10-17
Clean up the `MiniscriptKey` and `ToPublicKey` traits

tcharding:
Done while investigating the use of secret keys in descriptors, this is only loosely related but anyways ...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,8 +154,6 @@ pub trait MiniscriptKey: Clone + Eq + Ord + fmt::Debug + fmt::Display + hash::Ha
     fn is_uncompressed(&self) -> bool { false }
 
     /// Returns true if the key is an x-only pubkey, defaults to `false`.
-    // This is required to know what in DescriptorPublicKey to know whether the inner
-    // key in allowed in descriptor context
     fn is_x_only_key(&self) -> bool { false }
 
     /// Returns the number of different derivation paths in this key, defaults to `0`.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -210,7 +210,7 @@ impl MiniscriptKey for bitcoin::secp256k1::XOnlyPublicKey {
 }
 
 impl MiniscriptKey for String {
-    type Sha256 = String; // specify hashes as string
+    type Sha256 = String;
     type Hash256 = String;
     type Ripemd160 = String;
     type Hash160 = String;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,34 +148,39 @@ pub use crate::miniscript::satisfy::{Preimage32, Satisfier};
 pub use crate::miniscript::{hash256, Miniscript};
 use crate::prelude::*;
 
-///Public key trait which can be converted to Hash type
+/// Trait representing a key which can be converted to a hash type.
 pub trait MiniscriptKey: Clone + Eq + Ord + fmt::Debug + fmt::Display + hash::Hash {
-    /// Returns true if the pubkey is uncompressed. Defaults to `false`.
+    /// Returns true if the key is serialized uncompressed, defaults to `false`.
     fn is_uncompressed(&self) -> bool { false }
 
-    /// Returns true if the pubkey is an x-only pubkey. Defaults to `false`.
+    /// Returns true if the key is an x-only pubkey, defaults to `false`.
     // This is required to know what in DescriptorPublicKey to know whether the inner
     // key in allowed in descriptor context
     fn is_x_only_key(&self) -> bool { false }
 
-    /// Returns the number of different derivation paths in this key. Only >1 for keys
-    /// in BIP389 multipath descriptors.
+    /// Returns the number of different derivation paths in this key, defaults to `0`.
+    ///
+    /// Only >1 for keys in BIP389 multipath descriptors.
     fn num_der_paths(&self) -> usize { 0 }
 
-    /// The associated [`bitcoin::hashes::sha256::Hash`] for this [`MiniscriptKey`], used in the
-    /// sha256 fragment.
+    /// The associated [`sha256::Hash`] for this `MiniscriptKey`, used in the sha256 fragment.
+    ///
+    /// [`sha256::Hash`]: bitcoin::hashes::sha256::Hash
     type Sha256: Clone + Eq + Ord + fmt::Display + fmt::Debug + hash::Hash;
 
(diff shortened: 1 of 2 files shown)


tcharding:
And this PR led me to ask: https://github.com/rust-bitcoin/rust-miniscript/issues/619

apoelstra:
Yeah, this does look like a bug, though in normal use of the interpreter the Bitcoin consensus rules will protect us. Good catch!

I also believe that in the interpreter context compressedness doesn't matter (because we are letting the consensus rules have full reign, and trying not to impose any further restrictions). But I don't remember exactly where we landed with the interpreter context.

tcharding:
By "interpreter context" do you mean `NoChecks`?

tcharding:
I'm still really struggling to understand these hash types. Conceptually a script hashlock is satisfied by providing the hash pre-image, right? It doesn't have anything to do with a key? Why are they in the `MiniscriptKey` trait, and why are the conversions done in the `ToPublicKey` trait?

tcharding:
Perhaps you can link me (again) to some consumer code of `rust-miniscript` that flexes these traits because as we use them in `rust-miniscript` I'm coming up short.

tcharding:
This is all non-urgent, answer at your leisure please.

apoelstra:
> I'm still really struggling to understand these hash types. Conceptually a script hashlock is satisfied by providing the hash pre-image, right? It doesn't have anything to do with a key?

Correct, nothing to do with a key.

> Why are they in the `MiniscriptKey` trait, and why are the conversions done in the `ToPublicKey` trait?

Because otherwise we'd need 5 generics instead of 1, which is especially annoying because most users don't use generic hash preimages.

> Perhaps you can link me (again) to some consumer code of `rust-miniscript` that flexes these traits because as we use them in `rust-miniscript` I'm coming up short.

I'm not aware of anybody using hash preimages *at all*. But Miniscript supports them, and this library therefore needs to support them, and it would be a frustrating and surprising limitation if you could genericize public keys without genericizing hashes.

But for an example of how generic keys are useful, see [this file in ICBOC](https://github.com/apoelstra/icboc/blob/541194912158138d7cf96e5a84594470b4f00044/src/lib/wallet/mod.rs#L478-L517) which has a generic "cached public key" type which holds a descriptor, but when needed, queries the hardware wallet (a ~100ms operation) to obtain an actual key.

tcharding:
Ok, cool. Thanks for the explanation man.

tcharding:
I expected that to be harder since I did this 18 months ago.

src/lib.rs
```diff
@@ -147,34 +147,39 @@ pub use crate::primitives::absolute_locktime::{AbsLockTime, AbsLockTimeError};
 pub use crate::primitives::relative_locktime::{RelLockTime, RelLockTimeError};
 pub use crate::primitives::threshold::{Threshold, ThresholdError};
 
-/// Public key trait which can be converted to Hash type
+/// Trait representing a key which can be converted to a hash type.
 pub trait MiniscriptKey: Clone + Eq + Ord + fmt::Debug + fmt::Display + hash::Hash {
-    /// Returns true if the pubkey is uncompressed. Defaults to `false`.
+    /// Returns true if the key is serialized uncompressed, defaults to `false`.
```
apoelstra: In 12dd99d20223270524000ea9cd421fe7af7b132d:

This is a run-on sentence (and it was not before).

src/lib.rs
```diff
@@ -147,34 +147,39 @@ pub use crate::primitives::absolute_locktime::{AbsLockTime, AbsLockTimeError};
...
     fn is_uncompressed(&self) -> bool { false }
 
-    /// Returns true if the pubkey is an x-only pubkey. Defaults to `false`.
+    /// Returns true if the key is an x-only pubkey, defaults to `false`.
```
apoelstra: In 12dd99d20223270524000ea9cd421fe7af7b132d:

Run-on.

src/lib.rs
```diff
@@ -147,34 +147,39 @@ pub use crate::primitives::absolute_locktime::{AbsLockTime, AbsLockTimeError};
...
-    /// Returns the number of different derivation paths in this key. Only >1 for keys
-    /// in BIP389 multipath descriptors.
+    /// Returns the number of different derivation paths in this key, defaults to `0`.
+    ///
+    /// Only >1 for keys in BIP389 multipath descriptors.
     fn num_der_paths(&self) -> usize { 0 }
 
-    /// The associated [`bitcoin::hashes::sha256::Hash`] for this [`MiniscriptKey`], used in the
-    /// sha256 fragment.
+    /// The associated [`sha256::Hash`] for this `MiniscriptKey`, used in the sha256 fragment.
```
apoelstra: In 12dd99d20223270524000ea9cd421fe7af7b132d:

We should just drop these doclinks. They don't make any sense.

tcharding: hmm, because I shorten the type it looses the `bitcoin::hashes` info which was in the link, you happy with that? It seems reasonable to link to `hashes`, what makes you say it doesn't make sense?

tcharding: Que? Do you mean that the point of mentioning it is just that its the sha256 algo and has nothing to do with the `hashes` crate? Or that we should not mention it at all?

apoelstra: You can keep the reference to sha256, as in "The SHA-256 hash type used in the `sha256` fragment".

We can also drop the "this `MiniscriptKey`" which is redundant (it's alreday a method on the `MiniscriptKey` trait).

apoelstra:
In 765df8944dd22b5bfe1ddc5a56e5b7bcb1e3f3c7:

You said `ripemd256` instead of `ripemd160`.

apoelstra:
In 9207e2c04adabf29058ee3828a54c2781365edb2:

We should leave the default impl on `is_uncompressed`. Nobody should be using uncompressed keys. It would be nice not to remind users that they even exist.

If you are using uncompressed keys with this library and incorrectly set them to be compressed, you will notice because the produced scripts are wrong. It's not a subtle thing. (There *are* subtle things about uncompressed keys, which were broken for years and nobody noticed, which suggests that nobody is ever using uncompressed keys. See #620.)

tcharding:
Done.

apoelstra (approved):
ACK c3a02d25a9227f012de7ca77889abfad9497115e; successfully ran local tests

PR #621, opened 2023-10-19
POC: Add custom hash types to try and simplify all the associated hash type stuff

tcharding:
Hey @apoelstra this is another exploratory POC based on recent discussion about `MiniscriptKey` and `ToPublicKey`. It is also in a similar vein to #600.

Just the last patch, I think this works this time. Please take a look at your leisure.

From the last patch to save you pulling down the branch
```
    Add a custom Sha256 type
    
    It seems that we do a whole bunch of stuff with the hash types to keep
    the Rust compiler happy and be able the implement `MiniscriptKey` on
    `String`. I believe we are conflating two things:
    
    - Support for string "placeholders" in miniscript
    - The ability to mock in test code
    
    This all stems from the observation that `MiniscriptKey` does not
    _really_ describe a key and neither does `ToPublicKey` because of the
    hash types.
    
    To clarify that I understand how we got here; we have a generic key on
    most functions/types and also we need type information about hashes for
    terminals, so instead of carrying a bunch of generics we store it all in
    `MiniscriptKey`. Then we implement `MinisicriptKey` for `String` so we
    can use, for example, `pk(A)` but this breaks the hash type stuff so we
    add `ToPublicKey` for types that can be encoded as _real_ miniscript (I
    think).
    
    This patch is a POC that we can add a custom type, that way we can
    remove the associated hash types then we can do translation in the type
    because the type knows if it is a _real_ hash or a string placeholder
    thing.
    
    As this is a POC its only done for sha256.
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,11 @@ name = "htlc"
 required-features = ["std", "compiler"]
 
 [[example]]
+name = "htlc-string"
+required-features = ["std", "compiler"]
+
+
+[[example]]
 name = "parse"
 required-features = ["std"]
 
--- a/bitcoind-tests/tests/setup/test_util.rs
+++ b/bitcoind-tests/tests/setup/test_util.rs
@@ -182,11 +182,6 @@ impl<'a> Translator<String, DescriptorPublicKey, ()> for StrDescPubKeyTranslator
         }
     }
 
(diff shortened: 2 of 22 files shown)


tcharding:
> NACK. This dramatically reduces functionality.

Damn, I'll understand this crate one day, mark my word.

PR #630, opened 2023-12-02
fix: needless borrow (reference is immediately dereferenced by the rust-compiler)

wthrajat:
fixes:
```shell
this expression creates a reference which is immediately dereferenced by the compiler
   --> src/psbt/mod.rs:881:54
    |
881 |                     cache.p2wpkh_signature_hash(idx, &script_code, amt, hash_ty)?
    |                                                      ^^^^^^^^^^^^ 
```


apoelstra (approved):
ACK d8abb0805c5c006ecde80bf502650dc9fba70363

PR #631, opened 2023-12-05
Remove internals dependency

tcharding:
We do not need to depend on `internals` anymore, this was just needed until we sorted out `DisplayHex` in `hex-conservative`. This change should have been done as part of upgrading to `rust-bitcoin v0.31`.


apoelstra (approved):
ACK 198ebf34a7a5fae0aff03dfb3a368f889540a563

PR #632, opened 2023-12-05
Use `Ordered` and remove `AbsLockTime` type

tcharding:
This is not a merge candidate, its a demo, together with https://github.com/rust-bitcoin/rust-bitcoin/pull/2248

The sole purpose of the `AbsLockTime` type is to implement `PartialOrd` and `Ord`. We can use the `Ordered` type to wrap `absolute::LockTime` to achieve this purpose.

### DIFF
--- a/src/miniscript/decode.rs
+++ b/src/miniscript/decode.rs
@@ -11,6 +11,8 @@ use core::marker::PhantomData;
 use std::error;
 
 use bitcoin::hashes::{hash160, ripemd160, sha256, Hash};
+use bitcoin::locktime::absolute;
+use bitcoin::ordered::Ordered;
 use bitcoin::{Sequence, Weight};
 use sync::Arc;
 
@@ -22,7 +24,7 @@ use crate::miniscript::ScriptContext;
 use crate::prelude::*;
 #[cfg(doc)]
 use crate::Descriptor;
-use crate::{hash256, AbsLockTime, Error, Miniscript, MiniscriptKey, ToPublicKey};
+use crate::{hash256, Error, Miniscript, MiniscriptKey, ToPublicKey};
 
 /// Trait for parsing keys from byte slices
 pub trait ParseableKey: Sized + ToPublicKey + private::Sealed {
@@ -135,7 +137,7 @@ pub enum Terminal<Pk: MiniscriptKey, Ctx: ScriptContext> {
     RawPkH(hash160::Hash),
     // timelocks
     /// `n CHECKLOCKTIMEVERIFY`
-    After(AbsLockTime),
+    After(Ordered<absolute::LockTime>),
     /// `n CHECKSEQUENCEVERIFY`
     Older(Sequence),
     // hashlocks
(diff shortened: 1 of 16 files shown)


src/miniscript/decode.rs
```diff
@@ -135,7 +137,7 @@ pub enum Terminal<Pk: MiniscriptKey, Ctx: ScriptContext> {
     RawPkH(hash160::Hash),
     // timelocks
     /// `n CHECKLOCKTIMEVERIFY`
-    After(AbsLockTime),
+    After(Ordered<absolute::LockTime>),
```
Kixunil, line 140: Maybe this should be actually just `absolute::LockTime` and then `Terminal` should impl `ArbitraryOrd`.

tcharding: Interesting, with a big match that calls through to `cmp` or `arbitrary_cmp` respectively. Nice, I'll have a play with it. Thanks

apoelstra: Then we'd need to wrap `Terminal` in `Ordered` everywhere, which would affect everybody, not just locktime users. And as @tcharding says, it'll be more annoying to implement because there's a huge match. And it's much less obvious why `ArbitraryOrd` is needed.

I prefer this solution.

tcharding: With that, I wont bother playing with the `Terminal` idea, unless @Kixunil has some counter-argument to why it might be cleaner? Its not super important because this isn't a merge candidate anyways, just a demo.

Kixunil: My argument is that `Ordered<absolute::LockTime>` is supposed to be an arbitrary ordering which implies `Terminal` is as well. Although it'd be reasonable to specify that `ArbitraryOrd` on `LockTime` guarantees properties beyond what `ArbitraryOrd` requires and then use it in `Terminal` to guarantee specific canonical ordering.

And regarding match, a derive macro for `ArbitraryOrd` would be great, though I'd 100% understand that nobody wants to write it.

>  need to wrap Terminal in Ordered everywhere, which would affect everybody

That's the point - to expose that the ordering is semantically meaningless.

apoelstra: The nature of `Terminal` is that it has no natural semantically meaningful ordering. This isn't the case for `Locktime` where there is a natural way you might imagine locktimes are ordered, if you don't realize how subtle/stupid they are.

Kixunil: Oh, so you mean that because `Terminal` is so obviously unordered that you wouldn't use comparison operators on it and expect anything specific so there's no footgun but for `LockTime` there is ordering that looks fine but it's a footgun?

apoelstra: Yeah, that's my view.

tcharding: Flagging that the decision here may/could be taken as policy if/when we start removing `PartialOrd`/`Ord` from derives in `rust-bitcoin`, note that maybe not all cases will be as cut n dry. I guess we come up with a policy "err on the side of caution" or some such thing (when judging foot-gun-ness)?

apoelstra: I would prefer that we "err on the side of always implementing `PartialOrd` / `Ord`" unless there is clearly a canonical ordering that is impossible/inefficient to implement.

tcharding: You lost me, this is only about not implementing `PartialOrd`/`Ord` if there is a footgun with regards to users writing `a < b`. Nothing to do with efficiency, at least in my mind?

apoelstra: Imagine that the locktime ordering wasn't incoherent but it did require milliseconds to compute. Then we still wouldn't implement `Ord` since it is surprising to users that `<` or maps would kill their performance, and would still want to provide a "cheap ordering" via `Ordered` for people who don't need a semantically meaningful one.

Kixunil: OK, that makes sense, ACK this design.

In case of slow canonical ordering it should be provided somehow. `PartialOrd` is the most natural but `<` is indeed a footgun.

PR #634, opened 2024-01-04
fuzz: cover taproot in `roundtrip_miniscript_str`

brunoerg:


apoelstra:
Thanks! I'll ACK/merge this though I wonder if we'd be better off dropping the `else`. It would make the test a little less efficient but would let us test parsing taproot descriptors that are also valid non-taproot descriptors.

apoelstra:
CI failure is an unrelated breakage in one of our dependencies.

apoelstra (approved):
ACK 3b47c3cf1b22ac0432d1e1191250a6a9daf55688

PR #639, opened 2024-02-16
Update MSRV to 1.56.1

storopoli:
Following `rust-bitcoin` MSRV of 1.56.1, this pins MSRV to 1.56.1.

Quoting a recent merge in `rust-bitcoin`

> Rust version 1.56.0 introduced edition 2021. Shortly afterwards, on
> October 21 2021 Rust version 1.56.1 was released.
>
> Debian stable is currently shipping `rustc 1.63.0`.
>
> Our stated MSRV policy is: In Debian stable and at least 2 years old.
>
> Therefore our MSRV policy is met by Rust version 1.56.1 and we can strat
> to bump our MSRV org wide.
>
> Links:
> - https://blog.rust-lang.org/2021/11/01/Rust-1.56.1.html
> - https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html
> - https://packages.debian.org/stable/rust/rustc

Removing the pins in `contrib/test.sh` since they are not necessary
in 1.56.1.

EDIT: CI fails because of the `#[allow(dead_code)]` implemented in #638 on some Psbt traits.
Do I port these to here? Conceptually they belong to #638 and not here.

PR #638, opened 2024-02-15
Deprecate across the board `max_satisfaction_weight`

storopoli:
Adds a `since=10.0.0` to all `max_satisfaction_weight` deprecations. Adds a note telling users to check #476 for more details.

Closes #637.

PR #523, opened 2023-02-01
Fixing the Inclusivity of Z in StrKeyTranslator

Harshil-Jani:
Fixing `StrKeyTranslator` in `test_utils.rs` where Z was not included. 
Closes #496 
Signed-off-by: Harshil Jani <harshiljani2002@gmail.com>

PR #640, opened 2024-02-16
Clippy lints MSRV 1.56.1

storopoli:
Nothing out of the ordinary.
The most alarming was the non-inclusive range in
`test_utils.rs`.

In some test functions I allowed myself to use
`#[allow(clippy::type_complexity)]` because clippy was asking to create a type instead of using the big tuple. Since these were used just once, I thought it was overkill.

Note, depends on:

- #639 (we can also merge this into it instead of `master`)
- #638 (has some allows on dead_code)

### DIFF
--- a/src/psbt/mod.rs
+++ b/src/psbt/mod.rs
@@ -343,32 +343,32 @@ impl<'psbt, Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for PsbtInputSatisfie
         self.psbt.inputs[self.index]
             .hash160_preimages
             .get(&Pk::to_hash160(h))
-            .and_then(try_vec_as_preimage32)
+            .and_then(|x: &Vec<u8>| try_vec_as_preimage32(x))
     }
 
     fn lookup_sha256(&self, h: &Pk::Sha256) -> Option<Preimage32> {
         self.psbt.inputs[self.index]
             .sha256_preimages
             .get(&Pk::to_sha256(h))
-            .and_then(try_vec_as_preimage32)
+            .and_then(|x: &Vec<u8>| try_vec_as_preimage32(x))
     }
 
     fn lookup_hash256(&self, h: &Pk::Hash256) -> Option<Preimage32> {
         self.psbt.inputs[self.index]
             .hash256_preimages
             .get(&sha256d::Hash::from_byte_array(Pk::to_hash256(h).to_byte_array())) // upstream psbt operates on hash256
-            .and_then(try_vec_as_preimage32)
+            .and_then(|x: &Vec<u8>| try_vec_as_preimage32(x))
     }
 
     fn lookup_ripemd160(&self, h: &Pk::Ripemd160) -> Option<Preimage32> {
         self.psbt.inputs[self.index]
             .ripemd160_preimages
             .get(&Pk::to_ripemd160(h))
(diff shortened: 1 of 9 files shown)


src/psbt/mod.rs
```diff
@@ -343,32 +343,32 @@ impl<'psbt, Pk: MiniscriptKey + ToPublicKey> Satisfier<Pk> for PsbtInputSatisfie
...
-            .and_then(try_vec_as_preimage32)
+            .and_then(|x: &std::vec::Vec<u8>| try_vec_as_preimage32(x))
     }
 
     fn lookup_sha256(&self, h: &Pk::Sha256) -> Option<Preimage32> {
         self.psbt.inputs[self.index]
             .sha256_preimages
             .get(&Pk::to_sha256(h))
-            .and_then(try_vec_as_preimage32)
+            .and_then(|x: &std::vec::Vec<u8>| try_vec_as_preimage32(x))
```
tcharding, line 353: This won't work in no-std builds, just use plain old `Vec`, its imported already in this file using `use prelude::*`.

src/test_utils.rs
```diff
@@ -168,7 +168,7 @@ impl StrXOnlyKeyTranslator {
             .collect();
         let mut pk_map = HashMap::new();
         let mut pkh_map = HashMap::new();
-        for (i, c) in (b'A'..b'Z').enumerate() {
+        for (i, c) in (b'A'..=b'Z').enumerate() {
```
tcharding, line 171: This changes the behaviour of the loop to now include 'Z', from looking at the code it looks like this was a bug and the change is correct but this change warranted a patch on its own with a comment on the bug. Since its test code, I think we can leave it as is now though.

BTW I've got deja vu reviewing this, I've definitely removed these lint warnings before somewhere. I can't remember where but I remember hitting this 'Z' thing especially.

tcharding (approved):
Thanks for doing these man, appreciated - and the linter found a bug in the tests, linting for the WIN.

ACK bb29eccb4e8609a9f285c009e32a62698a51f0a7

tcharding:
Woops I acked, but needs the Vec change. CI should catch that though.

src/psbt/mod.rs
storopoli: Done, this seems obvious now in retrospect...

src/test_utils.rs
apoelstra: Yeah, I remember this too. https://github.com/rust-bitcoin/rust-miniscript/pull/523 It was even mixed in with CI crap.

apoelstra:
Hooray!

apoelstra:
Lol @ the `type_complexity` one. We probably should actually create a struct for that. But that's a job for another PR.

Thanks very much for fixing CI. Funny how every clippy update finds a new category of "unnecessary `&`" errors that it previously did not detect.

apoelstra (approved):
ACK e2636f390a41e43292b47a2eb7aeea1e3aa34a50

storopoli:
> Lol @ the `type_complexity` one. We probably should actually create a struct for that. But that's a job for another PR.

Probably a simple Tuple Struct



> Funny how every clippy update finds a new category of "unnecessary `&`" errors that it previously did not detect.

Yes, clippy lints are truly amazing. Never stops to impress me...

PR #641, opened 2024-02-19
WIP : Implement a `sortedmulti_a` descriptor

Harshit933:
This is mostly the copy of `sortedmulti` for now with some changes in the Terminal being `MultiA` instead of `Multi` and some `max_satisfaction_witness_element` and `max_satisfaction_size` as implemented in core.

Some Question that I have and I think we can implement it on the way in this PR. 

- How can we build_template and what exactly is a template?
- How to calculate the `script_size` of `sortedmulti_a`.

Some other questions that I have about implementing `sortedmulti_a` inside tr like `tr(c6...., sortedmulti_a(2, 5f...., 6e....)`. For now we build a taptree and consider the trees inside the tr like : `tr(c6..., {pk(), pk()}`. This means when I trying to build a tree for `tr(c6...., sortedmulti_a(2, 5f...., 6e....)` I am hitting an error with 
```
`Result::unwrap()` on an `Err` value: Unexpected("sortedmulti_a(3 args) while parsing Miniscript")
```

One solution that I for now think is to implement a `TrInner` as we have inside `Sh()`. Am I on the right path?

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -34,6 +34,7 @@ mod segwitv0;
 mod sh;
 mod sortedmulti;
 mod tr;
+mod sortedmulti_a;
 
 // Descriptor Exports
 pub use self::bare::{Bare, Pkh};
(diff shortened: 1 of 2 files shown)


Harshit933:
@sanket1729 @apoelstra sorry for the ping. But can you please tell me if this is the right way of approaching the porblem or is there a better aproach to this?
Thanks.

PR #642, opened 2024-02-26
Remove generics from Error by turing fragment into String

RCasatta:
This cause a reduction of IR lines emitted by the compiler of about 2% (using cargo llvm-lines), so this should improve compile times.

I've seen using the parse example to check the produced binary size, but it's a small snippet I am not sure it calls a representative portion of the API surface.

From a library perspective this is equivalent since the field is used only to be converted to string.

In theory externally someone could use the field in the error to do things so it comes at a cost.


apoelstra:
When no-std is on you need to explicitly import `alloc::string::String` I believe. This currently does not compile without std.

But concept ACK doing this. It seems like an improvement to me. Eventually we should go over all the error types like we have been doing with rust-bitcoin.

tcharding:
One day I'd love to get time (ie months) to focus on `miniscript`.

apoelstra:
I fixed CI in #645. I'd like to get that in and rebase this because it's the kind of thing that we want CI to be passing for.

RCasatta:
rebased to take #645

apoelstra:
Looks like you need to run `cargo fmt`.

RCasatta:
> Looks like you need to run cargo fmt.

done

It's broken on nightly which seems to have `alloc::string::String` automatically imported, I am not sure how to fix

apoelstra:
When `std` is on then `String`, `ToString` etc are always imported. It's just that nightly warns about this. There are half a dozen open issues on rustc related to this new lint triggering all the time in counterproductive ways.

Meanwhile to deal with it you just need to make sure your imports are feature-gated on nostd.

RCasatta:
ah right the same you did [here](https://github.com/rust-bitcoin/rust-miniscript/pull/645/commits/3725549e92fdaee24a23e1c96ada1e4c3a33e60c)

apoelstra (approved):
ACK 0afd8324dfb1fb0850efc6edfc3a554c4dcddd45

apoelstra:
cc @sanket1729 can you concept-ACK this?

PR #643, opened 2024-02-26
call reduce0 from reduce1 and 2

RCasatta:
reduce code duplication by calling the function that provide the same check and side effects


apoelstra (approved):
ACK 5c4851615a43373bb8b6bd15e1d144f26b9300d0 neat!

PR #645, opened 2024-02-27
remove sketchy `LikelyFalse` error

### LINKED ISSUE #633: bug: some miniscripts are being pointed as invalid, but are valid
Hi, recently I started developing [bitcoinfuzz](https://github.com/brunoerg/bitcoinfuzz) - differential fuzzing of Bitcoin implementations and libraries. One of the targets gets a string and checks whether it's a valid miniscript. The code I'm using to check it with `rust-miniscript` is:
```rust
#[no_mangle]
pub extern "C" fn rust_miniscript_from_str(input: *const c_char) -> bool {
    if let Ok(data) = unsafe { CStr::from_ptr(input) }.to_str() {
        if let Ok(_pol) = Miniscript::<String, Segwitv0>::from_str_insane(data) {
            return true
        } else if let Ok(_pol) = Miniscript::<String, Tap>::from_str_insane(data) {
            return true
        }
    }
    false
}
```

and `bitcoinfuzz` is crashing (rust-miniscript returning invalid) with the following miniscripts (and other ones):
```
nnnnnnnnnnnnnnnln:1
dv:0
lll:0
l:1
```

apoelstra:
This error would trigger on `l:0` on the basis that this is equivalent to `u:0` but always less efficient. There are no other "linting" errors like this in this library, and the C++ implementation doesn't have it, so we should drop it.

Fixes #633

### DIFF
--- a/examples/psbt_sign_finalize.rs
+++ b/examples/psbt_sign_finalize.rs
@@ -8,7 +8,7 @@ use miniscript::bitcoin::hashes::hex::FromHex;
 use miniscript::bitcoin::psbt::{self, Psbt};
 use miniscript::bitcoin::sighash::SighashCache;
 use miniscript::bitcoin::{
-    self, secp256k1, transaction, Address, Amount, Network, OutPoint, PrivateKey, Script, Sequence,
+    secp256k1, transaction, Address, Amount, Network, OutPoint, PrivateKey, Script, Sequence,
     Transaction, TxIn, TxOut,
 };
 use miniscript::psbt::{PsbtExt, PsbtInputExt};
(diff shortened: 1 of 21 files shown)


apoelstra:
CI is broken because of https://github.com/rust-lang/rust/issues/121684

apoelstra:
Added a pile of commits and filed an issue on rustc to fix CI.

RCasatta:
I can't properly ack because I don't feel super confident with my knowledge of the repo (talking about b87b4fba8ffb2712e0b5a67cddf77d5dfc690a6b) but

LGTM

tcharding (approved):
ACK 50db03c9620f052740c370362b3498a7ad483a8c

apoelstra:
cc @sanket1729 can you ACK at least the first commit here?

sanket1729:
Sure. Will do in 10 mins

sanket1729:
Sure. Will do in 10 mins

sanket1729 (approved):

sanket1729 (approved):
ACK 50db03c9620f052740c370362b3498a7ad483a8c

Issue #585, opened 2023-07-31
There are a lot of things we can do in the longer term to help this. We can split out the crates into multiple separate crates like `psbt`, `interpreter`, `descriptor` and `policy`. This would reduce the compilation time and memory usage by offering flexibility.


sanket1729:
There are a lot of things we can do in the longer term to help this. We can split out the crates into multiple separate crates like `psbt`, `interpreter`, `descriptor` and `policy`. This would reduce the compilation time and memory usage by offering flexibility. 

As pointed by @benma , this is because we use a lot of generics. In particular, almost every single function is being monomorphized for each `Ctx`. 
There are 4 types of Contexts (Bare, Legacyp2sh, Segwitv0, Taproot) that are compiled for each descriptor method. One idea might be to define `Terminal` without the Context. Guard `Miniscript` with appropriate Context while in the constructor `Miniscript::from_ast`. This should significantly reduce the code duplication and have a good impact on binary sizes.

_Originally posted by @sanket1729 in https://github.com/rust-bitcoin/rust-miniscript/issues/584#issuecomment-1657796630_

sanket1729:
Using a third party tool `cargo-bloat`, indeed monomorphization is the cause for this.
```
sanket1729:~/rust-miniscript/examples$ cargo bloat --release --example parse -n 100 --filter=miniscript --full-fn
    Finished release [optimized] target(s) in 0.02s
    Analyzing target/release/examples/parse

File .text     Size      Crate Name
0.2%  1.5%   9.9KiB miniscript miniscript::miniscript::astelem::<impl miniscript::expression::FromTree for miniscript::miniscript::decode::Terminal<Pk,Ctx>>::from_tree::hf7cabd7b5ae86e5f
0.2%  1.4%   9.7KiB miniscript miniscript::miniscript::astelem::<impl miniscript::expression::FromTree for miniscript::miniscript::decode::Terminal<Pk,Ctx>>::from_tree::h252a38afb9bfe9a7
0.2%  1.4%   9.5KiB miniscript miniscript::miniscript::astelem::<impl miniscript::expression::FromTree for miniscript::miniscript::decode::Terminal<Pk,Ctx>>::from_tree::hc8530801dfa8e9a7
[...]

PR #646, opened 2024-02-27
Introduce an example binary useful for profiling

RCasatta:
Tools like `cargo bloat` don't work on libraries but on final executable.

As seen in https://github.com/rust-bitcoin/rust-miniscript/issues/585 and https://github.com/rust-bitcoin/rust-miniscript/pull/584 the parse example is used as a base to profile binary bloat.

However, since the parse example is not using all the API surface, we may have good optimization that are not recognized as such because the optimized function are not in the tree of the functions used.

For benchmarking size optimization a specific binary that ideally touch all the API surface is needed and this MR introduce it.

More coverage will be added if this seem a good idea for reviewers.

### DIFF
--- a/examples/big.rs
+++ b/examples/big.rs
@@ -10,8 +10,8 @@
 //!
 
 use std::{collections::HashMap, str::FromStr};
-use bitcoin::ecdsa;
-use miniscript::{descriptor::Wsh, policy::{Concrete, Liftable}, psbt::PsbtExt, DefiniteDescriptorKey, Descriptor, DescriptorPublicKey, MiniscriptKey};
+use bitcoin::{ecdsa, XOnlyPublicKey};
+use miniscript::{descriptor::Wsh, policy::{Concrete, Liftable}, psbt::PsbtExt, translate_hash_fail, DefiniteDescriptorKey, Descriptor, DescriptorPublicKey, MiniscriptKey, TranslatePk, Translator};
 use secp256k1::Secp256k1;
 fn main() {
     let empty = "".to_string();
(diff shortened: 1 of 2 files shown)


apoelstra:
We may want a `maintainer-tools` repo in the org, similar to bitcoin-core/bitcoin-maintainer-tools, where we could put stuff like this. I'd also like to try unifying our test.sh scripts, API checking scripts, etc etc, there.

RCasatta:
Certainly it is useful in other crates too, but at the moment in particular this crate and its companion elements-miniscript have high percentage of binary size consumption in downstream libraries due to the high use of generics so it makes the most sense here.

apoelstra:
Sorry, I didn't actually look at the diff (because I am waiting for #645 to get in so that CI will work). I didn't realize it was actually an example in the `examples/` directory. I assumed it was a whole separate crate, similar to our embedded tests in some other repos.

Yes, this should certainly stay in this repo.

My comments about wanting a maintainer-tools repo are separate -- we should have this anyway so we can deduplicate our test scripts and write "end to end" tests of the whole ecosystem.

RCasatta:
Rebased and added some calls:

in debug big executable is 72Mb and the second biggest example is 51Mb (41% bigger)
in release big executable is 9Mb and the second biggest example is 7.3Mb (23% bigger)

examples/big.rs
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: CC0-1.0
+//! This is not an example and will surely panic if executed, the purpose of this is using the
```
apoelstra: In ec751fb06b7af82f20936208c377c61960f53648:

trailing space after `the`

apoelstra: Ah, the formatter got this in the last commit.

apoelstra (approved):
ACK c8d3b9ac1f9ec913eb30768439f0f5f4db9a02be thanks!

apoelstra:
Let's get this in. It's a super useful benchmark. I am planning to give "redo the error types" a shot in the coming week and hopefully I can get this size down.

PR #648, opened 2024-03-01
allow deprecated method calls in deprecated methods

RCasatta:
instead of https://github.com/rust-bitcoin/rust-miniscript/pull/647


apoelstra (approved):
ACK b96a1b53f834174edb8ec0fefcc7389cae78edd8

PR #649, opened 2024-03-03
Refactor out type_check

RCasatta:
This apply the same logic of https://github.com/ElementsProject/elements-miniscript/pull/74
removing a couple of unreachable, simplify things and very likely reduce binary bloat.

~~However, in doing so, I don't think this MR is equivalent to master for the logic used for `get_child`.~~

~~I suspect current master is broken and this fix it (or viceversa?), but I have to suspend this work for now.~~


apoelstra:
Oops, needs (another) rebase.

RCasatta:
Rebased!

apoelstra (approved):
ACK 5f41cb64b5927dcd06b0c1ecda23ed27518b16d9 this is great, thanks!

PR #650, opened 2024-03-03
Replace macros with traits; other cleanups

apoelstra:
There are several macros we use to implement functions when our `Pk` types satisfy a large bundle of trait conditions. There is a trick https://github.com/rust-lang/rust/issues/20671#issuecomment-1905186183 that we can use instead.

While I'm at it, do several other small cleanups.

As a weekend project I rewrote the expression module and was able to get a 10-15% speedup on parsing miniscripts, while eliminating recursion and simplifying the algorithms. I am still working on cleaning up the code and improving the error handling. This is the first set of commits from that branch, which should be simple and uncontroversial.


apoelstra:
Rebased.

src/descriptor/bare.rs
```diff
@@ -166,24 +166,22 @@ impl<Pk: MiniscriptKey> Liftable<Pk> for Bare<Pk> {
...
         BareCtx::top_level_checks(&sub)?;
         Bare::new(sub)
     }
-);
+}
 
-impl_from_str!(
-    Bare<Pk>,
-    type Err = Error;,
+impl<Pk: crate::FromStrKey> core::str::FromStr for Bare<Pk> {
```
tcharding, line 177: nit: I rekon we should import `FromStrKey` and use the more terse `<Pk: FromStrKey>` since inline trait bounds are noisy enough as it is.

tcharding (approved):
ACK f22dc3aa000438bdb2c38f4b7acf70d0903f5960

sanket1729 (approved):
ACK 164bde9a9b1ec04c4c545e4f24a7fdbae2281b17. Are you taking a performance refactor stab at rust-miniscript?

src/descriptor/bare.rs
apoelstra: Agreed. Will do this as a followup (or you are welcome to).

apoelstra:
> ACK [164bde9](https://github.com/rust-bitcoin/rust-miniscript/commit/164bde9a9b1ec04c4c545e4f24a7fdbae2281b17). Are you taking a performance refactor stab at rust-miniscript?

Sorta -- my original goal was to test a more-efficient tree structure, starting with `expression::Tree` because nobody directly uses that or cares if I wreck up its API. That was successful, and as a side effect of doing it I was able to get rid of recursion during parsing and also dramatically improve the error messages. (Not a hard thing to do; the existing errors are basically stringly-typed throughout the codebase :)).

From there I've gotten a bit sidetracked improving the errors in this crate and improving other things (like these macros) which I ran into. My next PR will bring in a new `WithSpan` error type which is able to keep track of locations in the original string being parsed, and either use it in the typechecker errors, or in the new `expression` module.

PR #654, opened 2024-03-06
Several locktime improvements

apoelstra:
This PR:
* Changes user-facing "check locktime" functions to use `bitcoin::relative::LockTime` rather than `Sequence` numbers.
* Introduces a new `RelLockTime` type analogous to `AbsLockTime` which is used to provide a locktime type that implements `Ord`.
* In both types, validate that the locktimes are legal for use in Miniscript during creation, rather than during typechecking or validity checks.
* In many places, replace manual code with the new `is_implied_by` method from rust-bitcoin.

This might be a little hard to review because it messes with locktime code which is so easy to get backward. But it appears that there are pretty extensive unit tests which caught several iterations of this where I did get them backward. Going forward things should be much easier to maintain because we have stronger types and an encapsulated `is_implied_by` method.

The next PR will do a similar thing with thresholds to validate the `0 < k <= n` invariant when thresholds are created, avoiding the repeated and inconsistent post-hoc error checking logic.


tcharding (approved):
ACK 9b3d45d8dd36b267f91e23b6d5220748f61b51cc

tcharding:
Gotta be happy with the new `is_implied_by`, feels good to get it right huh. Was a bit sad to see the error code at the top of the file \me shakes head, bangs table, and mumbles "have I taught you nothing" :)

apoelstra:
>Was a bit sad to see the error code at the top of the file \me shakes head, bangs table, and mumbles "have I taught you nothing" :)

In my later PRs there will be `errors.rs` modules, but for such short files with so few errors, it seemed fine to inline them :).

And if you want to feel disappointment, you should see the *existing* logic in this crate ;) there are stringly-typed errors everywhere and even places where we call `.to_string` on errors and then embed the strings in new errors!

src/policy/compiler.rs
```diff
@@ -1308,14 +1297,16 @@ mod tests {
     #[test]
     fn compile_timelocks() {
         // artificially create a policy that is problematic and try to compile
+        /*
```
sanket1729, line 1291: Creating an issue to fix this.

src/interpreter/stack.rs
```diff
@@ -236,14 +231,19 @@ impl<'txin> Stack<'txin> {
...
+            if n.is_implied_by(tx_locktime) {
+                self.push(Element::Satisfied);
+                Some(Ok(SatisfiedConstraint::RelativeTimelock { n: *n }))
+            } else {
+                Some(Err(Error::RelativeLocktimeNotMet(*n)))
+            }
         } else {
-            Some(Err(Error::RelativeLocktimeNotMet(n.to_consensus_u32())))
+            // BIP 112: if the tx locktime has the disable flag set, fail CSV.
+            Some(Err(Error::RelativeLocktimeNotMet(*n)))
```
sanket1729, line 246: RelativeLocktimeDisabled variant should be returned here.

sanket1729 (approved):
ACK 1f8af49144add6537044e58a4be393824a9921f1.

The error variant can be fixed in future commit. Does clippy have a  lint that checks that all error variants are constructed and used in atleast one place?

apoelstra:
@sanket1729 annoyingly no, Clippy, does not have a lint. But you can locally change `pub` on the error types to `pub(crate)` and then you'll get rustc warnings, so you can manually lint this way.

src/interpreter/stack.rs
apoelstra: Good catch. I'll just re-do the PR since this is a quick thing to fix.

src/policy/compiler.rs
apoelstra: I don't remember commenting out this code. I wonder if some unknown-to-me vim binding did it.

apoelstra: Oh, no, I remember now. The point is that this policy used to cause an error when you compiled it, but now you just can't construct it.

I should delete this test entirely.

apoelstra:
Addressed both comments. Also changed `Locktime` to `LockTime` in the interpreter errors, since this was the only place that this spelling occurred (in rust-bitcoin we standardized on `LockTime`).

There are some unused variants in src/lib.rs `Error` type but I'll get those in a later PR. Eventually I want to delete this because we've moved away from the "library has a single giant global error type" model in rust-bitcoin.

sanket1729 (approved):
ACK 422b83cb198976ee0932686dc7dd4bdf396a9c8b

PR #658, opened 2024-03-12
Experimenting reducing stack size of `Error`

RCasatta:
This draft PR reduce the stack size of the crate::Error type from 72 bytes to 32 bytes by boxing bigger variants.

I noticed a very noticeable reduction in binary size produced which I can't explain
in comparison to master the rlib produced is reduced by 0.5% and big example of about 2%.

At the same time llvm lines emitted are almost the same.

Can someone explain this?

### DIFF
--- a/src/descriptor/bare.rs
+++ b/src/descriptor/bare.rs
@@ -299,7 +299,7 @@ impl<Pk: MiniscriptKey + ToPublicKey> Pkh<Pk> {
             let witness = vec![];
             Ok((witness, script_sig))
         } else {
-            Err(Error::MissingSig(self.pk.to_public_key()))
+            Err(Error::MissingSig(Box::new(self.pk.to_public_key())))
         }
     }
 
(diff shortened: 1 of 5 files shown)


apoelstra:
If I had to guess, boxing error variants means that some of the error subtypes don't need to be compiled at all anymore.

RCasatta:
ah, okay...

It seems the savings are enough to justify the annoying boxing, should I polish this to pursue merging?

apoelstra:
@RCasatta I'd suggest not, unless you urgently want this work in, because I am working toward cleaning up the error types with my recent refactoring PRs. So the top-level `Error` type should hopefully go away and be replaced by smaller more specific errors.

Having said that -- in my own code, I haven't decided yet whether to use boxing. I think probably I should, in particular to deal with the generic parsing errors (`<Pk as FromStr>::Error` etc). Right now I think we convert these to strings lol. The alternatives are to box the errors (still requires alloc, plausibly loses some information users might want) or to make our error types generic (which will blow up codegen and is super annoying to use).

So my plan is also to box errors.

Conveniently, our existing error framework is so bad that even a "bad" solution like this is definitely an improvement :).

@sanket1729 do you have any opinions?

PR #659, opened 2024-03-17
Derives `Hash` for pub items

### LINKED ISSUE #226: Derive `Hash` for everything that can derive `Hash` for all structures in the lib

Harshit933:
This commit dervies `Hash` for enums and structs which also derives `Eq`.

Fixes #226

### DIFF
--- a/src/interpreter/inner.rs
+++ b/src/interpreter/inner.rs
@@ -52,7 +52,7 @@ fn script_from_stack_elem<Ctx: ScriptContext>(
 }
 
 /// Helper type to indicate the origin of the bare pubkey that the interpereter uses
-#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]
 pub enum PubkeyType {
     Pk,
     Pkh,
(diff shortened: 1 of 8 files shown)


apoelstra:
At first glance `DescriptorSecretKey` is missing.

Harshit933:
`DescriptorSecretKey` contains `SinglePriv` variant which have `bitcoin::PrivateKey` as one of the fields which does not implement `Hash`. Thus, can't derive `Hash` for it.

apoelstra:
Ok, that sounds reasonable. There are also a bunch of errors not done but that's fine, we need to do a separate pass over this library to fix all the error types. (A few errors *are* done, which is also fine I guess.)

`Token` in miniscript/lex.rs is missing.

There are some other things I think should be `Hash` but they currently don't derive `Eq` (some of them derive nothing, lol) so I think they're fine too. We can revisit them in a later pass.

So I think it's just `Token` I'd like you to add.

Harshit933:
> So I think it's just `Token` I'd like you to add.

Done.

apoelstra (approved):
ACK b3027a959523fca50e70d0f7903009f3a47e07b7

PR #660, opened 2024-03-19
Introduce Threshold type and use it in SortedMulti as a preview

apoelstra:
I have a big branch which uses this type throughout the codebase. I split it up into pieces (separate commit for its use in each policy type, for its use in multi/multi_a, for its use in Terminal::thresh, etc) but it's still a pretty big diff.

Despite the size of the diff I think this is definitely worth doing -- in my current branch we have eliminated a ton of error paths, including a ton of badly-typed ones (grepping the codebase for `errstr` shows I've reduced it from 30 instances to 18).

This PR simply introduces the new `Threshold` type and uses it in `SortedMultiVec` to show how the API works. Later PRs will do the more invasive stuff.


src/primitives/threshold.rs
```diff
@@ -0,0 +1,310 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Thresholds
+//!
+//! Miniscript
```
tcharding, line 5:
```suggestion
//! A generic (k,n)-threshold type.
```

Literally the only thing that was better from #605 :)

src/primitives/threshold.rs
```diff
@@ -0,0 +1,310 @@
...
+
+/// Structure representing a k-of-n threshold collection of some arbitrary
+/// object `T`.
+///
+/// If the constant parameter `MAX` is nonzero, it represents a cap on the
+/// `n` value; if `n` exceeds `MAX` then an error is returned on construction.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Threshold<T, const MAX: usize = 0> {
+    k: usize,
+    inner: Vec<T>,
```
tcharding, line 51: Slight preference for `v` (or something else) because this is not a wrapper type so `inner` seems slightly off.

apoelstra:
Will need to iterate a bit to get CI to pass. Rust's rules for redundant/unused imports are idiotic and incompatible with nostd.

src/descriptor/sortedmulti.rs
```diff
@@ -19,62 +19,55 @@ use crate::plan::AssetProvider;
...
-            .iter()
-            .map(|sub| expression::terminal(sub, Pk::from_str))
-            .collect();
-
-        pks.map(|pks| SortedMultiVec::new(k as usize, pks))?
+        let ret = Self {
+            inner: tree
+                .to_null_threshold()
+                .map_err(Error::ParseThreshold)?
+                .translate_by_index(|i| expression::terminal(&tree.args[i + 1], Pk::from_str))?,
```
tcharding, line 66: I wasn't able to convince myself that the index addition is good. Would be good to have a simple test to check for potential off-by-one error.

tcharding (commented):
Looks mad! Just commenting not acking because you force pushed while I was reviewing.

src/descriptor/sortedmulti.rs
apoelstra: I use this method all over the place in later PRs. It's definitely correct.

apoelstra: But sure, I can add a unit test that directly checks it.

apoelstra: Actually, I'm confused by what you mean by "a simple test". This method is called literally every time that a `SortedMulti` is parsed (and whenever any other type is parsed, after later PRs which use this method in the same way for the other types).

The addition has a +1 because the original expression starts with a k value that we need to skip.

tcharding (approved):
ACK 50bf40485654f9a32dc310e94b9e2c63a30e27c5

src/primitives/threshold.rs
apoelstra: Lol, oh, right :) I should finish this doccomment.

src/descriptor/sortedmulti.rs
tcharding: lolz, still some ways to go improving as a review huh. I just flicked through looking for a unit test, not looking for other call sites. Marking as resolved.

src/descriptor/sortedmulti.rs
```diff
@@ -19,62 +19,55 @@ use crate::plan::AssetProvider;
...
-        if pks.len() > MAX_PUBKEYS_PER_MULTISIG {
-            return Err(Error::BadDescriptor("Too many public keys".to_string()));
-        }
-
+    fn constructor_check(&self) -> Result<(), Error> {
         // Check the limits before creating a new SortedMultiVec
         // For example, under p2sh context the scriptlen can only be
         // upto 520 bytes.
-        let term: Terminal<Pk, Ctx> = Terminal::Multi(k, pks.clone());
+        let term: Terminal<Pk, Ctx> = Terminal::Multi(self.inner.k(), self.inner.data().to_owned());
```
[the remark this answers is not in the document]
apoelstra: Lemme check my followup PR and make sure that this allocation goes away. I'm pretty sure it does.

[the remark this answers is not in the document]
apoelstra: Yep, once `Terminal::Multi` also takes a `Threshold` this alllocation can go away.

PR #661, opened 2024-03-21
Upgrade bech32 dependency

tcharding:
Upgrade to the newest release of the `bech32` crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ rand = ["bitcoin/rand"]
 base64 = ["bitcoin/base64"]
 
 [dependencies]
-bech32 = { version = "0.10.0-beta", default-features = false }
+bech32 = { version = "0.11.0", default-features = false }
 bitcoin = { version = "0.31.0", default-features = false }
 
 # Do NOT use this as a feature! Use the `serde` feature instead.
--- a/src/descriptor/checksum.rs
+++ b/src/descriptor/checksum.rs
@@ -19,6 +19,7 @@ use crate::prelude::*;
 use crate::Error;
 
 const CHECKSUM_LENGTH: usize = 8;
+const CODE_LENGTH: usize = 32767;
 
 /// Compute the checksum of a descriptor.
 ///
@@ -142,6 +143,7 @@ const GEN: [u64; 5] = [0xf5dee51989, 0xa9fdca3312, 0x1bab10e32d, 0x3706b1677a, 0
 impl Checksum for DescriptorChecksum {
     type MidstateRepr = u64; // We need 40 bits (8 * 5).
     const CHECKSUM_LENGTH: usize = CHECKSUM_LENGTH;
+    const CODE_LENGTH: usize = CODE_LENGTH;
     const GENERATOR_SH: [u64; 5] = GEN;
     const TARGET_RESIDUE: u64 = 1;
 }


tcharding:
By "needs reviewing by a cryptographer" I meant you obviously @apoelstra :)

apoelstra:
The length is actually 32767.

This code is very interesting in that correcting it requires working in a degree-3 extension of the base field. It'll be a good test case for our error correcting logic.

tcharding:
> It'll be a good test case for our error correcting logic.

I read this first thing this an hour or so ago and didn't understand, you mean our "soon-to-be-implemented error correcting logic", right?

tcharding:
Do we want a comment on the code length magic numbers (1023 and 32767)? I for one can't remember where they come. Can you either tell me again or just give what ever tip you can off the top of your head for me to research a code comment.

apoelstra:
> I read this an hour or so ago and didn't understand, you mean our soon-to-be-implemented error correcting logic, right?

Yes.

>Do we want a comment on the code length magic numbers (1023 and 32767)? I for one can't remember where they come. Can you either tell me again or just give what ever tip you can off the top of your head for me to research a code comment.

They come from the mathematical properties of the error correcting code. Specifically it is the lcm of the order of the roots of the generating polynomial. I don't think there's a simple reason "why" that we could comment. It's kinda annoying to compute this value directly from the other code parameters. In this case I found it in a code comment block in src/script/descriptor.cpp in Bitcoin Core (and it was not labelled as "the length of the code", it was labelled as the order of a particular field element whose powers form the roots of the polynomial).

tcharding:
How about just "This value calculable from the mathematical properties of the error correcting code."

apoelstra:
I think that's a little misleading, since the value is a design parameter of the code. The mathematical properties come from the length (among other things), rather than the other way around.

tcharding:
How about (can you confirm I got lcm right?)
```rust
/// The Least Common Multiple (lcm) of the order of the roots of the generating polynomial (or if you don't understand that, a property of the error correcting code).
```

apoelstra:
This is correct, but I think it's way too technical for user-facing docs. What this parameter actually means is the maximum length of a string that can be error-corrected.

tcharding:
Ugh, well that is clearly the doc we want: "The maximum length of a string that can be error-corrected"

We got there in the end, you math nerds, I don't know :)

apoelstra:
Lol, well, I thought you were asking how I got the specific constant 32767, not about what the constant meant in the abstract.

tcharding:
Ha! I did want to know that but "I grepped Core for it" answered that for me. I do enjoy the full technical description but when I have to start looking up acronyms I know its got too deep, and basically anytime I read polynomial my eyes glaze over just a little bit :) I should have done more math at uni.

apoelstra:
BTW this is written in the docs: https://docs.rs/bech32/latest/bech32/trait.Checksum.html#associatedconstant.CODE_LENGTH

In any case, to do error correction I will need to implement a bunch more algebra that will hopefully clarify/sanity-check these values so we don't need to go digging around for them (we can just guess-and-check, or maybe I'll have code that can compute them for us).

tcharding:
> BTW this is written in the docs:

OMG, I literally laughed out loud at my initiating that whole conversation without reading that doc - EPIC FAIL.

sanket1729 (approved):
ACK 6bd6301641851d3bf55ad35ee88c890d21836638

PR #664, opened 2024-03-23
Return Weight type for max_weight_to_satisfy methods

yancyribbens:
Return Weight type for max_weight_to_satisfy methods.

### DIFF
--- a/embedded/src/main.rs
+++ b/embedded/src/main.rs
@@ -51,7 +51,7 @@ fn main() -> ! {
     assert!(desc.sanity_check().is_ok());
 
     // Estimate the satisfaction cost
-    assert_eq!(desc.max_weight_to_satisfy().unwrap(), 288);
+    assert_eq!(desc.max_weight_to_satisfy().unwrap().to_wu(), 288);
     // end miniscript test
 
     // exit QEMU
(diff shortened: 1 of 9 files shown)


yancyribbens:
I'm not sure if there should be a more specific error added or if there is a better one to use then the general one I picked here.

apoelstra:
> I'm not sure if there should be a more specific error added or if there is a better one to use then the general one I picked here.

Fine to use the general one for now. I have an ambitious PR in the works that will try to add span information etc to errors so any "extra" work you do now will just get overwritten when I finish that.

yancyribbens:
Rebased.

apoelstra:
You need to run `cargo +nightly fmt` in this repo.

yancyribbens:
> You need to run cargo +nightly fmt in this repo.

Thanks, done.

apoelstra:
Now this appears to be based on the tip of #661 rather than on master.

yancyribbens:
> Now this appears to be based on the tip of https://github.com/rust-bitcoin/rust-miniscript/pull/661 rather than on master.

Now sure how I managed that.  I'm now showing 4aaedf is now the previous commit.

apoelstra (approved):
ACK 7c5f3dc5b54373c12cad430d356f29bce33e9dc4

PR #665, opened 2024-03-25
Add weight to max satisfaction weight

yancyribbens:
Return Weight type for max_satisfaction_weight methods.

### DIFF
--- a/embedded/src/main.rs
+++ b/embedded/src/main.rs
@@ -51,7 +51,7 @@ fn main() -> ! {
     assert!(desc.sanity_check().is_ok());
 
     // Estimate the satisfaction cost
-    assert_eq!(desc.max_weight_to_satisfy().unwrap(), 288);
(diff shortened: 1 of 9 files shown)


apoelstra:
This builds on #664 and has the same issue with commit structure.

yancyribbens:
> This builds on https://github.com/rust-bitcoin/rust-miniscript/pull/664 and has the same issue with commit structure.

Oops, Rebase should be fixed now.

yancyribbens:
Looks like `max_satisfaction_weight` is deprecated, maybe it's not worth changing these.

apoelstra:
Ah, yeah, I think we should leave the deprecated methods alone. (Let's keep them in another release cycle though, since we're re-breaking the new methods..)

yancyribbens:
max_satisfaction_weight methods are depreciated.

PR #666, opened 2024-03-27
Manually implement is_uncompressed for BitcoinKey

tcharding:
The `interpreter::BitcoinKey` uses the default implementation of `MiniscriptKey`, which means `is_uncompressed` returns `false`. However if the full key is a `bitcoin::PublicKey` it may be compressed.

Manually implement `MiniscriptKey::is_uncompressed` for `BitcoinKey` and return the compressedness of the inner full key.

Originally done, and discussed, in #620.


apoelstra:
We may want to just remove the default impl. I think it may have been there to avoid a breaking change in the trait a long time ago.

apoelstra (approved):
ACK 52198dd62da7c4066f6f04fc38aa418480a5e5af

PR #669, opened 2024-03-28
Remove explicit link target

tcharding:
When building the rustdocs clippy emits:

  warning: redundant explicit link target

As suggested, remove the explicit link target. Do not fix column width because this rustdoc block is already a mess - just make the change while keeping the diff small.


apoelstra (approved):
ACK d1175a53e4ccf53c8680bfd5b6401449daf15b63

PR #668, opened 2024-03-28
Add justfile

tcharding:
Add a `just` config file with various useful commands. Note that a bunch of these commands do not succeed currently on master, specifically:

- just lint
- just docsrs
- just sane

Will fix as separate PRs

PR #670, opened 2024-03-28
Fix all lint errors and warnings

tcharding:
After doing #668 I wanted to be able to use the `just sane` command before pushing PRs, so fix the lint issues.


apoelstra (approved):
ACK 76346427f9f1ea8624ca6e1e1bb228a718c26e88 thanks! we should re-enabled Clippy in CI in this repo...I believe we disabled it due to a clippy bug in July

PR #671, opened 2024-04-01
CI: Lint the crate and examples

tcharding:
We are not currently running the linter in CI, do so.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,6 +15,8 @@ jobs:
           profile: minimal
           toolchain: nightly
           override: true
+      - name: Install clippy
+        run: rustup component add clippy
       - name: Running benchmarks
         env:
           DO_BENCH: true
@@ -26,6 +28,10 @@ jobs:
       - name: Running formatter
         env:
           DO_FMT: true
+        run: ./contrib/test.sh
+      - name: Running linter
+        env:
+          DO_LINT: true
         run: ./contrib/test.sh
 
   Int-tests:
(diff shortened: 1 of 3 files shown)


apoelstra:
Will `$clippy --all-targets` get the same effect?

I think we should try running
```
clippy --all-features --all-targets
clippy --all-targets
clippy --no-default-features --all-targets # maybe with --features=no-std if we require that here
```

Which I *think* will cover the examples (though let me check..) and then gets us a bit of feature-matrix coverage as well.

apoelstra:
Yeah, confirmed that --all-targets will flag `let x = u8::from(10u8)` in an example program.

tcharding:
Interesting, then we (I) got it wrong in `rust-bitcoin` - I thought I had checked this before.

IIUC running the linter with different feature flags only adds checking unused imports when features are turned off, and while its nice to have these correct I'm not sure its worth bothering contributors with failing CI runs for it?

tcharding:
I was wrong we get real errors with `cargo +nightly clippy --no-default-features --features=no-std --all-targets -- -D warnings` that we don't get with `--all-features` - WIN.

Here is an example (there are a few more)
```bash
error: this expression creates a reference which is immediately dereferenced by the compiler
   --> src/lib.rs:757:37
    |
757 |             fn deref(&self) -> &T { &self.lock.deref() }
    |                                     ^^^^^^^^^^^^^^^^^^ help: change this to: `self.lock.deref()`
```

tcharding:
Now includes two clippy fixes up front and uses the three feature gate combinations suggested.

apoelstra (approved):
ACK a410c06b3f1a452f4fa1095558d8f0618ca371af

PR #673, opened 2024-04-03
test: Add max_weight_to_satisfy test

yancyribbens:
Add a test for `max_weight_to_satisfy`.  Marking as a draft for now since I had to change the visibility of a few things which may not be desirable.  However, I'm looking to possibly use this in an external crate, so the visibility may need to be adjusted to make that viable.  Also of note that most key types return an error when tying to compute the satisfaction weight.

### DIFF
--- a/src/descriptor/bare.rs
+++ b/src/descriptor/bare.rs
@@ -410,3 +410,63 @@ where
         }
     }
 }
+
+#[cfg(test)]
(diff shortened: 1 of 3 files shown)


yancyribbens:
Testing `max_weight_to_satisfy` this way is too complicated.

PR #674, opened 2024-04-04
Use `Threshold` type in concrete policy and in Terminal::multi/multi_a

apoelstra:
Some more large but mostly-mechanical diffs. This is able to eliminate a bunch of error paths, though the actual error variants can't be removed until we also convert Terminal::thresh in the next PR(s). At that point we will start to see the real benefits of this type because fallible functions will become fallible and massive amounts of compiler error-checking can just go away entirely.

Also removes the allocation in `SortedMulti::constructor_check` that was pointed out in #660.

### DIFF
--- a/src/descriptor/sortedmulti.rs
+++ b/src/descriptor/sortedmulti.rs
@@ -36,7 +36,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> SortedMultiVec<Pk, Ctx> {
         // Check the limits before creating a new SortedMultiVec
         // For example, under p2sh context the scriptlen can only be
         // upto 520 bytes.
-        let term: Terminal<Pk, Ctx> = Terminal::Multi(self.inner.k(), self.inner.data().to_owned());
+        let term: Terminal<Pk, Ctx> = Terminal::Multi(self.inner.clone());
         let ms = Miniscript::from_ast(term)?;
         // This would check all the consensus rules for p2sh/p2wsh and
         // even tapscript in future
@@ -113,11 +113,8 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> SortedMultiVec<Pk, Ctx> {
     /// utility function to sanity a sorted multi vec
     pub fn sanity_check(&self) -> Result<(), Error> {
         let ms: Miniscript<Pk, Ctx> =
-            Miniscript::from_ast(Terminal::Multi(self.k(), self.pks().to_owned()))
-                .expect("Must typecheck");
-        // '?' for doing From conversion
-        ms.sanity_check()?;
-        Ok(())
+            Miniscript::from_ast(Terminal::Multi(self.inner.clone())).expect("Must typecheck");
+        ms.sanity_check().map_err(From::from)
     }
 }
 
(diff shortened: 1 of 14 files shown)


src/descriptor/sortedmulti.rs
```diff
@@ -32,15 +32,21 @@ pub struct SortedMultiVec<Pk: MiniscriptKey, Ctx: ScriptContext> {
...
         // Check the limits before creating a new SortedMultiVec
         // For example, under p2sh context the scriptlen can only be
         // upto 520 bytes.
-        let term: Terminal<Pk, Ctx> = Terminal::Multi(self.inner.k(), self.inner.data().to_owned());
+        let term: Terminal<Pk, Ctx> = Terminal::Multi(self.inner);
         let ms = Miniscript::from_ast(term)?;
         // This would check all the consensus rules for p2sh/p2wsh and
         // even tapscript in future
-        Ctx::check_local_validity(&ms).map_err(From::from)
+        Ctx::check_local_validity(&ms).map_err(Error::from)?;
```
tcharding, line 43:
```suggestion
        Ctx::check_local_validity(&ms)?;
```

tcharding (approved):
Sans the `map_err` thing, ACK 1345fb2de2caa7d2def4acbf26757c872d62c8d6

src/policy/compiler.rs
```diff
@@ -1613,11 +1598,13 @@ mod tests {
...
-        let keys: Vec<Arc<Concrete<bitcoin::PublicKey>>> = keys
-            .iter()
-            .map(|pubkey| Arc::new(Concrete::Key(*pubkey)))
-            .collect();
-        let thresh_res = Concrete::Thresh(keys.len() - 1, keys).compile::<Legacy>();
+        let thresh = Threshold::from_iter(
+            keys.len() - 1,
+            keys.iter().map(|pubkey| Arc::new(Concrete::Key(*pubkey))),
+        )
+        .unwrap();
```
yancyribbens, line 1605: I notice there's a few unwraps() in this PR.  Just curious if there's a reason to do this instead of returning an Error type.

sanket1729: These are tests. `unwraps` are common in test and that is how we test most things.

sanket1729 (approved):
ACK 1345fb2de2caa7d2def4acbf26757c872d62c8d6

src/policy/compiler.rs
yancyribbens: Oops, I had meant to place this comment on the  [`generate_combination()`](https://github.com/rust-bitcoin/rust-miniscript/pull/674/files/1345fb2de2caa7d2def4acbf26757c872d62c8d6#diff-c88a4b1ba51d4a4eb96d972ee202e5b484fc29f573aee6655bd3982bf4a37ee6R1041) changes.  Heh yeah, I get using unwrap on the tests though.

sanket1729: Yeah, that was one occurrence which was a safe unwrap because of miniscript invariants around `Threshold`. We could add a expect with a message with the reasoning, but there was a short comment the commit message explaining that.

src/descriptor/sortedmulti.rs
apoelstra: Yeah, I could drop it here. In future though I want to move away from using `from` and explicitly write variants (e.g. `map_err(Error::ScriptContext)` or whatever).

But agreed that `.map_err(Error::from)?` is the worst of both worlds and I should drop it.

apoelstra:
Rebased to fix the `map_err` thing and also repaced the `unwrap` in `generate_combinations` with an `expect` that makes the reasoning explicit.

src/policy/compiler.rs
apoelstra: I added an `expect`. There was also a `debug_assert` which would have triggered before the `unwrap` did, but better to be explicit.

sanket1729 (approved):
ACK 79d3f7e15c90de9caad70ef72e1fd7f9cd92cc8e

tcharding (approved):
ACK 79d3f7e15c90de9caad70ef72e1fd7f9cd92cc8e

PR #676, opened 2024-04-06
Use threshold in Terminal and Satisfaction

apoelstra:
This completes the "threshold" conversion and completely drops the various ad-hoc error paths related to threshold values (most of which were just random strings).

The next set of PRs will be related to cleaning up errors in general a bit, to strongly-type things and add span information. But I am struggling a bit with how to box the associated error types for the MiniscriptKey types so it may be a while. (Very hard to do this in a std/nostd compatible way and the final result will probably be inconvenient/annoying for users.)

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -1498,7 +1498,7 @@ mod tests {
     #[test]
     fn empty_thresh() {
         let descriptor = Descriptor::<bitcoin::PublicKey>::from_str("thresh");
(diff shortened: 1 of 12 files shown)


tcharding (approved):
ACK 6925ae66b9fc9183652c14f660ce3b62ba0707d6

sanket1729 (approved):
ACK 6925ae66b9fc9183652c14f660ce3b62ba0707d6

PR #677, opened 2024-04-30
Fix compilation of pk()-only policies into tr() descriptors

shesek:
Before this fix, the added test failed with: `Unexpected("Empty Miniscript compilation")`

### DIFF
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -1120,6 +1120,14 @@ mod compiler_tests {
             .collect::<Vec<_>>();
         assert_eq!(combinations, expected_comb);
     }
+
+    #[test]
+    fn test_tr_pk_only() {
+        let policy: Policy<String> = policy_str!("pk(A)");
+        let desc = policy.compile_tr(None).unwrap();
+        // pk(A) promoted to the internal key, leaving the script tree empty
+        assert_eq!(desc.to_string(), "tr(A)#xyg3grex");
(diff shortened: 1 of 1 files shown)


tcharding:
Thanks for the contribution!

Can you put the unit test as a separate patch after the fix please so that during review I can swap them around and verify the failure.

shesek:
@tcharding Yes for sure, I updated my branch.

tcharding:
Bother, the CI fail is unrelated to this PR. I hit the same thing elsewhere yesterday, I'll push a PR to fix it (it requires dependency pinning).

sanket1729:
It's been a while since I looked at MSRV things. What is the current rust-bitcoin MSRV?

The CI is failing at 
> error: package `cc v1.0.96` cannot be built because it requires rustc 1.63 or newer, while the currently active rustc version is 1.56.1

tcharding:
Will need rebasing once #678 goes in.

shesek:
Rebased.

Note that #732 turned this from an Err to a panic.

apoelstra (approved):
ACK ea3c523b1c0b52069ca2b80e1b07bb275a3ebe9a successfully ran local tests; good find!

apoelstra:
I'm gonna go ahead and merge this. My local tests pass and I've been recently reading this code (hence my breaking it :)) so I'm fairly confident I understand it.

PR #678, opened 2024-04-30
CI: Pin cc for MSRV build

tcharding:
The `cc` dependency just stopped working for MSRV build with error

    error: package `cc v1.0.96` cannot be built because it requires
    rustc 1.63 or newer, while the currently active rustc version is
    1.56.1

Pin `cc` back to an earlier version for MSRV build.

### DIFF
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -9,8 +9,15 @@ rustc --version
 
 # Cache the toolchain we are using.
 NIGHTLY=false
+MSRV=false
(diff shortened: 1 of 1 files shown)


apoelstra:
Thanks for being on top of this!

And the new cargo error messages are nice. It's clear that it's a deliberate decision by upstream and it's clear what the issue is.

apoelstra (approved):
ACK fb0fc1d0e2d5908cb1733de726ebec500f873acf

PR #683, opened 2024-05-02
CI: Run fuzzer daily

tcharding:
Update the `fuzz/generate-files.sh` script to mimic that in `rust-bitcoin` and generate a new fuzz job that runs the fuzz tests daily for an hour.

Run the fuzz job the time that `rust-bitcoin` fuzz job finishes - I don't know if that matters. I still don't know if there are org wide limits or just repo limits but it doesn't hurt to run at a different time.

PR #684, opened 2024-05-02
Follow up from cron-daily-fuzz PR

tcharding:
I royally botched up PR #683 which changed fuzzing to run daily, fix it up by doing:

- Use the correct yaml file name in `generate-files.sh`.
- Actually run `generate-files.sh` to generate the yaml file.
- Fix the time UTC comment (the other times were correctly set).


sanket1729 (approved):
utACK 1f4b007f08ff44d9743150d565334d570210afc5

apoelstra:
In 1f4b007f08ff44d9743150d565334d570210afc5:

I get a different sort order than you for the fuzztests. I think we need to set the locale.

apoelstra:
Actually we currently don't sort these at all.. we just run `find` and take the raw output. We should sort them.

tcharding:
Done as an additional patch

apoelstra:
Guess we need to pin the nightly compiler here too :(

tcharding:
I was trying to get away from doing that in ever repo, I am not at all keen to be manually merging an "update msrv" patch on every repo we have twice a week, that is a job for a robot, or just kill me now. I mean, I can write a script that just does it but then we may as well have bot run the script.

apoelstra:
@tcharding As a matter of policy, I don't want bots to merge things unfortunately. But I see your point. This is entirely mechanical and pretty annoying to do.

Maybe if we made the update be every 2 weeks or something?

tcharding:
I can handle once a week, I realised also that the first repo is the annoying one (requires installing the new toolchain, which is slow). The additional repos should be less painful.

apoelstra:
@tcharding in the "usual" case that CI passes, you don't need to install the toolchain.

But yes, when things break, at least you only need to do it once..

tcharding:
Huh? As soon as it merges everyone needs to install the new toolchain to be able to build with nightly (at least if they use the `justfile` and/or want to run the same nightly that CI will run).

apoelstra:
Oh, interesting. I did not realize this. In my local CI I am always using the latest nightly (which does mean that the first job of the day is a bit slow) and on the CLI I have just been using whatever nightly I happen to have laying around, and only updating it when things change.

tcharding:
Ah of course, I an just add `rustup update` to the script I run every morning when I first hit the keyboard.

tcharding:
Ready to go!

apoelstra (approved):
ACK bd1df30a7944133374c4696f652515a8cdd1fba2

PR #682, opened 2024-05-02
CI: Use run_task from maintainer tools

tcharding:
This PR slightly improves test coverage by linting the integration tests, no meaningful change of coverage really though just improvement to the CI setup.

PR #686, opened 2024-05-15
bitcoind_test: Fix clippy

tcharding:
All the clippy patches rom #682, only touches `bitcoind_test`.

### DIFF
--- a/bitcoind-tests/tests/setup/test_util.rs
+++ b/bitcoind-tests/tests/setup/test_util.rs
@@ -159,7 +159,7 @@ impl<'a> Translator<String, DescriptorPublicKey, ()> for StrDescPubKeyTranslator
     fn pk(&mut self, pk_str: &String) -> Result<DescriptorPublicKey, ()> {
         let avail = !pk_str.ends_with('!');
         if avail {
-            self.0 = self.0 + 1;
+            self.0 += 1;
             if pk_str.starts_with('K') {
                 Ok(DescriptorPublicKey::Single(SinglePub {
                     origin: None,
@@ -212,7 +212,7 @@ impl<'a> Translator<String, DescriptorPublicKey, ()> for StrTranslatorLoose<'a> 
     fn pk(&mut self, pk_str: &String) -> Result<DescriptorPublicKey, ()> {
         let avail = !pk_str.ends_with('!');
         if avail {
-            self.0 = self.0 + 1;
+            self.0 += 1;
             if pk_str.starts_with('K') {
                 Ok(DescriptorPublicKey::Single(SinglePub {
                     origin: None,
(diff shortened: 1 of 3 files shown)


bitcoind-tests/tests/setup/test_util.rs
```diff
@@ -86,8 +86,8 @@ fn setup_keys(
     let mut x_only_keypairs = vec![];
     let mut x_only_pks = vec![];
 
-    for i in 0..n {
-        let keypair = bitcoin::secp256k1::Keypair::from_secret_key(&secp_sign, &sks[i]);
+    for sk in sks.iter().take(n) {
```
apoelstra, line 89: In b328736efdc0e07e8605e37c1b7526e6b6542113:

`iter` and `take` continue to be unnecessary

storopoli:
Should we squash all these commits down?

apoelstra:
@storopoli I see no particular reason to. They are all independent and compile and pass all tests (I assume -- I have not tested them because the second one needs to be changed).

bitcoind-tests/tests/setup/test_util.rs
tcharding: Its like I didn't even read your review the first time :(

tcharding:
I go to painful lengths to put all these changes separately so that its quick and easy for reviewers to review them [0]. Also its easy to drop any particular patch if a reviewer doesn't like it. Even the commit messages should be so uniform that they are easy to read quickly (and mechanical for me to write).

[0] Definitely pull me up if there are changes in the wrong place.

bitcoind-tests/tests/setup/test_util.rs
tcharding: Will fix, and lol "continue to be unnecessary" - love it.

tcharding: I don't know what was the original intention of the code but as it reads the `take(n)` is required because `n` is a function parameter so we can't just iterate over all of `sk`.

apoelstra: In the iteration above we construct `sk` by looping `n` times and pushing stuff onto `sk`.

We could add a `debug_assert` that `sk.len()` equals `n` if it makes you nervous.

tcharding: My bad, I did not read the code properly. Used `for sk in sks {` as suggested.

tcharding:
Neat, TIL:

`for f in foo` desugars to `for f in foo.into_iter()` and `for f in &foo` desugars to `for f in foo.iter()`

apoelstra (approved):
utACK b9a207bfd58b8f961c44976e6ff72b8940bf9f21

apoelstra:
Oops, that utACk should have been a full ACK. I did test every commit, it just took some manual prodding to avoid opening too many files at once.

ErnestoLara (approved):

ErnestoLara (approved):

tcharding:
Is there anything we can do about bot accounts triggering notifications like this @ErnestoLara idiot above? (Mentioning him incase he is a real person :)

PR #689, opened 2024-05-19
ci: shellcheck checks

storopoli:
Following https://github.com/rust-bitcoin/rust-bitcoin/pull/2762,
adding CI shellcheck cheks here as well.

I also did all fixes that I could find with

```bash
shellcheck **/*.sh
```

If I've missed any please let me know.

### DIFF
--- a/fuzz/cycle.sh
+++ b/fuzz/cycle.sh
@@ -7,7 +7,8 @@
 
 set -e
 REPO_DIR=$(git rev-parse --show-toplevel)
-# shellcheck source=./fuzz-util.sh
+# can't find the file because of the ENV var
+# shellcheck source=/dev/null
 source "$REPO_DIR/fuzz/fuzz-util.sh"
 
 while :
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,16 @@
+
+name: Shellcheck
+on:
+  pull_request:
+    branches:
+      - master
+jobs:
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Run ShellCheck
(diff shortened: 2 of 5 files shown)


fuzz/cycle.sh
```diff
@@ -7,7 +7,8 @@
 
 set -e
 REPO_DIR=$(git rev-parse --show-toplevel)
-# shellcheck source=./fuzz-util.sh
+# can't find the file because of the ENV var
+# shellcheck source=/dev/null
```
tcharding, line 11: Why do you silence this like this man instead of? 
```
# shellcheck disable=SC1090
```

(I hit the same warning working on https://github.com/rust-bitcoin/rust-secp256k1/pull/699.)

tcharding (approved):
ACK 982e3115c7f7dc60ee9eea18d8b45a462f741ae1

fuzz/cycle.sh
storopoli: It was the same approach I did in the original `rust-bitcoin` PR and this is the way to adopt a Exception per the shellcheck wiki: https://www.shellcheck.net/wiki/SC1090

apoelstra:
Can you please rebase this? It is currently based on a version of rust-miniscript from February (prior to the rust-bitcoin upgrade) and my local lockfiles have changed a lot since then.

(This should become less of a problem soon once we bring lockfiles into the repo itself.)

storopoli:
@apoelstra done

apoelstra (approved):
ACK 926bb05cfba6c0c94375aab955e2b6f4616ba0ee

fuzz/cycle.sh
tcharding: No worries, cheers.

tcharding (approved):
ACK 926bb05cfba6c0c94375aab955e2b6f4616ba0ee

PR #697, opened 2024-06-15
plan: make LoggerAssetProvider constructible

apoelstra:
It looks like we have an example AssetProvider that is intended to be used for testing/debugging. It looks useful but evidently has never been used because it was impossible to construct.

Recent versions of rust nightly notice this and output a warning about this, which is causing CI to fail, so we need to fix it.

My solution is to make the inner `Assets` public (and also a reference, since there is no need to take ownership) so that people can easily construct and deconstruct the type.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -344,7 +344,7 @@ impl<E> TranslateErr<E> {
     /// - SegwitV0 does not allow uncompressed keys and x_only keys
     /// - Tapscript does not allow uncompressed keys
     /// - Translating into multi-path descriptors should have same number of path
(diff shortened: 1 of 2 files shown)


tcharding (approved):
ACK c79639d8ef016fe2d7291d1a826fdea8d47c91a2

apoelstra:
cc @sanket1729

sanket1729 (approved):
ACK c79639d8ef016fe2d7291d1a826fdea8d47c91a2

sanket1729:
@apoelstra, need to merge this before #704 .

PR #704, opened 2024-07-08
Track tree depth to avoid stack overflow

sanket1729:
We track tree depth at the time of parsing, but that is insufficient because wrappers also count towards the stack limit when creating `Miniscript` structure. 

We could also fix this by fixing the wrappers to be limited to size 10 or something(wrappers are only for transformation of types, they don't provide any additional functionality). 

Long term we want to move towards to complete non-recursive parsing, this is a good stop gap solution to address the current issues.

PR #706, opened 2024-07-11
Backport 11.x of #704 

sanket1729:


apoelstra:
Can you change the CHANGELOG to drop the thing about `LoggerAssetProvider` because I don't think that's true of the backports.

sanket1729:
@apoelstra done

sanket1729:
@apoelstra this is still unreleased.

apoelstra (approved):
ACK fd823e8ab3302b180673f1893d524d44fb0c54f2

apoelstra:
Yep, sorry -- was hacking on my local CI to get it to work with 11.x -- but I was able to succeed, which is great. We have far better QA now on our backports than we used to. And after #682 I think that going forward, actual github CI should work for 12.x forever.

Can we do a backport to 9.x? Then we should be done as far as we intend to go.

sanket1729:
> Can we do a backport to 9.x? 

On it. Will do in an hour or so

apoelstra:
Tagged and published.

PR #708, opened 2024-07-15
9.x backport for recursion tree depth

sanket1729:
I had to comment some test cases because `bitcoind` had a breaking change and checking for a compatible version would have resulted in pinning it's dependencies too.


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "miniscript"
-version = "9.0.2"
-authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>, Sanket Kanjalkar <sanket1729@gmail.com>"]
+version = "9.1.0"
+authors = ["Andrew PoelstrPa <apoelstra@wpsoftware.net>, Sanket Kanjalkar <sanket1729@gmail.com>"]
```
apoelstra, line 4: In 5b1df514dab84f016d0b7f6758b51c025eef5a9f:

You injected a typo into my name :)

apoelstra:
I can't actually test this one because the MSRV is too messed up (it seems like some deps of bitcoind have updated to no longer work on 1.41.1 **and** they yanked the old versions?). But the tests pass on 1.64.0 (not with latest stable because we have `#[deny(dead_code)]` on and there is newly-detected dead code..) so I'm ok with this. But I would like you to fix the typo.

sanket1729:
Force pushed

apoelstra (approved):
utACK 9922e37220300e8b5840887d6d2fb0620bf038de

apoelstra:
Tagged and published.

PR #710, opened 2024-07-18
rustfmt: Use show_parse_errors

jamillambert:
Following on from the [comment in rust-bitcoin #3006](https://github.com/rust-bitcoin/rust-bitcoin/pull/3006#issuecomment-2226342485), 
Quoting tcharding (link removed to prevent spamming rustfmt PR):

> There are no behaviour changes with this patch applied.
> 
> In rust-lang/rustfmt 5961 the `hide_parse_errors` was changed to `show_parse_errors` with default `true`.
> 
> Since we run the formatter with nightly we are getting the following warning
> 
> Warning: the `hide_parse_errors` option is deprecated. Use `show_parse_errors` instead
> 
> Rename the variable and keep the default value `true`.


tcharding (approved):
ACK 9e8505d8427446ef9fada457bce269635c1b23cf

PR #712, opened 2024-07-21
Fix panic while decoding large Miniscripts from Script

sanket1729:
Use private module to guard all constructors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# # 12.2.0 - July 20, 2024
+
+- Fix panics while decoding large miniscripts from script [#712](https://github.com/rust-bitcoin/rust-miniscript/pull/712)
+
 # 12.1.0 - July 9, 2024
 
 - Make `LoggerAssetProvider` constructible [#697](https://github.com/rust-bitcoin/rust-miniscript/pull/697)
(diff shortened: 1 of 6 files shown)


apoelstra:
The new lockfiles don't work with MSRV.

sanket1729:
@apoelstra this is ready for review.

apoelstra:
The new lockfile updates `cc` to 1.1.6  which still does not work  with MSRV.

Ok if I force-push to this PR to try to fix the lockfiles?

sanket1729:
@apoelstra I just tested it again with 1.56.1 and it works. 

image

apoelstra:
Re "CI happy means we're good", see #719 -- we are actually testing only with Cargo-minimal.lock, not Cargo-recent.lock (which is broken).

Furthermore, this lockfile commit makes several unnecessary changes and because it's in a separate commit from the one that changes Cargo.toml, also means that I cannot locally test every  commit of this PR.

apoelstra (commented):
Successfully ran local tests on 192c499232fe70f1a79433746223dcb91443e816.

apoelstra (approved):
ACK 192c499232fe70f1a79433746223dcb91443e816 successfully ran local tests

apoelstra:
Tagged and published.

PR #713, opened 2024-07-21
Fix panic while decoding large Miniscripts from Script

sanket1729:
Backport 11.x


apoelstra:
Tagged and published.

PR #719, opened 2024-08-05
Bump the MSRV

tcharding:
As we have done in `rust-bitcoin` bump the MSVR to Rust 1.63.0

Also, run the MSRV CI job with both lockfiles (which are currently different).


storopoli:
As you did with other bumps, you forget to update the README 😂

tcharding:
Damn, I have a grep alias that ignores md files - I have to remember not to use it when doing MSRV bumps. Thanks man.

storopoli (approved):
ACK add2fc3660683bd460ca168e0e98b60ebd8ccfcf

tcharding:
Force push is just a change to the docs on the MSRV job, sorry to require re-ack @storopoli (found while rebaseing #694).

apoelstra:
Could you add a commit to the start of this which bumps the honggfuzz dep in both lockfiles?

It doesn't really have anything to do with this PR but it's a "toolchain update" and it'd simplify my local CI runs.

tcharding:
I'm not sure why the hongfuzz change causes `just update-lock-files` to do so much. (Running it on master does not change the lock files.)

storopoli (approved):
ACK 3916522e416dc37e974b66b374c93e4115b74610

apoelstra:
6a56d371ef93c122c7c7d9977ba5020eec20993a seems weird, and it bumps the versions of stuff in Cargo-minimal.lock that should continue to work. I'd prefer you just updated `honggfuzz` and nothing else, using the normal `cargo update -p honggfuzz --precise 0.5.56` mechanism.

And the reason why is that honggfuzz is both a dependency and a binary that the developer needs to have installed on their system, and the versions of these two things need to match exactly, and my CI scripts have the latest version since we updated rust-bitcoin and others. And also the latest version fixes a bug with `memcpy` which was messing up my fuzzing in Simplicity and likely messing up fuzzing here.

tcharding:
Ah I see, thanks for the explanation. Will re-spin

apoelstra (approved):
ACK c95b41168933c118d0e30009ddb304f80c43f7b5 successfully ran local tests; glad I got #712 in first so I could do the release!

tcharding:
Want me to do a release PR?

apoelstra:
No, I just released 12.2 right before merging this.

PR #720, opened 2024-08-10
CI: Use specific revision of maintainer-tools

tcharding:
We don't want to pull the maintainer tools repo from master otherwise we cannot patch it, we should use a specific revision.

This was an oversite in the original work when introducing the maintainer tools `run_task` script.

Use commit:

b2ac115 Merge rust-bitcoin/rust-bitcoin-maintainer-tools#4: Add CI shell scripts

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,6 +23,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: rust-bitcoin/rust-bitcoin-maintainer-tools
+          rev: b2ac115
           path: maintainer-tools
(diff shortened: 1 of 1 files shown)


sanket1729 (approved):
ACK dc7ed26003869f36aad41bb0a3cb777cf30e723b

apoelstra (approved):
ACK dc7ed26003869f36aad41bb0a3cb777cf30e723b successfully ran local tests; though once it stabilizes a bit we may want to start adding tags to maintainer-tools (maybe with just the date) to make it more visually obvious when we get out of date

PR #722, opened 2024-08-18
Eliminate a bunch more recursion; expand the `TreeLike` iterator trait a bit

apoelstra:
This PR replaces the recursive derived implementations of `PartialEq`, `PartialOrd`, `Hash`, `fmt::Debug` and `fmt::Display`. Along the way it expands the `iter::TreeLike` trait to make it a bit more useful (adding a `Ternary` variant and making the existing `Nary` variant generic as long as you can get a length out of it and index into it).

The new `fmt::Debug` method works using a new `DisplayNode` wrapper around `Terminal` which allows iterating over a script in the same way as it's displayed, treating things like `or_i(0, X)` as `l:X` and implementing the `c:pk_k` aliases and so on. This seems generally useful but this PR does not expose it in the public API because I'd like to make some breaking changes to `Terminal` down the line and don't want to expand the API.

The new `Ord` impl orders things alphabetically (assuming it is bug-free at least) by using the same iterator as the display logic. This is both nonrecursive and a more useful ordering for anybody who cares about the exact ordering. On the other hand, if anybody is depending on the *existing* ordering this will break their code. I would assume not, since the existing ordering is a weird ad-hoc thing based on the order that `derive(PartialOrd)` happens to use.

Similarly, it adds a `Terminal::fragment_name` accessor which returns a `&'static str` representing the fragment name as displayed. This is also not put into the public API.

The block of type parameters used in the debug output of Miniscripts is moved into a `fmt::Display` impl on `Type` itself. This is part of the public API.

This PR does **not** replace the `Clone` impl, which is currently derived. This impl actually isn't recursive, since it just clones the `Arc`s in the first layer of the script. Maybe we want to implement this manually and do a "deep copy"? This would be useful for users who have keys with interior mutability or something. I have a followup which does this, but didn't include it because I felt it might be controversial and need its own discussion.

This PR also does **not** replace the implicit `Drop` impl, because that's hard to do and I haven't done it. But we're making progress.


apoelstra (commented):
Successfully ran local tests on d0c0c14ae26e353c950319182fd853ee90b8eb54.

sanket1729:
> Maybe we want to implement this manually and do a "deep copy"? This would be useful for users who have keys with interior mutability or something. I have a followup which does this, but didn't include it because I felt it might be controversial and need its own discussion.

I don't see any harm "deep copy". Or we can wait for someone to complain or explicitly request that feature. I can forsee users using keys with interior mutability when scanning funds from descriptors keys containing wildcards.

src/miniscript/display.rs
```diff
@@ -0,0 +1,314 @@
...
+        }
+    }
+}
+
+#[derive(Copy, Clone)]
+enum DisplayTypes {
+    /// Display no types.
+    None,
+    /// Display all types, including the initial type.
+    All(Type),
```
sanket1729, line 145: nit:

In cccaaec8bc189f5dcf8212c40d0f752ce90d9063, why do we need `Type` argument here? We are always calling this with `self.ty`?

sanket1729 (approved):
ACK d0c0c14ae26e353c950319182fd853ee90b8eb54. One nit, but we are good.

src/miniscript/display.rs
apoelstra: @sanket1729 yes, but the `self` of `self.ty` is a `Miniscript` while the `self` of `conditional_fmt` is a `Terminal`. So we do need to pass in `self.ty` explicitly like this. It's not available from `conditional_fmt` otherwise.

apoelstra:
> I don't see any harm in having a "deep copy". Or we can wait for someone to complain or explicitly request that feature. I can forsee users using keys with interior mutability when scanning funds from descriptors keys containing wildcards.

Nice. I'll PR it since I've already got the code written.

src/miniscript/display.rs
sanket1729: Ah, makes sense

PR #724, opened 2024-08-19
miniscript: non-recursive Clone

apoelstra:
This is mostly just annoying/mechanical code. But it needs to be written because the derive(Clone) logic, I think, was only doing a "shallow clone" by cloning the Arcs in the Miniscript.

In my view this is undesirable behavior. If users want to do a shallow clone, they can hold an Arc<Miniscript> and call Arc::clone on that. But there are some cases where they might want to do a deep clone, and currently there isn't really any way to do so (you can do it with the translator API and an "identity" translator, but this is awkward to do and will be slow because it redoes typechecking and rechecks for duplicate keys).

### DIFF
--- a/src/miniscript/decode.rs
+++ b/src/miniscript/decode.rs
@@ -115,7 +115,6 @@ enum NonTerm {
 ///
 /// The average user should always use the [`Descriptor`] APIs. Advanced users who want deal
 /// with Miniscript ASTs should use the [`Miniscript`] APIs.
-#[derive(Clone)]
 pub enum Terminal<Pk: MiniscriptKey, Ctx: ScriptContext> {
     /// `1`
     True,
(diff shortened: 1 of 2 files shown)


apoelstra (commented):
Successfully ran local tests on 0d1a59a570cde5eb8e365b25a83e91179ee400a1.

sanket1729 (approved):
Looks good to me. Will ACK in a min after testing compiler benchmarks. I don't know how much they rely on cheap clones.

sanket1729 (approved):
ACK 0d1a59a570cde5eb8e365b25a83e91179ee400a1. 


No difference in compiler performance. We are using Arc<AstElemExt> there, and not Arc<Miniscript>

PR #725, opened 2024-08-20
Eliminate more recursion, and Liftable impl for Terminal

apoelstra:
This PR does a few things which are mostly related:

* Adds unit tests and does a couple cleanups.
* Introduces a new right-to-left post-order iterator and uses that to eliminate a bunch of Arc::clones and extra memory usage from existing non-recursive algorithms
* Fixes a bug in `Miniscript::substitute_raw_pkh` where the function completely didn't work (this fell out of the right-to-left change)
* Converts `Miniscript::lift` to be non-recursive
* Eleminates `Terminal::lift` because this impl is conceptually confused (`Terminal` as a type should never be used except to construct `Miniscript`s; they are not typechecked or sanity-checked and have no capacity to store type information). It was originally implemented as a convenience to help the `Miniscript` implementation but we don't need it now

It also stops using duplicate keys in a couple unit tests. Right now our "sanity checking" logic for things like duplicate keys is super ad-hoc and inconsistent. For example, the duplicate-key check is implemented three times (in miniscript::analyzable, policy::semantic and policy::concrete) and the check is actually done in somewhat-random places. This PR doesn't change that, but since I was touching unit tests I figured I'd clean it up.

### DIFF
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -1409,6 +1409,23 @@ mod tests {
     }
 
     #[test]
+    fn duplicate_keys() {
+        // You cannot parse a Miniscript that has duplicate keys
+        let err = Miniscript::<String, Segwitv0>::from_str("and_v(v:pk(A),pk(A))").unwrap_err();
+        assert_eq!(err, Error::AnalysisError(crate::AnalysisError::RepeatedPubkeys));
+
+        // ...though you can parse one with from_str_insane
+        let ok_insane =
+            Miniscript::<String, Segwitv0>::from_str_insane("and_v(v:pk(A),pk(A))").unwrap();
+        // ...but this cannot be sanity checked.
+        assert_eq!(ok_insane.sanity_check().unwrap_err(), crate::AnalysisError::RepeatedPubkeys);
+        // ...it can be lifted, though it's unclear whether this is a deliberate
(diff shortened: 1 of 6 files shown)


apoelstra (commented):
Successfully ran local tests on 8e2d90c50105d03fdc3e52c18d99c488886aa410.

src/miniscript/mod.rs
```diff
@@ -1408,6 +1408,39 @@ mod tests {
         ms.translate_pk(&mut t).unwrap_err();
     }
 
+    #[test]
+    fn mixed_timelocks() {
```
sanket1729, line 1497: I am surprised that this did not exist before. Anyways more tests are always welcome.

sanket1729 (approved):
ACK 8e2d90c50105d03fdc3e52c18d99c488886aa410

sanket1729:
Thanks for adding the tests.

src/miniscript/mod.rs
apoelstra: Possibly one existed somewhere. I grepped for the timelock-specific errors and found nothing. But there are a ton of tests that just assert `.unwrap_err()` and don't check more specifically.

And I'm pretty confident that there is no test that covers all three separate checks (in `Miniscript`, `Semantic` and `Concrete`).

PR #729, opened 2024-08-26
miniscript: add an unit test for substitute_raw_pkh() 

pythcoiner:
miniscript: add an unit test for substitute_raw_pkh()

### DIFF
--- a/src/iter/tree.rs
+++ b/src/iter/tree.rs
@@ -104,6 +104,14 @@ pub trait TreeLike: Clone + Sized {
     fn post_order_iter(self) -> PostOrderIter<Self> {
         PostOrderIter { index: 0, stack: vec![IterStackItem::unprocessed(self, None)] }
     }
+
+    /// Obtains an iterator of all the nodes rooted at the DAG, in right-to-left post order.
+    ///
+    /// This ordering is useful for "translation" algorithms which iterate over a
+    /// structure, pushing translated nodes and popping children.
+    fn rtl_post_order_iter(self) -> RtlPostOrderIter<Self> {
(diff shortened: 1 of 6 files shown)


sanket1729:
@pythcoiner, I believe #725 already solves this. Can you try building your test on top of #725?

sanket1729:
We should still keep this PR with the test. The test is valuable and thanks for your contribution and bug report.

pythcoiner:
> @pythcoiner, I believe #725 already solves this. Can you try building your test on top of #725?

sure!

pythcoiner:
rebased on top of #725

apoelstra:
Running CI -- but probably going to close this because #725 already includes a fixed unit test for this function.

Though if CI fails here, that would be really interesting.

Thanks for noticing this and for writing the test though!

apoelstra:
Ok, CI is failing due to a new lint and a lack of pinning in this crate. That's not so interesting :).

The new unit test passes, which is what we expected.

apoelstra:
#725 is in which should fix the problem. @pythcoiner would you like me to backport the fix to the latest-released 12.x? We aren't planning a new major release for a little while.

pythcoiner:
> #725 is in which should fix the problem. @pythcoiner would you like me to backport the fix to the latest-released 12.x? We aren't planning a new major release for a little while.

up to you, i'm actually using my fork

apoelstra:
Given the choice, I'd rather not backport :). This function has been totally broken for years, and you are the first person to file a bug about it (curiously, only a few days after I stumbled into it myself, just reading the code).

PR #731, opened 2024-08-30
fix: possible absolute locktime overflow error

ChrisCho-H:
Maximum absolute locktime is 2^31 - 1, not 2^31 in miniscript.

reference: https://github.com/bitcoin/bitcoin/blob/master/src/script/miniscript.h#L2281


apoelstra:
Good catch!

apoelstra (approved):
ACK d01ad3f45414fb1f90aa546220b130535e412fc4 successfully ran local tests

PR #732, opened 2024-08-30
Refactor `extract_key` logic in Taproot compiler

apoelstra:
Right now our Taproot compiler does some complicated analysis to identify a key within a policy that may be extracted to use as an internal key. Specifically, if you look at the "disjunction tree" at the root of a policy, every leaf of this tree can be a tapbranch and any leaf which is just a pubkey-check and nothing else is a candidate to be the internal key.

However, our logic to do this is convoluted, has many unnecessary error paths, and lots of gratuitous allocations (which ironically make the code harder, not easier, to follow).

This PR smooths all that out by introducing a new iterator over the potential tapbranches of a policy (along with their probabilities). It also cleans up much of the error handling in the policy module, eliminating 3 calls to the stringly-typed `errstr` function.

### DIFF
--- a/.github/workflows/cron-daily-fuzz.yml
+++ b/.github/workflows/cron-daily-fuzz.yml
@@ -19,6 +19,7 @@ jobs:
         # over that limit with fuzzing because of the hour run time.
         fuzz_target: [
 compile_descriptor,
+compile_taproot,
 parse_descriptor,
 parse_descriptor_secret,
 roundtrip_concrete,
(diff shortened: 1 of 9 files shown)


sanket1729 (approved):
ACK a195c8197c7029d13110cfb9c71df2fe2db5e079

Awesome. Thanks for the code cleanup.

apoelstra (commented):
Successfully ran local tests on a195c8197c7029d13110cfb9c71df2fe2db5e079.

shesek:
> We have a few error returns that are impossible to hit:
> ...
> \* An error return from with_huffman_tree which can only happen if it's given an empty input (impossible)

This one was not actually impossible. :sweat_smile: See #677 for a test (that now panics) and a fix.

PR #733, opened 2024-08-30
Remove `TranslatePk` trait and clean up `Translator` trait

apoelstra:
This may be too much of a breaking change and an instance of https://github.com/rust-bitcoin/rust-bitcoin/issues/3166, but the existing traits are *really* annoying to use. But this change is not essential to any of my other work so feel free to concept NACK it.

There are two changes here:

* This drops the `TranslatePk` trait which had no business existing. It attempts to be generic over "things whose keys can be translated", but is missing impls (on keys themselves, for example) and anyway it is basically impossible to usefully be generic over this trait since it has an associated `Output` type which is unconstrained except by a comment saying "this must be `Self<Q>`. Really, the only purpose of this trait is to force users to write `use miniscript::TranslatePk` every time they want access to the `translate_pk` method.
* It moves two of the generics in `Translator` from generics to associated types. This makes it far more ergonomic to implement and require the trait, since you no longer need to write 3 separate generics everywhere when two are implied. (Actually, all three are implied, including the source `Pk` type, but in practice users want to constrain this to match an existing key type that they've got. So I left it as a generic rather than an associated type.)

### DIFF
--- a/bitcoind-tests/tests/setup/test_util.rs
+++ b/bitcoind-tests/tests/setup/test_util.rs
@@ -155,8 +155,11 @@ pub fn parse_insane_ms<Ctx: ScriptContext>(
 #[derive(Debug, Clone)]
 struct StrDescPubKeyTranslator<'a>(usize, &'a PubData);
 
-impl<'a> Translator<String, DescriptorPublicKey, ()> for StrDescPubKeyTranslator<'a> {
-    fn pk(&mut self, pk_str: &String) -> Result<DescriptorPublicKey, ()> {
+impl<'a> Translator<String> for StrDescPubKeyTranslator<'a> {
+    type TargetPk = DescriptorPublicKey;
+    type Error = core::convert::Infallible;
+
+    fn pk(&mut self, pk_str: &String) -> Result<Self::TargetPk, Self::Error> {
         let avail = !pk_str.ends_with('!');
         if avail {
             self.0 += 1;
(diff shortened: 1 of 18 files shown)


apoelstra (commented):
Successfully ran local tests on 269f44d2618c8ad0c6ef6475d68e68c45256bc93.

sanket1729 (dismissed):
ACK 269f44d2618c8ad0c6ef6475d68e68c45256bc93.

Definitely makes the translate API easier to use.

sanket1729:
@apoelstra, the integration tests also need to be updated.

apoelstra (commented):
Successfully ran local tests on 380a0d3a92fdd1a525015442b238d480a610f91f.

apoelstra:
Current CI failure is because of bad doctest links. My local CI should be catching this. Will investigate..

apoelstra:
Got it. I was forcing `no-std` on any feature list that didn't have `std` in it, including the `default`-only list. But I only run the doc-link check (and clippy and fmt and a couple other things) when the feature-list is exclusiely `default` (otherwise I'd be running these checks for every feature combination, which is pointless because none of these checks use the feature list).

Will confirm that my local checks fail here, and then fix the PR.

apoelstra (commented):
Successfully ran local tests on 36b065981a10c30f37548f5df792885d5397730f.

sanket1729 (approved):
ACK 36b065981a10c30f37548f5df792885d5397730f

PR #735, opened 2024-08-31
[backport] Several small fixes for 12.x

apoelstra:
We are incorrectly serializing and_n as "and_b", and incorrectly parsing `tr(<key>,)`.

In master the `and_b` thing is fixed by https://github.com/rust-bitcoin/rust-miniscript/pull/722 which rewrites the Display impl completely and the `tr(<key>,)` thing will be fixed as part of a coming series of PRs to clean up expression parsing.

Also includes #740 since it showed up in time.

PR #740, opened 2024-09-03
fix: zero sequence check on relative time

ChrisCho-H:
I've just found out RelLockTime misses zero sequence check, which must be disallowed for miniscript.


apoelstra:
Great catch! I will also copy this commit into #735 so it gets backported to 12.x.

apoelstra (approved):
ACK c128e79477a0d7cb27563765b843b3cae97d6311 successfully ran local tests

PR #741, opened 2024-09-03
Remove `PartialEq` impl from `Error`

apoelstra:
This is a short PR in terms of LOC, but may be controversial. I think it is necessary though.

Currently we have `PartialEq` on our giant `Error` enum (though amusingly not `Eq`) so that we can use `assert_eq` in unit tests. We can get away with this because our error enum does not contain any `std::io::Error`s anywhere, which can't be compared, nor does it contain any boxed errors.

In the next couple of PRs I am going to clean up the `expression` module to make the parser non-recursive and the error types strongly typed and detailed. To do this, I want to store `<Pk as FromStr>::Err`s, which to avoid a `Pk` parameter on `Error`, will need to be boxed. When we box the type we need to declare upfront what traits the error types have, and these traits will become requirements of the `MiniscriptKey` type.

It's not really reasonable to require `PartialEq` or `Eq` since these are not common traits for errors. Really the only things we *can* reasonaby require are `Debug`, `Display` and `'static`. (I would like to also require `std::error::Error` but as far as I can tell there is no way to do this while retaining both nostd compatibility and additive features. See this [users.rust-lang.org thread about this](https://users.rust-lang.org/t/how-to-box-an-error-type-retaining-std-error-only-when-std-is-enabled/)).

### DIFF
--- a/src/blanket_traits.rs
+++ b/src/blanket_traits.rs
@@ -17,6 +17,11 @@ use core::{fmt, hash};
 
 use crate::MiniscriptKey;
 
+/// Auxiliary trait indicating that a type implements both `Debug` and `Display`.
+pub trait StaticDebugAndDisplay: fmt::Debug + fmt::Display + 'static {}
+
+impl<T: fmt::Debug + fmt::Display + 'static> StaticDebugAndDisplay for T {}
+
 /// Blanket trait describing a key where all associated types implement `FromStr`,
 /// and all `FromStr` errors can be displayed.
 pub trait FromStrKey:
@@ -36,7 +41,7 @@ pub trait FromStrKey:
         + fmt::Debug
         + hash::Hash;
     /// Dummy type. Do not use.
-    type _Sha256FromStrErr: fmt::Debug + fmt::Display;
+    type _Sha256FromStrErr: StaticDebugAndDisplay;
     /// Dummy type. Do not use.
     type _Hash256: FromStr<Err = Self::_Hash256FromStrErr>
         + Clone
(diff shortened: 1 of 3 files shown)


apoelstra (commented):
Successfully ran local tests on 57bbde8d3de1fb14d3478ced87f01cc70715e4a5.

tcharding:
Interesting, I was wondering about `PartialEq` and `Eq` on error types myself yesterday in an unrelated repo. Is the plan here in miniscript to eventually break all the errors up like we have been doing in `rust-bitcoin`?

apoelstra:
Within reason, yes.

src/blanket_traits.rs
```diff
@@ -17,6 +17,11 @@ use core::{fmt, hash};
 
 use crate::MiniscriptKey;
 
+/// Auxiliary trait indicating that a type implements both `Debug` and `Display`.
```
tcharding, line 20:
```suggestion
/// Auxiliary trait indicating that a type implements `'static`, `Debug` and `Display`.
```

tcharding:
Then this seems like a reasonable step IMO.

src/blanket_traits.rs
apoelstra: "implements `'static`" isn't the right terminology ... if we want to say this would say it "is `'static`".

But IMHO it's fine as-is. The name of the trait exactly describes what it does.

tcharding: Works for me.

sanket1729 (approved):
ACK 57bbde8d3de1fb14d3478ced87f01cc70715e4a5.

>  which to avoid a Pk parameter on Error, will need to be boxed

It will be super annoying to have `Pk` on `Error`. If we have to remove `PartialEq` for this; I think is it is worth it.

This will break downstream users; but hopefully the fix is easy for them. If we get strong complaints we implement some reasonable manual version of `PartialEq` and make a non-breaking point release.

PR #743, opened 2024-09-05
feat: actually check push data bytes

ChrisCho-H:
Miniscript witness item can be maximum 72 bytes(except length prefix), but it only throws error when it's >= 4294967296 bytes as only dependant on `PushBytesError` from bitcoin crate.  I've changed the logic to check the maximum length internally so that actual check can be done.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -2,6 +2,7 @@
 
 use core::convert::TryFrom;
 
+use bitcoin::constants::MAX_SCRIPT_ELEMENT_SIZE;
 use bitcoin::hashes::Hash;
 use bitcoin::script::{self, PushBytes, ScriptBuf};
 use bitcoin::PubkeyHash;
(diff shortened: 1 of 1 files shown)


apoelstra:
Oh, of course, the script itself could be too large! Ok concept ACK.

apoelstra:
....which I think is also what's triggering your new tighter `expect` in `witness_to_scriptsig`. Interesting. I think this is an existing panicking bug in rust-miniscript. (Though kinda an academic one since to trigger it in our existing code you need to construct a Miniscript which encodes to more than 4Gb.) (But upstream we are considering tightening this to 32M to match Bitcoin consensus rules, and in that case we need to fix this!!)

ChrisCho-H:
While it's true that it's impossible to hit this error in descriptor as `satisfy` called before, it's possible to hit in miniscript like `miniscript::Legacy::check_witness(witness)`. My trigger is actually just error message of `All pushes in miniscript are < 73 bytes`. It seems weird not to check maximum witness bytes with such error message.

ChrisCho-H:
I've added the logic to check the size of redeem script and other push separately https://github.com/rust-bitcoin/rust-miniscript/pull/743/commits/9dbaad3bfd89508a388222ba9d84b7c5c3c93fac.

apoelstra:
I would suggest squashing these two commits because they have substantial overlap in the code that they change.

Secondly, I'm struggling to understand the point of calling `map_err` and `and_then` on something we're just going to unwrap.

My suggestion, if you want to make the assertion more precise, would be to add a `debug_assert` above the push line, then change the expect message to just say "checked above".

Your method of checking whether we are on the last push is also pretty inefficient. One better way to do this is to add `.enumerate()` to the iterator and then compare the index to `witness.len() - 1`.

apoelstra:
It can also be a real `assert` if you want. No need to `debug_assert`.

ChrisCho-H:
Thanks for great review! I've fixed following your suggestion.

apoelstra:
324b33d08f86c1cc99c15d740f0cbbda03804152 is looking great!

But now it doesn't look like `witness_to_scriptsig` actually returns an error anywhere :).

Could this PR be reduced to just adding the new assertions and not changing the signature of `witness_to_scriptsig` at all?

ChrisCho-H:
You are right. No need to change the original signature. Reduced down by https://github.com/rust-bitcoin/rust-miniscript/pull/743/commits/e2e9281fcfe634d003405a7b5f6b4e63abcf1eda!

apoelstra (approved):
ACK e2e9281fcfe634d003405a7b5f6b4e63abcf1eda successfully ran local tests; thanks for iterating!

PR #745, opened 2024-09-08
Rewrite (non-compiler) benchmarks

apoelstra:
This PR introduces a much larger and more comprehensive set of benchmarks for parsing miniscripts. It now covers:

* Balanced and deep segwit scripts
* Balanced and deep Taproot scripts inside a single leaf (which are allowed to be much bigger than segwit ones)
* All of the above using thresh(2,A,B) as a combinator (which forces a Vec allocation and a bunch of different codepaths vs any of the other combinators)
* Balanced and deep Taproot trees with pk() as leaves
* A wide range of sizes for all of the above.

This doesn't touch the compiler benchmarks, mainly because it seemed like a bunch of work and I don't have any immediate intentions of messing with the compiler. But also because I'm not sure how to approach these -- with the parser I can just do giant trees of pk()s but with the compiler I probably want to use all of pk/hash/timelock etc.


sanket1729 (approved):
ACK b0d95003eef44333f824b5ad5e0f0b3e16cda3b7. 

Sorry for the delay

PR #746, opened 2024-09-09
fix: sync max multi_a pubkey with core

ChrisCho-H:
Although this is checked indirectly by checking MAX_STACK_SIZE, it'd be better to check in threshold explicitly as core does.


apoelstra:
o.O I had no idea this limit existed.

Great find!

apoelstra:
A few extra thoughts:

* The "max stack size" limit is enforced during sanity checking (and can therefore be disabled and eventually adjusted) while the threshold limits are enforced during parsing and in the Rust type system (cannot be adjusted)
* The threshold limits also provide a hard barrier on how much memory an individual Miniscript node can take and I'd like them to be as small as possible. (Though this is a bit silly since we have no limit on the `thresh` combinator...maybe we should On the basis that <k> and ADD each take at least one byte and each child takes at least one byte, it could be limited to 2MB?)

apoelstra (approved):
ACK 96d8662b70d1ea40d8ee850da68dac3e434f1220 successfully ran local tests

sanket1729 (approved):
ACK 96d8662b70d1ea40d8ee850da68dac3e434f1220

PR #747, opened 2024-09-11
refactor: check the node syntax before size limit

ChrisCho-H:
I’m not 100% sure if it’s intended, but the behavior of `check_global_consensus_validity`, which checks the script size limit first and then node could confuse the user-side.
For example, If `sh` or `wsh` descriptor uses `multi_a` with 999 public key(e.g. `descriptor::Wsh::from_str("wsh(multi_a(pk1...pk999)")`), it will throw an error on the size limit, not the fragment `multi_a`. It might be a bit more efficient in case of size overflow, but

1.  As miniscript functions as a language, syntax or grammar check should precedes stack limit. While the most of syntax are checked before `check_global_consensus_validity`, Syntax of which validity differ on context(e.g. `multi` and `multi_a`) are not.
2. It’s more helpful to debug for user if blocked by wrong syntax first.
3. It’s weird to check the script size which could be possibly wrong script upon context.

While descriptor itself is kinda helper tool which uses miniscript, it would worth checking syntax first.

### DIFF
--- a/src/miniscript/context.rs
+++ b/src/miniscript/context.rs
@@ -394,11 +394,8 @@ impl ScriptContext for Legacy {
     fn check_global_consensus_validity<Pk: MiniscriptKey>(
         ms: &Miniscript<Pk, Self>,
     ) -> Result<(), ScriptContextError> {
-        if ms.ext.pk_cost > MAX_SCRIPT_ELEMENT_SIZE {
-            return Err(ScriptContextError::MaxRedeemScriptSizeExceeded);
-        }
-
-        match ms.node {
+        // 1. Check the node first, throw an error on the language itself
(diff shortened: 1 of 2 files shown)


apoelstra:
concept ACK. Agreed this is a better order of operations for the user.

Can you add a unit test? I'm planning to significantly rewrite this code in the next few months and without a unit test I'm likely to accidentally revert this behavior.

CI failure looks unrelated to you. We don't pin our nightly compiler here (yet) and it looks like there's a new lint.

apoelstra:
CI failures are unrelated.

apoelstra:
LOL at how bad the existing error messages are. But that's fine, it's unrelated to this PR.

PR #748, opened 2024-09-15
fix: zero sequence check for from_height

ChrisCho-H:
This is following https://github.com/rust-bitcoin/rust-miniscript/pull/740.
I missed `from_height` also needs to check if zero.
I'm afraid as it's API breaking change.

### DIFF
--- a/src/interpreter/mod.rs
+++ b/src/interpreter/mod.rs
@@ -1211,7 +1211,7 @@ mod tests {
         assert_eq!(
             older_satisfied.unwrap(),
             vec![SatisfiedConstraint::RelativeTimelock {
-                n: crate::RelLockTime::from_height(1000).into()
+                n: crate::RelLockTime::from_height(1000).unwrap().into()
             }]
         );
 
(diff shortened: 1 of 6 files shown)


apoelstra:
CI failure is unrelated.

Concept ACK. I'm tempted to also suggest we provide infallible conversions to/from `NonZeroU16`, but the API for that type is super annoying, and there's also no equivalent type that can express the allowable range of `AbsLockTime`s so it'd make our API asymmetric.

@ChrisCho-H can you add a `from_height_unchecked` method which has the original behavior? You don't need to use it anywhere (it's longer to type than `from_height(x).unwrap()` after all) but I think some users want to grep for `unwrap` and have the output be clean.

ChrisCho-H:
I think `NonZeroU16` could be better option for the explicit and infallible usage, but as you mention it would be very inconvenient to use. The only failing case is when zero is given(which is unlikely to happen for the most cases), so we might better not to tighten interface that much.

apoelstra (approved):
ACK d030b8f54f425555023a34099aa57ca2bc1e9d88 successfully ran local tests

PR #750, opened 2024-09-20
DEMO: Hack the plan module

tcharding:
This is not a merge candidate but is just an investigation of the benefits of having a new `psbt-v0` crate.

The `plan` module currently includes a method on `Plan` that takes a `psbt::Input` as parameter. We would like to be able to either deprecate or remove the `psbt` module. In order to do so we would need to solve this problem. Here is one quick and dirty solution:

- Make the private `Plan` fields public
- Delete the method and make a standalone function in `psbt-v0`

This is related to https://github.com/rust-bitcoin/rust-bitcoin/discussions/3210 and was created to test https://github.com/tcharding/rust-psbt-v0/pull/6 and https://github.com/bitcoindevkit/bdk/pull/1621

### DIFF
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -20,11 +20,11 @@ use bitcoin::hashes::{hash160, ripemd160, sha256};
 use bitcoin::key::XOnlyPublicKey;
 use bitcoin::script::PushBytesBuf;
 use bitcoin::taproot::{ControlBlock, LeafVersion, TapLeafHash};
-use bitcoin::{absolute, bip32, psbt, relative, ScriptBuf, WitnessVersion};
+use bitcoin::{absolute, bip32, relative, ScriptBuf, WitnessVersion};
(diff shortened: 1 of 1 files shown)


apoelstra:
I would prefer a solution that just depended on `psbt-v0` here and kept the API intact (other than changing the types).

tcharding:
Ha! I jumped out of bed this morning after having that exact thought, does that mean I'm brilliant or you are?

tcharding:
Would you be open to adding public re-exports of `bitcoin::psbt` stuff here from `miniscript` and doing a minor release? I wrote this idea here: https://github.com/bitcoindevkit/bdk/pull/1621#issuecomment-2364197120

Only if this whole thing gets traction though of course.

tcharding:
Decided not to import `miniscript` stuff.

PR #751, opened 2024-10-10
descriptor: introduce several Taproot accessors

apoelstra:
When working with Taproot descriptors you typically need to do an annoying (and hard to discover) `match` statement to get the `Tr` out of the descriptor, and then call accessors on that to get the actual data out.

Add two new methods to `Descriptor` that directly access the internal key and the taptree. Document that the actual leaves can be obtained by calling `.iter` on the taptree.

Next, when a user is trying to sign a Taproot branch, they need to obtain a TapLeafHash. We have internal code which does this (which I have pulled into a helper function since there is some room to optimize it there..) but no exposed code, forcing the user to go digging through the rust-bitcoin docs to figure it out (including knowing the standard Taproot leaf version, which is an arcane detail of the sort that Miniscript otherwise hides).

Add a new method `leaf_hash` on Taproot miniscripts, so that the user can directly obtain the leaf hashes.

Now you can write e.g.

```rust
for script in trdesc.tap_tree_iter() {
    let leaf_hash = script.leaf_hash();
    // Do whatever you want...
}
```

vs the previous code which was roughly

```rust
let tr = match trdesc {
    Descriptor::Tr(ref tr) => tr,
    _ => unreachable!("I know this is a Taproot descriptor"),
};
// Or tr.tap_tree().unwrap().iter() in case you miss the weirdly-named // Tr::iter_scripts
for script in tr.iter_scripts() {
    // Hope you know your rust-bitcoin docs by heart, and also that
    // .encode is the way to convert a Miniscript to a Script!
    let leaf_hash = TapLeafHash::from_script(
        LeafVersion::TapScript,
        script.encode(),
    );
}
```

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -241,6 +241,40 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
         Ok(Descriptor::Tr(Tr::new(key, script)?))
     }
 
+    /// For a Taproot descriptor, returns the internal key.
+    pub fn internal_key(&self) -> Option<&Pk> {
+        if let Descriptor::Tr(ref tr) = self {
+            Some(tr.internal_key())
+        } else {
+            None
+        }
+    }
+
+    /// For a Taproot descriptor, returns the [`TapTree`] describing the Taproot tree.
+    ///
+    /// To obtain the individual leaves of the tree, call [`TapTree::iter`] on the
+    /// returned value.
+    pub fn tap_tree(&self) -> Option<&TapTree<Pk>> {
+        if let Descriptor::Tr(ref tr) = self {
+            tr.tap_tree().as_ref()
+        } else {
+            None
+        }
+    }
+
+    /// For a Taproot descriptor, returns an iterator over the scripts in the Taptree.
+    ///
+    /// If the descriptor is not a Taproot descriptor, **or** if the descriptor is a
+    /// Taproot descriptor containing only a keyspend, returns an empty iterator.
+    pub fn tap_tree_iter(&self) -> tr::TapTreeIter<Pk> {
+        if let Descriptor::Tr(ref tr) = self {
+            if let Some(ref tree) = tr.tap_tree() {
+                return tree.iter();
+            }
+        }
+        tr::TapTreeIter::empty()
+    }
+
     /// Get the [DescriptorType] of [Descriptor]
     pub fn desc_type(&self) -> DescriptorType {
         match *self {
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -730,7 +730,7 @@ impl Descriptor<DescriptorPublicKey> {
 
         struct KeyMapWrapper<'a, C: secp256k1::Signing>(KeyMap, &'a secp256k1::Secp256k1<C>);
 
-        impl<'a, C: secp256k1::Signing> Translator<String> for KeyMapWrapper<'a, C> {
+        impl<C: secp256k1::Signing> Translator<String> for KeyMapWrapper<'_, C> {
             type TargetPk = DescriptorPublicKey;
             type Error = Error;
 
@@ -778,7 +778,7 @@ impl Descriptor<DescriptorPublicKey> {
     pub fn to_string_with_secret(&self, key_map: &KeyMap) -> String {
         struct KeyMapLookUp<'a>(&'a KeyMap);
 
-        impl<'a> Translator<DescriptorPublicKey> for KeyMapLookUp<'a> {
+        impl Translator<DescriptorPublicKey> for KeyMapLookUp<'_> {
             type TargetPk = String;
             type Error = core::convert::Infallible;
 
@@ -941,7 +941,7 @@ impl Descriptor<DefiniteDescriptorKey> {
     ) -> Result<Descriptor<bitcoin::PublicKey>, ConversionError> {
         struct Derivator<'a, C: secp256k1::Verification>(&'a secp256k1::Secp256k1<C>);
 
-        impl<'a, C: secp256k1::Verification> Translator<DefiniteDescriptorKey> for Derivator<'a, C> {
+        impl<C: secp256k1::Verification> Translator<DefiniteDescriptorKey> for Derivator<'_, C> {
             type TargetPk = bitcoin::PublicKey;
             type Error = ConversionError;
 
--- a/src/descriptor/checksum.rs
+++ b/src/descriptor/checksum.rs
@@ -177,7 +177,7 @@ impl<'f, 'a> Formatter<'f, 'a> {
     }
 }
 
-impl<'f, 'a> fmt::Write for Formatter<'f, 'a> {
+impl fmt::Write for Formatter<'_, '_> {
     fn write_str(&mut self, s: &str) -> fmt::Result {
         self.fmt.write_str(s)?;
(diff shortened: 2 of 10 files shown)


apoelstra:
One still-missing piece of API surface is that there is no accessor for the BIP340 "standard unspendable key".

This is difficult to implement because there are currently no const constructors for XOnlyPublicKey, so I didn't include it here. Probably it will depend on updates to rust-secp to do key-parsing in pure Rust.

sanket1729:
Nice, on my reviewlist for this weekend.

shesek:
How about a `Descriptor::tap() -> Option<Tr>` method? Personally I would prefer using an explicit `.tap()` and chaining methods on that over the shortcut methods.

src/descriptor/mod.rs
```diff
@@ -241,6 +241,40 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
         Ok(Descriptor::Tr(Tr::new(key, script)?))
     }
 
+    /// For a Taproot descriptor, returns the internal key.
+    pub fn internal_key(&self) -> Option<&Pk> {
```
shesek, line 245: Perhaps name it `tap_internal_key()` to keep it consistent with the other methods and to make it clearer that its taproot-related and taproot-only?

src/descriptor/mod.rs
```diff
@@ -241,6 +241,40 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
+        if let Descriptor::Tr(ref tr) = self {
+            tr.tap_tree().as_ref()
+        } else {
+            None
+        }
+    }
+
+    /// For a Taproot descriptor, returns an iterator over the scripts in the Taptree.
+    ///
+    /// If the descriptor is not a Taproot descriptor, **or** if the descriptor is a
```
shesek, line 267: I find this kind of strange, I would prefer to differentiate these two cases so I can handle the 'not Taproot descriptor' as an error case if I was expecting it to be a Taproot.

But I guess that if the distinction is important, one could also explicitly use (my suggested) `tap()`, handle the `None` case, then iter.

src/descriptor/mod.rs
```diff
@@ -241,6 +241,40 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
+    /// Taproot descriptor containing only a keyspend, returns an empty iterator.
+    pub fn tap_tree_iter(&self) -> tr::TapTreeIter<Pk> {
+        if let Descriptor::Tr(ref tr) = self {
+            if let Some(ref tree) = tr.tap_tree() {
```
shesek, line 271: Why not use `tr.iter_scripts()`?

src/descriptor/mod.rs
apoelstra: The other methods yield a `TapTree` and are named as such, with no prefix. It's just more obvious that a "taptree" is a taproot object while an "internal key" is less obvious.

I can stick a `taproot_` prefix onto here if you want, though that makes the method name longer and won't make it look consistent with the other methods. Since Taproot is the "default" output type for the forseeable future, I would almost rather stick `non_taproot_` onto everything else..

src/descriptor/mod.rs
apoelstra: You can differentiate them by calling `.tap_tree()`, dealing with the Option, then calling `iter_scripts`. I added this method specifically to get rid of the extra `unwrap` that would require.

apoelstra: Ah, actually to distinguish a non-taproot descriptor from a keyspend-only descriptor, but without matching and without being able to distinguish further, you would need to call `.internal_key().is_some()`.

apoelstra:
@shesek I am not going to add a `.tap()` method that returns a `Tap`. Users have no idea what a `Tap` is because it's an awkwardly named struct that exists in the API purely to hide some implementation details of the `Descriptor::Tr` variant. From a user POV it consists of exactly two things: an internal key and an (optional) taptree. I have added accessors for both of those.

src/descriptor/mod.rs
shesek: If Taproot is considered the 'default' then not having a prefix definitely makes sense. I didn't think of it as such.

apoelstra:
Ah, good point. Also, I realize that the structure we're talking about is [`Tr`](https://docs.rs/miniscript/latest/miniscript/descriptor/struct.Tr.html), not `Tap`. And I like the name `Tr` much better since it reflects "the data of a `tr` descriptor". Ok, I will add a `tr` accessor. But I would still like to keep the set of accessors that I've added here, since they let you do common things on known-tr descriptors without needing unwraps.

I also see that there isn't a direct accessor for the external key. To get this from a descriptor you need to call `descriptor.tr().unwrap().spend_info().output_key()` which is pretty hard to discover.

Will you be in Atlanta for TABConf? Maybe we should try to whiteboard out all the Taproot data structures and try to come up with a sensible API to organize them.

src/descriptor/mod.rs
shesek: > I added this method specifically to get rid of the extra `unwrap` that would require.

In what scenarios would it make sense to have a code path that handles both taproot and non-taproot descriptors in that way? It seems to me that using `iter_tap_tree` on non-taproot descriptors would typically be a coding mistake, can't really think of a case where I'd want to do it and have it return a empty iterator

shesek: Ah, I see from your other comment that you were thinking about already known-taproot descriptors, say if the same code initializes the tr descriptor and calls methods on it. It can make sense for such cases.

apoelstra: Also, if you are trying to "just sign a transaction" you might take an arbitrary descriptor, grab its scriptcode to produce a pre-Taproot sighash, then also iterate through all the Taproot branches (if any) to produce Taproot sighashes. In this case you could use the same code for both Taproot and non-Taproot descriptors.

shesek:
> I also see that there isn't a direct accessor for the external key.

Same is also true for the merkle root. 

> Will you be in Atlanta for TABConf? Maybe we should try to whiteboard out all the Taproot data structures and try to come up with a sensible API to organize them.

Was not planning to... but maybe :eyes: 

I do have some ideas and some taproot stuff I implemented as extension traits that would be nice to have as part of rust-{miniscript,bitcoin}. We could also discuss it online (IRC?)

apoelstra:
Yep, I'm on IRC. Let's move there ##miniscript or #rust-bitcoin on Libera to discuss the bigger picture.

sanket1729 (approved):
ACK 86305f0ad499b486d00bea8e7a9dca14cfe7a836

apoelstra:
I'm going to merge this because I think it's a strict improvement. Will try to meet with @sanket1729 this week at TABConf and we can brainstorm the Taproot API.

apoelstra:
But for the record -- @shesek would like an accessor for the `Tr` object that doesn't require the user `match` themselves, and this PR does not provide that.

apoelstra:
At TABConf @sanket1729 and I worked out a new API which has all of the functionality of the existing `Tr` struct, its iterator, and the ``SpendInfo` structure, while being flatter and more discoverable. We also think it will be more efficient since the `SpendInfo` structure has a weird map that doesn't align well with Miniscript but is costly to instantiate.

I will open a new PR this week with a detailed description and implementation.

apoelstra:
miniscript-api

PR #752, opened 2024-10-11
 Document an error case for at_derivation_index()

shesek:


apoelstra (approved):
ACK 3d9acd2231a4a7560dcd871d1476831b2e6b0531 successfully ran local tests; thanks!

PR #753, opened 2024-10-14
Fix decoding of WIF with BIP32 origin

shesek:
A small fix for a small oversight.

The test has to pattern match the inner `SinglePriv` because there's no other way to get derivation path (no `full_derivation_path(s)` method like `DescriptorPublicKey` has, which would be nice to have ^.^)


apoelstra (approved):
ACK 9416098447ca816d08224804e40ed3c6f52b7edb successfully ran local tests; good catch!

PR #756, opened 2024-10-16
feat: add MaxBareScriptSizeExceeded err for bare consensus check

ChrisCho-H:
`BareCtx` consensus validity check returns `MaxWitnessScriptSizeExceeded` err when exceeding script size limit, which is wrong and confusing. Adding `MaxBareScriptSizeExceeded` and using it instead can clearly give a correct message on compile error.


apoelstra (approved):
ACK 834bd3dcc9bbcecd8603d85fc17ddb82f2d4ba65; successfully ran local tests; Yeah, this seems like a good idea

PR #757, opened 2024-10-17
Support conversion of multi-Xprivs into multi-Xpubs

shesek:
Possible when all hardened derivation steps are shared among all paths (or if there are none). Errors otherwise.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -227,20 +227,13 @@ impl DescriptorXKey<bip32::Xpriv> {
         let xpub = bip32::Xpub::from_priv(secp, &xprv);
 
         let origin = match &self.origin {
-            Some((fingerprint, path)) => Some((
-                *fingerprint,
-                path.into_iter()
-                    .chain(hardened_path.iter())
-                    .cloned()
-                    .collect(),
-            )),
-            None => {
-                if hardened_path.is_empty() {
-                    None
-                } else {
-                    Some((self.xkey.fingerprint(secp), hardened_path.into()))
-                }
+            Some((fingerprint, path)) => {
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 7d3e1e846ce75c86af3fed79300df5ce6247d404; successfully ran local tests; the stringly-typed errors are OK since this module is full of them

apoelstra:
Would like @sanket1729 to take a quick look at this before acking to sanity-check it.

In particular this forbits multixpubs where there are hardened derivations after the split. I assume this is a requirement to be a legal multixpub? Because there isn't any obvious reason we should be forced to have this limit.

shesek:
>  assume this is a requirement to be a legal multixpub? Because there isn't any obvious reason we should be forced to have this limit.

Correct, this is a limitation imposed by the current `DescriptorMultiXKey` structure (can only have one concrete pubkey as the base for derivations) and not something inherent. But supporting it would require some pretty invasive changes and it seemed to me that in the typical case hardened derivation steps are usually shared, and that it can be good enough with that limitation.

shesek:
Added another commit with some formatting changes that also updates the existing `DescriptorXKey::to_public()`. Let me know if you'd like me to squash or drop that commit.

sanket1729 (approved):
LGTM

sanket1729 (approved):
utACK ae64ce617d0e1bed0e25edda1fa13108c37e27c0

apoelstra (approved):
ACK ae64ce617d0e1bed0e25edda1fa13108c37e27c0; successfully ran local tests; the stringly-typed errors are OK since this module is full of them

PR #758, opened 2024-10-17
fix: differentiate max witness script size upon context

ChrisCho-H:
`MaxWitnessScriptSizeExceeded` is used in the context of `SegwitV0` and `Tap`, where each of max witness script size differs. Moreover, even in the same context of `SegwitV0`, max witness script size differ whether it's standard or consensus rule. I just let `MaxWitnessScriptSizeExceeded` receive param `usize` to differentiate max witness script size upon context, which can fix the wrong err message of `"The Miniscript corresponding Script would be larger than MAX_STANDARD_P2WSH_SCRIPT_SIZE  bytes."` when `SegwitV0` consensus and `Tap` context

### DIFF
--- a/src/miniscript/context.rs
+++ b/src/miniscript/context.rs
@@ -47,8 +47,9 @@ pub enum ScriptContextError {
     /// than `MAX_OPS_PER_SCRIPT`(201) opcodes.
     MaxOpCountExceeded,
     /// The Miniscript(under segwit context) corresponding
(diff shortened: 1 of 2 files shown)


apoelstra:
In 5a3989c1d1ad2ad89f93242e2b31dbbcb69df479:

Can you instead add two named fields to the error variant, `max` and `got`?

ChrisCho-H:
Good idea! I've changed err message including `got` bytes.

apoelstra (approved):
ACK 13a131819a1bea0088cce03e272a5980ced46413; successfully ran local tests

PR #760, opened 2024-10-19
feature: improve context err

ChrisCho-H:
919bc4b6c2ad37e94c446871de10a96c84b29f29
- use `ImpossibleSatisfaction` when op_count is None instead of `MaxOpCountExceeded`.

51a50ae3e81b02c4e8f4759df62186b53c5459ba
- unify error message for satisfaction-kind error and scriptPubKey-kind error so that these error can contain max(limit) with what it got(actual).


ChrisCho-H:
faf16fd75c3340e561ae694d0db7390601ab0d0f
- changed the err message which might be misleading!

apoelstra (approved):
ACK faf16fd75c3340e561ae694d0db7390601ab0d0f; successfully ran local tests

PR #764, opened 2024-10-28
Make xpub and xpriv identifiers match rust-bitcoin

tcharding:
Recently in `rust-bitcoin` we agreed that 'xpub' and 'xpriv' are well known terms in the Bitcoin vernacular so, being "words" should be written as `Xpub` and `Xpriv` when translated to Rust identifiers.

Make the descriptor key identifiers match `rust-bitcoin`.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -22,9 +22,9 @@ pub enum DescriptorPublicKey {
     /// Single public key.
     Single(SinglePub),
     /// Extended public key (xpub).
-    XPub(DescriptorXKey<bip32::Xpub>),
+    Xpub(DescriptorXKey<bip32::Xpub>),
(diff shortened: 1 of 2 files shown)


tcharding:
Jeeebus, I didn't realize that in descriptors we use `xprv` not xpriv - what a mess.

apoelstra:
This isn't a descriptor thing. `xprv` and `xpub` come from the base58 encoding of these objects, as defined in BIP32 in 2012 https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#serialization-format

apoelstra:
I don't mind changing the capitalization here.

But FWIW I don't think anybody uses the rust-bitcoin types if they have access to the rust-minscript types. So it's not the end of the world if they're inconsistent.

Issue #709, opened 2024-07-17
Implement GetKey for KeyMap


LLFourn:
See the trait here:

https://docs.rs/bitcoin/latest/bitcoin/psbt/trait.GetKey.html

I think it'd make sense if KeyMap (or something like it) implemented this trait so you could sign PSBTs with the secret keys extracted from a private descriptor.

apoelstra:
Oo, neat. Concept ACK. May be a little bit because we are rejigging the relationship between rust-bitcoin and rust-miniscript and I'm unsure what happens to PSBT during that transition.

tcharding:
Once `bitcoin` depends on `miniscript` hopefully we move [`psbt_v2`](https://crates.io/crates/psbt-v2) into `bitcoin` but that shouldn't effect this issue, we can solve this in either `miniscript::psbt` and port it or in `psbt_v2` if we have done the move already. (`psbt_v2` contains all the code from `psbt::miniscript` anyways.)

PR #765, opened 2024-10-29
Implement `GetKey` for `KeyMap`

tcharding:
Create a `KeyMay` type to replace the current `BTreeMap` alias and implement `bitcoin::psbt::GetKey` for it.

Close: #709


tcharding:
cc @LLFourn

apoelstra:
Nice! Yeah, this is probably the right approach.

In 00342a180cc6ad8906c9a905bf6dfa9d7412ef88 you have a bunch of `unreachable`s because you assume that public keys map consistently to secret keys. I think this is a reasonable assumption, but you need to enforce it in the `insert` method. Or perhaps you could change `insert` to just take a secret key, and do the pubkey computation itself.

Alternately, you could make `insert` be non-`pub` so that the type can only be constructed by calling `Descriptor::parse_descriptor`. This might be the better approach. (Though in this case I'd still add an assertion to `insert` checking consistency of the public and secret keys.)i

LLFourn:
Thanks @tcharding. It might be nice to be able to `append` your key maps onto each other so you keep one big jumble of secret keys across all your descriptors in the application. This is possible now but would be harder to do if you make the field private (which does seem like the right thing in general). So maybe an `.extend` with another keymap. If you make it private you probably want to make a way to iterate it too.

tcharding:
Yes I threw the `pub` inner field so as not to have to implement the whole `BTreeMap` API but private inner is probably better as you say - I can do this now we a have loose concept ack.

ValuedMammal:
Concept ACK

sanket1729:
> It might be nice to be able to append your key maps onto each other so you keep one big jumble of secret keys across all your descriptors in the application.

Maybe in the longer term, a cleaner solution is to implement rust-bitcoin to implement `GetKey` for a tuple, vec, set of things that already implement GetKey.

```rust
impl_get_key_tuple!((T1, T2))
impl GetKey for Vec<T> where T: GetKey {
}
impl GetKey for BtreeSet<T>...
```

sanket1729:
Concept ACK.

tcharding:
Implemented `iter::IntoIterator` and `iter::Extend`, would love someone else to give a few minutes to thinking if this is all of the iterator traits we need please.

Also made new take only the secret key. I did not remove any `unreachable` calls, nor did I add an assertion on the pk/sk since pk in insertion is now controlled.

src/descriptor/key_map.rs
```diff
@@ -0,0 +1,322 @@
...
+    ) -> Result<Option<bitcoin::PrivateKey>, Self::Error> {
+        Ok(self.map.iter().find_map(|(k, v)| {
+            match k {
+                DescriptorPublicKey::Single(ref pk) => match key_request {
+                    KeyRequest::Pubkey(ref request) => match pk.key {
+                        SinglePubKey::FullKey(ref pk) => {
+                            if pk == request {
+                                match v {
+                                    DescriptorSecretKey::Single(ref sk) => Some(sk.key),
+                                    _ => unreachable!("Single maps to Single"),
```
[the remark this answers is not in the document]
apoelstra: Lol, no

apoelstra:
Too bad. copilot has been around long enough that you'd think it would work, but 100% of its suggestions are wrong and it's full of irritating woke language like "harmful" and "problematic".

src/descriptor/key_map.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    fn get_key<C: Signing>(
+        &self,
+        key_request: KeyRequest,
+        secp: &Secp256k1<C>,
+    ) -> Result<Option<bitcoin::PrivateKey>, Self::Error> {
+        Ok(self.map.iter().find_map(|(k, v)| {
+            match k {
+                DescriptorPublicKey::Single(ref pk) => match key_request {
+                    KeyRequest::Pubkey(ref request) => match pk.key {
```
apoelstra, line 91: In d2c954aae82f23119bb5a9380bbcaeb77e3589d7:

I think if you replace

```
match k {
    Thing => match key_request {
        Thing2 => ...
```
with 
```
    match (k, key_request) {
        (Thing, Thing) => {
```

you can reduce the amount of indentation and should be able to reduce the number of branches.

sanket1729:
> Too bad. copilot has been around long enough that you'd think it would work, but 100% of its suggestions are wrong and it's full of irritating woke language like "harmful" and "problematic".

:) . Still too far. Let's try it in another 6 months.

src/descriptor/key_map.rs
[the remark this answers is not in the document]
tcharding: AI is so awesome I think we are all going to be out of a job soon.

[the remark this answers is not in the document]
yancyribbens: Not if your job is to have fun :)

tcharding:
FTR I had totally forgotten about this PR until @ValuedMammal tagged me. I tried but am unable to engage my brain right now to implement the review suggestion from Andrew (https://github.com/rust-bitcoin/rust-miniscript/pull/765#discussion_r1887219793).

tcharding:
> @apoelstra @tcharding wdyt about [#765 (comment)](https://github.com/rust-bitcoin/rust-miniscript/pull/765#issuecomment-2467421004)? I think that would solve future API requests like extend, merge etc

I have no opinion ATM sorry mate simply because I haven't managed to context switch into miniscript properly right now.

oleonardolima:
@tcharding I've been working on similar changes for https://github.com/bitcoindevkit/bdk_wallet/issues/70 in https://github.com/bitcoindevkit/bdk_wallet/pull/235, and I can continue the initial work from here and open another candidate PR. Is that fine?

I discussed with @ValuedMammal, and it seems that we could implement the `GetKey` for `DescriptorSecretKey` instead. Any previous thoughts on this?

tcharding:
I've no real opinion mate, I don't work much on miniscript. I actually don't even know if this PR can go in or requires changes.

src/descriptor/key_map.rs
```diff
@@ -0,0 +1,322 @@
...
+    /// Inserts secret key into key map returning the associated public key.
+    #[inline]
+    pub fn insert<C: Signing>(
+        &mut self,
+        secp: &Secp256k1<C>,
+        sk: DescriptorSecretKey,
+    ) -> Result<DescriptorPublicKey, DescriptorKeyParseError> {
+        let pk = sk.to_public(secp)?;
+        if !self.map.contains_key(&pk) {
+            self.map.insert(pk.clone(), sk);
```
apoelstra, line 40: In d2c954aae82f23119bb5a9380bbcaeb77e3589d7:

Both the `contains_key` check and the clone are unnecessary here.

src/descriptor/key_map.rs
```diff
@@ -0,0 +1,322 @@
...
+                                if let Ok(Some(sk)) = xpriv.xkey.get_key(key_request.clone(), secp)
+                                {
+                                    Some(sk)
+                                } else {
+                                    None
+                                }
+                            }
+                            _ => unreachable!("XPrv maps to XPrv"),
+                        },
+                        _ => unreachable!("rust-bitcoin v0.32"),
```
apoelstra, line 145: In d2c954aae82f23119bb5a9380bbcaeb77e3589d7:

This might not be `unreachable` for every version of 0.32.x. You should just return `None` here.

tcharding: Also are you ok with TODO in the codebase here in miniscript?

src/descriptor/key_map.rs
```diff
@@ -0,0 +1,324 @@
...
+                        }
+                        KeyRequest::Bip32(..) => match v {
+                            DescriptorSecretKey::XPrv(xpriv) => {
+                                // This clone goes away in next release of rust-bitcoin.
+                                if let Ok(Some(sk)) = xpriv.xkey.get_key(key_request.clone(), secp)
+                                {
+                                    Some(sk)
+                                } else {
+                                    None
+                                }
```
oleonardolima, line 141: The existing `get_key` implementation for `Xpriv` does not cover the scenario when the `KeyRequest::Bip32(..)` uses the key origin information, and it'd need to be covered here (it's a scenario that's being tested by BDK).

I handled it this way in my implementation, see:  https://github.com/oleonardolima/rust-miniscript/commit/4f304d1205a9bd0a87d8aafd81415d7716fb4bd5#diff-9b4da42f4bda05e0c63a0bae838b0eae32dfc2d5511a0c8142db8acfa9337aa2R142-R160 it's tested by https://github.com/oleonardolima/rust-miniscript/commit/4f304d1205a9bd0a87d8aafd81415d7716fb4bd5#diff-9b4da42f4bda05e0c63a0bae838b0eae32dfc2d5511a0c8142db8acfa9337aa2R391-R422

src/descriptor/key_map.rs
oleonardolima: Yes, at least it looks better. I tried this, see: https://github.com/oleonardolima/rust-miniscript/commit/4f304d1205a9bd0a87d8aafd81415d7716fb4bd5

src/descriptor/key_map.rs
apoelstra: Oh, yeah, the `clone` may be necessary.

But the check isn't doing anything except a pointless lookup.

src/descriptor/key_map.rs
apoelstra: Fine to have a TODO in the codebase.

oleonardolima:
> Thanks for the review @oleonardolima, I'll look into your suggestions.

@tcharding Any update on this one? I can open another PR and take over from there if it would be helpful.

tcharding:
Yes please that would be awesome. Take or leave anything from this PR mate, no attribution required but feel free to blame me for any bugs.

tcharding:
Picked up in #851

PR #766, opened 2024-11-02
Adding taptree_of_horror example.

miketwenty1:
Adding taptree_of_horror example.
~~Adding bitcoin_hashes as dev dependency.~~
~~Made some formatting consistency corrections in Cargo.toml.~~

PR #769, opened 2024-11-10
Remove no-std and actual-serde features

### LINKED ISSUE #681: We should remove "no-std" feature
In `rust-bitcoin v0.32.0` the "no-std" feature is gone, we can remove it here as well once we use that version.

JoseSK999:
Closes #681

The `no-std` feature is no longer needed with `bitcoin` version 0.32.0

- Update README.md (in Building I have copy-pasted part of the same section from `rust-bitcoin`)
- Disable `actual-serde` implicit feature

I think `FEATURES_WITH_NO_STD` now could be removed from the maintainer tools. Let me know if there's something missing or to improve!


apoelstra (dismissed):
ACK 8722d58228ceac23c0b48eb3af048d09972cd3b2; successfully ran local tests; nice!

Cargo.toml
```diff
@@ -14,19 +14,17 @@ rust-version = "1.63.0"
...
 rand = ["bitcoin/rand"]
 base64 = ["bitcoin/base64"]
 
 [dependencies]
-bech32 = { version = "0.11.0", default-features = false }
+bech32 = { version = "0.11.0", default-features = false, features = ["alloc"] }
 bitcoin = { version = "0.32.0", default-features = false }
 
-# Do NOT use this as a feature! Use the `serde` feature instead.
 actual-serde = { package = "serde", version = "1.0.103", optional = true }
```
tcharding, line 28: By looking at `rust-bitcoin` it looks like this can be renamed to just plain old `serde` (the `actual-serde` was only used because we didn't have `"dep:serde"` available).

Will also need 
```diff
 #[cfg(feature = "serde")]
-pub use actual_serde as serde;
+pub use serde;
```

README.md
```diff
@@ -27,24 +27,38 @@ are convertible to `bitcoin::PublicKey`
...
+```
+
+You can run tests with:
+
+```
+cargo test
+```
+
+Please refer to the [`cargo` documentation](https://doc.rust-lang.org/stable/cargo/) for more
+detailed instructions.
```
tcharding, line 54: These changes are unrelated to this PR, right? (FTR I never liked these docs in `rust-bitcoin`, if one doesn't know how to run Rust unit tests they probably shouldn't be writing financial software - and `rust-miniscript` is even more hardcore than `rust-bitcoin`.

README.md
```diff
@@ -27,24 +27,38 @@ are convertible to `bitcoin::PublicKey`
...
 
 ## Minimum Supported Rust Version (MSRV)
 
 This library should always compile with any combination of features on **Rust 1.63.0**.
 
 Some dependencies do not play nicely with our MSRV, if you are running the tests
-you may need to pin some dependencies. See `./contrib/test.sh` for current pinning.
+you may need to pin some dependencies. See `./contrib/pin.sh` for current pinning.
```
tcharding, line 62: This change is good, its an unrelated change for this PR though. (Sorry for being picky, we just uncovered a bug in `rust-bitcoin` that slipped through review so I'm trying to review more ~anally~ diligently.

contrib/test_vars.sh
```diff
@@ -7,12 +7,8 @@
 # Test all these features with "std" enabled.
 FEATURES_WITH_STD="compiler trace serde rand base64"
 
-# Test all these features with "no-std" enabled.
-# rust-miniscript only: https://github.com/rust-bitcoin/rust-miniscript/issues/681
-FEATURES_WITH_NO_STD="compiler trace serde rand base64"
-
 # Test all these features without "std" enabled.
```
tcharding, line 10: Mad! If this merges we can patch `rust-bitcoin-maintainer-tools` to remove the logic that supports this.

tcharding (commented):
Thanks man! Apart from the comments on the docs stuff this is awesome, appreciate the improvement.

Cargo.toml
JoseSK999: Right, good point!

README.md
JoseSK999: Thanks, no worries about being thorough. It is unrelated, but I figured it was okay since it’s just an outdated path in the README.

README.md
JoseSK999: Yeah. The **Building** section had more relevant info about how to select the `no-std`/`std` features, but now it is just one line (I added that to make it more complete).

Technically the `no_std` support is already explained in **High-Level Features** and doesn't require more in-depth explanation, so based on your comment I think it makes sense to remove **Building**.

JoseSK999:
Updated with @tcharding comments

README.md
apoelstra: They're not totally unrelated because they replace docs that talked about nostd.

Though agreed that we probably don't need to tell people to run `cargo test` if there's not even any env variables or anything :). If this were end-user software then sure, but this is a Rust library that you can't use without being familiar with the language and its ecosystem.

apoelstra:
CI failure is unrelated (will be fixed by #766).

apoelstra (approved):
ACK 8372e8c7713c17acbf08021f7bcb990e3fa1e571; successfully ran local tests; even nicer!

tcharding (approved):
ACK 8372e8c7713c17acbf08021f7bcb990e3fa1e571

tcharding:
Thanks bro.

sanket1729:
ACK. Can you rebase to get the CI fix in? Or maybe can @apoelstra directly merge after running tests on merge commit?

JoseSK999:
Rebased!

apoelstra (approved):
ACK ccc2815595b026ab4ac0c4ecc6523d2450c59eff; successfully ran local tests; even nicer!

PR #770, opened 2024-11-10
Update deprecated field in rustfmt.toml

JoseSK999:
This fixes the "Warning: the `version` option is deprecated. Use `style_edition` instead." that we see when running `cargo +nightly fmt`.

The 2021 edition is the last one matching the version "One".


apoelstra (approved):
ACK b5a8374eb0f59dac0616a6f2600a2cf08fd8f98b; successfully ran local tests; thanks!

PR #773, opened 2024-11-12
refactor expression parsing and checksum checking

apoelstra:
As a step toward rewriting the expression parser to be non-recursive, add a pre-parsing well-formedness check, which verifies that an expression is well-formed, uses only the correct character set, and that the checksum (if present) is valid.

Along the way, replace the error types returned from the `expression` module with a new more-precise one which can identify the location of parse errors (and identify what the error was in a more correct way). This improves our error reporting and drops many instances of the stringly-typed `BadDescriptor` error.

There is currently a bunch of special logic for Taproot descriptors which have the extra characters `{` and `}`. To the extent possible, this PR doesn't touch that logic. It will be addressed in a later PR.

The benchmarks show a slight slowdown since we essentially added new validation logic without removing the old logic. Later PRs will improve things.

### DIFF
--- a/src/descriptor/checksum.rs
+++ b/src/descriptor/checksum.rs
@@ -14,7 +14,6 @@ use core::iter::FromIterator;
 use bech32::primitives::checksum::PackedFe32;
 use bech32::{Checksum, Fe32};
 
-pub use crate::expression::VALID_CHARS;
 use crate::prelude::*;
 
 const CHECKSUM_LENGTH: usize = 8;
--- a/src/descriptor/checksum.rs
+++ b/src/descriptor/checksum.rs
@@ -16,44 +16,116 @@ use bech32::{Checksum, Fe32};
 
 pub use crate::expression::VALID_CHARS;
 use crate::prelude::*;
-use crate::Error;
 
 const CHECKSUM_LENGTH: usize = 8;
 const CODE_LENGTH: usize = 32767;
 
-/// Compute the checksum of a descriptor.
+/// Map of valid characters in descriptor strings.
 ///
-/// Note that this function does not check if the descriptor string is
-/// syntactically correct or not. This only computes the checksum.
-pub fn desc_checksum(desc: &str) -> Result<String, Error> {
-    let mut eng = Engine::new();
-    eng.input(desc)?;
-    Ok(eng.checksum())
+/// The map starts at 32 (space) and runs up to 126 (tilde).
+#[rustfmt::skip]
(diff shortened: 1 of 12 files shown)


src/descriptor/checksum.rs
```diff
@@ -16,44 +16,116 @@ use bech32::{Checksum, Fe32};
...
 /// Checks and verifies the checksum if it is present and returns the descriptor
 /// string without the checksum.
-pub(super) fn verify_checksum(s: &str) -> Result<&str, Error> {
-    for ch in s.as_bytes() {
-        if *ch < 20 || *ch > 127 {
-            return Err(Error::Unprintable(*ch));
+pub fn verify_checksum(s: &str) -> Result<&str, Error> {
+    let mut last_hash_pos = s.len();
+    for (pos, ch) in s.char_indices() {
+        if !(32..127).contains(&u32::from(ch)) {
```
sanket1729, line 97: There is a logic change here. From 20..127 to 32..127.

sanket1729: Which is correct.

src/descriptor/checksum.rs
```diff
@@ -16,44 +16,116 @@ use bech32::{Checksum, Fe32};
...
-        let expected_sum = desc_checksum(desc_str)?;
-        if checksum_str != expected_sum {
-            return Err(Error::BadDescriptor(format!(
-                "Invalid checksum '{}', expected '{}'",
-                checksum_str, expected_sum
-            )));
+        let expected = eng.checksum_chars();
+        let mut actual = ['_'; CHECKSUM_LENGTH];
+        for (act, ch) in actual.iter_mut().zip(checksum_str.chars()) {
+            *act = ch;
```
sanket1729, line 120: From chatgpt; with MSRV [1.63](https://doc.rust-lang.org/beta/core/array/fn.from_fn.html); you can avoid the mut and creating the actual with all '_' as 
```rust
let mut iter = checksum_str.chars()
let actual: [char; CHECKSUM_LENGTH] = std::array::from_fn(|_| iter.next().expect("Len checked eariler))
```

sanket1729 (approved):
ACK eb4f8c4dce3c45d6e510075a63d2115f7198f7b4

I love how this is shaping out to be. Thanks for detailed benchmarks.

src/descriptor/checksum.rs
apoelstra: Nice! I will add a commit to my next PR doing this. I'm thrilled to have `from_fn` now.

PR #781, opened 2024-12-03
feat: optimize pushbytes by removing unreachable minimal check

ChrisCho-H:
In [rust-bitcoin](https://github.com/rust-bitcoin/rust-bitcoin/blob/aa25adfc64ea0d742ca202991140f280024aba1b/bitcoin/src/blockdata/script/push_bytes.rs#L312), `read_scriptint` checks `NonMinimalPush`, which makes an additional minimal check needless and `PushBytes` inefficient.

### DIFF
--- a/src/miniscript/lex.rs
+++ b/src/miniscript/lex.rs
@@ -213,14 +213,9 @@ pub fn lex(script: &'_ script::Script) -> Result<Vec<Token<'_>>, Error> {
                     33 => ret.push(Token::Bytes33(bytes.as_bytes())),
                     65 => ret.push(Token::Bytes65(bytes.as_bytes())),
                     _ => {
+                        // check minimality of the number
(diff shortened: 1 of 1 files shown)


apoelstra:
We should add a fuzztest which tries to decode scripts both with master and 0.12.

Can wait til after #780 gets in, since that already includes a similar fuzztest for parsing expression strings.

apoelstra (approved):
ACK 84d1ad5932c7e1527a1aa1b46c1efde68a0f0244; successfully ran local tests; yep, I think you are right

sanket1729:
> Can wait til after https://github.com/rust-bitcoin/rust-miniscript/pull/780 gets in, since that already includes a similar fuzztest for parsing expression strings.

Working on it this week

apoelstra:
I'm gonna go ahead and merge this. Will add the fuzztest in a later PR.

PR #798, opened 2025-03-26
cleanup: refactor a bunch of concatenation-related code in `satisfy`

apoelstra:
There is a bunch of essentially-repeated "combine two satisfactions" code throughout satisfy.rs. It's hard to tell at a glance what order the concatenantions are in, even though this is essential, or whether the combination is actually done correctly. 

In particular, we routinely use the opposite order for Witness::combine and for ||'ing the has_sig, and we routinely assume that timelock conditions are None for dissatisfactions, which is true, but it's still a lot to keep in your head.

Also adds a fuzz test to sanity check our satisfactions (just checks for crashes; we don't check whether satisfaction "should" succeed since that's hard to tell in general.)

### DIFF
--- a/.github/workflows/cron-daily-fuzz.yml
+++ b/.github/workflows/cron-daily-fuzz.yml
@@ -20,6 +20,7 @@ jobs:
         fuzz_target: [
 compile_descriptor,
 compile_taproot,
+miniscript_satisfy,
 parse_descriptor,
 parse_descriptor_priv,
 parse_descriptor_secret,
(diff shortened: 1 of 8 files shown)


apoelstra:
Heh, there's more clippy stuff to fix (I think I need to update my nightly..) but I am signing off for today.

apoelstra:
This one should be good to go and have a green CI. I'll start working on a followup which ports the rust version pinning from rust-bitcoin which should make CI stop breaking.

Then since I'm thinking about this crate I'll take a look at what the status of my "eliminate recursion / fix errors / replace ctx" projects are and see if I can't do another PR..

apoelstra (commented):
On 426479d436c6a10a2e444542ada293bcef8ffbd1 successfully ran local tests

sanket1729 (approved):
utACK 426479d436c6a10a2e444542ada293bcef8ffbd1

PR #801, opened 2025-03-31
Fix: make `StaticDebugAndDisplay` `Send` and `Sync`

### LINKED ISSUE #800: Unreleased v13.0.0 is incompatible with `anyhow` crate
The [`anyhow`](https://crates.io/crates/anyhow) crate is a popular error handling crate for application development. 

It makes handling `Error` types from various dependencies easy by adding blanket `From` implementations to `anyhow::Error`, which is internally represented as a form of `Box<dyn std::error::Error + Send + Sync + 'static>`.

In `v12.3.0` of `rust-miniscript`, the `ParseError` type was able to be automatically converted to `anyhow::Error`. 
However, the unreleased `v13.0.0` breaks this compatibility by introducing the `StaticDebugAndDisplay` trait and blanket implementation.

This is because the  `StaticDebugAndDisplay` trait does not implement `Send` / `Sync` and the `ParseError::FromStr` variant wraps `Box<dyn StatusDebugAndDisplay`, also making the `ParseError` type not `Send` / `Sync`.

https://github.com/rust-bitcoin/rust-miniscript/blob/a39725dab59051dc08d747eb6d2f2da768d9f27d/src/blanket_traits.rs#L20-L23
https://github.com/rust-bitcoin/rust-miniscript/blob/a39725dab59051dc08d747eb6d2f2da768d9f27d/src/error.rs#L15-L32

The following definition of the `StaticDebugAndDisplay` trait retains compatibility at the cost of restricting `FromStr::Err` types to being thread-safe:

```rust
/// Auxiliary trait indicating that a type implements both `Debug`, `Display`, `Send` and `Sync`.
// NOTE: `Send` / `Sync` is required to maintain compatibility with downstream error handling libraries.
pub trait StaticDebugAndDisplay: fmt::Debug + fmt::Display + Send + Sync + 'static {}

impl<T: fmt::Debug + fmt::Display + Send + Sync + 'static> StaticDebugAndDisplay for T {}
```

chris-ricketts:
Fixes #800

### DIFF
--- a/src/blanket_traits.rs
+++ b/src/blanket_traits.rs
@@ -17,10 +17,11 @@ use core::{fmt, hash};
 
 use crate::MiniscriptKey;
 
-/// Auxiliary trait indicating that a type implements both `Debug` and `Display`.
-pub trait StaticDebugAndDisplay: fmt::Debug + fmt::Display + 'static {}
+/// Auxiliary trait indicating that a type implements both `Debug`, `Display`, `Send` and `Sync`.
+// NOTE: `Send` / `Sync` is required to maintain compatibility with downstream error handling libraries.
+pub trait StaticDebugAndDisplay: fmt::Debug + fmt::Display + Send + Sync + 'static {}
 
-impl<T: fmt::Debug + fmt::Display + 'static> StaticDebugAndDisplay for T {}
+impl<T: fmt::Debug + fmt::Display + Send + Sync + 'static> StaticDebugAndDisplay for T {}
 
 /// Blanket trait describing a key where all associated types implement `FromStr`,
 /// and all `FromStr` errors can be displayed.
--- a/src/error.rs
+++ b/src/error.rs
@@ -18,7 +18,7 @@ use crate::Box;
 pub enum ParseError {
     /// Invalid absolute locktime
     AbsoluteLockTime(AbsLockTimeError),
-    /// Invalid absolute locktime
+    /// Invalid relative locktime
     RelativeLockTime(RelLockTimeError),
     /// Failed to parse a public key or hash.
     ///


apoelstra:
Nice. Yeah, I'm ok with this. In general I don't wanna throw every trait under the sun into this, but `Send` and `Sync` seem very reasonable, and if it makes the difference between `anyhow` compatibility and not we should do it.

Maaaybe I had some specific error types in mind that couldn't be `Send`/`Sync`, in which case I'll need to revisit this the next time I dust off my error improvement branch. But I don't think so.

chris-ricketts:
Thanks for the quick review and merge @apoelstra!

> Maaaybe I had some specific error types in mind that couldn't be Send/Sync, in which case I'll need to revisit this the next time I dust off my error improvement branch. But I don't think so.

On the topic of error improvement, I have a [commit](https://github.com/rust-bitcoin/rust-miniscript/compare/master...chris-ricketts:rust-miniscript:master#diff-f0f5d337fc7d5fe38728c709556ed7167675d09249806da1cf889d0c8a635da6R333) which attempts to refactor the `descriptor::key::DescriptorKeyParseError` by turning it into an enum as suggested in the `FIXME` comment:

https://github.com/rust-bitcoin/rust-miniscript/blob/d0da327417cb177884c5d77edffa0891ba9b6969/src/descriptor/key.rs#L327-L330

Would it be worth a PR? 

My main motivation is to add support for Ypubs in descriptors and this was a prerequisite to do it right.

My plan is the follow the advice in https://github.com/rust-bitcoin/rust-bitcoin/issues/2190 and change the version bytes before calling `Xpub::decode` in the [`parse_bip32_xpub` function](https://github.com/chris-ricketts/rust-miniscript/blob/75f720aaeb7639f184def16621d300ce67db6c1e/src/descriptor/key.rs#L935-L937).

apoelstra:
>Would it be worth a PR?

This would be amazing. I've taken a few cracks at improving the global `Error` enum and never bothered touching `DescriptorKeyParseError` because it's self-contained and more of a chore than something that's blocking progress on the rest of the codebase.

> My plan is the follow the advice in rust-bitcoin/rust-bitcoin#2190 and change the version bytes before calling Xpub::decode in the parse_bip32_xpub function.

Yeah, I still think that's the most reasonable way to approach this. (Given that I am still opposed to having any direct ypub/zpub support in this library. :))

PR #799, opened 2025-03-27
ci: pin nightly rust compiler version

apoelstra:
Turns out we were most of the way there already.

I added `secrets.APOELSTRA_CREATE_PR_TOKEN` to this repo so the weekly cronjob should work. I just copy/pasted the code from rust-bitcoin.

PR #803, opened 2025-03-31
[backport] cleanup: refactor a bunch of concatenation-related code in satisfy

apoelstra:
This backports the main commit in #798 so that the code in (released) 12.x is substantially the same as the code in (unreleased) 13.x. This makes it easier to write and maintain regression tests.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -316,7 +316,7 @@ dependencies = [
 
 [[package]]
 name = "miniscript"
-version = "12.3.0"
+version = "12.3.1"
 dependencies = [
  "bech32",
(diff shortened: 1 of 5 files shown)


apoelstra:
I backported one lint fix that I couldn't even compile with (a change to `deny(missing_docs)` behavior) but we still expect the Docs/Lint jobs to fail on this branch.

I may later backport #799 with an old version of rust nightly to fix this as well.

sanket1729 (approved):
ACK 9d81dcfde822925fc85e5046bf6b9cd82c6b006b

apoelstra:
Tagged and published.

PR #804, opened 2025-03-31
Refactor: make descriptors::key::DescriptorKeyParseError an enum

chris-ricketts:
This PR has a go at refactoring `DescriptorKeyParseError` into an enum, as suggested in the `FIXME` comment:

https://github.com/rust-bitcoin/rust-miniscript/blob/d0da327417cb177884c5d77edffa0891ba9b6969/src/descriptor/key.rs#L327-L330

I have separated the error variants by different inner parse error types and result types (e.g, public key, private key, etc) where I think it made sense to retain the inner error.

I left a variant that wraps a `&'static str` and its used for the various ad-hoc parsing errors instead of adding the boilerplate for variants only created once and without retained data.

The generic `parse_xkey_deriv` function now takes a parse function as its first parameter to allow specific key parsing errors:

https://github.com/chris-ricketts/rust-miniscript/blob/790b08fcf46f53ce6ee80355a7db8e5108ed860f/src/descriptor/key.rs#L935-L952

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -223,7 +223,8 @@ impl DescriptorXKey<bip32::Xpriv> {
         let xprv = self
             .xkey
             .derive_priv(secp, &hardened_path)
-            .map_err(|_| DescriptorKeyParseError("Unable to derive the hardened steps"))?;
+            .map_err(DescriptorKeyParseError::DeriveHardenedKey)?;
+
         let xpub = bip32::Xpub::from_priv(secp, &xprv);
 
         let origin = match &self.origin {
@@ -276,7 +277,9 @@ impl DescriptorMultiXKey<bip32::Xpriv> {
                         if child_num.is_normal() {
                             Ok(*child_num)
                         } else {
-                            Err(DescriptorKeyParseError("Can't make a multi-xpriv with hardened derivation steps that are not shared among all paths into a public key."))
+                            Err(DescriptorKeyParseError::MalformedKeyData(
+                                MalformedKeyDataKind::InvalidMultiXKeyDerivation,
+                            ))
                         }
                     })
                     .collect()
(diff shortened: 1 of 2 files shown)


apoelstra:
>instead of adding the boilerplate for variants only created once and without retained data.

Can you add the boilerplate? I know it's annoying to write but it makes things much more maintainable since it forces a visible API change whenever we add or change variants.

apoelstra:
In 895da06d4c4c6daeb459ca2587e08f79bd0e4a5f:

You need to re-export the new error type `MalformedKeyDataKind` from src/descriptor/mod.rs. Unfortunately I think the compiler will then immediately whine that you need a doccomment on every single variant. I'd be ok with just sticking `#[allow(missing_docs)]` on the enum since it's a C-like enum where every variant is named what it is.

apoelstra:
In 895da06d4c4c6daeb459ca2587e08f79bd0e4a5f and 790b08fcf46f53ce6ee80355a7db8e5108ed860f:

Can you run through all your error messages and change them to start with a lowercase letter?

apoelstra:
Other than those two comments 895da06d4c4c6daeb459ca2587e08f79bd0e4a5f looks great! Thanks for grinding through this.

chris-ricketts:
Thanks for the review @apoelstra

- Re-export `MalformedKeyDataKind`: https://github.com/rust-bitcoin/rust-miniscript/pull/804/commits/bf058c96390e73b03993e81b72fc057491e113bd

- Allow missing docs for `MalformedKeyDataKind`: https://github.com/rust-bitcoin/rust-miniscript/pull/804/commits/a5c1785fe7a6fc497273f05dadcf693401df76aa

- Start all error messages will lower case letter: https://github.com/rust-bitcoin/rust-miniscript/pull/804/commits/0e98dd2c2b02337b83d759888b0960a9eeab52c1

Let me know if you would prefer all the commits in this PR to be squashed into one.

apoelstra:
Yes please -- can you just squash everything?

apoelstra (approved):
ACK 497c5a31974e9a19f8b3b5286edb323281957147; successfully ran local tests; thanks!

PR #807, opened 2025-04-10
[Taproot API project] move several files and tighten up iterator API

apoelstra:
This PR mostly just moves files around, but it includes two nontrivial API changes:

* It renames the `iter_scripts` iterator to `leaves` on `Tr` and `Taptree`
* It yields a dedicated struct rather than just a `(u8, Script)` tuple

In the next PRs we will completely rewrite this iterator, and `Taptree` and these preparatory commits help to reduce the diff throughout the rest of the codebase.


apoelstra:
Drafting for a bit -- I need to test all feature combinations.

apoelstra (commented):
On 8e5d809b1f2a8d55799476b20fbbb2c783ecd72d successfully ran local tests

src/descriptor/tr/taptree.rs
```diff
@@ -0,0 +1,58 @@
...
+    #[inline]
+    pub fn leaf_version(&self) -> LeafVersion { LeafVersion::TapScript }
+}
+
+impl<Pk: ToPublicKey> TapTreeIterItem<'_, Pk> {
+    /// Computes the Bitcoin Script of the leaf.
+    ///
+    /// This function is potentially expensive.
+    #[inline]
+    pub fn compute_script(&self) -> bitcoin::ScriptBuf { self.node.encode() }
```
sanket1729, line 198: I like this idea of potentially naming all functions that could be expensive with `compute` prefix.

sanket1729 (approved):
code review ACK 8e5d809b1f2a8d55799476b20fbbb2c783ecd72d.

Thanks for the helpful commit messages.

PR #808, opened 2025-04-14
[Taproot API Project] flatten internal representation of `TapTree`

apoelstra:
This PR completely rewrites the `TapTree` module to store a taptree as a flat list of leaves and heights. For a non-`ToPublicKey` key this is the only useful information that we ever extract from this, so storing a big recursive structure is wasteful. We also eliminate recursive code from everything `TapTree` related, including the implicit `Drop` impl.

In the next PR we'll introduce a new tree structure that contains control block information -- but `TapTree` can't be this structure because `TapTree` doesn't assume that its keys can be serialized.

Also breaks the `iter_scripts` iterator API in a couple small ways.

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -256,7 +256,7 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
     /// returned value.
     pub fn tap_tree(&self) -> Option<&TapTree<Pk>> {
         if let Descriptor::Tr(ref tr) = self {
-            tr.tap_tree().as_ref()
+            tr.tap_tree()
         } else {
             None
         }
@@ -268,7 +268,7 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
     /// Taproot descriptor containing only a keyspend, returns an empty iterator.
     pub fn tap_tree_iter(&self) -> tr::TapTreeIter<Pk> {
         if let Descriptor::Tr(ref tr) = self {
-            if let Some(ref tree) = tr.tap_tree() {
+            if let Some(tree) = tr.tap_tree() {
                 return tree.leaves();
             }
         }
@@ -988,7 +988,7 @@ impl<Pk: FromStrKey> FromStr for Descriptor<Pk> {
             // FIXME preserve weird/broken behavior from 12.x.
             // See https://github.com/rust-bitcoin/rust-miniscript/issues/734
             ret.sanity_check()?;
-            for item in inner.iter_scripts() {
+            for item in inner.leaves() {
                 item.miniscript()
                     .ext_check(&crate::miniscript::analyzable::ExtParams::sane())?;
             }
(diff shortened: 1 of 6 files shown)


apoelstra:
Ooo, it looks like I broke our satisfaction tests somehow.

apoelstra:
The satisfaction tests seem to require bitcoin 22, which is a bit of a PITA for me to obtain since it needs boost 1.75 to build. I will work on updating it to a more modern version of Core in a separate PR before moving forward on this.

apoelstra:
Actually I'll just do it independently. Since I was able to get the integration tests running it was easy for me to find the bug -- I had introduced an off-by-one error in our maximum Taproot depth checking.

It turns out that [BIP 341 specifies valid depths from 0 to 128 *inclusive*](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#cite_note-6) so actually there are 129 depths that might be used. This is surprisingly inconvient for something that sipa designed in parallel with a C++ implementation, but it is what it is.

Anyway I extended my `u128`-based bitmap in `TapTreeBuilder` with an extra bool, changed a `>=` check to a `>`, and this should be good to go now.

sanket1729:
> The satisfaction tests seem to require bitcoin 22

Interesting, is this because of some deprecated RPCs?

apoelstra:
It seems to be some change in how default wallets are created (or not).

sanket1729 (approved):
ACK adf3164bb4eda5ab30236c8f78fcf60df562691d

> Maaybe we should backport the rename to 12.x as a pure rename, so that
people will update their code and get nicer error messages.

Would be nice, but we can also add this in Changelog. 


> depths_leaves: Vec<(u8, Arc<Miniscript<Pk, Tap>>)>,

As discussed in person, this does change the structure of tree. But does not really change the satisfaction or the merkle paths to the leaves. Any spend paths from previous constructions should be spendable from this. 

Surprisingly, we did not have any fixed tests that check the merkle root or script pubkey for taproot descriptor. Our integration tests create new descriptor and spend from them. Cross tested a simple test below that this PR does not any of the that logic.

```diff
diff --git a/src/descriptor/mod.rs b/src/descriptor/mod.rs
index 1b6bc78..f822b11 100644
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -1519,17 +1519,17 @@ mod tests {
             "tr({},{{pk({}),{{pk({}),or_d(pk({}),pkh({}))}}}})",
             p1, p2, p3, p4, p5
         ))
-        .unwrap()
-        .to_string();
+        .unwrap();
 
         // p5.to_pubkeyhash() = 516ca378e588a7ed71336147e2a72848b20aca1a
         assert_eq!(
-            descriptor,
+            descriptor.to_string(),
             format!(
                 "tr({},{{pk({}),{{pk({}),or_d(pk({}),pkh({}))}}}})#tvu28c0s",
                 p1, p2, p3, p4, p5
             )
-        )
+        );
+        assert_eq!(descriptor.spend_info().merkle_root().unwrap().to_string(), "e1597abcb76f7cbc0792cf04a9c2d4f39caed1ede0afef772064126f28c69b09");
     }
 
     #[test]
```

apoelstra:
Thanks! I'll pull your unit test into my next PR.

PR #810, opened 2025-04-21
Backport of #809: fix crash in `Descriptor::parse_desc`

apoelstra:
Backports #809 and does another patch release.

PR #809, opened 2025-04-21
fix crash in Descriptor::parse_desc found by fuzzer

apoelstra:
Thanks to Bruno Garcia and i-am-yuvi who independently found this crash and reported it to me.

Fixes #806

Needs backport.

PR #814, opened 2025-04-29
Backport #809 to 0.10.x

apoelstra:
Backports #809 to 10.x. This is a direct rebase of #810 except that I added a commit with some lint fixes. (I had been avoiding this, but we've been having a lot of backports lately so I think I ought to just address it.)

After this I will do 11.x.


apoelstra:
I was able to get my local CI to pass on this. It's ready for review.

sanket1729 (approved):
utACK 06400b136ac6409d8eb8a698a7eb69612c9ed9b1

Issue #817, opened 2025-05-12
Only public API for constructing DefiniteDescriptorKey or DefinitePublicKey is `from_str(...)`?


philipr-za:
I am using rust-miniscript to work with taproot multi-signature scripts. I have a `Descriptor<String>` which defines the descriptor with a human readable placeholder keys and then I use a Translator with a map I maintain between these Strings and a concrete XOnlyPublicKey. When I apply the Translator I know have a `Descriptor<XOnlyPublicKey>`.

However, now if I want to use a PSBT to construct a transaction spending these inputs I need to the `.update_with_descriptor_unchecked(...)` interface which expects a descriptor of the form `Descriptor<DefiniteDescriptorKey>`.

My problem is how does `rust-miniscript` expect me to turn a `Descriptor<XOnlyPublicKey>` or `Descriptor<PublicKey>` into a `Descriptor<DefiniteDescriptorKey>`?

The only way I could figure out to do it was to serialize my `Descriptor<XOnlyPublicKey>` and use the `FromStr` interface to re-parse it as a `Descriptor<DefiniteDescriptorKey>`.

Similarly `DescriptorPublicKey` doesn't have a way to externally construct its `Single` variant other than with `FromStr`.

It feels like there needs to be an interface for constructing `DefiniteDescriptorKey` and `DescriptorPublicKey` directly from the rust-bitcoin key data structures rather than requiring a serialization-deserialization roundtrip.

PR #818, opened 2025-05-12
Add helpers to convert to DescriptorPublicKey and Descriptor<DefinitePublicKey> data structures

philipr-za:
closes https://github.com/rust-bitcoin/rust-miniscript/issues/817

I added the common cases of converting from `Descriptor<PublicKey>` and `Descriptor<XOnlyPublicKey>` to `Descriptor<DefinitePublicKey>`. Not sure if this is considered too specific but it feels like a use case that will be very common.

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -8,7 +8,7 @@ use std::error;
 
 use bitcoin::bip32::{self, XKeyIdentifier};
 use bitcoin::hashes::{hash160, ripemd160, sha256, Hash, HashEngine};
-use bitcoin::key::XOnlyPublicKey;
+use bitcoin::key::{PublicKey, XOnlyPublicKey};
 use bitcoin::secp256k1::{Secp256k1, Signing, Verification};
 
 use crate::prelude::*;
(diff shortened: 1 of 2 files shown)


apoelstra:
In 593709060bf32f35c2d16639fe0f6e83183005c1:

I think this is too specific. You're the first person I've encountered who wound up with `XOnlyPublicKey`s (which is a low level secp256k1 type) and wanted to use it in a PSBT like this.

The `From` impls on the keys themselves are fine.

I agree that it would be nice if we had a `.map_pk` function or something that was a simpler version of `.translate_pk` that assumed the hash types were all the same and that the mapping function were infallible. Then you could do `descriptor.map_pk(DefiniteDescriptorKey::from)`.

philipr-za:
> In [5937090](https://github.com/rust-bitcoin/rust-miniscript/commit/593709060bf32f35c2d16639fe0f6e83183005c1):
> 
> I think this is too specific. You're the first person I've encountered who wound up with `XOnlyPublicKey`s (which is a low level secp256k1 type) and wanted to use it in a PSBT like this.
> 
> The `From` impls on the keys themselves are fine.
> 
> I agree that it would be nice if we had a `.map_pk` function or something that was a simpler version of `.translate_pk` that assumed the hash types were all the same and that the mapping function were infallible. Then you could do `descriptor.map_pk(DefiniteDescriptorKey::from)`.

Fair enough, I will drop those impl's and run rustfmt

philipr-za:
Removed the Descriptor impl's. Had to make the `new` constructor public on `DefiniteDescriptorKey` to allow for the `translate_pk` to work  I left in the unit test to prove the application of these From impl's to the use case in the issue.

apoelstra (approved):
ACK a00c99312c216a61280b7513a3c1f1f12bffa478; successfully ran local tests

apoelstra:
BTW if you want to use this today, can you do a backport PR to the 12.x branch (along with a minor version bump)?

PR #824, opened 2025-06-04
plan: add some clarifying docs

benma:
"The size in bytes of the script sig that" is ambiguous and could refer to the actual scriptSig, exlcuding the varint prefix that indicates its size.

The code comment in witness_size is promoted to the docstring, as its helpful to anyone trying to compute the right size of a transaction.

### DIFF
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -248,7 +248,8 @@ impl Plan {
     /// the script sig weight and the witness weight)
     pub fn satisfaction_weight(&self) -> usize { self.witness_size() + self.scriptsig_size() * 4 }
 
-    /// The size in bytes of the script sig that satisfies this plan
+    /// The size in bytes of the script sig that satisfies this plan, including the size of the
+    /// var-int prefix.
     pub fn scriptsig_size(&self) -> usize {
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK e89e226ec6cfee5985c92a215282c552790f36ba; successfully ran local tests; nice, good call

benma:
@apoelstra speaking of, I just noticed that plan.satisfy() returns the scriptSig as a ScriptBuf, which does not include the var-int prefix, so it's a inconsistent that the size of the var-int prefix is included in sigscript_size(). 

Furthermore, it's not possible to know the size of the scriptsig without the var-int prefix during planning. 

I think it would make sense to offer a function to return the size excluding the var-int prefix. WDYT?

apoelstra:
Oh, yeah, I like this. Definitely agreed that `plan.satisfy().len()` ought to return the same value as `plan.???_size()`.

I am torn between deprecating the version that has the length prefix, or not. (We definitely should not just silently change the behavior of `scriptsig_size()`! We need a new name. Maybe `satisfy_len`?)

PR #592, opened 2023-08-18
Add the planning module (updated PR)

danielabrozzoni:
This PR builds on top of #481, fixing all the review comments.

I didn't squash my last commits on purpose to make review easier, I can squash them before merging if preferred.

PR #826, opened 2025-06-05
plan: remove unused AssetProvider::provider_lookup_tap_control_block_map

benma:
Was not sure if/when/how to implement it, when I noticed it's not needed at all.


apoelstra:
Heh, yeah, strange. This method dates to the original plan PR and AFAICT was never used. https://github.com/rust-bitcoin/rust-miniscript/pull/592

This functionality also doesn't seem to make sense coming from the asset provider (we have methods on `Descriptor` to get taproot-related stuff now), and the doccomment is inconsistent with the actual API.

apoelstra (dismissed):
ACK 42052941e3c43494df97e7d8d195c078a550131b; successfully ran local tests

PR #827, opened 2025-06-11
satisfy: change `lookup_tap_key_spend_sig` to take the public key

### LINKED ISSUE #825: `Satisfier::lookup_tap_key_spend_sig(&self)` should take the pubkey as an argument
https://github.com/rust-bitcoin/rust-miniscript/blob/f023c39a1e69a699c74ecfce742289558375b45e/src/miniscript/satisfy.rs#L38

Compare to AssetProvider in plan.rs:

https://github.com/rust-bitcoin/rust-miniscript/blob/f023c39a1e69a699c74ecfce742289558375b45e/src/plan.rs#L50

Seeing that AssetProvider mirrors Satisfier, they should be consistent, and it seems helpful to include the pubkey in the satisfier function so one can look up the signature based on it.

apoelstra:
There is only one public key possible, but this makes the API more consistent between keyspends and script-spends, and makes it much easier for users who have a key->sig lookup table to implement satisfiers.

Fixes #825


apoelstra (commented):
On 1c2800e2c5eeb96ae779fb5d1f506e6d53bffdd0 successfully ran local tests

apoelstra:
Because this is API-breaking I can't backport it to 12.x. It will be available in rust-miniscript 13.

sanket1729 (approved):
ACK 1c2800e2c5eeb96ae779fb5d1f506e6d53bffdd0.

I am not excited by this change. But I can see how this helps ergonomics. ACK

PR #828, opened 2025-06-20
Various cleanups, mostly around the compiler

apoelstra:
This PR is a grab back of prepatory commits for larger changes. It includes a couple of followups to #815 in the first commits. Then it cleans up some error types and paths.

I don't believe anything in here will be tricky or controversial, though there are API breaks because of the cleaned up error types.


apoelstra:
Need to investigate why jj is not signing these commits.

apoelstra:
I had authored them with the wrong email address set.

apoelstra (commented):
On c317cd7d0cf6089fea33949a145fe32e80d7fd99 successfully ran local tests

apoelstra:
cc @sanket1729 this one is a real PR, but just pinging you in case you missed it.

PR #830, opened 2025-07-02
Properly forbid multipath keys in definite descriptor keys

apoelstra:
This PR essentially just fixes a logic bug in `DefiniteDescriptorKey::new` but involves several commits that just rearrange error types.

I will backport the logic fix without the error refactoring, since for 12.x and earlier we didn't really have error types and were just returning `Error::Unexpected("arbitrary string")` style errors.


sanket1729 (dismissed):
ACK 5df4d5b2ed704857ffc297a3e4e49a59bcde9a6d

PR #832, opened 2025-07-02
backport #830 to rust-miniscript 12.x

apoelstra:
Just the logic fix; the error refactoring doesn't apply.

PR #833, opened 2025-07-02
backport #830 to rust-miniscript 10.x 

apoelstra:
Just the logic fix; the error refactoring doesn't apply.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "miniscript"
-version = "10.2.1"
+version = "10.2.2"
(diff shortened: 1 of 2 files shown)


apoelstra:
This one is good to go once I have an ACK from you @sanket1729 . The 11.x one I am still hacking on my local CI, but you can see from the range-diff that this PR and #832 are identical, so if one is good then the other should be too.

I will do some followup PRs, maybe tomorrow, which try to fix CI on 9.x, 10.x, 11.x and 12.x, since all these versions are used by software on crates.io.

sanket1729 (approved):
tACK 780d6bc49836751c15150e42e9c795b144fcc018. 

Tested on my local machine on stable.

apoelstra:
Tagged and published.

PR #834, opened 2025-07-03
0.12: fix CI

apoelstra:
Should get us green CI, and pins the nightly compiler version so that it stays green.


apoelstra:
This one should be good to go (though I intend to run my local CI through it in full, which should now pass).

apoelstra (commented):
On 2019e97ba4426f31491fc8dc5e0478383759bc6a successfully ran local tests

PR #837, opened 2025-07-05
9.x: fix CI

apoelstra:
According to crates.io there are a bunch of things still using 9.x. We really haven't made an effort to backport stuff this far, mainly because the crate is nearly impossible to build with the ancient `bitcoind` dev-dependency. This PR removes that and adds a lockfile, which should make this version maintainable going forward.

I don't intend to backport any of the missed backports, at least not this weekend, but going forward I'll try to remember to include it in our backport set.


sanket1729 (approved):
ACK bda13719b3c2c0dc7ef91c54397f71d88bf796d5

apoelstra:
Only change was to update the nightly version since the one I chose was affected by https://github.com/rust-lang/rust-clippy/issues/6776. Gonna go ahead and merge this if CI (and my local CI) passes.

apoelstra:
Ok, I had to add a whole new commit which changes a bunch of doc syntax. So I guess I need a re-ACK @sanket1729 
/

sanket1729 (approved):
utACK 0679c7ac9270b69ef6a84e69443a82e0856b1c52

PR #839, opened 2025-07-05
more `DefiniteDescriptorKey` fixes and cleanups

apoelstra:
Again this one changes logic in a way that I need to backport, but the backport PRs will be much simpler since they are constrained not to change the API.

This one makes it impossible to construct a `DefiniteDescriptorKey` with hardened derivation steps. Previously this was usually impossible, but not always, which led to inconsistent error checks and a proliferance of `unwrap`s and generally hard-to-reason-about behavior.


apoelstra (commented):
On f09bf48597ce58c16a2eaf52d455fb6749843691 successfully ran local tests

sanket1729 (approved):
ACK f09bf48597ce58c16a2eaf52d455fb6749843691.

Love the cleanup with core::Infalliable and removal of unwraps! Much cleaner

PR #846, opened 2025-07-25
Invert malleability test in `best_tap_spend`

Ademan:
I *really* don't feel like I have enough context to be confident in this change, but at the same time this seems pretty obviously backwards to the best of my (currently lacking) knowledge.


apoelstra:
Hah, I think you're correct. Nice find! Though unfortunate that this breaks no tests..

apoelstra (approved):
ACK 36349ba51bd4ae3f51cd3b4768e024da0a376935; successfully ran local tests

PR #805, opened 2025-04-05
Automated daily update to rustc (to nightly-2025-04-04)

apoelstra:
Automated update to Github CI workflow `rust.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #849, opened 2025-08-07
several improvements and bugfixes to uncompressed key handling and sortedmulti

apoelstra:
Removes the weird and incorrect `MalleablePkH` script context rule; fixes several bugs related to key size estimation in pkh and multisig fragments; improves tests of sortedmulti; uses the `Thresh` type in sortedmulti constructors.

These bugs are obscure and only visible when using uncompressed pubkeys and I don't think they're worth backporting.

This PR is starting to make "real" changes toward unifying our validation parameters. In particular we separate out validation errors in the `sortedmulti` constructors from threshold-construction errors (which the user is forced to deal with before calling a `sortedmulti` constructor). The PR is bigger than you might think because when testing this separation I found some bugs.

This PR also includes a clippy fix which should unstick #805 (update nightly version) which has been stalled for months because the weekly retries don't trigger Github notifications and I didn't notice it.

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -23,12 +23,13 @@ use sync::Arc;
 
 use crate::expression::FromTree as _;
 use crate::miniscript::decode::Terminal;
+use crate::miniscript::limits::MAX_PUBKEYS_PER_MULTISIG;
 use crate::miniscript::{satisfy, Legacy, Miniscript, Segwitv0};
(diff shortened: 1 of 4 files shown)


apoelstra (commented):
On b5db340809c62d11dce26958bab51b57ae44408c successfully ran local tests

sanket1729 (approved):
utACK b5db340809c62d11dce26958bab51b57ae44408c

PR #853, opened 2025-09-09
ci: switch to cargo-fuzz from honggfuzz

apoelstra:
honggfuzz is increasingly hard for me to maintain in my local CI. It requires an exact match between the version of honggfuzz in the lockfile and the binary installed in the nix derivation, and these exact versions come with specific requirements about which rustc versions are allowable (and in particular, honggfuzz 0.5.56 does not work with rustc 1.74.0, which is the MSRV for rust-bitcoin). I also have to bend over backward to set compiler flags and make source directories writable.

Once we get this working here, I will open the same PR to rust-bitcoin. But I wanted to start here because it's an easier crate to maintain and also our fuzztests are in much better shape.

### DIFF
--- a/fuzz/README.md
+++ b/fuzz/README.md
@@ -1,29 +1,20 @@
 # Fuzzing
 
-`miniscript` has a fuzzing harness setup for use with honggfuzz.
+`miniscript` has a fuzzing harness setup for use with cargo-fuzz.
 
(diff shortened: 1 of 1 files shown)


apoelstra:
The issue seems to be with nightly-2025-03-21, which this crate is stuck on since #805 isn't passing. I will add some commits which update to the latest nightly here.

PR #859, opened 2025-09-28
typeck: clean up data structures

apoelstra:
These two simple commits just collect some related data into common data structures, allowing satisfaction/dissatisfaction data to all be `Option`al at once without any debug_asserts, and simplifying a bunch of code.

This makes a couple obscure fixes that are not detected by tests and we have never received reports about. See the commit message for the first commit for details.

### DIFF
--- a/src/miniscript/context.rs
+++ b/src/miniscript/context.rs
@@ -467,8 +467,7 @@ impl ScriptContext for Legacy {
     }
 
     fn max_satisfaction_size<Pk: MiniscriptKey>(ms: &Miniscript<Pk, Self>) -> Option<usize> {
-        // The scriptSig cost is the second element of the tuple
-        ms.ext.max_sat_size.map(|x| x.1)
+        ms.ext.sat_data.map(|data| data.max_script_sig_size)
     }
 
     fn pk_len<Pk: MiniscriptKey>(pk: &Pk) -> usize {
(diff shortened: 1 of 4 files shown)


tcharding (approved):
ACK 5f0072cb76c409babcb0fa4c7793671a0897978d

tcharding:
FWIW

apoelstra:
Added a commit with unit tests that show the new changes.

@sanket1729 when you review this, I think it will be easiest to review the new tests first. They were written by Claude but they are small and have clear comments explaining the change in behavior (and have example descriptors that you can compute these values by hand to confirm the new ones arecorrect).

apoelstra:
After studying the new `test_threshold_exec_stack_count_max_not_sum` unit test I think maybe I also need to add a +1  to `sat_data.max_exec_stack_count` for a threshold.

The reason is that in addition to the stack elements inherited from each threshold child, we also have the accumulator value which is `ADD`ed after each child.

apoelstra:
cc @sanket1729 do you think you can review this in the next week or two? It's blocking the 13.0 release.

sanket1729:
@apoelstra looking at this now. 
I have some availability for the next week to review more PRs here if you have more upcoming.

sanket1729 (approved):
utACK f3ea32c0ea899038bf5e7192fde5c6c54e0cf10b

apoelstra:
Thanks! Ok, I'll merge this, open the next one, and assess whether we want it in before 13.0 or not.

PR #851, opened 2025-08-22
feat: implement `GetKey` for `KeyMap` and `DescriptorSecretKey`

oleonardolima:
fixes https://github.com/rust-bitcoin/rust-miniscript/issues/709

### Description

I've built upon Tobin's work in #765 (see https://github.com/rust-bitcoin/rust-miniscript/pull/765#issuecomment-3208797801) to implement the `GetKey` trait for `KeyMap`, a feature that will enable us to entirely remove the signers API from `bdk_wallet`, see: https://github.com/bitcoindevkit/bdk_wallet/issues/70 and https://github.com/bitcoindevkit/bdk_wallet/pull/235.

### Notes to the reviewers

I followed some of Poelstra's previous comments to reduce the number of required matching branches, and also some VM's suggestions on implementing the main `GetKey` logic for `DescriptorSecretKey` instead.

Let me know if you have any other suggestions or comments. 

### Changelog notice

```md
# Added:
 - implemented `GetKey` trait for `KeyMap`
 - implemented `GetKey` trait for `DescriptorSecretKey`

# Changed
 - moved `KeyMap` to it's own `key_map` module, replacing the `BTreeMap` alias.
```

PR #860, opened 2025-09-29
feat(descriptor): backport `GetKey` impl from #851

oleonardolima:
backports #851 to release/12.x

### Description

It's an initial attempt to backport the implementation of GetKey for KeyMap introduced in #851.  

### Notes to the reviewers

I noticed that the type `KeyMapWrapper` is already being used internally for parse_descriptor,   though, as it's not part of the public API, there's no conflict, but I'm fine with changing it to another name if needed.

### Changelog Notice
```
# Added
- add `KeyMapWrapper` with `GetKey` trait for PSBT signing
- implement `GetKey` for `DescriptorSecretKey`
- add `From<KeyMap>` conversion
- export new `KeyMapWrapper` for downstream crate usage
```

### DIFF
--- a/src/descriptor/key_map.rs
+++ b/src/descriptor/key_map.rs
@@ -1,0 +1,276 @@
+use bitcoin::psbt::{GetKey, GetKeyError, KeyRequest};
+use bitcoin::secp256k1::{Secp256k1, Signing};
+use bitcoin::PrivateKey;
+
+use crate::descriptor::{DescriptorSecretKey, KeyMap};
+use crate::BTreeMap;
+
+/// A wrapper around KeyMap that implements GetKey for PSBT signing.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct KeyMapWrapper {
+    map: KeyMap,
+}
+
+impl From<KeyMap> for KeyMapWrapper {
+    fn from(map: KeyMap) -> Self { KeyMapWrapper { map } }
+}
+
+impl GetKey for KeyMapWrapper {
+    type Error = GetKeyError;
+
+    fn get_key<C: Signing>(
+        &self,
+        key_request: KeyRequest,
+        secp: &Secp256k1<C>,
+    ) -> Result<Option<bitcoin::PrivateKey>, Self::Error> {
+        Ok(self
+            .map
+            .iter()
+            .find_map(|(_desc_pk, desc_sk)| -> Option<PrivateKey> {
+                match desc_sk.get_key(key_request.clone(), secp) {
+                    Ok(Some(pk)) => Some(pk),
+                    Ok(None) | Err(_) => None,
+                }
+            }))
+    }
+}
+
+impl GetKey for DescriptorSecretKey {
+    type Error = GetKeyError;
+
+    fn get_key<C: Signing>(
+        &self,
+        key_request: KeyRequest,
+        secp: &Secp256k1<C>,
+    ) -> Result<Option<PrivateKey>, Self::Error> {
+        match (self, key_request) {
+            (DescriptorSecretKey::Single(single_priv), key_request) => {
+                let sk = single_priv.key;
+                let pk = sk.public_key(secp);
+                let pubkey_map = BTreeMap::from([(pk, sk)]);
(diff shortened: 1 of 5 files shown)


apoelstra:
In c7d41712ccc581a8fb72eb8e4e1861e4f56b00d1:

Wait, why can't we directly implement `GetKey` on `KeyMap` without the wrapper?

apoelstra:
Ohh, derp. Yes, I forgot that `KeyMap` is a type alias, not one of our types. (And that `GetKey` is from rust-bitcoin, which *always* throws me off and I wonder if we can revisit it..)

src/descriptor/key_map.rs
```diff
@@ -0,0 +1,276 @@
...
+        key_request: KeyRequest,
+        secp: &Secp256k1<C>,
+    ) -> Result<Option<bitcoin::PrivateKey>, Self::Error> {
+        Ok(self
+            .map
+            .iter()
+            .find_map(|(_desc_pk, desc_sk)| -> Option<PrivateKey> {
+                match desc_sk.get_key(key_request.clone(), secp) {
+                    Ok(Some(pk)) => Some(pk),
+                    Ok(None) | Err(_) => None,
```
apoelstra, line 32: In c7d41712ccc581a8fb72eb8e4e1861e4f56b00d1:

Here (and below) you swallow the error from `DescriptorSecretKey::get_key` and turn it into a "key not found" by returning `None`.

This is an issue with the code on master. I will open a fix PR and then I'd like you to also backport that.

apoelstra: Lemme PR some unit tests and I'll just ask you to backport those :P.

oleonardolima: Thanks, I've cherry-picked the commits and updated the required types for `KeyMapWrapper`.

oleonardolima:
> > > Wait, why can't we directly implement `GetKey` on `KeyMap` without the wrapper? it's our type and our trati.
> > 
> > 
> > As it's a type alias for `BTreeMap` and `GetKey` is defined in `rust-bitcoin::psbt` we can't implement it directly.
> 
> I'm late to the party but on `master` `KeyMap` is a struct? What am I missing? The docs are wrong too
> 
> ```rust
> /// Alias type for a map of public key to secret key.
> ///
> /// This map is returned whenever a descriptor that contains secrets is parsed using
> /// [`Descriptor::parse_descriptor`], since the descriptor will always only contain
> /// public keys. This map allows looking up the corresponding secret key given a
> /// public key from the descriptor.
> #[derive(Debug, Clone, Eq, PartialEq)]
> pub struct KeyMap {
>     map: BTreeMap<DescriptorPublicKey, DescriptorSecretKey>,
> }
> ```

Yes, on master it's already the correct type because of #851. But in release/12.x it's still the old type alias. Indeed, the docs are wrong, will create a commit to fix it.

tcharding:
Is the next release `12.3.6` or `12.4.0`? I ask because we can't bump `bitcoin` point version since its pre-1.0 in a point release here, right?

tcharding:
> Also, I've discussed with Tobin, and should I go ahead and add a commit already deprecating the type, or should it be done when the v13.0.0 release is closer?

Yeah I suggested deprecating the `KeyMapWrapper` type with a note "will be removed in the next major release" because its not needed once master is released (assuming next release of master is a major release).

tcharding:
FWIW everything else looks fine to me, I did not review closely that the backport patch mirrors the original.

oleonardolima:
> Is the next release `12.3.6` or `12.4.0`? I ask because we can't bump `bitcoin` point version since its pre-1.0 in a point release here, right? (Since point version pre-1.0 is a minor release.)

I think we would need 12.3.6 then, but I'm not sure what @apoelstra had in mind, so I'll leave it to him to answer.

apoelstra:
Interesting question. Yeah, let's do 12.4.0. I think that's the right thing to do anyway since we are adding functionality.

apoelstra (approved):
ACK 08fa87a6b0d3c2f1d75536fe760d2b6f0a872236; successfully ran local tests

PR #861, opened 2025-09-30
key_map.rs: propagate "invalid key" errors from multipath keys

apoelstra:
Since #851 we return a `NotSupported` error when attempting to request an xpriv from an x-only key. I'm not sure if this is the correct semantics, but it was deliberately written, so we'll keep it. **However** the behavior was different for multipath xprivs. There we returned `None`, i.e. "key not found/mismatch" rather than an error.

This PR returns an error in both cases.

### DIFF
--- a/src/descriptor/key_map.rs
+++ b/src/descriptor/key_map.rs
@@ -92,6 +92,8 @@ impl GetKey for KeyMap {
             .find_map(|(_desc_pk, desc_sk)| -> Option<PrivateKey> {
                 match desc_sk.get_key(key_request.clone(), secp) {
                     Ok(Some(pk)) => Some(pk),
+                    // When looking up keys in a map, we eat errors on individual keys, on
+                    // the assumption that some other key in the map might not error.
                     Ok(None) | Err(_) => None,
                 }
             }))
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK df8d86011d51731c7d0d8af97263b90db22c2384

src/descriptor/key_map.rs
```diff
@@ -336,4 +336,81 @@ mod tests {
...
+        let sk =
+            PrivateKey::from_str("cMk8gWmj1KpjdYnAWwsEDekodMYhbyYBhG8gMtCCxucJ98JzcNij").unwrap();
+        let xonly_pk = sk.public_key(&secp).inner.x_only_public_key().0;
+        let request_x = KeyRequest::XOnlyPubkey(xonly_pk);
+
+        let result = descriptor_sk.get_key(request_x.clone(), &secp);
+        assert!(matches!(result, Err(GetKeyError::NotSupported)));
+
+        // Also test with KeyMap; as in the previous test, the error turns to None.
+        let descriptor_s = "wpkh([d34db33f/84h/1h/0h]tprv8ZgxMBicQKsPd3EupYiPRhaMooHKUHJxNsTfYuScep13go8QFfHdtkG9nRkFGb7busX4isf6X9dURGCoKgitaApQ6MupRhZMcELAxTBRJgS/*)";
```
oleonardolima, line 408: question: any reason not to test with the multipath here too?

oleonardolima (commented):
Thanks! Sorry for the oversight on the multipath descriptors.

Overall it looks good to me, just a minor comment on the test. I'll apply the tests to the backport PR.

apoelstra (commented):
On 89b5048c49c905fa3717200081aa75011ad43599 successfully ran local tests

oleonardolima (commented):
ACK 89b5048c49c905fa3717200081aa75011ad43599

src/descriptor/key_map.rs
oleonardolima: Although I suggested the changes, I noticed that I wont' be able to backport this to the release/12.x as it didn't converted Xpriv to Xpub properly for multipath: https://github.com/rust-bitcoin/rust-miniscript/pull/757, and get this error.

PR #871, opened 2025-10-26
Remove a bit of validation-related cruft

apoelstra:
This PR contains two mostly-unrelated commits.

The first removes the `Ctx::check_witness` function, which is a no-op for the overwhelmingly common case that users are working with sane Miniscripts, and probably wrong in uncommon cases.

The second one removes the `decode_insane` method, replacing it with `decode_consensus` which allows much more stuff. The premise here is that "insane" Miniscripts are about allowing well-formed scripts that might not be analyzable because they violate some sanity rules, while "consensus" Miniscripts are anything the library can parse, no matter how half-baked or incomplete it is. When parsing strings we want "insane", when parsing on-chain scripts we want "consensus". The difference, for now, is that "consensus" Miniscripts can have raw pkhs while "insane" ones cannot.

More details in the commit messages.

### DIFF
--- a/src/interpreter/inner.rs
+++ b/src/interpreter/inner.rs
@@ -8,7 +8,7 @@ use bitcoin::Witness;
 use super::{stack, BitcoinKey, Error, Stack};
 use crate::miniscript::context::{NoChecks, ScriptContext, SigType};
 use crate::prelude::*;
-use crate::{BareCtx, ExtParams, Legacy, Miniscript, Segwitv0, Tap, ToPublicKey, Translator};
+use crate::{BareCtx, Legacy, Miniscript, Segwitv0, Tap, ToPublicKey, Translator};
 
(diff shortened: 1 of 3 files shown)


apoelstra:
@sanket1729 this one should be easy to review -- the diff is big but it's mostly about updating test cases. The actual code changes are to remove obscure functionality that I doubt any user has ever (intentionally) used.

Since it removes the `decode_insane` method, which we just renamed a couple PRs ago, I'd like to get this in before the 13.0 release.

apoelstra (commented):
On bd8fca6cd7724f8e275d72d9c7816683fbe823b0 successfully ran local tests

tcharding (approved):
ACK bd8fca6cd7724f8e275d72d9c7816683fbe823b0 - My review does not hold much weight.

PR #867, opened 2025-10-17
fix(keymap): backport use of correct `derivation_path` for keys with origin

oleonardolima:
backports #872

### Description

It backports the fix for deriving the private key correctly in scenarios where the key origin is available, see #872.

The GetKey for DescriptorSecretKey (and thus KeyMap) returned the wrong key when an extended private key with key origin info was queried via KeyRequest::Bip32.

The key request path is master-relative, but the xkey is anchored at its origin (e.g. [d34db33f/84h/1h/0h]). The old code stripped the path returned by matches(); the fix strips the origin prefix instead (and uses the full path when there is no origin) before deriving from the xkey.

cherry-picks 20ae466ec598f7cf973027cefbb433a46b3a59cf and a2f4497598c9b388d22e5a4dececd99896774116.

### Changelog notice

```
### Fixed
- keymap: fix `GetKey` derivation path for `Xprv` with key origin info [#872](https://github.com/rust-bitcoin/rust-miniscript/pull/872)
```

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [x] I've added tests to reproduce the issue which are now passing
* [x] I'm linking the issue being fixed by this PR

PR #872, opened 2025-10-31
fix(key_map): derivation_path for `Xprv` with key origin info

oleonardolima:
### Description

The `GetKey` for `DescriptorSecretKey` (and thus `KeyMap`) returned the wrong key when an extended private key with key origin info was queried via `KeyRequest::Bip32`.

The key request path is master-relative, but the `xkey` is anchored at its origin (e.g. `[d34db33f/84h/1h/0h]`). The old code stripped the path returned by `matches()`; the fix strips the origin prefix instead (and uses the full path when there is no origin) before deriving from the `xkey`.

--

Also, this PR introduces the test `get_key_xpriv_with_key_origin` covering the scenarios: bare wildcard; single fixed step; fixed step then wildcard, and the matching/non-matching cases build requests from explicit `(fingerprint, path)` key sources.

### Notes

If you'd like an overview on the issue, see: https://github.com/rust-bitcoin/rust-miniscript/pull/863#discussion_r2418425341

### Changelog notice

```​
### Fixed
- keymap: fix `GetKey` derivation path for `Xprv` with key origin info [#872](https://github.com/rust-bitcoin/rust-miniscript/pull/872)
```

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [x] I've added tests to reproduce the issue which are now passing
* [ ] I'm linking the issue being fixed by this PR

### DIFF
--- a/src/descriptor/key_map.rs
+++ b/src/descriptor/key_map.rs
@@ -174,7 +174,7 @@ impl GetKey for DescriptorSecretKey {
 mod tests {
     use core::str::FromStr;
 
-    use bitcoin::bip32::{ChildNumber, DerivationPath, IntoDerivationPath, Xpriv};
+    use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, IntoDerivationPath, Xpriv};
 
     use super::*;
     use crate::Descriptor;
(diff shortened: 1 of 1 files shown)


oleonardolima:
@notmandatory I updated this PR and added new test scenarios, it should be ready to review.

cc'in @noahjoeris as you had interest in this and a review would be appreciated.

notmandatory (approved):
ACK a2f4497598c9b388d22e5a4dececd99896774116

Changes look good, tests easy to follow.

noahjoeris (approved):
tACK a2f4497

Looking good, thanks.

oleonardolima:
ping @tcharding @apoelstra for awareness.

apoelstra:
Oops, missed this somehow. Will review it and the backport today.

apoelstra (approved):
ACK a2f4497598c9b388d22e5a4dececd99896774116; successfully ran local tests

evanlinjin:
@oleonardolima I noticed that `KeyMapWrapper` of the `release-12.x` branch also has the same bug. Would you be able to port the fix there since `bdk_tx` depends on that version of miniscript? Let me know if you would like me to do it instead. Thanks in advance.

noahjoeris:
@evanlinjin it's here https://github.com/rust-bitcoin/rust-miniscript/pull/867

oleonardolima:
> @oleonardolima I noticed that `KeyMapWrapper` of the `release-12.x` branch also has the same bug. Would you be able to port the fix there since `bdk_tx` depends on that version of miniscript? Let me know if you would like me to do it instead. Thanks in advance.

@evanlinjin It's already covered in #867 as Noah commented above.

PR #880, opened 2025-11-21
ci: free bloat for MSRV job

nyonson:
The MSRV job is running out of memory in CI due to a recent change to Github's underlying runner. Sounds like this will be resolved at some point in the future, actions/runner-images#13315, but until then remove the extra bloat.

Copying over from rust-bitcoin: https://github.com/rust-bitcoin/rust-bitcoin/pull/5310


apoelstra (approved):
ACK dfe17ba3a34ae07e5dc5c8eb275d9c1ba62f09b8; successfully ran local tests

PR #884, opened 2025-12-13
Fix: lifted policy produces invalid non-binary AND/OR 

adyshimony:
## Summary

The `normalized()` function in semantic policy was incorrectly flattening nested AND/OR thresholds, producing non-binary `AND`/`OR` which violates the Miniscript spec where these are binary operators (exactly 2 arguments).

## The Bug

When lifting a Miniscript like:
`or_d(pk(Alice),and_v(and_v(v:hash160(H),v:pk(Bob)),older(144)))`

**Step 1 - lift() creates correct binary structure:**
`or(pk(Alice),and(and(hash160(H),pk(Bob)),older(144)))`

**Step 2 - normalized() incorrectly flattens it:**
`or(pk(Alice),and(hash160(H),pk(Bob),older(144)))`
The `AND` now has 3 arguments, which is invalid per the Miniscript spec.

## Root Cause

In `normalized()`, when the parent threshold is an AND and the child is also an AND (or OR/OR), the code was flattening by extending the parent's children with the child's children. 

## The Fix

Remove the flattening logic and always preserve nested Thresh structures. This maintains binary AND/OR as required by the Miniscript spec.  

## Testing

All existing tests pass after the fix.

Added lift_or_d test to verify nested AND/OR remain binary after lift.

It can easily be reproduced by Miniscript Studio [here](https://adys.dev/miniscript#miniscript=or_d(pk(Alice)%2Cand_v(and_v(v%3Ahash160(6c60f404f8167a38fc70eaf8aa17ac351023bef8)%2Cv%3Apk(Bob))%2Colder(144)))) - lift it with the Miniscript ⬆️ button to see that the lifted Policy is flattened by `normalized()`.  Miniscript Studio uses rust-miniscript latest version, 12.0. 
 
Expected:  `or(pk(Alice),and(and(hash160(6c60f404f8167a38fc70eaf8aa17ac351023bef8),pk(Bob)),older(144)))`
  
Rust Miniscript returns:  `or(pk(Alice),and(hash160(6c60f404f8167a38fc70eaf8aa17ac351023bef8),pk(Bob),older(144)))`

### DIFF
--- a/src/policy/mod.rs
+++ b/src/policy/mod.rs
@@ -233,6 +233,8 @@ impl<Pk: MiniscriptKey> Liftable<Pk> for Arc<Concrete<Pk>> {
 mod tests {
     use core::str::FromStr;
 
+    use bitcoin::hashes::hash160;
+
     use super::*;
     #[cfg(feature = "compiler")]
     use crate::descriptor::Tr;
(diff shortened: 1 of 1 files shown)


apoelstra:
I find this LLM output extremely grating to read, and there is no spec for the abstract policy language produced by the `lift` function so I don't see how this can be "out of spec".

apoelstra:
Yeah this "fix" is just wrong.

If you disagree please provide a human response explaining why.

adyshimony:
Sorry but every single line in this description is mine, only the structure was LLM. 
I think it took me 2 hours just to prepare this PR description, first PR here so I really tried to be clear and to provide examples, I’m sorry it came across as the opposite.

Miniscript Lift to Policy creates an expression that is not compiling back, I provided an example and a test.

1- lift() creates correct binary structure:
or(pk(Alice),and(and(hash160(H),pk(Bob)),older(144)))

2 - normalized()  flattens it:
or(pk(Alice),and(hash160(H),pk(Bob),older(144)))
The AND now has 3 arguments, which is not compiling by the library:

Error : "And policy fragment must take 2 arguments"

Does my fix breaks things? 
Maybe the right fix is to support AND with 3 parameters?

apoelstra:
I apologize for my aggressive reply, and I appreciate the effort you put into finding examples. But the heading/paragraph structure, the names of the headings, etc., are written in an LLM tone; the bug description conflates miniscripts and policies; and the "fix" itself simply deletes logic with no commentary about why that logic might have been there.

>Miniscript Lift to Policy creates an expression that is not compiling back,

It sounds like this is the real issue -- that you expect the "semantic policy" produced by `lift` to be a compilable policy. It will not. The "semantic policy" produced by `lift` is a completely different object than the "Miniscript policy" used by the compiler, on the demo website, on Miniscript Studio, etc.

Neither policy language has a specificatation but I guess the "policy language" has a de facto spec based on what this library and sipa's website do.

>I provided an example and a test.

It looks like this is actually a bug in Miniscript Studio -- I'm not sure why it provides a "lift" button. This object this function produces is to support analysis passes and it's not really clear why you would show it to a user.

>1- lift() creates correct binary structure:
> `or(pk(Alice),and(and(hash160(H),pk(Bob)),older(144)))`

It is an accident that `lift` produces a compileable policy.

> Maybe the right fix is to support AND with 3 parameters?

No, the right fix is to remove the string format for the semantic policy, and/or change it so it's clearly not a compilable policy.

apoelstra:
> Yes, and this feature is so nice. I really liked the fact that the user can lift the script back to Miniscript and Policy together, and understand the clean Policy. 

Yeah, agreed, this does seem to have educational/informative value. I think we should support it somehow. But there is the question of probabilities -- do we want to attempt to infer probabilites from a Miniscript by e.g. deciding why one construction was chosen over another?

I think that could be interesting -- it would also mean that you could e.g. compile a 99%/1% policy, get a miniscript that reverses to a 67%/33% policy, and then learn "how much" your original probabilities influenced the script. I dunno.

In any case, should have a separate issue/discussion about it.

>Thanks again.

Sorry again. I feel bad about directing LLM scorn at a human.

PR #891, opened 2026-01-13
 Use stack allocation in TrSpendInfo

lmorett1:
Replaces the heap-allocated `Vec` in [nodes_from_tap_tree](cci:1://file:///d:/friendly-project/rust-miniscript/src/descriptor/tr/spend_info.rs:103:4-176:5) with a fixed-size stack array, addressing an existing FIXME.

Since Taproot tree depth is strictly consensus-limited to 128, using a small stack array (~2KB) eliminates unnecessary heap allocations during tree traversal without any risk of overflow.

### DIFF
--- a/src/descriptor/tr/spend_info.rs
+++ b/src/descriptor/tr/spend_info.rs
@@ -47,6 +47,39 @@ impl BitStack128 {
     }
 }
 
+/// Stack of `(bool, usize)` elements, with a maximum capacity of 128.
+///
+/// Used to replace `Vec<(bool, usize)>` in recursive algorithms to avoid heap allocation.
+struct ParentStack {
+    data: [(bool, usize); 128],
+    len: usize,
+}
+
+impl ParentStack {
+    fn new() -> Self { Self { data: [(false, 0); 128], len: 0 } }
+
+    fn push(&mut self, val: (bool, usize)) {
+        // We know that the depth is limited to 128 by the bounds check in `TapTree::combine`
+        // and other places, so this strictly shouldn't happen with valid trees.
+        debug_assert!(self.len < 128, "ParentStack overflow");
+        if self.len < 128 {
+            self.data[self.len] = val;
+            self.len += 1;
+        }
+    }
+
+    fn pop(&mut self) -> Option<(bool, usize)> {
+        if self.len > 0 {
+            self.len -= 1;
+            Some(self.data[self.len])
+        } else {
+            None
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK 94eea850493e7ec610675f0b06ca4447140c8051

Code review only, I do not know about the 128 depth limit thing.

tcharding (approved):
ACK 90b1be3809dcfad54d191e2bce0f827702cc4694

apoelstra:
We don't have an `arrayvec` dependency here. (There is one which is inherited from hex-conservative 0.2 but it's not exposed and won't obviously be retained when we move to 1.0.)

This code looks fine to me but I'd prefer that it used `TAPROOT_CONTROL_MAX_NODE_COUNT` rather than 128, as in taptree.rs, so that the two bounds checks are comparable.

src/descriptor/tr/spend_info.rs
```diff
@@ -47,6 +47,39 @@ impl BitStack128 {
...
+    fn new() -> Self { Self { data: [(false, 0); 128], len: 0 } }
+
+    fn push(&mut self, val: (bool, usize)) {
+        // We know that the depth is limited to 128 by the bounds check in `TapTree::combine`
+        // and other places, so this strictly shouldn't happen with valid trees.
+        debug_assert!(self.len < 128, "ParentStack overflow");
+        if self.len < 128 {
+            self.data[self.len] = val;
+            self.len += 1;
+        }
```
vadim-anfv, line 68: The depth is already bounded when the tree is built (`taptree.rs` rejects depths above `TAPROOT_CONTROL_MAX_NODE_COUNT - 1`), so both the `debug_assert!` and the `if` are dead code here. Writing unconditionally would be simpler: the array is 128 long, so an out-of-range write panics on the bounds check anyway, in any profile.

vadim-anfv (commented):
@lmorett1 would you mind if I picked this up? I would keep your commits and address the open review comments.

apoelstra:
Sorry for not reviewing this. I have two comments:

* @lmorett1 is right that we can drop the overflow checks (though we should leave the comment in place explaining why we don't need them)
* Inside `ParentStack` can we split this into a `BitStack` and a `[usize; 128]`? This will nearly halve the size of the stack (a `(bool, usize)` will get padded to 16 bytes on a typical 64-bit platform vs `usize` alone which will stay a 8).

I'll leave it up to you whether you want to address these or allow @lmorett1 to take over.

Issue #734, opened 2024-08-31
When are sanity_checks supposed to be run?


apoelstra:
On the `Miniscript` type, which users are not really supposed to use directly, we run sanity checks (e.g. all branches are "safe") on `from_str`. You have to use `from_str_ext` or `from_str_insane` to override this.

However, we don't do the same for `Descriptor::from_str`, which simply parses a tree then calls `Miniscript::from_tree`, bypassing all sanity checks. So this seems backward -- by default for the users-shouldn't-use-this type we have sanity checks, while for descriptors, you have to manually call the `sanity_check` method.

However^2, for Taproot descriptors we **do** run the sanity checks on parsing, because we have bizarre special-purpose parsing logic which actually calls `Miniscript::from_str` for individual tapbranches rather than calling `Miniscript::from_tree`.

You can see this with the following unit test

```diff
commit 966524264e948b3e68f68a10eba513692f2839e6
Author: Andrew Poelstra <apoelstra@wpsoftware.net>
Date:   Sat Aug 31 20:36:29 2024 +0000

    f unit test

diff --git a/src/descriptor/mod.rs b/src/descriptor/mod.rs
index 0b0bd02b..72f533f7 100644
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -2003,6 +2003,12 @@ pk(03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8))";
         Descriptor::<DescriptorPublicKey>::from_str("wsh(andor(pk(tpubDEN9WSToTyy9ZQfaYqSKfmVqmq1VVLNtYfj3Vkqh67et57eJ5sTKZQBkHqSwPUsoSskJeaYnPttHe2VrkCsKA27kUaN9SDc5zhqeLzKa1rr/0'/<0;1;2;3>/*),older(10000),pk(tpubD8LYfn6njiA2inCoxwM7EuN3cuLVcaHAwLYeups13dpevd3nHLRdK9NdQksWXrhLQVxcUZRpnp5CkJ1FhE61WRAsHxDNAkvGkoQkAeWDYjV/8/<0;1;2>/*)))").unwrap_err();
     }

+    #[test]
+    fn taproot_s_check() {
[...]

PR #895, opened 2026-02-01
miniscript: rewrite satisfier to be non-recursive

apoelstra:
Currently the `Miniscript::satisfy` function is implemented by the function `Satisfaction::satisfy_helper` which is mutually recursive with the function `Satisfaction::dissatisfy_helper`. There are a few issues with this:

1. Rust does not handle recursion well, and this may stack overflow.
2. We need to fish several parameters (closures for computing min/thresh based on whether we allow malleability; the Taproot leaf hash which might be garbage for non-Taproot outputs; a boolean indicating whether we are hassig) through every single recursive call, which results in a *lot* of noise and makes the code very hard to maintain. (In particular, I want to fix the Taproot leaf hash thing, but without rewriting the satisfier, this requires changing several dozen call sites.)
3. The logic is very hard to follow and mixes somewhat-complicated recursion logic with actual satisfaction logic, which might be wrong (there is one logic change I made related to timelocks; I have a FIXME to find a test case for this because I am pretty sure the old logic was wrong.)

So I rewrote the whole function. I haven't benchmarked but I suspect the new one is a little slower, though it is asymptotically the same. The slowness comes from using heap-based rather than stack-based recursion, and also because the old algorithm could sometimes skip computing dissatisfactions while this one never does. I think the increase in code clarity justifies the loss of performance (if there is one).

See the last commit message for more details.


apoelstra:
Hrpmh. Apparently I am using a [FromIterator impl from 1.79](https://doc.rust-lang.org/std/iter/trait.FromIterator.html#impl-FromIterator%3C(T,)%3E-for-(ExtendT,)).

apoelstra (commented):
On 2b9c724d7d0bbdd5ac7a5b6f99ed7078b94b63a8 successfully ran local tests

apoelstra:
Rebased after #915

src/miniscript/satisfy/sat_dissat.rs
```diff
@@ -0,0 +1,477 @@
...
+        } else {
+            Self::minimum
+        };
+        let thresh_fn = if malleable {
+            Self::thresh_mall
+        } else {
+            Self::thresh
+        };
+
+        let mut stack = vec![];
```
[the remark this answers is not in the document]
apoelstra: Yeah, not a bad idea. I added `with_capacity(128)`. As I said above, I haven't benchmarked this (it's a little tricky/annoying to benchmark satisfaction in a realistic way).

apoelstra:
Rebased after #920 and #921.

trevarj:
Can't provide much for a review. Just admiring the cleanup!

tcharding:
ACK 14a79f22a5489c4b535fbf555c3d46a782920e1f

I can't help with the last patch because its over my head. The first 5 look good.

tcharding:
How do you want to move forward with this @apoelstra? Wait for Sanket or want me and/or @trevarj to dig into something in particular that would give you confidence in our review? (Assuming Trevar wants to do that.)

src/miniscript/satisfy/sat_dissat.rs
```diff
@@ -0,0 +1,487 @@
...
+                        Witness::combine(acc, Witness::Stack(sig))
+                    }),
+                    has_sig: true,
+                    ..Self::TRIVIAL
+                },
+            )
+        }
+    }
+
+    /// The (dissatisfaction, satisfaction) pair for a `multi` fragment.
```
trevarj:
```suggestion
    /// The (dissatisfaction, satisfaction) pair for a `multi_a` fragment.
```

src/miniscript/satisfy/sat_dissat.rs
```diff
@@ -0,0 +1,487 @@
...
+    where
+        S: AssetProvider<Pk>,
+    {
+        (
+            Self { stack: Witness::hash_dissatisfaction(), ..Self::TRIVIAL },
+            Self { stack: Witness::hash160_preimage(stfr, h), ..Self::TRIVIAL },
+        )
+    }
+
+    /// The (dissatisfaction, satisfaction) pair for a `ripemd256` fragment.
```
trevarj:
```suggestion
    /// The (dissatisfaction, satisfaction) pair for a `sha256` fragment.
```

src/miniscript/satisfy/sat_dissat.rs
```diff
@@ -0,0 +1,487 @@
...
+    where
+        S: AssetProvider<Pk>,
+    {
+        (
+            Self { stack: Witness::hash_dissatisfaction(), ..Self::TRIVIAL },
+            Self { stack: Witness::sha256_preimage(stfr, h), ..Self::TRIVIAL },
+        )
+    }
+
+    /// The (dissatisfaction, satisfaction) pair for a `ripemd256` fragment.
```
trevarj:
```suggestion
    /// The (dissatisfaction, satisfaction) pair for a `hash256` fragment.
```

apoelstra:
Hi @tcharding. I reached out to @sanket1729 out of band and he said it's okay if we push forward on this without him.

I would appreciate a cursory review from you and/or @trevarj. You don't need to verify correctness of the rewrite. Hopefully the tests are thorough enough to catch any mistakes. Miniscript 12 is the "reference version" and we have pretty thorough fuzztests against it, although I don't think we're fuzzing the satisfier.

In parallel to this project we should try to improve those fuzztests -- in particular, finding a way to improve the satisfier. I think we want to have a "fuzz library" similar to what exists in rust-simplicity, which is able to create fuzzer-guided complete transaction chains.

src/miniscript/satisfy/sat_dissat.rs
```diff
@@ -0,0 +1,487 @@
...
+                        min_fn(l_sat, Self::concatenate_rev(l_dis, r_sat)),
+                    )
+                }
+                Terminal::OrI(_, _) => {
+                    let (r_dis, r_sat) = stack.pop().unwrap();
+                    let (l_dis, l_sat) = stack.pop().unwrap();
+                    // FIXME existing code sets timelocks to None instead of propagating them,
+                    // in the dissat case. This dates to the original plan module and was not
+                    // comment on. https://github.com/rust-bitcoin/rust-miniscript/pull/592
+                    // Was likely a mistake. Need to find a test case that distinguishes before
```
trevarj: Will we follow up on this test?

Staring at this for a long time wondering which dissatisfaction actually has a timelock that isn't IMPOSSIBLE.

trevarj (commented):
Did another review and tried to study sat_dissat.rs harder. Few doc nits found.

src/miniscript/satisfy/sat_dissat.rs
apoelstra: Can you indicate which commit you are talking about? Github provides no hint except 4 lines of diff.

apoelstra: Appears to be 14a79f22a5489c4b535fbf555c3d46a782920e1f

trevarj: Yes, sorry, I did this review on c31154d4a53cba6279a592a6ff0191237449f2e7 but it seems that it was introduced in https://github.com/rust-bitcoin/rust-miniscript/commit/14a79f22a5489c4b535fbf555c3d46a782920e1f

src/miniscript/satisfy/sat_dissat.rs
apoelstra: Let me follow up on this. Now that we have good LLMs I should be able to gin up a test case.

>Staring at this for a long time wondering which dissatisfaction actually has a timelock that isn't IMPOSSIBLE.

For example, to dissatisfy an `and_v(<timelock>, <something else>)` you need to satisfy the timelock but dissatisfy the `<something else>`.

This may be the only example.

apoelstra: I have constructed an example. Amusingly, while the old code would produce an invalid dissatisfaction, the new code produces no dissatisfaction at all, which is arguably no better. But fixing *that* is nontrivial and out of scope for this PR.

I will create a test case with a detailed comment analyzing this situation and add it to this PR.

apoelstra: > the new code produces no dissatisfaction at all,

Actually I think it produces no dissatisfaction at all because of an unrelated bug (the comment "Dissatisfactions don't need to non-malleable." is false). Probably I need to fix that separately and backport it.

apoelstra: Ah, the "unrelated bug" is a combination of https://github.com/rust-bitcoin/rust-miniscript/issues/734 which is fine to just leave alone for now, and a misunderstanding I had about malleability of dissatisfactions.

apoelstra:
@trevarj 

* addressed your copy/paste errors in comments
* replaced a call to `min_fn` with `Self::minimum_mall` to match the old code; replaced the FIXME comment with a detailed comment justifying the use of `minimum_mall`
* added a commit with a regression test that covers both the FIXME and the new `minimum_mall` thing

This should be good to go now.

trevarj (approved):
ACK a662453

apoelstra:
On a6624537b5f19b0964f9571b1f0c10eed6044efb successfully ran local tests

tcharding:
The module is called `sat_dissat` but the function is `dissat_sat` and returns `(dissat, sat)`. Can we have them the same? And here there are around the other way

```rust
            // Satisfaction is more complicated.
            sat_dissat_vec.push((sub.sat_data, sub.dissat_data));
```

We could do a separate PR to change `extra_props` to use `(dissait, sat)` before or after this?

tcharding (approved):
ACK a6624537b5f19b0964f9571b1f0c10eed6044efb

apoelstra:
@tcharding sure, in my next PR I'll add a commit that swaps them.

Probably I should just make a utility struct for these. When writing and reviewing this code I *repeatedly* got the sat and dissat branches switched in my head.

PR #897, opened 2026-02-03
Append witness script for P2WSH in `Plan::satisfy()`

### LINKED ISSUE #896: `Plan::satisfy()` does not append witnessScript for P2WSH descriptors
## Description

`Plan::satisfy()` does not append the witnessScript as the final witness stack item for `Wsh`/`WshSortedMulti` (and by extension `ShWsh`/`ShWshSortedMulti`) descriptor types. This produces an incomplete witness that fails validation with "Witness program hash mismatch".

Per BIP 141, a P2WSH witness must be `[...satisfaction_items..., witnessScript]`.

## Expected behavior

`Descriptor::get_satisfaction()` handles this correctly — `Wsh::get_satisfaction()` explicitly appends `self.inner_script()` to the witness stack before returning.

`Plan::satisfy()` should do the same for `Wsh`/`WshSortedMulti`/`ShWsh`/`ShWshSortedMulti`.

## Actual behavior

`Plan::satisfy()` groups `Wsh` and `WshSortedMulti` together with `Wpkh` and `Tr`, returning only the raw satisfaction items without appending the witnessScript. `Wpkh` and `Tr` don't need a trailing script, but `Wsh` and `WshSortedMulti` do.

Similarly, the `ShWsh`/`ShWshSortedMulti` arm returns the correct P2SH scriptSig but is also missing the witnessScript in the stack.

Broadcasting the resulting transaction fails with "Witness program hash mismatch".

## Workaround

After calling `plan.satisfy()`, append the witnessScript from the PSBT input (which is correctly populated by `plan.update_psbt_input()`).

## Version

`miniscript 13.0.0`

evanlinjin:
Fix Plan::satisfy() to correctly append the witnessScript as the final witness stack element for P2WSH descriptor types (Wsh, WshSortedMulti, ShWsh, ShWshSortedMulti).

Previously, these types were incorrectly grouped with Wpkh and Tr, which don't require a trailing witness script. This caused transactions built using Plan::satisfy() to fail validation with "Witness program hash mismatch" when broadcast.

The fix separates the descriptor type handling:
- Wpkh/Tr: return stack as-is (no witness script needed)
- Wsh/WshSortedMulti: append witness script, empty script_sig
- ShWpkh: return stack with unsigned_script_sig (no witness script)
- ShWsh/ShWshSortedMulti: append witness script and unsigned_script_sig

Closes #896

### DIFF
--- a/bitcoind-tests/tests/test_desc.rs
+++ b/bitcoind-tests/tests/test_desc.rs
@@ -5,6 +5,7 @@
 //!
 
 use std::collections::BTreeMap;
+use std::str::FromStr;
 use std::{error, fmt};
 
 use actual_rand as rand;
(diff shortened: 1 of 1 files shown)


evanlinjin:
The other option is to update `Plan::satisfy` docs to mention that `witnessScript` will not be pushed to the witness stack. However, I don't think that makes any sense.

bitcoind-tests/tests/setup/test_util.rs
```diff
@@ -25,8 +25,7 @@ use bitcoin::hex::DisplayHex;
 use bitcoin::secp256k1;
 use miniscript::descriptor::{SinglePub, SinglePubKey};
 use miniscript::{
-    bitcoin, hash256, Descriptor, DescriptorPublicKey, Error, Miniscript, ScriptContext,
-    Translator,
+    bitcoin, hash256, Descriptor, DescriptorPublicKey, Error, Miniscript, ScriptContext, Translator,
```
tcharding: nit: If you feel like it can you throw a patch at the front of this PR that runs the formatter in `bitcoind-tests` please mate so that these fromatting changes aren't mixed in with your changes.

bitcoind-tests/tests/test_desc.rs
```diff
@@ -419,3 +424,160 @@ fn test_satisfy() {
     let cl = &setup::setup().client;
     test_descs(cl, &testdata);
 }
+
+/// Test that Plan::satisfy() correctly constructs witness for P2WSH descriptors.
+/// This is a regression test for https://github.com/rust-bitcoin/rust-miniscript/issues/896
```
tcharding: I rekon we can trim a bit more LLM noise out of this patch. I see no value in these comments.

tcharding:
Thanks @evanlinjin. I did a more thorough read of the test code. Its not perfect but I retract my statement that there is just too much test code. Thanks for iterating.

My review doesn't hold too much weight in this repo but FWIW this LGTM.

One nit: can we format with the nightly toolchain like we do for the rest of the repo?

Reviewed: 026943a2617dbce9742e085ed150b024bd8a2940

tcharding (approved):
ACK 871e953f6ea193d0b927c42a11d3f72a9ff296a3

apoelstra (approved):
ACK 871e953f6ea193d0b927c42a11d3f72a9ff296a3; successfully ran local tests

PR #906, opened 2026-03-03
Add compile_tr_native() for Taptree-native policy compilation

### LINKED ISSUE #905: compile_tr: emit Or branches as separate Tapleaves instead of OP_IF/NOTIF
The policy compiler's Taproot path currently compiles Or branches into single Tapleaves using OP_IF/NOTIF, carrying over the P2WSH pattern. In Tapscript, the Taptree itself provides a branching mechanism via separate leaves.

For simple policies, decomposing Or branches into separate leaves can improve privacy (only the spent leaf is revealed on-chain). For complex policies (e.g. large conjunctions of ors, or high-k thresholds), full decomposition is impractical due to combinatorial explosion, so a `max_leaves` cap is provided.

Proposed approach: add `compile_tr_native()` as an opt-in alternative that decomposes Or/Thresh at the policy level, compiles each branch to an IF-free leaf using the existing Tap compiler, then assembles an optimal Taptree via Huffman coding.

kwsantiago:
Closes #905

Adds `compile_tr_native()` which decomposes all `Or` and `Thresh` branches into separate TapTree leaves instead of using `OP_IF`/`OP_NOTIF` within a single leaf. Each leaf is compiled to branch-free Tapscript and assembled into a Huffman-weighted TapTree.

Also refactors `enumerate_policy_tree` into a shared `enumerate_leaves` helper to avoid duplication.

### DIFF
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -64,6 +64,11 @@ pub enum CompilerError {
         /// Maximum allowed number of Tapleaves.
         max: usize,
     },
+    /// Native Taproot compilation produced a leaf containing OP_IF/NOTIF.
+    IfFragmentInNativeLeaf {
+        /// Index of the leaf that contains branching fragments.
+        leaf_index: usize,
+    },
     ///Policy related errors
     PolicyError(policy::concrete::PolicyError),
 }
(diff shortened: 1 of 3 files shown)


apoelstra (approved):
ACK ae5c7d303d944fd7a3d5ac96b66d439b36da0aa7; successfully ran local tests

apoelstra (approved):
ACK a46031c160c5d24392e3f527635778ab26c4d0b3; successfully ran local tests

JeremyRubin:
"posthumous nack"*; i think this creates somewhat surprising behavior.

Do i understand correctly that something like:

`<bob> checksigverify IF 12 CSV <Alice> ELSE <Carol> ENDIF Checksig`

would compile to

`<bob> checksigverify 12 CSV <Alice> Checksig`, `<bob> checksigverify <Carol>  Checksig`?

These are not functionally the same because the former guarantees that _one_ signature from Bob is sufficient for either Alice or Carol to sign, whereas the latter you need one signature per script leaf from Bob.


* not a real nack -- this is good behavior in a lot of use cases, but can be surprising if not tightly documented.

kwsantiago:
@JeremyRubin Yes, sig commits to the tapleaf hash, so a key in multiple leaves signs each one. Still expressible, just per-leaf sigs at sign time. This tradeoff is why it's opt-in.

apoelstra:
Let's move this discussion to https://github.com/rust-bitcoin/rust-miniscript/issues/1000

Issue #885, opened 2025-12-14
Should remove `Display` impl for `semantic::Policy`


apoelstra:
The `Display` impl for semantic policies is misleading to users because it resembles a compileable policy (and the objects have very similar names). This, combined with the fact that there's no way to go from a Miniscript back to a compileable policy except by abusing `lift`, encourages people to do this.

Apparently Miniscript Studio has exposed a "lift" button that produces a semantic policy and then stringifies it into a box labeled "policy" that users can attempt to compile from. This will appear to successfully round-trip in many cases but not others, leading to the impression of bugs.

It may be that people want to present the semantic policies to their users in some way, such as a graphical representation. But presenting it as a flat string that resembless a policy is bound to lead to confusion, and we shouldn't support that.

----

EDIT: To close this issue do bullet points 1-3 from comment below: https://github.com/rust-bitcoin/rust-miniscript/issues/885#issuecomment-3665732418

tcharding:
@apoelstra please check/ack the edit to the PR description and pulling point 4 out into #926. And pull me up if I'm doing too much stuff here, I've been known to be like a chook (chicken) preparing its dirt bath, you know scratching and flapping its wings around getting everything just like it likes it ...

apoelstra:
I believe this is resolved by #914.

apoelstra:
Lol the issue description says "To close this issue do bullet points 1-3 from comment below" and #914 only does bullet point 4.

PR #914, opened 2026-04-01
Use mathematical notation for semantic policy Display

### LINKED ISSUE #926: Fix `Display` impl for `Semantic`
Change the `Display` impl for `Semantic` to output something goofy, like instead of `and(a, b, c)` it does `(a ∧ b ∧ c)`. Whoever PRs to do this should provide several examples so we can debate this, especially the threshold notation. (Maybe we want `(2 = a + b + c)` for `thresh(2, a, b, c)`?)

This issue was first described in #885 (https://github.com/rust-bitcoin/rust-miniscript/issues/885#issuecomment-3665732418) after a PR was up already (#914). Opening this issue so #914 can close it.

febyeji:
## Summary
- Change `semantic::Policy` `Display` impl to use mathematical notation:
  - `and(a,b)` → `(a ∧ b)`
  - `or(a,b)` → `(a ∨ b)`
  - `thresh(k,a,b,c)` → `#{a, b, c} = k`

Close #926


tcharding:
Do we want to close the issue if this merges?

tcharding (approved):
ACK 73ee81877d2c8a45610fe62ab0bd135f681493bc

fuzz/fuzz_targets/regression_descriptor_parse.rs
```diff
@@ -14,31 +14,29 @@ fn do_test(data: &[u8]) {
...
-                (Ok(new), Ok(old)) => {
-                    assert_eq!(
-                        old.to_string(),
-                        new.to_string(),
-                        "lifted input {} (left is old, right is new)",
-                        data_str
-                    )
+            // Lifted semantic policy Display format has intentionally changed
+            // to mathematical notation. Cross-crate structural comparison is not
+            // possible, so we only verify both crates agree on liftability.
```
apoelstra: In 73ee81877d2c8a45610fe62ab0bd135f681493bc:

Ouch. This is actually a critical check and removing it completely guts our regression testing.

I think we need to preserve the old serialization logic somewhere. Maybe add a new `to_policy_like_string()` or something and document that it's only there for legacy usage.

trevarj:
> Do we want to close the issue if this merges?

@tcharding Don't close https://github.com/rust-bitcoin/rust-miniscript/issues/885 until issues are made for these at least

> * We rename `policy::Concrete` to `policy::Policy`, leaving a type alias to minimize code breakage.
> * We move `policy::Semantic` into its own module so it becomes `semantic::Semantic` (we can also leave a type alias at the old location but we should deprecate it). We also move the `Lift` trait, which I think is in policy/mod.rs, to this own module (we can re-export it at the old location, but sadly we cannot deprecate re-exports).
> * We stop referring to `policy::Semantic` as a "semantic policy" or a "policy" at all anywhere in our docs
> * ~We change the `Display` impl for `Semantic` to output something goofy, like instead of `and(a, b, c)` it does `(a ∧ b ∧ c)`. Whoever PRs to do this should provide several examples so we can debate this, especially the threshold notation. (Maybe we want `(2 = a + b + c)` for `thresh(2, a, b, c)`?)~
> 
> Furthermore, I propose adding a method to "uncompile" from a `Miniscript` to a `Policy` (or even a `Semantic` to a `Policy`), but we should move that to a separate issue because I'm quite conflicted about what to do about percentages.

apoelstra:
No, to close the issue we've also got to do the renames and doc changes to stop referring to semantic policies as "policies".

tcharding:
@febyeji I've changed your PR description to close the new #926. Please check you are happy with that.

febyeji:
@tcharding Looks good, thanks for updating the PR body. Happy with that.

src/policy/semantic.rs
```diff
@@ -248,6 +248,13 @@ impl<Pk: MiniscriptKey> fmt::Debug for Policy<Pk> {
     }
 }
 
+/// Displays the policy using mathematical notation for readability.
+///
+/// - `and(a, b)` is displayed as `(a ∧ b)`
+/// - `or(a, b)` is displayed as `(a ∨ b)`
+/// - `thresh(k, a, b, c)` is displayed as `#{a, b, c} = k`
+///
+/// Note: this format is not parseable.
```
apoelstra: In 73ee81877d2c8a45610fe62ab0bd135f681493bc:

This format is absolutely parseable, and you should fix `from_str` to parse it.

fuzz/fuzz_targets/regression_descriptor_parse.rs
febyeji: Good point. I added `to_policy_syntax_string()` and restored the structural `assert_eq!` tests.

src/policy/semantic.rs
febyeji: Thanks for the review! I fixed the `from_str` to parse the format.

apoelstra:
In eb6011eb6a9b215cf553f4bf5a1a042bcc17beb7:

This parser is gratuitously inefficient. You should be able to just parse the new syntax directly. It is a simple CFG and I don't think it even needs lookahead.

You definitely do not need to rewrite the string into the old syntax, and even if you did, this manages to be both recursive (we should not be introducing any new recursive code) and still manages to rescan the entire string on every iteration.

src/policy/semantic.rs
```diff
@@ -261,24 +270,272 @@ impl<Pk: MiniscriptKey> fmt::Display for Policy<Pk> {
...
+/// Parses a semantic policy from either the mathematical Display form
+/// (`(a ∧ b)`, `(a ∨ b)`, `#{a, b, ..} = k`) or the legacy function-call
+/// form (`and(..)`, `or(..)`, `thresh(k, ..)`).
+///
+/// This is a single linear pass: each byte of input is visited at most
+/// twice (once by the outer scan and, for bytes that belong to a terminal
+/// or to a nested function-call atom, once by [`expression::Tree::from_str`]
+/// which is itself non-recursive). The parser uses explicit frame and
+/// operand stacks rather than recursion.
+fn parse_policy<Pk: FromStrKey>(s: &str) -> Result<Policy<Pk>, Error> {
```
trevarj: I'd like to see the context-free grammar for `Policy` in the doc comments. I think it might be worthwhile to explore a shift-reduce parser since it can be single pass and use a stack. 

Will wait for Andrew's input before reviewing more.

apoelstra: I think it's okay not to bother with the grammar. It's pretty self-explanatory (and I don't want to have to review a pile of BNC and confirm it meets my intuition :P).

I think this parser architecture is okay, except that it parses the old format and appears to allow weird hybrid constructions between the new and old formats. It shouldn't do that.

febyeji:
I simplified the parser and added hybrid rejection. The math parser now refuses atoms whose root name is `and`/`or`/`thresh`, so strings like `(a ∧ or(b, c))` and similar won't parse.

apoelstra:
That solution is insane. I feel like I'm just arguing with a LLM by proxy here. I am not going to provide detailed feedback anymore.

febyeji:
@apoelstra Apologies for the back-and-forth. I wrote the parser myself, but without carefully understanding the codebase. I was trying to be conservative and reuse `expression::Tree`, which compounded the mess since I had also misunderstood your intent and feedback.

I've rewritten the code. If the result isn't what you had in mind, please feel free to close the PR. I will just leave one note in case it helps resolving the issue later.
>  In the current code, `FromStr` accepts the legacy form as well to avoid breaking existing tests and users, but I'm not sure if `FromStr` should accept only the new format.

src/policy/semantic.rs
```diff
@@ -261,24 +266,308 @@ impl<Pk: MiniscriptKey> fmt::Display for Policy<Pk> {
...
+        if s.contains('∧') || s.contains('∨') || s.contains("#{") {
+            parse_math_policy(s)
+        } else {
+            let tree = expression::Tree::from_str(s)?;
+            expression::FromTree::from_tree(tree.root())
+        }
+    }
+}
+
+/// Grammar:
```
trevarj: Andrew said that we don't need a BNF grammar in the doc comment. 
It's a pretty simple grammar, but I prefer a more simplified notation anyway (like regex style for human readability).

I don't like how the grammar includes whitespace. I think it would be better to always skip whitespace to accept `pk(A)       ∧  pk(B)`.

src/policy/semantic.rs
```diff
@@ -261,24 +266,308 @@ impl<Pk: MiniscriptKey> fmt::Display for Policy<Pk> {
...
+    enum Kind {
+        Group(Option<Op>),
+        Thresh,
+    }
+    struct Frame<Pk: MiniscriptKey> {
+        subs: Vec<Arc<Policy<Pk>>>,
+        kind: Kind,
+    }
+
+    fn match_sep(s: &str) -> Option<Op> {
```
trevarj: Drop whitespace and then you can have a `impl From<&str> for Op` and have the `_` case be `unreachable!()` instead of this.

src/policy/semantic.rs
```diff
@@ -261,24 +266,308 @@ impl<Pk: MiniscriptKey> fmt::Display for Policy<Pk> {
...
+        if s.starts_with(" ∧ ") {
+            Some(Op::And)
+        } else if s.starts_with(" ∨ ") {
+            Some(Op::Or)
+        } else {
+            None
+        }
+    }
+
+    let bytes = s.as_bytes();
```
trevarj: I would prefer using a `s.chars().peekable()` (or `char_indices()`) Iterator.
I think it will make things more readable and you won't have to index into the byte slice and skip utf8 bytes. It's not a huge problem, but I think it's just nicer Rust.

After that, then you can have really nice looking `match` statements on the main `while` loop. It's more idiomatic Rust IMO.

You'll probably need some "traditional" parser/lexer helper functions, like skipping whitespace and reading until some char, so it may be worth it to wrap the iterator in a `MathParser` struct or something.

src/policy/semantic.rs
```diff
@@ -261,24 +266,308 @@ impl<Pk: MiniscriptKey> fmt::Display for Policy<Pk> {
...
+        if let Some(new_op) = match_sep(&s[i..]) {
+            match &mut frame.kind {
+                Kind::Group(slot @ None) => *slot = Some(new_op),
+                Kind::Group(Some(prev)) if *prev == new_op => {}
+                _ => return Err(malformed_math()),
+            }
+            frame.subs.push(cur.take().unwrap());
+            i += 5; // " ∧ " / " ∨ " in UTF-8
+        } else if bytes[i] == b',' && bytes.get(i + 1) == Some(&b' ') {
+            if !matches!(frame.kind, Kind::Thresh) {
```
trevarj: Derive `PartialEq, Eq` on `Kind` and you can get rid of `matches!` like this and just use `!=` here and in the other areas.

src/policy/semantic.rs
```diff
@@ -261,24 +266,308 @@ impl<Pk: MiniscriptKey> fmt::Display for Policy<Pk> {
...
+                _ => return Err(malformed_math()),
+            };
+            frame.subs.push(cur.take().unwrap());
+            i += 1;
+            if !s[i..].starts_with(" = ") {
+                return Err(malformed_math());
+            }
+            i += 3;
+            let k_start = i;
+            while i < bytes.len() && bytes[i].is_ascii_digit() {
```
trevarj: For example, this is copied a couple times and could be replaced with some helper to consume until some char, I think.

trevarj (commented):
Leaving some starting feedback on the parser. If the suggestions get implemented then I will do another review and see if we can refine it more.

You may want to not force push and start stacking some commits in case Andrew or someone else doesn't like a change and then you can drop that commit and rebase later.

src/policy/semantic.rs
febyeji: Thanks for the review. I changed the code to accept whitespaces except for the digraph `#{`. I dropped the BNF grammar from the doc comment as well.

src/policy/semantic.rs
febyeji: I refactored the code, but used `From<char>` instead of `From<&str>` since `Op` are of a single literal and we can avoid constructing `&str` from the call site. Happy to switch `&str` if it's a better option to consider.

src/policy/semantic.rs
febyeji: Thank you for the guidance. I refactored the code to wrap the parser state in a `MathParser` struct with a few helpers like `peek, bump, ...`. Most of the byte-index manipulation is now hidden, though some `match` arms may still not be ergonomic to read. Please let me know if there's room to push this further.

src/policy/semantic.rs
febyeji: Thanks, `Kind` now derives the traits and directly uses `== / !=`.

src/policy/semantic.rs
```diff
@@ -364,123 +364,165 @@ impl<Pk: FromStrKey> str::FromStr for Policy<Pk> {
...
     let mut cur: Option<Arc<Policy<Pk>>> = None;
-    let mut i = 0;
 
-    while i < bytes.len() {
+    loop {
+        parser.skip_ws();
+        let c = match parser.peek() {
+            Some(c) => c,
+            None => break,
+        };
```
trevarj: Does this still work if it's changed to?: 
```rust
parser.skip_ws();
while let Some(c) = parser.peek() {
  ...
  parser.skip_ws(); // <- last statementin block
}
```

I'm not actually sure what is better...just a nit and can ignore.

src/policy/semantic.rs
```diff
@@ -364,123 +364,165 @@ impl<Pk: FromStrKey> str::FromStr for Policy<Pk> {
...
-///                | "ripemd160" | "hash160"
-///     group    ::= policy ((" ∧ " policy)+ | (" ∨ " policy)+)
-///     thresh_body ::= policy ("," " " policy)+
-/// ```
-fn parse_math_policy<Pk: FromStrKey>(s: &str) -> Result<Policy<Pk>, Error> {
-    #[derive(Copy, Clone, PartialEq, Eq)]
-    enum Op {
-        And,
-        Or,
+struct MathParser<'a> {
```
trevarj, line 439: Can we get doc comments here and in the impl? Thanks!

src/policy/semantic.rs
febyeji: For this one, I added `consume_while(predicate)` helper to `MathParser`

src/policy/semantic.rs
```diff
@@ -364,123 +364,165 @@ impl<Pk: FromStrKey> str::FromStr for Policy<Pk> {
...
+}
 
-    let bytes = s.as_bytes();
+struct Frame<Pk: MiniscriptKey> {
+    subs: Vec<Arc<Policy<Pk>>>,
+    kind: Kind,
+}
+
+/// Parses the mathematical-notation form produced by `Display`.
+fn parse_math_policy<Pk: FromStrKey>(s: &str) -> Result<Policy<Pk>, Error> {
```
trevarj: Let's move this and the other `parse_*` functions into `MathParser`.
Then we can give it a good doc comment explaining the high level overview of how the parser works.

src/policy/semantic.rs
```diff
@@ -364,123 +364,165 @@ impl<Pk: FromStrKey> str::FromStr for Policy<Pk> {
...
+}
+
+#[derive(PartialEq, Eq)]
+enum Kind {
+    Group(Option<Op>),
+    Thresh,
+}
 
-    let bytes = s.as_bytes();
+struct Frame<Pk: MiniscriptKey> {
```
trevarj, line 503: Another doc comment here explaining what the use of Frame is used for.

src/policy/semantic.rs
```diff
@@ -276,11 +355,248 @@ impl<Pk: MiniscriptKey> fmt::Display for Policy<Pk> {
...
+fn parse_arg<T: core::str::FromStr>(arg: &str) -> Result<T, Error>
+where
+    T::Err: crate::blanket_traits::StaticDebugAndDisplay,
+{
+    arg.parse::<T>()
+        .map_err(ParseError::box_from_str)
+        .map_err(Error::Parse)
+}
+
+fn malformed_math() -> Error {
```
trevarj: I think it would be better to add a `context: &str` arg and display that so the user knows exactly where the policy is malformed. Need to update callers appropriately.

src/policy/semantic.rs
```diff
@@ -276,11 +355,248 @@ impl<Pk: MiniscriptKey> fmt::Display for Policy<Pk> {
...
+    s: &'a str,
+    iter: core::iter::Peekable<core::str::CharIndices<'a>>,
+}
+
+impl<'a> MathParser<'a> {
+    fn new(s: &'a str) -> Self { Self { s, iter: s.char_indices().peekable() } }
+
+    fn peek(&mut self) -> Option<char> { self.iter.peek().map(|&(_, c)| c) }
+
+    fn bump(&mut self) -> Option<char> { self.iter.next().map(|(_, c)| c) }
```
trevarj: nit: I'd rather have `next()` unless you think it's better to use `bump`. Some people use `chomp`...

```suggestion
    fn next(&mut self) -> Option<char> { self.iter.next().map(|(_, c)| c) }
```

src/policy/semantic.rs
```diff
@@ -276,11 +355,248 @@ impl<Pk: MiniscriptKey> fmt::Display for Policy<Pk> {
...
+
+    fn pos(&mut self) -> usize { self.iter.peek().map(|&(i, _)| i).unwrap_or(self.s.len()) }
+
+    fn skip_ws(&mut self) {
+        while matches!(self.peek(), Some(c) if c.is_whitespace()) {
+            self.bump();
+        }
+    }
+
+    fn try_consume(&mut self, want: char) -> bool {
```
trevarj:
```suggestion
    fn expect(&mut self, want: char) -> Result<(), Error> {
```

This way you can get rid of the `return Err(malformed_math());` that is scattered around and just use `parser.expect('(')?;`

trevarj (changes requested):
I think the parser is turning out pretty good, I like it.
Probably my last round of review since everything else seems ok to me.
Thanks!

src/policy/semantic.rs
febyeji: I've thought about it and would work if we put `parser.skip_ws()` before `continue` as well. For now, I kept as it is to have `loop -> skip_ws -> peek-or-break` format.

src/policy/semantic.rs
trevarj: You are correct, should be char!

src/policy/semantic.rs
febyeji: Right, it's much clearer. Thanks!

src/policy/semantic.rs
febyeji: Thanks, I updated this along with `expect`.

src/policy/semantic.rs
```diff
@@ -261,26 +266,355 @@ impl<Pk: MiniscriptKey> fmt::Display for Policy<Pk> {
...
+        Ok(Arc::try_unwrap(root).unwrap_or_else(|arc| (*arc).clone()))
+    }
+
+    fn parse_terminal<Pk: FromStrKey>(&mut self) -> Result<Policy<Pk>, Error> {
+        let name = self.consume_while(|c| c.is_ascii_alphanumeric());
+        if name.is_empty() {
+            return Err(malformed_math("expected terminal name"));
+        }
+
+        if !self.try_consume('(') {
```
trevarj: I would remove the `try_consume` function fully and then use:

```suggestion
        if let Err(e) = self.expect('(') {
```

febyeji: Oh, thank you. I missed this line.

trevarj (approved):
ACK 511e81b40ec9cf185af62fd85c8b89f6979dc76c

Don't forget to squash commits later. Probably last 5 into a single commit?

trevarj (approved):
ACK 9aeffc72ab5142874855dbb091a08710d21dac52

apoelstra:
Thanks for keeping up with this @febyeji! I apologize for getting frustrated earlier. This looks good. I have a few comments:

* Your grammar forbids "singletons" like `(pk(A))` -- this is good, I think, because otherwise we'd have stuff that we parse but then re-serialize differently.
* However, currently the parser gives an error of the form `Parse(Tree(UnknownName { name: "" }))` which is pretty ugly. Also fine, but if it's easy to detect this and provide a dedicated "no singletons" error we should do that.
* Similarly, if the user provides a malformed number, we eat the error from `expression::parse_num` and return a "malformed math" error. I don't want to hold this PR up on errors (which are bad throughout the library) but if you're feeling motivated you could improve this.
* Similarly you are super strict about spacing (exactly one space between operators). Elsewhere in the library we are super strict about having exactly *zero* spaces after `,`s etc. I think this is fine for now, I propose that in a followup we change this to just ignore whitespace entirely when parsing.
* Relatedly to parsing/reserializing, I'd like to see a fuzztest in this PR which tries to parse then re-serialize, checking that the result is identical (up to whitespace). It would also be good to have a fuzztest which generates an arbitrary `Policy` then tries to serialize then re-parse, but it's okay to skip this for now (I think all the mechanisms to "generate an arbitrary `Policy`" in this library currently allow you to stack overflow or otherwise do absurd things that won't re-parse, and fixing that is out of scope.)
* There are two commits here, one whose description is "update `Display`" and the other whose description is "update parsing". But the first one also does a bunch of parsing-related stuff. Please move code between the commits, or squash them, so that they make independent sense.

Done reviewing 9aeffc72ab5142874855dbb091a08710d21dac52. Getting closer.

febyeji:
I finished the update per @apoelstra 's review, thanks! 

> but if it's easy to detect this and provide a dedicated "no singletons" error we should do that.

However, I've got one question on the error handling. I wasn't sure whether this comment meant a clearer string message or a dedicated enum variant. I went with the variant approach, so I added `ParseError::Math(MathSyntaxError)` with `MathSyntaxError` as an enum of `SingletonGroup / MixedOperators / UnknownTerminal / Other(String)`. This changed `src/error.rs` which was outside the scope of previous reviews, so I'm happy to drop the variant if that was not the original intent.

apoelstra (approved):
ACK 8030c0bffa58e23e0a17ede30a5f662f16a7d10e; successfully ran local tests

PR #915, opened 2026-04-04
refactor: remove SortedMultiVec and use Terminal::SortedMulti

trevarj:
- Utilize the Miniscript parsing to handle sortedmulti as a Terminal.
- Deleted sortedmulti.rs (SortedMultiVec)
- Refactor Wsh to only wrap a Miniscript now that SortedMultiVec isn't used.
- Refactor ShInner to remove SortedMulti variant and only use the Ms variant for
  sortedmulti scripts

--- 

Now that sortedmulti_a is supported as a Terminal, I think it makes sense to
move sortedmulti over in the same way by following what multi does and applying
the sorting functions that were introduced in eba1aff.

Like sortedmulti_a, sortedmulti is sorted upon encoding and cannot be decoded
into from Script.


apoelstra:
(CI is failing because the integration tests don't compile)

Hmmmm, okay, this seems reasonable enough to me. Does Core allow `sortedmulti` in arbitrary positions or only as a top-level descriptor?

trevarj:
> (CI is failing because the integration tests don't compile)
> 
> Hmmmm, okay, this seems reasonable enough to me. Does Core allow `sortedmulti` in arbitrary positions or only as a top-level descriptor?

Working on fixing the integration tests now! Will ping you when it's ready, my bad.

bitcoind-tests/tests/test_desc.rs
```diff
@@ -222,29 +222,15 @@ pub fn test_desc_satisfy(
                 Descriptor::Pkh(pk) => find_sk_single_key(*pk.as_inner(), testdata),
                 Descriptor::Wpkh(pk) => find_sk_single_key(*pk.as_inner(), testdata),
                 Descriptor::Sh(sh) => match sh.as_inner() {
-                    miniscript::descriptor::ShInner::Wsh(wsh) => match wsh.as_inner() {
-                        miniscript::descriptor::WshInner::SortedMulti(ref smv) => {
-                            let ms = Miniscript::from_ast(smv.sorted_node()).unwrap();
```
trevarj, line 230: The key point here was to call `sorted_node()` and now we don't have something so convenient. See the `HACK` in `find_sks_ms`

src/descriptor/mod.rs
```diff
@@ -256,18 +254,31 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
...
             Descriptor::Wpkh(ref pk) => PkIter::from_key(pk.as_inner().clone()),
             Descriptor::Sh(ref sh) => match *sh.as_inner() {
-                ShInner::Wsh(ref wsh) => match wsh.as_inner() {
-                    WshInner::SortedMulti(ref sorted) => PkIter::from_sortedmulti(sorted.pks()),
-                    WshInner::Ms(ref ms) => PkIter::from_miniscript_segwit(ms),
-                },
+                ShInner::Wsh(ref wsh) => {
+                    let ms = wsh.as_inner();
+                    if let Terminal::SortedMulti(ref thresh) = ms.node {
+                        PkIter::from_sortedmulti(thresh.data())
```
trevarj: Cannot call `thresh.into_sorted_bip67()` due to lacking `ToPublicKey` on `Pk` here. ~It would clean up the integration test changes (`HACK`)~. I'm sure it would be useful for users as well.

src/descriptor/mod.rs
```diff
@@ -2698,4 +2721,26 @@ pk(03f28773c2d975288bc7d1d205c3748651b075fbc6610e58cddeeddf8f19405aa8))";
             );
         }
     }
+
+    #[test]
+    fn too_many_pubkeys_for_p2sh() {
```
trevarj, line 2892: FYI: migrated from deleted `sortedmulti.rs` file.

src/descriptor/segwitv0.rs
```diff
@@ -189,9 +154,10 @@ impl<Pk: MiniscriptKey + ToPublicKey> Wsh<Pk> {
     where
         S: Satisfier<Pk>,
     {
-        let mut witness = match self.inner {
-            WshInner::SortedMulti(ref smv) => smv.satisfy(satisfier)?,
-            WshInner::Ms(ref ms) => ms.satisfy_malleable(satisfier)?,
+        let mut witness = if let Terminal::SortedMulti(..) = self.ms.node {
```
trevarj: All the `if let Terminal::SortedMulti(..)`'s are kind of bothering me.

Does it make sense to add a function on Miniscript called `is_sortedmulti()` or something?

trevarj:
@apoelstra did a self-review with some questions. Let me know what you think, thanks!

bitcoind-tests/tests/test_desc.rs
apoelstra: Interesting. So, the issue is that we want to get all the keys in the descriptor, *in the order they'd be needed*, and we're using `pk_iter` for this. Since `pk_iter` can't sort the keys, we have to sort them after the fact.

This is a bit of a weird thing to desire -- does this test actually fail without this? It seems like we never use the ordering of `sks_reqd`. In fact, we *lose* the ordering because we iterate over it, produce signatures, then stick them into the `BTreeMap` `psbt.inputs[0].partial_sigs`.

Also, since #910 I think the logic in this example is actually wrong, or at least inconsistent -- it would sort `sortedmulti` but not `sortedmulti_a`. And your hack doesn't change this.

So can we just drop the sortedmulti logic here and treat sortedmulti like multi?

src/descriptor/mod.rs
apoelstra: Right -- I get the desire that `pk_iter` return the keys in the correct order. But as you note, without a `ToPublicKey` bound, we don't actually know the correct order (and the user might not either, e.g. if there are wildcards in the keys).

We also can't make this do the right thing for `ToPublicKey` keys because Rust lacks specialization. (There are **tons** of places in this crate where we suffer from this, and you'll find that if you keep trying to clean stuff up, you will run into missing language features basically every single time.)

I think we need to just live with this.

apoelstra:
816620a564e83fcaefbb249903c40c6c6e4ff81c needs rebase

src/descriptor/segwitv0.rs
apoelstra: Probably yes, but I'm curious what the heck is going on here. Why are we calling the non-malleable satisfier for sortedmultis? (What I suspect is: the old `SortedMultiVec` type had a `satisfy` method but no `satisfy_malleable` method, presumably because for multisigs, malleability is irrelevant. We needed to branch because `SortedMultiVec` is a distinct type, and then the branch wound up looking sus with one side saying `satisfy` and the other side saying `satisfy_malleable`, just because of this API gap.)

SO I think that in this PR, we just always want to call `self.ms.satisfy_malleable`, and not branch at all.

apoelstra:
f9d0bddd0295a7425f65c04546edd0da577380fd needs rebase

src/descriptor/mod.rs
trevarj: > There are tons of places in this crate where we suffer from this, and you'll find that if you keep trying to clean stuff up, you will run into missing language features basically every single time.

@apoelstra don't worry, in 2035 when specialization stabilizes we can get to it!

src/descriptor/segwitv0.rs
trevarj: I based it off of this, which is from the deleted `SortedMultiVec`.

```diff
-    /// Attempt to produce a satisfying witness for the
-    /// witness script represented by the parse tree
-    pub fn satisfy<S>(&self, satisfier: S) -> Result<Vec<Vec<u8>>, Error>
-    where
-        Pk: ToPublicKey,
-        S: Satisfier<Pk>,
-    {
-        let ms = Miniscript::from_ast(self.sorted_node()).expect("Multi node typecheck");
-        ms.satisfy(satisfier)
```

> We needed to branch because SortedMultiVec is a distinct type, and then the branch wound up looking sus with one side saying satisfy and the other side saying satisfy_malleable, just because of this API gap.

I don't think I could have guessed that lol...trusting you and changing to only call `satisfy_malleable`

src/miniscript/iter.rs
```diff
@@ -29,7 +29,13 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
...
     pub fn branches(&self) -> Vec<&Miniscript<Pk, Ctx>> {
         match self.node {
-            Terminal::PkK(_) | Terminal::PkH(_) | Terminal::RawPkH(_) | Terminal::Multi(_) => {
+            Terminal::PkK(_)
+            | Terminal::PkH(_)
+            | Terminal::RawPkH(_)
+            | Terminal::Multi(_)
+            | Terminal::SortedMulti(_)
+            | Terminal::MultiA(_)
+            | Terminal::SortedMultiA(_) => {
```
trevarj, line 38: Saw SortedMultiA was also missing from this.

src/miniscript/iter.rs
```diff
@@ -96,6 +102,7 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx> {
         match (&self.node, n) {
             (Terminal::PkK(key), 0) | (Terminal::PkH(key), 0) => Some(key.clone()),
             (Terminal::Multi(thresh), _) => thresh.data().get(n).cloned(),
+            (Terminal::SortedMulti(thresh), _) => thresh.data().get(n).cloned(),
```
trevarj, line 105: I missed this on the first iteration since type-driven development failed me with the `_ => ` fallback.

@apoelstra It resolves the `HACK` that was mentioned in [this comment](https://github.com/rust-bitcoin/rust-miniscript/pull/915/changes#r3058554938).

src/miniscript/mod.rs
```diff
@@ -648,7 +660,9 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> ForEachKey<Pk> for Miniscript<Pk, Ct
                 }
                 // These branches cannot be combined since technically the two `thresh`es
                 // have different types (have different maximum values).
-                Terminal::Multi(ref thresh) if !thresh.iter().all(&mut pred) => {
+                Terminal::Multi(ref thresh) | Terminal::SortedMulti(ref thresh)
+                    if !thresh.iter().all(&mut pred) =>
+                {
```
trevarj, line 665: After rebasing one of the tests from 1062144 caught that I missed this! Thanks @febyeji

febyeji: Great!

src/miniscript/context.rs
febyeji: `Terminal::SortedMulti(ref thresh)` falls through `_ => Ok(())`, skipping the compressed-key check.

febyeji: I think `Terminal::SortedMulti(...)` should lead to `Err(ScriptContextError::TaprootMultiDisabled)` as well.

febyeji: `Terminal::SortedMulti(ref thresh)` too?

febyeji: `Terminal::SortedMulti(ref thresh)` misses the `n   <= 3` non-standard check.

src/descriptor/sh.rs
```diff
@@ -129,7 +123,7 @@ impl<Pk: MiniscriptKey> Sh<Pk> {
     pub fn new_sortedmulti(thresh: Threshold<Pk, MAX_PUBKEYS_PER_MULTISIG>) -> Result<Self, Error> {
         // The context checks will be carried out inside new function for
```
febyeji: The comment looks stale now because `SortedMultiVec` is no longer used. (Same at L152)

febyeji (changes requested):
I left a couple of comments. One on missing `SortedMulti` arms in `context.rs` and one stale comment nit.

trevarj:
> I left a couple of comments. One on missing `SortedMulti` arms in `context.rs` and one stale comment nit.

Thanks again, @febyeji ! Can't believe I missed so many match arms 😞

src/descriptor/mod.rs
apoelstra: In a65deea53871b032a8490e238f4b5f15679170cb:

Here, and below, you can just delete the `from_sortedmulti` method and get rid of this branch. It is now covered by the `from_miniscript_*` methods.

src/descriptor/mod.rs
```diff
@@ -313,18 +324,29 @@ impl<Pk: MiniscriptKey> Descriptor<Pk> {
             Descriptor::Pkh(ref _pkh) => DescriptorType::Pkh,
             Descriptor::Wpkh(ref _wpkh) => DescriptorType::Wpkh,
             Descriptor::Sh(ref sh) => match sh.as_inner() {
-                ShInner::Wsh(ref wsh) => match wsh.as_inner() {
-                    WshInner::SortedMulti(ref _smv) => DescriptorType::ShWshSortedMulti,
-                    WshInner::Ms(ref _ms) => DescriptorType::ShWsh,
-                },
+                ShInner::Wsh(ref wsh) => {
+                    if let Terminal::SortedMulti(..) = wsh.as_inner().node {
+                        DescriptorType::ShWshSortedMulti
```
apoelstra, line 309: In a65deea53871b032a8490e238f4b5f15679170cb:

As a followup let's maybe delete this from `DescriptorType` since it's no longer a distinct kind of descriptor. But let's not do it here since it'd be an API-breaking change and I'm not certain that we want to do it.

apoelstra:
Done reviewing a65deea53871b032a8490e238f4b5f15679170cb. All good except the iterpk thing. Frustrating how many `match`es there are in this crate that have `_` fields. In general it seems super hard to find them all.

apoelstra:
Nice :) great to see such a red diff.

apoelstra (approved):
ACK 01d25c17457d5025abc7ea6f82def05eceefb9b6; successfully ran local tests

src/miniscript/satisfy.rs
```diff
@@ -1478,7 +1478,7 @@ impl<Pk: MiniscriptKey + ToPublicKey> Satisfaction<Placeholder<Pk>> {
                     thresh_fn(thresh, stfr, root_has_sig, leaf_hash, min_fn)
                 }
             }
-            Terminal::Multi(ref thresh) => {
+            Terminal::Multi(ref thresh) | Terminal::SortedMulti(ref thresh) => {
```
apoelstra, line 1481: Oops, I missed this -- in this case we need to change the logic. In `SortedMultiA` we have special logic to sort keys, but in `SortedMulti` we don't.

We should also add a test case for this.

apoelstra: I will fix this in #895 -- though if you have an earlier fix I'd appreciate since that PR is likely to be in limbo for at least several more weeks.

trevarj: Yes, big oversight on my part...I will fix that and a test. I guess we test it in the bitcoind tests but not using our own satisfier.

Issue #902, opened 2026-02-25
`compile_tr` and `compile_tr_private_experimental` fail for `pk(A)`


simongltr:
Both `compile_tr` and `compile_tr_private_experimental` fail for `pk(A)`. Expected result in both cases is `tr(A)`, `extract_key` correctly promotes `A` to the taproot internal key, but the remaining `Policy::Unsatisfiable` is not handled the same as `Policy::Trivial`, causing it to fall through to leaf compilation with nothing to compile.

`compile_tr_private_experimental` also panics for `and(pk(A),pk(B))` (`compile_tr` is fine there).

I discovered this while trying to push for a privacy first taproot compiler in bdk-cli: https://github.com/bitcoindevkit/bdk-cli/issues/244

simongltr:
I re-coded the test and I get the same as you for `pk(A)` now.

Did the same for `and(pk(A), pk(B))` here we need to provide the unspendable key or both fail with `ERROR: Taproot compilation had no internal key available`. I got this:

```
compile_tr(and(pk(A),pk(B))): OK: tr(UNSPENDABLE,and_v(v:pk(A),pk(B)))#787dpsca

thread 'main' (28780209) panicked at /Users/simon/src/rust-miniscript/src/policy/concrete.rs:505:25:
assertion failed: tapleaf_prob_vec.remove(&(Reverse(*prev_prob), policy.clone()))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
compile_tr_private(and(pk(A),pk(B))): PANIC!
```

The code (your code with adaptations):
```rust
use std::fmt;
use std::str::FromStr;

use miniscript::policy::Concrete;

fn run<T: fmt::Display, E: fmt::Display>(
    label: &str,
    f: impl FnOnce() -> Result<T, E> + std::panic::UnwindSafe,
) {
    print!("{label}: ");
    match std::panic::catch_unwind(f) {
        Ok(Ok(v)) => println!("OK: {v}"),
        Ok(Err(e)) => println!("ERR: {e}"),
        Err(_) => println!("PANIC!"),
    }
}

fn main() {
    let and_ab = || Concrete::<String>::from_str("and(pk(A),pk(B))").unwrap();

    run("compile_tr(and(pk(A),pk(B)))", || {
        and_ab().compile_tr(Some("UNSPENDABLE".to_string()))
    });
    run("compile_tr_private(and(pk(A),pk(B)))", || {
[...]

PR #916, opened 2026-04-05
tests: add regression tests for `compile_tr_private_experimental` panic

va-an:
Relates #902

Add `#[should_panic]` regression tests for `compile_tr_private_experimental` - it panics on `pk(A)` and `and(pk(A),pk(B))`.

Expected descriptor values verified against @febyeji's fix and match `compile_tr` output from @simongltr's [reproduction](https://github.com/rust-bitcoin/rust-miniscript/issues/902#issuecomment-4093596333). 

Don't forget to drop `#[should_panic]` when the fix lands!


va-an:
@febyeji added test in #917 fix for #902, so this PR might be closed.

PR #917, opened 2026-04-06
Fix panic in compile_tr_private_experimental for non-expandable policies

### LINKED ISSUE #902: `compile_tr` and `compile_tr_private_experimental` fail for `pk(A)`

febyeji:
Closes #902.

- Keep `pol_prob_map` in sync with `tapleaf_prob_vec` by removing entries from both data structures together. 
- Previously, removing a node from `tapleaf_prob_vec` without removing it from `pol_prob_map` caused a panic when `expand_fn` returned the same policy (e.g. for `pk(A)` or `and(pk(A),pk(B))`).
- Adds regression tests for both cases.

### DIFF
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -614,10 +614,12 @@ impl<Pk: MiniscriptKey> Policy<Pk> {
 
             // Remove current node
             assert!(tapleaf_prob_vec.remove(&(prob, curr_policy.clone())));
+            pol_prob_map.remove(&curr_policy);
 
             // OPTIMIZATION - Move marked nodes into final vector
             for (p, pol) in to_del {
                 assert!(tapleaf_prob_vec.remove(&(Reverse(OrdF64(p)), pol.clone())));
+                pol_prob_map.remove(&pol);
                 ret.push((p, pol.clone()));
             }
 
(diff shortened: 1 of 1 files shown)


src/policy/concrete.rs
```diff
@@ -1276,6 +1278,21 @@ mod compiler_tests {
...
     }
+
+    #[test]
+    fn compile_tr_private_non_expandable() {
+        let unspendable = "UNSPEND".to_string();
+
+        // Single key: extract_key promotes it, leaving Unsatisfiable
+        let pol = Policy::<String>::from_str("pk(A)").unwrap();
+        pol.compile_tr_private_experimental(Some(unspendable.clone()))
+            .expect("pk(A) should compile without panic");
```
va-an: I suggest adding a comparison with the expected result instead of checking for a panic as I did in #916.

febyeji: Thanks for letting me know!

apoelstra (approved):
ACK f728d78ceafb3abc640fccc777fecf42234b9734; successfully ran local tests

PR #920, opened 2026-04-13
refactor: remove *SortedMulti variants from DescriptorType

### LINKED ISSUE #919: Delete DescriptorType::ShWshSortedMulti
In a65deea53871b032a8490e238f4b5f15679170cb:

As a followup let's maybe delete this from `DescriptorType` since it's no longer a distinct kind of descriptor. But let's not do it here since it'd be an API-breaking change and I'm not certain that we want to do it.

_Originally posted by @apoelstra in https://github.com/rust-bitcoin/rust-miniscript/pull/915#discussion_r3065501966_

trevarj:
After 01d25c1, these variants no longer serve a purpose since descriptors
containing sortedmulti are not handled as a distinct type, but as an underlying
Miniscript.

Closes #919

### DIFF
--- a/bitcoind-tests/tests/test_desc.rs
+++ b/bitcoind-tests/tests/test_desc.rs
@@ -501,10 +501,7 @@ fn test_plan_satisfy(
     let sighash_type = sighash::EcdsaSighashType::All;
     let desc_type = derived_desc.desc_type();
     let sighash_msg = match desc_type {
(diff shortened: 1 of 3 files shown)


apoelstra (approved):
ACK ce18ba2e989f207677cb0c3eafc2c73c2ab10965; successfully ran local tests

PR #921, opened 2026-04-13
fix: add sorting to satisfy_helper for sortedmulti

trevarj:
This logic was done in sortedmulti_a but accidentally overlooked when
refactoring sortedmulti.

Wrote a test that proves two differently ordered sortedmulti descriptors produce
the same witness.

Found by Andrew in https://github.com/rust-bitcoin/rust-miniscript/pull/915#discussion_r3073591603


apoelstra (approved):
ACK 3f20a6903799e1cc37d6ac6fd982d637657296f8; successfully ran local tests

PR #932, opened 2026-04-28
wallet_policy: fix set_key_info silently dropping keys for repeated placeholders

sanket1729:
`set_key_info()` counted key-expression occurrences instead of unique placeholder indices. A template reusing `@0` with disjoint paths (e.g. `@0/**,@0/<2;3>/*`) would accept extra keys that are silently dropped during descriptor translation. Fix by counting unique placeholder indices via `BTreeSet`.

I'm not super familiar with wallet_policy/BIP-388, but this looks like a bug to me. Credit to jmeccom for reporting this.

### DIFF
--- a/src/descriptor/wallet_policy/mod.rs
+++ b/src/descriptor/wallet_policy/mod.rs
@@ -379,4 +379,27 @@ mod tests {
         template_only.set_key_info(&keys).unwrap();
         assert!(template_only.clone().into_descriptor().is_ok());
     }
+
+    // Regression test for a bug where set_key_info() counted key-expression
+    // occurrences instead of unique placeholder indices. A template reusing @0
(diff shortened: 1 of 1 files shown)


trevarj (approved):
@sanket1729 thanks for fixing. I overlooked this when doing the implementation - it is in the BIP:

> A KI (key index) @i for some number i represents the i-th key in the vector of key information items (which must be of size at least i + 1, or the wallet policy is invalid).

ACK 1f655a91a44561914e19028d39a91b4dc950dc7c

apoelstra:
Oo, good catch.

apoelstra (approved):
ACK 1f655a91a44561914e19028d39a91b4dc950dc7c; successfully ran local tests

PR #933, opened 2026-04-30
doc: Add some high level notes about rust-miniscript design

tcharding:
Add a file to `doc/` giving some high level notes about the design of the repo. This is Sanket's work picked up from #603.


tcharding:
The only changes to the original PR are the review suggestions I made back then. cc @sanket1729 just FYI.

trevarj:
Nice. Would like to see one bullet describing the `Miniscript` type and another bullet describing the `Descriptor` type. Maybe another bullet on how they relate?

tcharding:
Fair. Will require some thought if I'm to write those.

apoelstra:
I don't think we want to foreground the two types of policies. We should definitely cover `Descriptor` and `Miniscript` and their relationship. We should also talk about Taproot and maybe the other kinds of descriptors (and why e.g. `Descriptor::address` returns a `Result`).

PR #935, opened 2026-04-30
descriptor: support hash terminals in WalletPolicy translator

bg002h:
`WalletPolicyTranslator` rejects hash terminals (`sha256` / `hash256` / `ripemd160` / `hash160`) because both directions use `translate_hash_fail!`. BIP 388 places no restriction on them; HTLC patterns are a primary use case. This PR adds explicit implementations bridging `KeyExpression`'s hex `String` storage and `DescriptorPublicKey`'s binary `Hash` types, with a new `WalletPolicyError::TranslatorInvalidHashHex` variant for the parse direction.

Found while building [md-codec](https://github.com/bg002h/descriptor-mnemonic), a BIP 388 wallet-policy encoder.

Authored by Claude Opus 4.7 under my direction.

### DIFF
--- a/src/descriptor/wallet_policy/mod.rs
+++ b/src/descriptor/wallet_policy/mod.rs
@@ -3,9 +3,11 @@
 use core::fmt::{self, Display};
 use core::str::FromStr;
 
+use bitcoin::hashes::{hash160, ripemd160, sha256};
+
 use super::key::XKeyParseError;
 use super::{DerivPaths, DescriptorKeyParseError, Wildcard};
-use crate::{BTreeSet, Descriptor, DescriptorPublicKey, String, Translator, Vec};
+use crate::{BTreeSet, Descriptor, DescriptorPublicKey, String, ToString, Translator, Vec};
 
 mod key_expression;
 
@@ -61,7 +63,30 @@ impl Translator<KeyExpression> for WalletPolicyTranslator {
             .ok_or(WalletPolicyError::KeyInfoInvalidKeyIndex(idx))
     }
 
-    translate_hash_fail!(KeyExpression, DescriptorPublicKey, Self::Error);
+    // Hash terminals: KeyExpression stores hashes as hex `String` (for
+    // template round-tripping), DescriptorPublicKey uses the concrete
+    // `bitcoin::hashes::*::Hash` types. Parse the hex string into the
+    // binary form during materialization.
+
+    fn sha256(&mut self, s: &String) -> Result<sha256::Hash, Self::Error> {
+        s.parse::<sha256::Hash>()
+            .map_err(|_| WalletPolicyError::TranslatorInvalidHashHex("sha256", s.clone()))
+    }
+
+    fn hash256(&mut self, s: &String) -> Result<crate::hash256::Hash, Self::Error> {
+        s.parse::<crate::hash256::Hash>()
+            .map_err(|_| WalletPolicyError::TranslatorInvalidHashHex("hash256", s.clone()))
+    }
+
+    fn ripemd160(&mut self, s: &String) -> Result<ripemd160::Hash, Self::Error> {
+        s.parse::<ripemd160::Hash>()
+            .map_err(|_| WalletPolicyError::TranslatorInvalidHashHex("ripemd160", s.clone()))
+    }
+
+    fn hash160(&mut self, s: &String) -> Result<hash160::Hash, Self::Error> {
+        s.parse::<hash160::Hash>()
+            .map_err(|_| WalletPolicyError::TranslatorInvalidHashHex("hash160", s.clone()))
+    }
 }
 
 impl Translator<DescriptorPublicKey> for WalletPolicyTranslator {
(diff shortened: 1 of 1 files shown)


trevarj (commented):
Another oversight by me in the BIP388 implementation! 

This does make sense and should not error when hashes are in the descriptor template.

I wonder if the tests can be made a little more concise. They seem overly verbose to test such a simple thing...perhaps I'm just getting replaced by Claude...

trevarj:
Oh yeah...it would be nice if this is committed by you instead of Claude. Just my opinion though, I don't know how Andrew feels about that.

bg002h:
I can apologize via Claude too, and I’m sure it would do it better!

That said, Claude was totally comfortable posting as if it were me…I added the human tag at the bottom by telling Claude it can’t pretend to be human. 

That said, as best I can tell this is a real enough issue to be passed along in a “don’t just complain, do something” fashion.

apoelstra:
If Claude writes stuff I strongly prefer it identify as itself rather than as a human. (Identifying as a human "helped by Claude" or "co-authored by Claude" is also fine.)

But can you reduce the amount of text in your commit message and in your PR description? It may be sufficient to just tell Claude to be concise and to assume that the reader is an expert on the codebase.

CI failures are real -- and annoyingly you will have to wait for me to show up and click "Approve CI run" to get that feedback -- can you fix the formatting and the no-std gating?

bg002h:
I told claude AP had some feedback and it literally ingested your words as commands and is busily applying your feedback. 

I’ll let Claude give a whack at this revision and if it continues to spit out text walls with kernels of truth disguised therein, I’ll bow out and just let Claude prepare locally applied patches in Claude-crappy fashion to make my proof of concept project function. 

I’ve never programmed in rust and I’m at best a high school level pascal programmer, and tokens are cheaper than time in my current world…so I apologize for not recognizing it’s crappy output on account of my vague understanding!

bg002h:
Thanks both. Addressed:

- **CI fmt + no-std:** ran `cargo fmt`; switched the module to `use crate::prelude::*` so `ToString` is in scope under `--no-default-features` (matches the pattern in `descriptor/{key,bare,checksum}.rs`). `cargo build --no-default-features` clean locally.
- **Test verbosity:** collapsed the 6 hash-terminal tests into 3 — a parameterized round-trip across all 4 hash kinds, one mixed-pair guard against cross-type interference, and a compact macro-driven invalid-hex variant pin. Net −189 lines in the diff.
- **Commit message + PR body:** trimmed both; assumed expert reader.
- **Attribution:** Claude is named in the `Co-Authored-By` trailer and the PR body. Happy to adjust the wording further if you'd prefer a different form.

Ready for another CI run when you have a moment.

— Brian (with Claude Opus 4.7)

src/descriptor/wallet_policy/mod.rs
```diff
@@ -3,9 +3,12 @@
 use core::fmt::{self, Display};
 use core::str::FromStr;
 
+use bitcoin::hashes::{hash160, ripemd160, sha256};
+
 use super::key::XKeyParseError;
 use super::{DerivPaths, DescriptorKeyParseError, Wildcard};
-use crate::{BTreeSet, Descriptor, DescriptorPublicKey, String, Translator, Vec};
+use crate::prelude::*;
```
trevarj: This import is unneeded. Compiles without it for me.

apoelstra:
8c101fbd7c94cc42c1bef5611d34426a86d50c84 needs rebase

src/descriptor/wallet_policy/mod.rs
```diff
@@ -366,6 +410,67 @@ mod tests {
...
+    #[test]
+    fn translator_invalid_hash_hex_errors() {
+        let mut t = WalletPolicyTranslator { key_info: Vec::new() };
+        let bad = String::from("not_hex");
+
+        macro_rules! assert_bad {
+            ($method:ident, $kind:literal) => {{
+                let err =
+                    <WalletPolicyTranslator as Translator<KeyExpression>>::$method(&mut t, &bad)
+                        .unwrap_err();
```
trevarj: nit:
```suggestion
                let err = Translator::<KeyExpression>::$method(&mut t, &bad).unwrap_err();
```

trevarj (commented):
> I’ve never programmed in rust and I’m at best a high school level pascal programmer, and tokens are cheaper than time in my current world…so I apologize for not recognizing it’s crappy output on account of my vague understanding!

No need to get existential, i'm already worried about my future in programming 😅 

Looks pretty good now.

tcharding:
> Authored by Claude Opus 4.7 under my direction.

Something like this might actually be useful to enforce. If any PR comes in and the author has no clue what they are doing and states as such up front it is way easier for us to review because we can decide, before looking, if it would be quicker/easier to do the work ourself than go through all the reviews. Especially if its a bug or feature that seems legit, we don't want to just ban uneducated devs wielding LLMs we just want to get the best out of everyones time, right?

trevarj:
> > Authored by Claude Opus 4.7 under my direction.
> 
> Something like this might actually be useful to enforce. If any PR comes in and the author has no clue what they are doing and states as such up front it is way easier for us to review because we can decide, before looking, if it would be quicker/easier to do the work ourself than go through all the reviews. Especially if its a bug or feature that seems legit, we don't want to just ban uneducated devs wielding LLMs we just want to get the best out of everyones time, right?

Indeed...probably would make sense to enforce having Claude/LLM write an issue first and then it can be triaged appropriately.

apoelstra:
Please make Clippy happy.

bg002h:
> > > Authored by Claude Opus 4.7 under my direction.
> > 
> > 
> > Something like this might actually be useful to enforce. If any PR comes in and the author has no clue what they are doing and states as such up front it is way easier for us to review because we can decide, before looking, if it would be quicker/easier to do the work ourself than go through all the reviews. Especially if its a bug or feature that seems legit, we don't want to just ban uneducated devs wielding LLMs we just want to get the best out of everyones time, right?
> 
> Indeed...probably would make sense to enforce having Claude/LLM write an issue first and then it can be triaged appropriately.

+1 on the triage idea. Maybe with a mandatory 20 word summary (AI gets wordy when user requests max effort). A repo specific set of initial commands for model would help (simple commands like “Do not file an issue without first searching PRs and Issues” and “do not file a pull request unless specifically requested or other condition.”)

also, as best I can tell “clippy” is happy status post last push…Claude is telling me to say “clippy clean locally now (default + all-features + no-default-features); can you approve a CI re-run?”

apoelstra (approved):
ACK 2092faaffe1e5e6b959af533b2942bf0fe9d1470; successfully ran local tests

apoelstra (approved):
ACK 2092faaffe1e5e6b959af533b2942bf0fe9d1470; successfully ran local tests

Issue #934, opened 2026-04-30
WalletPolicy::set_key_info doesn't enforce AST-order correspondence with template


bg002h:
Found this while implementing a downstream consumer of `WalletPolicy` in [md-codec](https://github.com/bg002h/descriptor-mnemonic). Filing as an issue rather than a PR because the fix shape is a design call; happy to PR your preference.

Context: I have a related accessor PR (#936) that exposes `template()` and `key_info()` on `WalletPolicy`. Both that PR and this `set_key_info` concern surfaced together while consuming the WalletPolicy API for [md-codec](https://github.com/bg002h/descriptor-mnemonic).

## The contract

After consuming `template()` and `key_info()` accessors (proposed in #936), downstream code relies on an implicit contract:

> `key_info()[i]` is the concrete key at AST position `i`, matching the order yielded by `template().iter_pk()`.

This holds by construction for `from_descriptor[_unchecked]` (which sets `key_info = descriptor.iter_pk().collect()`). It does **not** hold after `set_key_info(&[...])`, which only length-checks:

```rust
pub fn set_key_info(&mut self, keys: &[DescriptorPublicKey]) -> Result<(), WalletPolicyError> {
    if keys.len() != self.template.iter_pk().count() {
        return Err(WalletPolicyError::WalletPolicyInvalidKeyInfo);
    }
    self.key_info = keys.to_vec();
    Ok(())
}
```

The caller can pass keys in any order; the API accepts. Downstream `into_descriptor()` then materializes a descriptor with keys at the wrong AST positions.

## How `key_info` gets populated

| Constructor / mutator | `key_info` order | Contract holds? |
|---|---|---|
| `from_descriptor` / `from_descriptor_unchecked` | AST order | ✅ by construction |
| `from_str(template_string)` | empty | ✅ vacuous |
| `from_str(full_descriptor_string)` | falls through to `from_descriptor` | ✅ |
| `set_key_info(&keys)` post-construction | whatever caller passed | ⚠️ caller's responsibility, no validation |

## Why this matters in practice

md-codec's `to_bytecode` calls `set_key_info(&dummy_keys(placeholder_count))` to materialize a descriptor for tree encoding. The dummies are placeholder-indexed, not AST-indexed. For non-multipath-shared templates the two coincide (AST position == placeholder index), so it works by happy accident.

[...]

PR #936, opened 2026-04-30
descriptor: expose WalletPolicy template + key_info accessors

bg002h:
Read accessors for `WalletPolicy`'s private `template` and `key_info` fields. Motivating case: multipath-shared `@N` placeholders (e.g. `sh(multi(1,@0/**,@0/<2;3>/*))`) where `into_descriptor()` erases placeholder identity but the template's `KeyExpression::index` preserves it. Re-exports `KeyExpression` and `KeyIndex` so the return type is namable.

Found while building [md-codec](https://github.com/bg002h/descriptor-mnemonic), a BIP 388 wallet-policy encoder. Cross-link: #934 (about `set_key_info` AST-order correspondence; not a blocker for this PR).

Authored by Claude Opus 4.7 under my direction.

### DIFF
--- a/src/descriptor/wallet_policy/mod.rs
+++ b/src/descriptor/wallet_policy/mod.rs
@@ -9,7 +9,7 @@ use crate::{BTreeSet, Descriptor, DescriptorPublicKey, String, Translator, Vec};
 
 mod key_expression;
 
-use key_expression::{KeyExpression, KeyIndex};
+pub use key_expression::{KeyExpression, KeyIndex};
 
 /// A wallet policy as described in BIP-388
 ///
@@ -107,6 +107,15 @@ impl WalletPolicy {
     ) -> Result<WalletPolicy, WalletPolicyError> {
         WalletPolicy::from_descriptor_unchecked(descriptor).and_then(WalletPolicy::validate)
     }
+
+    /// Returns a reference to the template descriptor.
+    ///
+    /// Each [`KeyExpression`] yielded by `iter_pk()` carries a public
+    /// `index` field giving the BIP-388 placeholder index.
+    pub fn template(&self) -> &Descriptor<KeyExpression> { &self.template }
+
+    /// Returns the concrete keys, indexed in `template().iter_pk()` order.
+    pub fn key_info(&self) -> &[DescriptorPublicKey] { &self.key_info }
 
     /// Convert a `WalletPolicy` into a `Descriptor<DescriptorPublicKey>` using
     /// the underlying template and key information.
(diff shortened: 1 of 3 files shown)


src/descriptor/wallet_policy/mod.rs
```diff
@@ -108,6 +108,26 @@ impl WalletPolicy {
...
+    /// Returns a reference to the template descriptor.
+    ///
+    /// The template uses [`KeyExpression`] in place of concrete keys.
+    /// Each `KeyExpression` carries a public `index` field identifying
+    /// which entry in the key-information vector is referenced at that
+    /// AST position. Callers needing the
+    /// `(AST position) -> (placeholder index)` mapping can iterate
+    /// `policy.template().iter_pk()` and read `ke.index.0` for each
+    /// yielded `KeyExpression`. For the resolved descriptor with
+    /// concrete keys, see [`Self::into_descriptor`] (consumes `self`).
```
trevarj: Lol, why did the LLM feel the need to say this here?

```suggestion
    /// concrete keys, see [`Self::into_descriptor`].
```

trevarj (approved):
Seems reasonable.

ACK a95a61a124919e64f55f0eed59c9837bbd2561dd

src/descriptor/wallet_policy/mod.rs
bg002h: > Lol, why did the LLM feel the need to say this here?

Maybe a prohibition against self consumption :)

bg002h:
https://github.com/rust-bitcoin/rust-miniscript/pull/932

I’m not sure if there is a use case outside of shared @N index, which 932 is addressing (https://github.com/rust-bitcoin/rust-miniscript/pull/932)…

apoelstra:
Can you dramatically reduce the amount of text in the commit description, PR description, and doccomments for the new methods?

Concept ACK having these accessors.

bg002h:
Thanks. Trimmed:

- **Doccomments** on `template()` and `key_info()` cut to a sentence each, dropping the formula-style guidance for callers.
- **Commit message** and **PR description** trimmed; assumed expert reader.
- Tests left as-is — three of them, each pinning a specific invariant. Happy to collapse if you'd prefer.

`cargo test --lib` 155 / `--all-features` 173 / `clippy -D warnings` / `fmt --check` / `build --no-default-features` all clean locally.

— Brian (with Claude Opus 4.7)

apoelstra:
83e8d36565e73832f38c81638a1b5963d753213b needs rebase

apoelstra:
Heh, thanks for the short and clear motivation. But as you say, #932 eliminates the proposed usecase.

I think we should either

* close this; or
* update the motivation to "these might be useful"

I'm happy to go in either direction. Though I guess we should lean toward closing, to avoid constraining our API.

bg002h:
(Brian says): Theoretically these accessors could be useful for template inspection without specified keys, but my use case doesn't actually require that — #932 closed the concrete gap I was worried about. Perhaps it's wise to not expose these methods until a genuine use case arises from a real programmer, as their needs may hint at a better form or way of accomplishing the same goal than what I've sketched. Confirm closing as the wisest move at this stage of the game?

trevarj:
+1 for closing if no immediate use-case

PR #937, opened 2026-04-30
Re-name types in `policy`

tcharding:
As described here: https://github.com/rust-bitcoin/rust-miniscript/issues/885#issuecomment-3665732418

Big diff but this is mostly re-names and code moves. I split it up in steps. First might be best to look at the final state of `policy`, especially I'm not sure if we want the re-exports left in there considering deprecation doesn't work.


apoelstra:
concept ACK these changes. But can you split this into two or three PRs:

* one that renames `semantic::Policy` to `semantic::Semantic` (and reexports it from the crate root and removes all the stutters, if you want)
* one that moves the `Lift` trait to the `semantic` module (can be combined with the above)
* one that renames `policy::Concrete` to `policy::Policy` (and moves the code into `policy/mod.rs`, if you want) and deals with the fallout

Even if the deprecations don't work we should retain the aliases. Then people who are surprised by the breakage in a future version will at least have the ability to fix things by doing the renames *then* doing the upgrade.

apoelstra:
Oh -- to clarify the above. in this PR you move `Lift` into the `policy` module. It should be in the `semantic` module instead.

tcharding:
No worries. One question came to me. Isn't this PR going to wreak havoc on that big local branch you have? Is it worth doing now or should we just wait a bit since this is easy enough to do at any time?

apoelstra:
Heh, yeah, it will cause me some chaos, but I suspect I could jury-rig `jj fix` to deal with it for me. Or I can just fix the specific places where there are problems. For the most part I didn't make big changes to the policy module.

PR #938, opened 2026-04-30
Implement `MiniscriptKey` for each variant of `DescriptorPublicKey`

Velnbur:
This adds ability to specify `SinglePub`, `DescriptorXKey` and `DescriptorMultiXKey` directly in `Descriptor` without using `enum DescriptorPublicKey`.

In one of the projects we're using only `DescriptorMultiXKey` with fixed structure, so matching `DescriptorPublicKey` is kind of a pain. This PR fixes that

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: CC0-1.0
 
 use core::convert::TryInto;
-use core::fmt;
 use core::str::FromStr;
+use core::{fmt, hash};
 #[cfg(feature = "std")]
 use std::error;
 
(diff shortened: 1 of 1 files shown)


apoelstra:
Neat! Yeah, good idea. I think for `SinglePub` you can also implement `ToPublicKey`.

apoelstra (approved):
ACK eeb49eb2de3a6e9aef89be89fb33c724cfed815e; successfully ran local tests; can do `ToPublicKey` in a followup if you like

Velnbur:
> can do ToPublicKey in a followup if you like

I would prefer doing it in the next PR. I can take care of that if there are no objections, but if you want, I can do it here with the next commit.

apoelstra:
@Velnbur I'd appreciate a new PR! If not, I will probably do it "eventually".

PR #930, opened 2026-04-23
Followup to #895: rename `dissat_sat` function and use struct rather than tuple for return values

apoelstra:
Makes naming more consistent and reduces confusion caused by having a tuple with two entries of the same type.

PR #941, opened 2026-05-05
miniscript: change `tap_leaf_internal` to return an `Option` in the satisfier

apoelstra:
Now that we have #895 and #930 we have room to change the API for helpers to the satisfaction functions.

Change the method `tap_leaf_internal`, which currently computes the "tapleaf hash" for an arbitrary Miniscript, even non-Taproot ones, to detect whether this is a Taproot script and only compute the tapleaf hash for Taproot scripts.

Once we are returning an `Option` from `tap_leaf_internal`, we can change a bunch of satisfaction code that currently takes a `Ctx` paramater and introspects it to determine whether we are in Taproot mode or not, and have it simply use the `Option` to decide.

Aside from speeding up and simplifying the code, this removes several instances of the `Ctx` parameter, which we want to eventually eliminate.

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -1222,7 +1222,7 @@ mod tests {
 
         let mut checksum_eng = checksum::Engine::new();
         checksum_eng.input(&normalize_aliases).unwrap();
-        assert_eq!(format!("{}#{}", &normalize_aliases, checksum_eng.checksum()), output);
+        assert_eq!(format!("{}#{}", normalize_aliases, checksum_eng.checksum()), output);
     }
 
     #[test]
--- a/src/miniscript/context.rs
+++ b/src/miniscript/context.rs
@@ -167,7 +167,7 @@ impl fmt::Display for ScriptContextError {
 /// context under which the script is used.
 /// For example, disallowing uncompressed keys in Segwit context
 pub trait ScriptContext:
-    fmt::Debug + Clone + Ord + PartialOrd + Eq + PartialEq + hash::Hash + private::Sealed
+    fmt::Debug + Clone + Ord + PartialOrd + Eq + PartialEq + hash::Hash + private::Sealed + 'static
 where
     Self::Key: MiniscriptKey<Sha256 = sha256::Hash>,
     Self::Key: MiniscriptKey<Hash256 = hash256::Hash>,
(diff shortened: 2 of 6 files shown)


apoelstra:
On 1f9e81845ad6278b7a1a304fbf540a27ab1873cf  successfully ran local tests

src/blanket_traits.rs
```diff
@@ -100,3 +100,27 @@ where
     type _Hash160FromStrErr = <<T as MiniscriptKey>::Hash160 as FromStr>::Err;
     type _FromStrErr = <T as FromStr>::Err;
 }
+
+/// Internal trait used to convert a generic &Miniscript<Pk, Ctx> to a specific
+/// value of Ctx.
+pub(crate) trait DowncastMiniscript<Pk: MiniscriptKey> {
+    fn downcast<Ctx: ScriptContext>(&self) -> Option<&Miniscript<Pk, Ctx>>;
+}
```
tcharding: Any reason not to make this a method on `Miniscript` (in the `impl<Pk: MiniscriptKey, Ctx: ScriptContext> Miniscript<Pk, Ctx>` block)?

This builds for me. `NarrowedCtx` came from an AI suggestion and is allegedly nomenclature from Typescript.

```rust
        #[allow(unsafe_code)]
        pub(super) fn downcast<NarrowedCtx: ScriptContext>(&self) -> Option<&Miniscript<Pk, NarrowedCtx>> {
            use core::any::TypeId;
            if TypeId::of::<Ctx>() == TypeId::of::<NarrowedCtx>() {
                // SAFETY: this pointer cast is a no-op, as guaranteed by the if guard
                // In Rust 1.76 we can use core::ptr::from_ref in place of this cast.
                Some(unsafe { &*(self as *const Self).cast() }) // cast needed til Rust 1.76
            } else {
                None
            }
        }
```

tcharding (approved):
ACK 1f9e81845ad6278b7a1a304fbf540a27ab1873cf

src/blanket_traits.rs
apoelstra: Yeah, I like this. Let me change it.

apoelstra:
ACK 6fd1cc3ed609f16f7e960446ee4fc4174beb691a; successfully ran local tests

tcharding (approved):
ACK 6fd1cc3ed609f16f7e960446ee4fc4174beb691a

PR #942, opened 2026-05-13
just: Use cargo rbmt in lint task

tcharding:
We have `cargo rbmt` in CI but not in the justfile. Use it for the lint job because currently `just lint` is failing, did not investigate why.

At the same time make the command pass. I do not know why the lint job in CI is working but not the patched `just lint`?

The fixes are just allow `hex` duplicate dep in `minisicript` and in `fuzz` allow duplicates of `miniscript` and `hex`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,3 +73,8 @@ required-features = ["compiler"]
 [workspace]
 members = ["fuzz"]
 exclude = ["embedded", "bitcoind-tests"]
+
+[package.metadata.rbmt.lint]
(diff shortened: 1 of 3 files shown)


trevarj:
Semi-related, but I noticed a bit ago that we should also bump the rbmt version and move the config to Cargo.toml too

apoelstra (approved):
ACK 4b625a74ad5cc6b6122869f29c3a70704083912c; successfully ran local tests

PR #946, opened 2026-05-13
refactor: return `DerivationResult` from `derive_at_index`

### LINKED ISSUE #918: Have better ergonomics for `derive_at_index` on a descriptor with no wildcards
> * Might be nice if the `NoWildcard` error variant had the original descriptor converted to a definite descriptior, so the user could unwrap the error rather than if-gating and calling a different method.

This reminds me of what I was originally expecting, in terms of ergonomics.

It would be nice to be able to do something like:

(don't implement this)
```rust

enum DerivationResult<'a, T, E> {
    Ok(T),
    Err { error: E, original: &'a T },
}

desc.derive_at_index(0).into_result()?; // returns `DerivationResult<T, E>` -> `Result<T, E>` then errors if no wildcards
desc.derive_at_index(0)?; // same as above but nicer
// returns `DerivationResult<T, E>`, then ignores no wildcard and returns original, not sure about `fallback()` naming.
desc.derive_at_index(0).fallback();
```

_Originally posted by @trevarj in https://github.com/rust-bitcoin/rust-miniscript/issues/911#issuecomment-4187396684_

trevarj:
This allows a caller of `derive_at_index` to fall back on the original
descriptor (as a `Descriptor<DefiniteDescriptorKey>`) without having to `match`
or `if-let` on the `Result` when there is no wildcard.

We introduce a new result type called `DerivationResult` to handle this, but
since the `Try` trait is not stabilized in our MSRV we cannot have `?` and must
use `into_result()?` instead.

Closes #918

I ultimately settled on `or_fallback` cause I feel that it reads nicely with
`derive_at_index(0)`.

Not sure if I really like this solution when we can't use `Try` and then just
`derive_at_index(0)?`

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -650,6 +650,50 @@ impl<Pk: MiniscriptKey> ForEachKey<Pk> for Descriptor<Pk> {
     }
 }
 
+/// Result of [`Descriptor::derive_at_index`].
+///
+/// When the source descriptor has no wildcard,
+/// [`or_fallback`](DerivationResult::or_fallback) recovers the definite
+/// descriptor instead of returning an error.
+#[derive(Debug)]
+#[must_use]
+pub enum DerivationResult {
+    /// Derivation succeeded.
+    Ok(Descriptor<DefiniteDescriptorKey>),
+    /// The descriptor has no wildcard. Carries the original for fallback.
+    WithoutWildcard(Descriptor<DescriptorPublicKey>),
+    /// An error other than `NonDefiniteKeyError::NoWildcard`.
+    Error(NonDefiniteKeyError),
+}
+
+impl DerivationResult {
+    /// Converts into a `Result`.
+    ///
+    /// Use [`or_fallback`](DerivationResult::or_fallback) if you want
+    /// `NonDefiniteKeyError::NoWildcard` errors to be recovered automatically.
+    pub fn into_result(self) -> Result<Descriptor<DefiniteDescriptorKey>, NonDefiniteKeyError> {
+        match self {
+            DerivationResult::Ok(d) => Ok(d),
+            DerivationResult::WithoutWildcard(_) => Err(NonDefiniteKeyError::NoWildcard),
+            DerivationResult::Error(e) => Err(e),
+        }
+    }
+
+    /// Converts into a `Result`, recovering from `NonDefiniteKeyError::NoWildcard` by calling
+    /// [`into_definite`](Descriptor::into_definite) on the original descriptor.
+    pub fn or_fallback(self) -> Result<Descriptor<DefiniteDescriptorKey>, NonDefiniteKeyError> {
+        match self {
+            DerivationResult::Ok(d) => Ok(d),
+            DerivationResult::WithoutWildcard(original) => original.into_definite(),
+            DerivationResult::Error(e) => Err(e),
+        }
+    }
+
+    /// Unwraps the result, panicking on non-recoverable errors. Ignores
+    /// `NonDefiniteKeyError::NoWildcard` errors by calling [`or_fallback`](DerivationResult::or_fallback).
+    pub fn unwrap(self) -> Descriptor<DefiniteDescriptorKey> { self.or_fallback().unwrap() }
+}
+
 impl Descriptor<DescriptorPublicKey> {
     /// Whether or not the descriptor has any wildcards i.e. `/*`.
     pub fn has_wildcard(&self) -> bool { self.for_any_key(|key| key.has_wildcard()) }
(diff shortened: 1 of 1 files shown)


apoelstra:
Yeah, concept ACK. I was skeptical of this design at first but I think it leads to pretty nice-looking code.

src/descriptor/mod.rs
```diff
@@ -703,15 +749,12 @@ impl Descriptor<DescriptorPublicKey> {
...
-    ) -> Result<Descriptor<DefiniteDescriptorKey>, NonDefiniteKeyError> {
+    #[allow(deprecated)]
+    pub fn derive_at_index(&self, index: u32) -> DerivationResult {
         if !self.has_wildcard() {
-            return Err(NonDefiniteKeyError::NoWildcard);
+            return DerivationResult::new(Err(NonDefiniteKeyError::NoWildcard), Some(self.clone()));
         }
-        #[allow(deprecated)]
-        self.at_derivation_index(index)
+        DerivationResult::new(self.at_derivation_index(index), None)
```
tcharding: I'm having trouble following this code now. An undeprecated function calls a deprecated function (`at_derive_index`) that has a deprecation message that suggests calling the original function?

Unrelated nit, I rekon its cleaner without the private `new` function. Amusingly also, I argued exactly _for_ this style of early return recently and am now suggesting functional style is clearer. 

Feel free to do as you think best.

```rust
        if self.has_wildcard() {
            DerivationResult { result: self.at_derivation_index(index), original: None }
        } else {
            DerivationResult { result: Err(NonDefiniteKeyError::NoWildcard), original: Some(self.clone()) }
        }
```

tcharding: Its kind of weird, at least to me, that the `None` is coupled with the error in the `else` clause. I.e I had to chase the code to see why `original: None` was correct and it is exactly, and only, because the error in the other clause is `NoWildcard`). Just an observation.

tcharding:
Is `deprecated` used differently in this repo @apoelstra? is the `derive_at_index` function going to be deleted or is it going to stay and just be made private? For my understanding also, how long is the deprecation cycle in `miniscript`?

src/descriptor/mod.rs
trevarj: I should've stated more clearly that this stuff is a follow up to https://github.com/rust-bitcoin/rust-miniscript/pull/911#issuecomment-4187396684 

`derive_at_index` is new (from the parent PR) but it still uses the deprecated function for now, which I assume is because of this review comment : https://github.com/rust-bitcoin/rust-miniscript/pull/911#issuecomment-4183330098

> I had to chase the code to see why original: None was correct and it is exactly, and only, because the error in the other clause is NoWildcard). Just an observation.

Also to prevent cloning in the happy path. I don't actually like that either...I think changing it to an enum could make it nicer. Let me try that.

trevarj:
> is the `derive_at_index` function going to be deleted

this is the new function that was recently added in #911. `at_derivation_index` is the deprecated one.

apoelstra:
> For my understanding also, how long is the deprecation cycle in miniscript?

Things stay for at least two versions, but maybe (much) longer if they never get in the way of anything.

src/descriptor/mod.rs
```diff
@@ -650,6 +650,50 @@ impl<Pk: MiniscriptKey> ForEachKey<Pk> for Descriptor<Pk> {
...
+/// descriptor instead of returning an error.
+#[derive(Debug)]
+#[must_use]
+pub enum DerivationResult {
+    /// Derivation succeeded.
+    Ok(Descriptor<DefiniteDescriptorKey>),
+    /// The descriptor has no wildcard. Carries the original for fallback.
+    WithoutWildcard(Descriptor<DescriptorPublicKey>),
+    /// An error other than NoWildcard.
+    Error(NonDefiniteKeyError),
```
tcharding, line 666: Either doc or variant name needs changing, right?

src/descriptor/mod.rs
```diff
@@ -650,6 +650,50 @@ impl<Pk: MiniscriptKey> ForEachKey<Pk> for Descriptor<Pk> {
     }
 }
 
+/// Result of [`Descriptor::derive_at_index`].
+///
+/// When the source descriptor has no wildcard,
+/// [`or_fallback`](DerivationResult::or_fallback) recovers the definite
```
tcharding, line 656:
```suggestion
/// [`or_fallback`](DerivationResult::or_fallback) recovers the original
```

I think?

src/descriptor/mod.rs
tcharding: In docs below also.

tcharding:
Everything else looks good to me. I raised #956 while reviewing this, please holla at me if I was confused in writing it.

Reviewed: 84dd4dfb3b3ff416bb4e7e19a7344ab9e2f6616a

src/descriptor/mod.rs
trevarj: It returns the `DefiniteDescriptor` derived from the original descriptor (`Descriptor<DescriptorPublicKey>`)

src/descriptor/mod.rs
trevarj: It's referring to `NonDefiniteKeyError::NoWildcard`, so I think it's correct still.

trevarj: Updated all `NoWildcard` references to full name - `NonDefiniteKeyError::NoWildcard`

tcharding (approved):
ACK a15ce35646f202a76b000e14eaff3286d578fdf0

src/descriptor/mod.rs
tcharding: Ah my bad, hard to get good reviews huh.

src/descriptor/mod.rs
tcharding: No worries, thanks.

apoelstra (approved):
ACK a15ce35646f202a76b000e14eaff3286d578fdf0; successfully ran local tests

PR #949, opened 2026-05-17
plan: add Plan::tap_sighash_default accessor

evanlinjin:
## Summary

Adds a public accessor on `Plan` that exposes the existing per-Schnorr-placeholder size assumption — recorded when the plan was built from `TaprootCanSign::sighash_default` — as a single `Option<bool>`.

```rust
impl Plan {
    /// Returns whether the witness was sized assuming `SIGHASH_DEFAULT` (the
    /// 64-byte Taproot signature encoding).
    pub fn tap_sighash_default(&self) -> Option<bool>;
}
```

- `Some(true)` — every Schnorr placeholder is sized for a 64-byte signature (the `SIGHASH_DEFAULT` encoding).
- `Some(false)` — every Schnorr placeholder is sized for a 65-byte signature (any non-Default Taproot sighash flag).
- `None` — non-Taproot plans (signature size is invariant under the sighash flag for ECDSA), or mixed plans whose Schnorr placeholders have differing sizes — these cannot be expressed by a single uniform sighash policy per BIP-174.

## Motivation

Callers building a PSBT from a `Plan` need to populate `PSBT_IN_SIGHASH_TYPE` consistently with the plan's witness-size assumption — otherwise weight estimates drift from reality. The information already lives inside the plan (on every Schnorr placeholder's `size` field); this just surfaces it as a single readable answer.

Concretely: `bdk-tx` is reworking per-input sighash handling and needs to derive the PSBT sighash policy from the plan rather than carrying separate state on its `Input` type (see bitcoindevkit/bdk-tx#69 for the full design). Without this accessor it would have to walk `Plan::witness_template()` and infer the bool from placeholder sizes, which leaks internals.

## Test plan

- [x] `cargo test --lib tap_sighash_default` — new test covers Default-Taproot, non-Default-Taproot, and non-Taproot (wsh) plans.
- [x] `cargo test` — full suite still passes (139 / 0 / 0).

### DIFF
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -234,6 +234,44 @@ impl Plan {
     /// the script sig weight and the witness weight)
     pub fn satisfaction_weight(&self) -> usize { self.witness_size() + self.scriptsig_size() * 4 }
 
+    /// Returns whether the witness was sized assuming `SIGHASH_DEFAULT` — the 64-byte
+    /// Taproot signature encoding.
+    ///
+    /// * `Some(true)` — every Schnorr placeholder in this plan is sized for a 64-byte
+    ///   signature; the witness assumes `SIGHASH_DEFAULT`.
+    /// * `Some(false)` — every Schnorr placeholder is sized for a 65-byte signature;
+    ///   the witness assumes any non-Default Taproot sighash flag.
+    /// * `None` — this plan has no Schnorr placeholders (e.g. non-Taproot plans, where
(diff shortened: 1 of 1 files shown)


evanlinjin:
Fully AI generated - I need to ensure it didn't screw up before I take it out of draft.

evanlinjin:
Closing — wrong target branch. Superseded by #951 (targets master) and #950 (12.x backport). Sorry for the noise.

PR #951, opened 2026-05-17
plan: add Plan::tap_sighash_default accessor

evanlinjin:
## Summary

Adds a public accessor on `Plan` that exposes the existing per-Schnorr-placeholder size assumption — recorded when the plan was built from `TaprootCanSign::sighash_default` — as a single `Option<bool>`.

```rust
impl<Pk: MiniscriptKey + ToPublicKey> Plan<Pk> {
    /// Returns whether the witness was sized assuming `SIGHASH_DEFAULT` (the
    /// 64-byte Taproot signature encoding).
    pub fn tap_sighash_default(&self) -> Option<bool>;
}
```

- `Some(true)` — every Schnorr placeholder is sized for a 64-byte signature (the `SIGHASH_DEFAULT` encoding).
- `Some(false)` — every Schnorr placeholder is sized for a 65-byte signature (any non-Default Taproot sighash flag).
- `None` — non-Taproot plans (signature size is invariant under the sighash flag for ECDSA), or mixed plans whose Schnorr placeholders have differing sizes — these cannot be expressed by a single uniform sighash policy per BIP-174.

## Motivation

Callers building a PSBT from a `Plan` need to populate `PSBT_IN_SIGHASH_TYPE` consistently with the plan's witness-size assumption — otherwise weight estimates drift from reality. The information already lives inside the plan (on every Schnorr placeholder's `size` field); this just surfaces it as a single readable answer.

Concretely: `bdk-tx` is reworking per-input sighash handling and needs to derive the PSBT sighash policy from the plan rather than carrying separate state on its `Input` type (see bitcoindevkit/bdk-tx#69 for the full design). Without this accessor it would have to walk `Plan::witness_template()` and infer the bool from placeholder sizes, which leaks internals.

A 12.x backport is up at #950.
A 13.x backport is up at #952.

## Test plan

- [x] `cargo test --lib tap_sighash_default` — new test covers Default-Taproot, non-Default-Taproot, and non-Taproot (wsh) plans.
- [x] `cargo test --lib` — full suite still passes (157 / 0 / 0).

### DIFF
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -234,6 +234,44 @@ impl<Pk: MiniscriptKey + ToPublicKey> Plan<Pk> {
     /// the script sig weight and the witness weight)
     pub fn satisfaction_weight(&self) -> usize { self.witness_size() + self.scriptsig_size() * 4 }
 
+    /// Returns whether the witness was sized assuming `SIGHASH_DEFAULT` — the 64-byte
+    /// Taproot signature encoding.
+    ///
+    /// * `Some(true)` — every Schnorr placeholder in this plan is sized for a 64-byte
+    ///   signature; the witness assumes `SIGHASH_DEFAULT`.
+    /// * `Some(false)` — every Schnorr placeholder is sized for a 65-byte signature;
+    ///   the witness assumes any non-Default Taproot sighash flag.
+    /// * `None` — this plan has no Schnorr placeholders (e.g. non-Taproot plans, where
+    ///   signature size is invariant under the sighash flag), or its Schnorr placeholders
+    ///   have differing sizes — a mixed plan that cannot be expressed by a single uniform
+    ///   sighash policy per BIP-174.
+    ///
+    /// This mirrors the `sighash_default` flag recorded on each
+    /// [`TaprootCanSign`] when the plan was built. Callers constructing a PSBT
+    /// can use this to populate `PSBT_IN_SIGHASH_TYPE` consistently with the
+    /// witness-size assumption baked into [`Plan::satisfaction_weight`].
+    pub fn tap_sighash_default(&self) -> Option<bool> {
+        let mut acc: Option<bool> = None;
+        for placeholder in &self.template {
+            let size = match placeholder {
+                Placeholder::SchnorrSigPk(_, _, size) => *size,
+                Placeholder::SchnorrSigPkHash(_, _, size) => *size,
+                _ => continue,
+            };
+            let is_default = match size {
+                64 => true,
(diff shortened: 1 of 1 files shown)


nymius (commented):
This is a guess based on information which was available at the moment of providing the assets. Why not surfacing directly the SIGHASH value obtained from the assets?

evanlinjin:
> This is a guess based on information which was available at the moment of providing the assets. Why not surfacing directly the SIGHASH value obtained from the assets?

I agree we should expose what's available directly. But the asset model only tracks sighash to the extent it affects witness size and the detailed sighash type isn't carried. Look at `CanSign: ecdsa`. This is a plain `bool` with no sighash field at all, because every ECDSA sighash variant produces a same-sized DER signature. The Taproot side has `sighash_default: bool` only because Taproot is the one place sighash affects sig length (64 vs 65 bytes).

To expose without inferring, I can bubble `TaprootCanSign::sighash_default: bool` information up as a variant field changes in `Placeholder`. However that will be a breaking change and adds more complexity imo.

apoelstra:
I agree that this method does not seem useful for the claimed motivation. `PSBT_IN_SIGHASH_TYPE` is supposed to have an actual sighash type and this does not provide one.

Also, if there are multiple sighash types, we ought to return a vector of them. Maybe we should return `Result<sighashtype, Vec<sighashtype>>` or something like that.

evanlinjin:
> I agree that this method does not seem useful for the claimed motivation. `PSBT_IN_SIGHASH_TYPE` is supposed to have an actual sighash type and this does not provide one.

Fair argument, let me try rehash the motivation of this PR.

In transaction building, we do coin selection, then we would build a PSBT. During coin selection, we can use `Plan` to give us exactly the input size. When we create a PSBT, we don't want to put in put a `PSBT_IN_SIGHASH_TYPE` that goes against the input size of plan. In `bdk_tx`, we would like to error if the caller attempts to do this.

Why this is dangerous? Let's say the caller set's a feerate target of 0.1 sats/B for coin selection. Their wallet has all single-sig taproot outputs. Their plan assumes sighash type of DEFAULT. Now if they create a PSBT with `PSBT_IN_SIGHASH_TYPE` with ALL, this will create a transaction with a weight greater than what coin selection expected. Now they can't broadcast the transaction (feerate below 0.1 sats/B).

Relooking at this PR, I agree that the usecase is very niche. I also just realized that we have `pub fn witness_template(&self) -> &Vec<Placeholder<Pk>> { &self.template }` so we can figure this out without upstream changes!

Closing.

PR #944, opened 2026-05-13
Re-name `semantic::Policy` to `Semantic`

tcharding:
First patch introduces `use_self` lint. Its massive, are you ok with that?

Second patch does the rename. For reference, this was prompted by the first item in https://github.com/rust-bitcoin/rust-miniscript/pull/937#issuecomment-4364016259

Patch two is only 100 lines of red/green.

PR #957, opened 2026-05-21
Introduce `ValidationParams` structure; use in `Miniscript` constructors

apoelstra:
This PR begins the process of overhauling the validation framework of this library. It introduces a `ValidationParams` struct which is a huge struct containing knobs for every single validation check that this library does. The goal is to centralize all these knobs, document them, apply them consistently, and make it possible for the user to choose any set of them.

Currently validation is spread across a variety of `sanity_check` methods, inconsistently applied in various constructors for `Miniscript` and `Descriptor`, and only configurable in some ad-hoc inconsistent ways via methods like `from_str_insane`. There is no place where these are all listed; the closest thing is the `Ctx` type parameter, which has a bunch of methods like `check_local_policy_validity` which have inscrutable names and are all implemented by calling each other in slightly-different ways.

In short, it's a mess.

As a side-effect of this PR, we introduce some new validation errors which let us eliminate two variants from the top-level `Error` mega-enum. We also delete the `ExtParams` struct, which was similar in spirit to `ValidationParams`, except that it was incomplete and all its names sucked.

This PR also introduces a soft-deprecation of the term "insane", which was used to mean "validate a bunch of rules, but not the ones that we decided are the boundary between a Miniscript that is 'sane' and not". We now use the term "consensus", i.e., "validate the consensus rules but no additional rules". Note that the new notion allows raw pkh fragments while the old one did not. I am open to reverting this change, but my feeling is that nobody will notice because nobody is pushing the boundaries of any of these validation rules. (If they were, they'd be filing a lot more bugs.)


apoelstra:
Happy to rebase on #944 if that goes in first. I don't think they'll conflict significantly.

apoelstra:
CI is failing because the policy size limits for `Tap` are wrong, and because I need to fixup one of the tests. I will do this later.

apoelstra:
On f7756f5e595b5a65f1459eecea9bd69dc71f030a successfully ran local tests

src/miniscript/context.rs
```diff
@@ -283,9 +282,6 @@ where
 
     /// Check whether the top-level is type B
     fn top_level_type_check<Pk: MiniscriptKey>(ms: &Miniscript<Pk, Self>) -> Result<(), Error> {
-        if ms.ty.corr.base != types::Base::B {
-            return Err(Error::NonTopLevel(format!("{:?}", ms)));
-        }
```
trevarj, line 288: So if I understand correctly, this will be fixed later to once again check if it's type `B`?

apoelstra: No, this function will be totally deleted. We now check for type `B` in `Miniscript::validate`. The commit message describes this.

PR #963, opened 2026-05-22
ci: enable workspace lint rules

### LINKED ISSUE #959: Linting isn't configured to work with `cargo rbmt` ATM
And with 5 minutes playing on a Friday I couldn't get it working. Probably something to do with this repo having a workspace but also a `src/` directory. I tried throwing "." in the `members` list to no avail.

To close this issue probably decide on what lints we want in this repo (@apoelstra) and then hack it up (anyone).

FYI, in my environment I have.

```bash
$ cargo rbmt version
0.3.0 (unknown)
```

nyonson:
First patch defines the workspace lint rules and enables them in the root package. While I believe the rules could just be defined in the root package directly, might be nice to share them with the fuzz package (and any future packages)?

Closes #959

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,11 +70,17 @@ name = "taptree_of_horror"
 path = "examples/taptree_of_horror/taptree_of_horror.rs"
 required-features = ["compiler"]
 
-[workspace]
-members = ["fuzz"]
-exclude = ["embedded", "bitcoind-tests"]
-
 [package.metadata.rbmt.lint]
 allowed_duplicates = [
     "hex-conservative",
 ]
+
+[lints]
+workspace = true
+
+[workspace]
+members = ["fuzz"]
+exclude = ["embedded", "bitcoind-tests"]
+
+[workspace.lints.clippy]
+use_self = "warn"


nyonson:
I'll tack on a commit to fix up the lints if cool with this approach.

trevarj:
Should we bump the rbmt-version commit?

nyonson:
Yea, let me see how easy it is to update to v0.3.0

apoelstra:
concept ACK.

Interesting to just turn on `use_self` completely. In my version of that patch I'd ignored a lot of instances (e.g. for `Pkh` which is shorter to type than `Self`). But I noticed Tobin just fixed every instance and that's fine too (and maybe easier to maintain).

The lint is particularly useful here because we have a lot of generics, and `Self` sets them all while just `Miniscript` lets the type inference engine set them, which can be confusing for both the reviewer and the compiler.

Cargo.toml
```diff
@@ -70,11 +70,17 @@ name = "taptree_of_horror"
...
-members = ["fuzz"]
-exclude = ["embedded", "bitcoind-tests"]
-
 [package.metadata.rbmt.lint]
 allowed_duplicates = [
     "hex-conservative",
 ]
+
+[lints]
+workspace = true
```
tcharding, line 79: Gee wiz, so this is the only thing I was missing. Face palm.

tcharding:
How about I grab this patch then the ones from #944 sans the re-name and we push that up first? (And I'll include the 'stop using ::*` thing as suggested on that PR.)

trevarj:
I think this can be closed cause it was merged in [bdf50e3](https://github.com/rust-bitcoin/rust-miniscript/commit/bdf50e3bad8130989fde5a014ea6c0306a7e36cf)

Also resolves #959

apoelstra:
Superceded by #965.

PR #965, opened 2026-05-25
`Self` lint and cleanup

tcharding:
This is Nick's patch from #963 + the `Self` stuff from #944 + a patch on top to remove some of the `use Policy::*` instances. There are some in there still because I can't be bothered right now working out why `Self::` doesn't work everywhere. This is a step in the right direction.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,11 +70,17 @@ name = "taptree_of_horror"
 path = "examples/taptree_of_horror/taptree_of_horror.rs"
 required-features = ["compiler"]
 
-[workspace]
-members = ["fuzz"]
-exclude = ["embedded", "bitcoind-tests"]
-
 [package.metadata.rbmt.lint]
 allowed_duplicates = [
     "hex-conservative",
 ]
+
+[lints]
+workspace = true
+
+[workspace]
+members = ["fuzz"]
+exclude = ["embedded", "bitcoind-tests"]
+
+[workspace.lints.clippy]
+use_self = "warn"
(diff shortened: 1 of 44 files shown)


tcharding:
Formatting in separate patches to ease review.

tcharding:
I went ahead and poached your patch @nyonson.

trevarj:
> [Run the forgatter](https://github.com/rust-bitcoin/rust-miniscript/pull/965/commits/7af8cb980e616eb8b72c4c5582e6c0bcda3d65d6)

g and m aren't even close on any keyboard layout!

Will you squash the formatting commits into their parents?

ACK 7af8cb980e616eb8b72c4c5582e6c0bcda3d65d6

apoelstra:
> Will you squash the formatting commits into their parents?

For the `use_self` one I think it'd make the combo-commit much harder to read.

For the "no wildcards" commit sure, there's not a ton of benefit in separating that from the formatter.

tcharding:
Everyone knows hard core hackers call it a forgatter, keep up.

tcharding:
Squashed ... which also removed the ~slang~ typo.

apoelstra:
@tcharding Could you re-order these commits?

Because the first commit (which enables lints but doesn't make them pass) is the tip of #963, my CI system is trying to test it in "tip" mode, which runs clippy, which fails.

Rather than re-architecting my system to understand "test this commit as though it was part of #965 and *not* part of #963", I think it's easier to make this your problem.

tcharding:
> I think it's easier to make this your problem.

Lol, love it. Can do.

apoelstra (approved):
ACK 2cc105c4fa7fcf98cdee71b5e008d07c4944c6d8; successfully ran local tests

tcharding:
I thought of this joke the other day but couldn't find a 'master' quote then coincidentally re-read Eric Raymond's seminal (to me anyways) paper on how to be a hacker and it has the quote in it I wanted. So here it is, for your amusement.

Poelstra says, "I think it's easier to make this your problem."

I drift back to the days of being an apprentice and remember:

> Look to the master,
> Follow the master,
> Walk with the master,
> See through the master,
> Become the master.

Then think, "mad, I'm saying that from now on -  I think it's easier to make this your problem."

PR #960, opened 2026-05-22
fix: preserve previously finalized PSBT inputs

trevarj:
Skip re-finalizing PSBT inputs that already have `final_script_sig` or
`final_script_witness` which could cause a failure if re-finalized.

See https://github.com/bitcoindevkit/bdk_wallet/issues/488

Fixes #958 and requires backport to v13.x

PR #966, opened 2026-05-25
12.x: Preserve finalized PSBT inputs

noahjoeris:
This is a back-port of #960 to the 12.x branch. 
BDK still uses miniscript v12.x

It preserves PSBT inputs that already have `final_script_sig` or `final_script_witness`, avoiding re-finalization failures when finalized inputs no longer contain fields such as `witness_script` or `partial_sigs`.


apoelstra (approved):
ACK 7d2c3aefae7fe49af29a4818935cde224d1dbbd0; successfully ran local tests

apoelstra:
Would you be willing to open another PR that bumps the version in Cargo.toml (and the two Cargo.lock files) so we can cut a new release of 12.x?

Issue #785, opened 2024-12-18
Failure to parse `pk(xprv.../*)` descriptor


brunoerg:
See https://github.com/brunoerg/bitcoinfuzz/issues/70. 

The following descriptor `pk(xprv9s21ZrQH143K31xYSDQpPDxsXRTUcvj2iNHm5NUtrGiGG5e2DtALGdso3pGz6ssrdK4PFmM8NSpSBHNqPqm55Qn3LqFtT2emdEXVYsCzC2U/*)` fails to be parsed due to "Public keys must be 64/66/130 characters in size"/"pubkey string should be 66 or 130 digits long, got: 113".

apoelstra:
I believe that this is a gap in the definition for `DescriptorPublicKey`.

It may be that we want to reorganize the type to always have a derivation path (or maybe a vector of derivation paths or something).

The relevant spec is BIP 380 https://github.com/bitcoin/bips/blob/master/bip-0380.mediawiki#user-content-Key_Expressions

brunoerg:
https://github.com/bitcoinfuzz/bitcoinfuzz/issues/70#issuecomment-4415379828

cc @trevarj

trevarj:
@brunoerg thanks for  the reminder, forgot about this.

Quoted from the bitcoinfuzz issue:

> I could hack this to work by adding a case to FromStr for DescriptorPublicKey that checks if it's a xprv key and then just derive the xpub, but I'm not sure if that's what @apoelstra has in mind.
>
> I think someone might also expect parsing as a Descriptor<DescriptorSecretKey> (or even Descriptor<DescriptorXKey<Xprv>> to work, but they don't.

I will open up a draft PR so we don't lose my hacky fix.

trevarj:
Hi @brunoerg, can you change the fuzz to use `Descriptor<DescriptorPublicKey>::parse_descriptor()` instead?

This is how we are going to mitigate the issue: https://github.com/rust-bitcoin/rust-miniscript/pull/970#issuecomment-4583463195

brunoerg:
> Hi [@brunoerg](https://github.com/brunoerg), can you change the fuzz to use `Descriptor<DescriptorPublicKey>::parse_descriptor()` instead?
> 
> This is how we are going to mitigate the issue: [#970 (comment)](https://github.com/rust-bitcoin/rust-miniscript/pull/970#issuecomment-4583463195)

Sure, will do it.

brunoerg:
[...]

PR #970, opened 2026-05-27
fix: return error when parsing xprv into DescriptorPublicKey

trevarj:
Refactor FromStr for DescriptorPublicKey so that it can parse a xprv/tprv but
then error with a new variant `DescriptorKeyParseError::UnexpectedXPrivateKey`
so that the user knows to use
`Descriptor<DescriptorPublicKey>::parse_descriptor()` instead.

Added test that reflects what was reported in the issues:
https://github.com/bitcoinfuzz/bitcoinfuzz/issues/70
https://github.com/rust-bitcoin/rust-miniscript/issues/785

### DIFF
--- a/src/descriptor/key.rs
+++ b/src/descriptor/key.rs
@@ -534,6 +534,8 @@ pub enum DescriptorKeyParseError {
     XonlyPublicKey(bitcoin::secp256k1::Error),
     /// XKey parsing error
     XKeyParseError(XKeyParseError),
+    /// Unexpected XPrivateKey when parsing [`Descriptor<DescriptorPublicKey>`][crate::Descriptor].
+    UnexpectedXPrivateKey,
 }
 
 impl fmt::Display for DescriptorKeyParseError {
@@ -553,6 +555,9 @@ impl fmt::Display for DescriptorKeyParseError {
             Self::WifPrivateKey(err) => err.fmt(f),
             Self::XonlyPublicKey(err) => err.fmt(f),
             Self::XKeyParseError(err) => err.fmt(f),
+            Self::UnexpectedXPrivateKey => f.write_str(
+                "unexpected private extended key. Use `Descriptor::parse_descriptor` instead to preserve private key data in the returned KeyMap",
+            ),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


apoelstra:
This feels a little dangerous to me. If the user provides secret data and then we irreversibly throw it away, could that cause a loss of the secret?

trevarj:
> we irreversibly throw it away

Yeah, that's a good point. Using `parse_descriptor` seems like the best route since you get back the `KeyMap`. 
Wonder if `FromStr` should just be removed. I feel like most people will see `parse_descriptor` in the API before they consider calling `parse()`, or maybe that's just me.

apoelstra:
> Wonder if `FromStr` should just be removed. I feel like most people will see `parse_descriptor` in the API before they consider calling `parse()`, or maybe that's just me.

`parse` is a stdlib function that users will reach for before looking at our API for even a second.

And given that descriptors are, at their core, a string representation of an output, `FromStr` is a pretty fundamental trait.

Having said this, we may want to add a special error variant to `FromStr` for `DescriptorPublicKey` saying that it appears we're trying to parse an xprv, and maybe even suggesting to use `parse_descriptor`.

apoelstra:
We can also add an `allow_private_keys` boolean to `ValidationParams` in #957, which is explicitly documented to convert private keys to public ones, and then people can use `from_str_with_params` to explicitly opt into this behavior.

That could be used for the fuzztests so that they would pass.

trevarj:
> That could be used for the fuzztests so that they would pass.

Why not just change the fuzz tests to use `parse_descriptor`? Then return a new error variant in `FromStr` with a warning to use  `parse_descriptor`. No need for the extra `from_str_with_params` IMO

apoelstra:
Yeah, good idea.

apoelstra (approved):
ACK 1d2208612ee690fb76b16eb7bda624f4d09c2664; successfully ran local tests

apoelstra:
1d2208612ee690fb76b16eb7bda624f4d09c2664 needs rebase

trevarj:
Rebased

apoelstra (approved):
ACK d7deb69e53e82af4c775c84f44a5cd993e6b8b1d; successfully ran local tests

PR #972, opened 2026-06-01
Model OrdF64 as a non-NaN float

### LINKED ISSUE #445: Model `OrdF64` as `NonNan`
It'd be better to model `OrdF64` as `NonNan` because it brings panic (or error) closer to the cause of the issue making debugging easier. Potential inspiration: https://docs.rs/ordered-float/3.0.0/ordered_float/struct.NotNan.html

The change should be simple: make the field private and constructor panics or returns `Result`.

Abeeujah:
`PositiveReal` newtype for probability handling

Summary:
Replaces  `f64` and `OrdF64` with `PositiveReal` newtype throughout the policy module, making invalid probability states unrepresentable at the type level.

Changes:
1. Patch 1: Rename `OrdF64` to `PositiveReal`.
2. Patch 2: Extract `PositiveReal` into its own module.
3. Patch 3: Implement (`Add`, `Mul`), display, and `normalize` for `PositiveReal`, `ratio` for `Threshold`.
4. Patch 4: Add `parse_probability` helper.
5. Patch 5: Use `PositiveReal` in the policy module.
6. Patch 6: Fix lint warning.

closes #445

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -1671,7 +1671,7 @@ mod tests {
     #[test]
     fn tr_roundtrip_key() {
         let script = Tr::<String>::from_str("tr()").unwrap().to_string();
-        assert_eq!(script, format!("tr()#x4ml3kxd"))
+        assert_eq!(script, "tr()#x4ml3kxd".to_string())
     }
 
     #[test]
--- a/src/expression/mod.rs
+++ b/src/expression/mod.rs
@@ -681,7 +681,7 @@ impl<'a> Tree<'a> {
 }
 
 /// Parse a string as a u32, forbidding zero.
-pub fn parse_num_nonzero(s: &str, context: &'static str) -> Result<u32, ParseNumError> {
+fn parse_num_nonzero(s: &str, context: &'static str) -> Result<u32, ParseNumError> {
     if s == "0" {
         return Err(ParseNumError::IllegalZero { context });
     }
--- a/src/expression/mod.rs
+++ b/src/expression/mod.rs
@@ -34,7 +34,9 @@ pub use self::error::{ParseNumError, ParseThresholdError, ParseTreeError};
 use crate::blanket_traits::StaticDebugAndDisplay;
 use crate::descriptor::checksum::verify_checksum;
 use crate::prelude::*;
-use crate::{AbsLockTime, Error, ParseError, RelLockTime, Threshold, MAX_RECURSION_DEPTH};
+use crate::{
+    AbsLockTime, Error, ParseError, PositiveF64, RelLockTime, Threshold, MAX_RECURSION_DEPTH,
+};
 
 /// Allowed characters are descriptor strings.
 pub const INPUT_CHARSET: &str = "0123456789()[],'/*abcdefgh@:$%{}IJKLMNOPQRSTUVWXYZ&+-.;<=>?!^_|~ijklmnopqrstuvwxyzABCDEFGH`#\"\\ ";
@@ -689,6 +691,12 @@ pub fn parse_num_nonzero(s: &str, context: &'static str) -> Result<u32, ParseNum
         }
     }
     u32::from_str(s).map_err(ParseNumError::StdParse)
+}
+
+/// Parse a string as a PositiveF64, forbidding zero.
+pub fn parse_probability(s: &str, context: &'static str) -> Result<PositiveF64, ParseNumError> {
+    let parsed = parse_num_nonzero(s, context)?;
+    PositiveF64::try_from(parsed).map_err(|_| ParseNumError::IllegalZero { context })
 }
 
 /// Parse a string as a u32, for timelocks or thresholds
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -140,6 +140,7 @@ pub use crate::miniscript::satisfy::{Preimage32, Satisfier};
 pub use crate::miniscript::{hash256, Miniscript};
 use crate::prelude::*;
 pub use crate::primitives::absolute_locktime::{AbsLockTime, AbsLockTimeError};
+pub use crate::primitives::positive_f64::PositiveF64;
 pub use crate::primitives::relative_locktime::{RelLockTime, RelLockTimeError};
 pub use crate::primitives::threshold::{Threshold, ThresholdError};
 pub use crate::validation::{Error as ValidationError, ValidationParams};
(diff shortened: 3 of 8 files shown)


apoelstra:
We actually have a much stronger invariant available than "not `NaN`". It's actually true that our numbers are always positive and finite. So we can rename this type from `OrdF64` to `PositiveReal` or even `Probability`.

Rather than making a ton of functions like `insert_elem` return errors, change them to take `OrdF64`s instead of `f64`s. Then we can implement `Add`/`Mul` (but not `Div` and `Sub`!) on `OrdF64` and basically do everything in terms of these.

Similarly we should change `cost_1d` to take and return `OrdF64`s.

The only places that a NaN can actually occur are where we do division -- I believe these are all in `best_compilations` (which already returns a `Result`) and in `threshold` (which is only called by `best_compilations`. These divisions should be encapsulated: 

* the `threshold` one involves division by `n` which is guaranteed to be nonzero by the `Threshold` type, so we should refactor it to take a `thresh` rather than `k` and `n` separately
* the other case is division by `total` where we're normalizing two values so they sum to one. We should encapsulate this into some sort of `normalize` function on `OrdF64`

apoelstra:
I would propose splitting this into several commits:

* One renames the type from `OrdF64` to `PositiveReal` (or `PosReal` or `Positive` if you don't like typing)
* One moves the type from `compiler.rs` up to its own module in `src/primitives`
* One adds `Add`, `Mul` impls to the type, a `normalize` method which takes two values and divides both by their sum, and a method on `Threshold` which returns `k/n` as a `PositiveReal`. **None of these need to return a Result**
* One adds a `parse_probability` method to `src/expression/mod.rs` which wraps `parse_num_nonzero` but returns a `PositiveReal` instead of a `u32`
* Then you can update the compiler to use all this new functionality, which will enforce the invariant without needing to change any compiler methods

Abeeujah:
@apoelstra Thank you so much for the detailed spec you provided, I believe this is ready for review, although you didn't mention, using `parse_probability` resulted into me updating the `Policy::Or` variant, if this isn't what you had in mind, I can revert.

apoelstra:
Thanks! A few comments that should probably be addressed in followups:

* (in 1f25a35bf957b07827a6e854d924d2af9eb409a2) change the `pub` to `pub(super)` on the `f64` in `PositiveReal`; the internals are only needed by this module and the adjacent `threshold` module
* (in 7ae706a0bfcf3a99b8f9bf2118ca2d6a93e207f5) please add `#[must_use]` to `PositiveReal::normalize`; also I think the method should be called `normalized` though I'm not sure
* (in 7ae706a0bfcf3a99b8f9bf2118ca2d6a93e207f5) the `Threshold::ratio` method I think should be called `k_over_n` rather than `ratio`.
* (in 5d83c23ff214fec8b444685bff32e9127fc3e842) rather than having `parse_probability` directly construct a `PositiveReal`, it should use `try_from` and `unwrap` -- or better, `parse_num_nonzero` should be refactored into a private method that returns a `NonZeroU32`, and we can add a `From<NonZero<T>> where T: Into<u64>` impl on `PositiveReal`. And then use `from` with no `unwrap`

apoelstra:
Actually, given the quantity of these I think it'd be better to update the PR rather than doing a folowup.

Abeeujah:
> * From<NonZero<T>> where T: Into<u64>

This was only made stable in `1.79.0`, and to use a generic `T`, it has to implement `ZeroablePrimitive` which isn't available on stable rust. I did `TryFrom<u32>`, and mapping the error to `IllegalZero` .

Every other changes were addressed as stated.

apoelstra:
In cfb0336ee81687dd37261cb479b5bbf6beb2b934:

You cannot have a `new` method that takes an arbitrary `f64` without checking whether it's positive.

You can call it `new_unchecked` but by the end of the PR you should be able to delete it.

In 3b0e7f97df2f57552c7c865b6c1a9851c561799a:

You definitely should not be calling `PositiveReal::new(f64::INFINITY)` or `PositiveReal::new(0.0)`. Neither zero nor infinity are positive reals.

To handle 0 you will need to add some sort of `conditional_add` method to `PositiveReal` which takes an `Option` and either adds a value or doesn't.

To handle infinity I suspect you want to just back up, call the type `PositiveF64` instead of `PositiveReal`, and we can drop the invariant that it be finite.

Abeeujah:
I don't think we can entirely eliminate a constructor for this type, for the `new` method check/validation, I took the `assert!` route as propagating errors would need it to be handled by the caller, and you specifically mentioned that none of the compiler methods needs to be modified to return a Result.

All cases where we were having Zero has been handled, the INFINITY extracted to a constant as `f64` would.

apoelstra:
I think I will need to take over this PR. I see no reason that `PositiveReal`s should be constructed inside compiler.rs at all.

apoelstra:
Yeah, I see the complexity here. There are a bunch of changes that need to happen, possibly at once (though I'll try to untangle them into intellegible commits) if we want to eliminate all the constructors in compiler.rs.

In particular, `Policy::Or` needs to stop using `usize` and start using `PositiveF64`, which affects the entire file (and then some).

apoelstra:
Further complicating things is that the compiler sometimes uses 0 probablities to mean "unsatisfiable". But I think this is confused logic that comes from the compiler attempting to be overly generic over different fragments.

As always, as soon as I pull on one thing with this library it turns out there's a whole ediface of weird hacks that come tumbling down..

Abeeujah:
> Further complicating things is that the compiler sometimes uses 0 probablities to mean "unsatisfiable". But I think this is confused logic that comes from the compiler attempting to be overly generic over different fragments.

I think I handled the 0-probability case in 0243267 those were PositiveF64::new(0.0) placeholders, now replaced with Option + conditional_add.

The real blocker going forward for me isindicating a fragment is free to satisfy. Unlike probabilities, the cost domain already encodes impossibility separately as INFINITY/f64::MAX, so zero isn't a sentinel there; it's a real number PositiveF64 can't represent.

> As always, as soon as I pull on one thing with this library it turns out there's a whole ediface of weird hacks that come tumbling down..

To be fair, the initial commit was 8 years ago.. and a lot has evolved overtime.

apoelstra:
See https://github.com/rust-bitcoin/rust-miniscript/pull/988

apoelstra:
Costs cannot be `PositiveF64`s. Costs are routinely 0 (and as you say, in one case they are infinity). Only probabilities are guaranteed to be positive.

But even probabilities are set to 0 sometimes in the current compiler, notably when handling the andor construction.

Abeeujah:
Superceded by #988

PR #973, opened 2026-06-02
Remove `--features` flag from `sane` just recipe

nymius:
While executing `just sane` I received the following error message:

```bash
cargo test --quiet --all-targets --no-default-features --features > /dev/null || exit 1
error: a value is required for '--features <FEATURES>' but none was supplied

For more information, try '--help'.
error: Recipe `sane` failed on line 34 with exit code 1
```

I removed the flag because if fixes the issue and the flag is not receiving a feature list.

### DIFF
--- a/justfile
+++ b/justfile
@@ -31,7 +31,7 @@ docsrs:
 
 # Quick and dirty CI useful for pre-push checks.
 sane: fmt-check lint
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK a5728482ecde4c8931970c242b65f4ec70f17aff; successfully ran local tests

PR #974, opened 2026-06-02
Prepare `Plan::update_psbt_input` for migration to `rust-psbt`

nymius:
This is the simplest way I found to use the public `Plan` API in the `Plan::update_psbt_input` method so It can later be migrated to the `rust-psbt` project.

**Notes to the reviewers**:
- There is any reason to not expose the `descriptor` and `template` fields directly?
- There is a **TODO** in the code, but I'm lacking the context of it. It is something that should be fixed or a leftover?

### DIFF
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -363,8 +363,6 @@ impl Plan<DefiniteDescriptorKey> {
                     data
                 });
 
-            // TODO: TapTree. we need to re-traverse the tree to build it, sigh
-
             let leaf_hash = match data.spend_type {
                 Some(SpendType::KeySpend { internal_key }) => {
                     input.tap_internal_key = Some(internal_key);
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -217,8 +217,8 @@ pub struct Plan<Pk: MiniscriptKey> {
     pub absolute_timelock: Option<absolute::LockTime>,
     /// The relative timelock this plan uses
     pub relative_timelock: Option<relative::LockTime>,
-
-    pub(crate) descriptor: Descriptor<Pk>,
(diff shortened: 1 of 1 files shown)


apoelstra:
> * There is any reason to not expose the `descriptor` and `template` fields directly?

Nope, go for it.

> * There is a **TODO** in the code, but I'm lacking the context of it. It is something that should be fixed or a leftover?

The context appears to be a few lines below where we do a loop `for (pk, key_source) in data.key_origins` which then builds a new tree in `input.tap_key_origins`. I think the TODO is that afillini imagined that he could somehow directly memcpy the tree structure over instead of doing a tree traversal?

Probably fine to delete the TODO. It dates to #592, like much of the "weird" stuff in this module, and likely wasn't reviewed since the new module was too big.

nymius:
Thanks for clarifying _@apoelstra , I've removed the TODO as the functionality seems covered and the possible explanation is related to a performance improvement.

I've publicly exposed the `descriptor` field as it was the only field that I didn't have access to.

As removing the `witness_template` method would be a breaking change, and exposing the `template` field is duplicating the functionality provided by this method, I'm leaving as it.

I've copied the code from here into `rust-psbt` to verify it helps moving towards removing PSBT stuff from miniscript. You can check it in: https://git.rust-bitcoin.org/nymius/rust-psbt/compare/push-yormpqokyrxz...nymius/rust-psbt:push-xstopvrvruzp

nymius:
I've removed the changes that weren't needed by `rust-psbt` to port `update_psbt_input` there.

apoelstra (approved):
ACK 01834204a84a8739ac88c3fe182e016a4cab796a; successfully ran local tests

PR #978, opened 2026-06-03
Backport rust-bitcoin/rust-miniscript#974 to `rust-miniscript@13.x`

nymius:
Getting the changes from #974 into `rust-miniscript@13.x` is easier and faster than waiting for the next major release.

This is a back port of #974.

### DIFF
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -360,8 +360,6 @@ impl Plan {
                     data
                 });
 
-            // TODO: TapTree. we need to re-traverse the tree to build it, sigh
-
             let leaf_hash = match data.spend_type {
                 Some(SpendType::KeySpend { internal_key }) => {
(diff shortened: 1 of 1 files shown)


src/plan.rs
```diff
@@ -326,7 +326,7 @@ impl Plan {
             input.tap_merkle_root = spend_info.merkle_root();
 
             let data = self
-                .template
+                .witness_template()
```
nymius: This isn't necessary for `rust-psbt`, I will remove it here and in #974, to keep the changes minimal.

apoelstra (approved):
ACK 9e1ced68ed00e33f07dd448b482ddc6a71088eb9; successfully ran local tests

apoelstra:
I don't think we have any more backports coming down the pike. Feel free to open a PR to do a point release.

Issue #979, opened 2026-06-05
Can we drop `Ord` on `AbsLockTime` and `RelLockTime`?


apoelstra:
These types implement `Ord` so that we could `derive` `Ord` on `Terminal` (which in turn is used to order descriptors, but the ordering on descriptors is supposed to be lexicographic since descriptors are strings).

Upstream's locktimes avoid implementing this since it's rarely meaningful to compare locktimes that mix times and heights. In #537, when upgrading to rust-bitcoin 0.30, we just bodged our own `Ord` impls onto our locktime types, since we had them anyway, since Miniscript enforces locktimes to be nonzero. Indeed, our use of `Ord` on locktimes has caused some pathological behavior. See the `regression_895` unit test which demonstrates this. (Despite the name, this "regression test" does not test that any bug has been fixed; it only documents the particular form of bug that exists.)

Since #722 we no longer derive `Ord` on `Terminal`. Instead we have a manual implementation which uses our display logic to get a lexicographic order without using recursive functions.

So the original motivation for having `Ord` on locktimes no longer applies, and we should attempt to eliminate it as a way of shaking out any ordering-related locktime bugs.

apoelstra:
Ah we still `derive` `Ord` on the policy types. So to fix this bug, we will need to first manually implement `Ord` for those types.

151henry151:
Hey @apoelstra , I've just finished some work on this. PR here: https://github.com/rust-bitcoin/rust-miniscript/pull/982

I based it on your 2026-06/mallsat branch so it would only make sense to merge after #980.

This was a bit more involved than the "short and sweet" LowerHex PR on rust-bitcoin, but I'm hopeful that it's the right approach and would address this issue.

PR #980, opened 2026-06-05
miniscript: don't use malleable satisfier, even for or_i dissatisfactions

### LINKED ISSUE #976: Satisfier makes malleable satisfactions in some contrived cases
In https://github.com/rust-bitcoin/rust-miniscript/blob/bdf50e3bad8130989fde5a014ea6c0306a7e36cf/src/miniscript/satisfy/sat_dissat.rs#L445-L454 we have this giant blockcomment justifying the use of malleable satisfaction logic when computing dissatisfactions:

```rust
// Regarding use of `minimum_mall` for dissatisfactions: this is correct, because
// dissatisfactions for individual fragments are not required to be unique. (We
// track this property using the `ty.malleability.dissat` field.)
//
// This is because usually dissatisfactions are eventually dropped, e.g. by the
// `j:` wrapper or because at the top-level a dissatisfaction is simply invalid.
// In cases where non-unique dissatisfactions would cause malleability, e.g. in
// the `or_c` or `or_b` fragments which use dissatisfactions as part of their
// satisfactions, we set `ty.malleability.non_malleable` appropriately (and
// potentially reject the whole script at creation time).
```

This is just wrong. There's a lot packed into the claim "we set `ty.malleability.non_malleable` appropriately".

It **is** true that if there are multiple non-`s` dissatisfactions for a fragment, and it's possible that a top-level satisfaction would include a dissatisfaction for that fragment, we mark the total script as malleable.

But this is **not** true if there is a `s` dissatisfaction and also a non-`s` dissatisfaction. In this case the script may still be considered non-malleable. It is the satisfier's responsibility not to choose the `s` dissatisfaction, even if it's cheaper.

I noticed this trying to fix the existing `regression_895` to work with a non-malleable miniscript. After a couple hours of work I finally got one.....and then realized that the satisfier was producing a malleable satisfaction!

I will open a test case for this and a fix.

apoelstra:
Significantly cleans up the `regression_895` unit test, which I need to do to support the next PR in the `ValidationParams` series. Along the way I found multiple issues (#976 and #979).

Fixes #976

### DIFF
--- a/src/miniscript/satisfy/mod.rs
+++ b/src/miniscript/satisfy/mod.rs
@@ -563,29 +563,73 @@ mod tests {
 
     #[test]
     fn regression_895() {
-        // Tests a pathological descriptor whose cheapest satisfaction would require mixing
-        // timelocks, although a more expensive satisfaction is available that avoids the
-        // timelock mixing. This descriptor is accepted by rust-miniscript but its satisfier
-        // cannot produce a satisfaction for it.
+        // Tests a pathological descriptor whose cheapest satisfaction involves computing a
+        // dissatisfaction containing a timelock. Such dissatisfactions exist because the
+        // `and_v` fragment, uniquely, has a dissatisfaction that includes the satisfaction
+        // of its left child. (Check sat_dissat.rs and search 'dissat: ' to see how the
+        // dissatisfactions of every fragment are computed. You will see that exactly one,
+        // and_v, uses the satisfaction of a child.)
         //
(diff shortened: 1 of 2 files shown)


apoelstra:
On 99501a5e52b039f40971c73ffea2ab2cb5bd5014 successfully ran local tests

apoelstra:
@tcharding I accidentally merged this by merging #982 which builds on top of it.

Would you mind doing a cursory post-merge review of it? I don't think it makes sense to do a "deep" review since this stuff is way in the weeds, and no user has ever noticed the bugs it fixes/describes.

PR #981, opened 2026-06-07
refactor: rename safe to signed in the type checker

### LINKED ISSUE #977: `safe` in the typechecker should be called `signed`
Also the doccomment for this field should be changed. It currently says:

```
true if satisfactions cannot be created by any 3rd party who has not yet seen a satisfaction.

Hash preimages and signature checks are safe; timelocks are not. Affects malleability.
```

This is inaccurate. Hash preimages are not "safe" (which is why we renamed this term to "signed").

The code is correct, just the doccomment is wrong.

trevarj:
- simple find and replace, plus doc comment fixes
- changed CompilerError::TopLevelNonSafe to TopLevelSigless to coincide with the
  rename above

Fixes #977

### DIFF
--- a/src/miniscript/types/malleability.rs
+++ b/src/miniscript/types/malleability.rs
@@ -57,12 +57,11 @@ impl Dissat {
 pub struct Malleability {
     /// Properties of dissatisfying inputs
     pub dissat: Dissat,
-    /// `true` if satisfactions cannot be created by any 3rd party
-    /// who has not yet seen a satisfaction.
-    ///
-    /// Hash preimages and signature checks are safe; timelocks are not. Affects
-    /// malleability.
-    pub safe: bool,
+    /// `true` if every satisfaction, if one exists, requires a
+    /// signature. `false` for hash preimages and timelocks, since
+    /// they can be satisfied without a signature.
+    /// Affects malleability.
+    pub signed: bool,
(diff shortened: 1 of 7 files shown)


src/miniscript/types/malleability.rs
```diff
@@ -57,23 +57,22 @@ impl Dissat {
...
-    /// `true` if satisfactions cannot be created by any 3rd party
-    /// who has not yet seen a satisfaction.
-    ///
-    /// Hash preimages and signature checks are safe; timelocks are not. Affects
-    /// malleability.
-    pub safe: bool,
+    /// `true` if every satisfaction, if one exists, requires a
+    /// signature. `false` for hash preimages and timelocks, since
+    /// they can be satisfied without a signature.
+    /// Affects malleability.
```
trevarj, line 63: Is this wording ok?

apoelstra (approved):
ACK 8a552cd7ca93ac3920203bfe6b4bd3390ee0a046; successfully ran local tests

src/miniscript/types/malleability.rs
apoelstra: Yes, though it might be too long for a single line. We should make some effort down the line to audit our docs for stuff like that.

There is some nuance where "every satisfaction" can mean "dissatisfaction" in some pathological cases but I don't think it's worth hacking up the wording to try to cover that.

PR #982, opened 2026-06-07
primitives: remove Ord from AbsLockTime and RelLockTime

### LINKED ISSUE #979: Can we drop `Ord` on `AbsLockTime` and `RelLockTime`?

151henry151:
This PR implements #979 by removing `Ord` and `PartialOrd` from `AbsLockTime` and `RelLockTime`. Those impls compared height- and time-based locktimes as raw consensus `u32` values, which is not a meaningful total order. The wrapper types exist because Miniscript locktimes cannot be zero; ordering was incidental.

Policy types still need a total order, so we replace `derive(Ord)` on concrete and semantic `Policy` with explicit variant-name ordering and consensus-u32 comparison of matching `After`/`Older` nodes via `cmp_by_consensus` on the locktime types. `Terminal::cmp` uses the same method, and `Miniscript::Ord` picks that up through `node.cmp`. Policy and Miniscript ordering may change; that ordering was never a stable public contract.

In `concatenate_rev`, same-unit locktimes merge via `max` (bitcoin `partial_cmp`); mixed height/time on one path yields `Witness::Impossible`, so plan building fails or picks another branch. Composite satisfiers may still expose both locktime kinds. `regression_895` expects `into_plan_mall` to fail on the height-vs-time case while `into_plan` still succeeds with the time-based locktime.

Stacked on PR #980 (`2026-06/mallsat`) — merge #980 first. Breaking changes; changelog and version bump left to maintainers.

Fixes #979

**Tested with:**
`cargo test --all-features` — passed
`just fmt-check` — passed
`just sane` — passed

### DIFF
--- a/src/miniscript/satisfy/mod.rs
+++ b/src/miniscript/satisfy/mod.rs
@@ -93,18 +93,20 @@ pub trait Satisfier<Pk: MiniscriptKey + ToPublicKey> {
     /// Given a HASH160 hash, look up its preimage
     fn lookup_hash160(&self, _: &Pk::Hash160) -> Option<Preimage32> { None }
 
-    /// Assert whether an relative locktime is satisfied
+    /// Returns whether the given relative locktime is satisfied.
     ///
-    /// NOTE: If a descriptor mixes time-based and height-based timelocks, the implementation of
-    /// this method MUST only allow timelocks of either unit, but not both. Allowing both could cause
-    /// miniscript to construct an invalid witness.
+    /// Composite satisfiers may expose both height- and time-based locktimes.
+    /// If a single witness path would require incompatible timelock kinds
+    /// simultaneously, satisfaction for that path becomes [`Witness::Impossible`]
+    /// and plan building fails or selects another path.
     fn check_older(&self, _: relative::LockTime) -> bool { false }
 
-    /// Assert whether a absolute locktime is satisfied
+    /// Returns whether the given absolute locktime is satisfied.
     ///
-    /// NOTE: If a descriptor mixes time-based and height-based timelocks, the implementation of
-    /// this method MUST only allow timelocks of either unit, but not both. Allowing both could cause
-    /// miniscript to construct an invalid witness.
+    /// Composite satisfiers may expose both height- and time-based locktimes.
+    /// If a single witness path would require incompatible timelock kinds
+    /// simultaneously, satisfaction for that path becomes [`Witness::Impossible`]
+    /// and plan building fails or selects another path.
     fn check_after(&self, _: absolute::LockTime) -> bool { false }
 }
 
@@ -674,18 +676,10 @@ mod tests {
         // Same descriptor as above, except that now we use a time-based timelock rather than a
         // lower height-based one.
         //
-        // Again, we wind up taking both timelock branches. Prior to #895, we would do this because
-        // we did not track the after(144) at all. After #895, we do it because our timelock logic
-        // handles mixed timelocks by silently clobbering one of them. (See the implementation of
-        // `concatenate_rev`, which calls `cmp::max` on the timelocks. This clobbers a locktime.
-        // See #979 for an explanation of this.)
-        //
-        // This time the "ideal" behavior would be that we track best satisfactions for both time-
-        // and height-based locktimes, and whenever we are forced into a conflict we throw one away.
-        // (The type system guarantees that we will always have at least one satisfaction left;
-        // otherwise the whole script would be flagged as mixing timelocks.)
-        //
-        // For now we just use this unit test to document the behavior.
+        // Again, both timelock branches are considered. When `concatenate_rev` must merge a
+        // height-based and a time-based absolute locktime on the same path, that satisfaction
+        // becomes `Witness::Impossible`. Plan building then fails or selects an alternate path.
+        // Parallel per-kind tracking remains future work (see #979 discussion).
         let descriptor_str = format!(
             "wsh(or_b(n:or_i(and_v(v:after(144),and_v(v:pk({}),pk({}))),{expensive_threshold}),ajt:and_v(v:after(1000000000),v:pk({}))))",
             available_keys[0], unavailable_keys[0], available_keys[1],
@@ -693,10 +687,16 @@ mod tests {
         let descriptor =
             Descriptor::<crate::DefiniteDescriptorKey>::from_str(&descriptor_str).unwrap();
 
-        let plan = descriptor.into_plan_mall(&satisfier).unwrap();
+        // Both `or_b` arms hit a height/time conflict in `concatenate_rev`, so no malleable
+        // plan exists.
+        assert!(descriptor.clone().into_plan_mall(&satisfier).is_err());
+        // Non-malleable plan still succeeds: `into_plan` dissatisfies the expensive threshold
+        // branch and satisfies the time-based `after(1000000000)` branch without merging
+        // incompatible locktimes on one path.
+        let plan = descriptor.into_plan(&satisfier).unwrap();
         assert_eq!(
             plan.absolute_timelock,
-            Some(absolute::LockTime::from_time(1000000000).unwrap()),
+            Some(absolute::LockTime::from_time(1_000_000_000).unwrap()),
         );
     }
 }
@@ -1029,6 +1029,13 @@ pub struct Satisfaction<T> {
 }
 
 impl<Pk: MiniscriptKey + ToPublicKey> Satisfaction<Placeholder<Pk>> {
+    const IMPOSSIBLE: Self = Self {
+        stack: Witness::Impossible,
+        has_sig: false,
+        relative_timelock: None,
+        absolute_timelock: None,
+    };
+
     /// The empty satisfaction.
     ///
     /// This has the property that, when concatenated on either side with another satisfaction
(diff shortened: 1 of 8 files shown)


apoelstra:
concept ACK. Will wait for #980 to merge before reviewing in detail.

> later merges can overwrite an earlier `None` with a `Some`, so a height/time conflict may not appear in the final `Plan` timelock fields.

lmao. This timelock-mixing thing (and `regression_895` which explores it) really is the gift that keeps on giving.

I think instead we should return `None` for the whole satisfaction in this case. Which means changing the return value for `concatenate_rev` and causing a big diff. But I think it's the right thing to do.

If you think it's too big a diff, we can just accept this PR as-is, update the regression test with even more explanation/history, and file a bug to improve it later.

**Eventually** I kinda think we should do two parallel satisfactions: one with height-based timelocks and one with time-based ones. This will be a little bit tricky to implement since there are four possibilities (time vs height for both relative and absolute locktimes) and we don't want to waste time or space in the case of a "normal descriptor" that has only one kind (or no kinds) of timelocks.

> Branch selection is unaffected (`minimum`/`minimum_mall` choose by witness cost, not timelocks), and the scenario requires impossible satisfier conditions anyway.

I'm don't think it's actually impossible. The docs for [`Satisfier::check_older` and `check_after`](https://docs.rs/miniscript/latest/miniscript/miniscript/satisfy/trait.Satisfier.html#method.check_older) have a warning saying "don't accept both time-based and height-based timelocks!" but (a) nobody is going to read this, and (b) it's hard to avoid doing this, since you can just take a tuple of two satisfiers and then it'll accept the union of the two.

I'd like to just remove this warning, or at least weaken it.

src/miniscript/satisfy/mod.rs
```diff
@@ -694,10 +694,26 @@ mod tests {
...
+        None => None,
+    }
+}
+
+fn max_rel_locktime(a: RelLockTime, b: RelLockTime) -> Option<RelLockTime> {
+    use cmp::Ordering::*;
+    match relative::LockTime::from(a).partial_cmp(&b.into()) {
+        Some(Greater) | Some(Equal) => Some(a),
+        Some(Less) => Some(b),
+        None => None,
```
trevarj: Possibly derive `PartialOrd` on `AbsLockTime` and `RelLockTime` (i know the manual impls were just removed) and then

```rust
fn max_locktime<T: PartialOrd + Copy>(a: T, b: T) -> Option<T> {
    use cmp::Ordering::*;
    match a.partial_cmp(&b)? {
        Greater | Equal => Some(a),
        Less => Some(b),
    }
}
```

src/miniscript/satisfy/mod.rs
```diff
@@ -1051,8 +1067,14 @@ impl<Pk: MiniscriptKey + ToPublicKey> Satisfaction<Placeholder<Pk>> {
     fn concatenate_rev(self, other: Self) -> Self {
```
trevarj, line 1064: > One caveat: later merges can overwrite an earlier None with a Some, so a height/time conflict may not appear in the final Plan timelock fields

does it mean that the satisfaction should be Witness::Impossible in case of a conflict?

src/miniscript/satisfy/mod.rs
apoelstra: TBH I think we should remove `PartialOrd` as well. Because Rust is dumb the `<` and `>` operators are connected to `PartialOrd` which tricks users into thinking they have a total ordering anyway.

src/miniscript/satisfy/mod.rs
apoelstra: Oh, yeah, good idea. This is what I was getting at above with "`concatenate_rev` should return `None`" but using `Witness::Impossible` is simpler and more correct.

src/miniscript/satisfy/mod.rs
trevarj: Probably not worth making a trait...maybe stick each fn into the respective type's impl then.

151henry151:
Pushed 2c0d98b: mixed height/time in concatenate_rev now yields Witness::Impossible, with max_same_kind on the locktime types and updated regression_895 / satisfier docs. PR description updated to match.

src/primitives/absolute_locktime.rs
```diff
@@ -68,6 +68,16 @@ impl AbsLockTime {
 
     /// Whether this is a time-based locktime.
     pub fn is_block_time(&self) -> bool { self.0.is_block_time() }
+
+    /// Returns the later of two locktimes of the same unit, or `None` if units differ.
+    pub(crate) fn max_same_kind(a: Self, b: Self) -> Option<Self> {
```
trevarj: why not just `max`?

151henry151: Done in 6c3a514. Had used `max_same_kind` to make the height/time `None` case obvious at the call site; `max` is fine on these types with the doc comment.

apoelstra:
In 1140b69ce269bc275513a4e6428cbc0d0f42a5bb:

I think we should sort locktimes numerically by their u32 representation. I could be convinced either way...but the numeric comparison is cheaper to implement and I suspect it matches user intuition a bit better.

In this commit we should also update the ord impls for `Miniscript` to use the new locktime comparison function.

151henry151:
Addressed in 66b99c2: same-kind locktime ordering in Policy::Ord and Terminal::cmp now uses consensus u32 instead of display strings. Miniscript::Ord already delegates to Terminal::cmp via self.node.cmp, so no separate change there.

apoelstra:
Please don't add commits which change other commits. Use git-rebase to edit the original commits in-place. This helps us review code commit by commit.

151henry151:
> Please don't add commits which change other commits. Use git-rebase to edit the original commits in-place. This helps us review code commit by commit.

Rebased so that review feedback is folded into the original commits, not added as fix-ups on top. Each commit builds cleanly.

apoelstra:
In bd68e352d76b6df4a877e10c8ab9f4fb87cdded1:

It would be a better abstraction to have `locktime_consensus_cmp` take a locktime and do the conversion itself. You will need two methods, one for `AbsLockTime` and one for `RelLockTime`. Also, it should be a method (still (pub(crate))) rather than a standalone function. I suggest calling it `cmp_by_consensus` or `cmp_by_consensus_encoding`.

151henry151:
> In [bd68e35](https://github.com/rust-bitcoin/rust-miniscript/commit/bd68e352d76b6df4a877e10c8ab9f4fb87cdded1):
> 
> It would be a better abstraction to have `locktime_consensus_cmp` take a locktime and do the conversion itself. You will need two methods, one for `AbsLockTime` and one for `RelLockTime`. Also, it should be a method (still (pub(crate))) rather than a standalone function. I suggest calling it `cmp_by_consensus` or `cmp_by_consensus_encoding`.

Done — cmp_by_consensus on AbsLockTime and RelLockTime, folded into the original commit.

I was about to step out the door to go to my wedding; we're leaving in two hours! I'll be gone about a week, no service, camping out in the wilderness for the celebration. So if it seems like I drop off the map, don't give up on me, I'll be back in seven days. =)

apoelstra:
Nice! Almost there. In c170d55adbe90a5e07a7eac262fe2fdd4d8014de can you update the doccomments (where you removed "which implements `Ord`" to say "that cannot be zero"?

The nonzero invariant is ultimately why we have these wrapper types. The Ord thing was just a bonus.

151henry151:
Updated the struct docs in c170d55 to say "that cannot be zero."

trevarj:
Henry Romp ***@***.***> writes:

> I was about to step out the door to go to my wedding; we're leaving in two hours! I'll be gone about a week, no service, camping out in the wilderness for the celebration. So if it seems like I drop off the map, don't give up on me, I'll be back in seven days. =)

LOL what the heck...congratulations and enjoy!

151henry151:
While updating the PR description I noticed `just fmt-check` failing on long `cmp_by_consensus` lines in `display.rs`. Ran `cargo fmt` and folded the result into the `Terminal::cmp` commit.

apoelstra:
queued up an ack/merge. Enjoy your wedding! If CI passes it'll be merged when you're back, or else it'll wait for you.

apoelstra (approved):
ACK 476dac5ec4c9d93dc94ab2e3e4fc97e29b4223df; successfully ran local tests

apoelstra:
Lol, dammit, I didn't realize that this built on #980. I did not intend to merge #980 since it hasn't been reviewed by anyone but its author (me).

Interesting process failure here... my reviewing system saw that the two commits from 980 were "already reviewed" and skipped over them, so when I ran `local-ci 982 next` it only showed me the new commits.

Issue #945, opened 2026-05-13
CI: Further improve `cargo-rbmt` usage


tcharding:
I did #942 and #943 just for fun. We should investigate if there is more that needs doing.

At a minimum this stuff: https://github.com/rust-bitcoin/rust-miniscript/pull/942#issuecomment-4437232526

PR #983, opened 2026-06-08
ci: improve rbmt usage

trevarj:
Commit by commit is the best way to understand what was done.

Bumped rbmt to the rust-bitcoin version and then copied from
relevant rust-bitcoin's CI files.

I think there are more CI files that could be copied, but not related
to rbmt (maybe weekly-update-stable?)

Relates to #945


trevarj:
cc: @nyonson & @tcharding

apoelstra (approved):
ACK 9ffd35e6119d584c4a19fb18385ee7abd793c8de; successfully ran local tests

PR #988, opened 2026-06-15
Replace `OrdF64` type in the compiler with a first-class `PositiveF64` type

apoelstra:
Supercedes #972.

The `OrdF64` type in the compiler is weaker than it needs to be. Nominally it exists just to force an ordering on `f64`, with a code comment that it "will panic if given NaN". But it doesn't actually enforce that you can't put NaN into it, or any other value. There are also lots of places where we do reckless division that could plausibly turn into 0.0 / 0.0, and where we add values as `u32`s (which might overflow and panic) before converting to f64 or OrdF64. Plus there are 'as f64` casts all over the place, which is code smell.

In fact, all the probabilities used by the compiler, and parsed from policies, are required to be positive numbers (the policy values are even integers). So we can greatly improve this situation by renaming `OrdF64` to `PostiveF64`, moving it to its own module, and enforcing that the contained value is always positive. This lets us:

* freely add these values without worrying about cancellation
* freely multiply and divide these values without worrying about zeroing them out or dividing by zero (although `Inf` is a permissible value for `PositiveF64`)
* eliminate tons of casts

However, to do this we need to significantly refactor the compiler logic, which currently uses `Option<f64>` to represent "branch probabilities" and does so in an incoherent way, where `None` represents "not yet set or unneeded" and `Some(0.0)` roughly represents "required to be set but unneeded" and there are `.expect`s all over the place. The compiler also stores these branch probabilities in the `AstElemExt` structure, which is totally unnecessary and results in a ton of mutation and arc-cloning.

### DIFF
--- a/src/expression/mod.rs
+++ b/src/expression/mod.rs
@@ -27,6 +27,7 @@
 
 mod error;
 
+use core::num::NonZeroU32;
 use core::ops;
 use core::str::FromStr;
 
@@ -679,7 +680,7 @@ impl<'a> Tree<'a> {
 }
 
 /// Parse a string as a u32, forbidding zero.
-pub fn parse_num_nonzero(s: &str, context: &'static str) -> Result<u32, ParseNumError> {
+pub fn parse_num_nonzero(s: &str, context: &'static str) -> Result<NonZeroU32, ParseNumError> {
     if s == "0" {
         return Err(ParseNumError::IllegalZero { context });
     }
@@ -688,7 +689,7 @@ pub fn parse_num_nonzero(s: &str, context: &'static str) -> Result<u32, ParseNum
             return Err(ParseNumError::InvalidLeadingDigit(ch));
         }
     }
-    u32::from_str(s).map_err(ParseNumError::StdParse)
+    NonZeroU32::from_str(s).map_err(ParseNumError::StdParse)
 }
 
 /// Parse a string as a u32, for timelocks or thresholds
@@ -697,7 +698,7 @@ pub fn parse_num(s: &str) -> Result<u32, ParseNumError> {
         // Special-case 0 since it is the only number which may start with a leading zero.
         return Ok(0);
     }
-    parse_num_nonzero(s, "")
+    parse_num_nonzero(s, "").map(u32::from)
 }
 
 #[cfg(test)]
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -5,6 +5,9 @@
 //! Optimizing compiler from concrete policies to Miniscript
 //!
 
+#![allow(dead_code)] // will be removed in next commit
+#![allow(unused_variables)] // will be removed in next commit
+
 use core::num::NonZeroU32;
 use core::{f64, fmt, mem};
 #[cfg(feature = "std")]
@@ -98,62 +101,80 @@ fn best_compilations_or<Pk: MiniscriptKey, Ctx: ScriptContext>(
     sat_prob: f64,
     dissat_prob: Option<f64>,
 ) -> Result<(), CompilerError> {
-    macro_rules! compile_tern {
-        ($a:expr, $b:expr, $c: expr, $w: expr) => {
-            compile_tern(policy_cache, policy, ret, $a, $b, $c, $w, sat_prob, dissat_prob)?
-        };
-    }
-
     let total = u32::from(subs[0].0) as f64 + u32::from(subs[1].0) as f64;
     let lw = u32::from(subs[0].0) as f64 / total;
     let rw = u32::from(subs[1].0) as f64 / total;
 
     //and-or
+    let mut insert_ternary = |policy_cache: &mut _,
+                              a: &BTreeMap<_, _>,
+                              b: &BTreeMap<_, _>,
+                              c: &BTreeMap<_, _>,
+                              lw: f64,
+                              rw: f64|
+     -> Result<(), CompilerError> {
+        for a in a.values() {
+            for b in b.values() {
+                for c in c.values() {
+                    if let Ok(new_ext) = AstElemExt::and_or(a, b, c, lw, rw) {
+                        insert_best_wrapped(
+                            policy_cache,
+                            policy,
+                            ret,
+                            new_ext,
+                            sat_prob,
+                            dissat_prob,
+                        )?;
+                    }
+                }
+            }
+        }
+        Ok(())
+    };
+
     if let (Concrete::And(x), _) = (subs[0].1.as_ref(), subs[1].1.as_ref()) {
-        let mut a1 = best_compilations(
+        let a1 = best_compilations(
             policy_cache,
             x[0].as_ref(),
             lw * sat_prob,
             Some(dissat_prob.unwrap_or(0 as f64) + rw * sat_prob),
         )?;
-        let mut a2 = best_compilations(policy_cache, x[0].as_ref(), lw * sat_prob, None)?;
+        let a2 = best_compilations(policy_cache, x[0].as_ref(), lw * sat_prob, None)?;
 
-        let mut b1 = best_compilations(
+        let b1 = best_compilations(
             policy_cache,
             x[1].as_ref(),
             lw * sat_prob,
             Some(dissat_prob.unwrap_or(0 as f64) + rw * sat_prob),
         )?;
-        let mut b2 = best_compilations(policy_cache, x[1].as_ref(), lw * sat_prob, None)?;
+        let b2 = best_compilations(policy_cache, x[1].as_ref(), lw * sat_prob, None)?;
 
-        let mut c =
-            best_compilations(policy_cache, subs[1].1.as_ref(), rw * sat_prob, dissat_prob)?;
+        let c = best_compilations(policy_cache, subs[1].1.as_ref(), rw * sat_prob, dissat_prob)?;
 
-        compile_tern!(&mut a1, &mut b2, &mut c, [lw, rw]);
-        compile_tern!(&mut b1, &mut a2, &mut c, [lw, rw]);
+        insert_ternary(policy_cache, &a1, &b2, &c, lw, rw)?;
+        insert_ternary(policy_cache, &b1, &a2, &c, lw, rw)?;
     };
     if let (_, Concrete::And(x)) = (&subs[0].1.as_ref(), subs[1].1.as_ref()) {
-        let mut a1 = best_compilations(
+        let a1 = best_compilations(
             policy_cache,
             x[0].as_ref(),
             rw * sat_prob,
             Some(dissat_prob.unwrap_or(0 as f64) + lw * sat_prob),
         )?;
-        let mut a2 = best_compilations(policy_cache, x[0].as_ref(), rw * sat_prob, None)?;
+        let a2 = best_compilations(policy_cache, x[0].as_ref(), rw * sat_prob, None)?;
 
-        let mut b1 = best_compilations(
+        let b1 = best_compilations(
             policy_cache,
             x[1].as_ref(),
             rw * sat_prob,
             Some(dissat_prob.unwrap_or(0 as f64) + lw * sat_prob),
         )?;
-        let mut b2 = best_compilations(policy_cache, x[1].as_ref(), rw * sat_prob, None)?;
+        let b2 = best_compilations(policy_cache, x[1].as_ref(), rw * sat_prob, None)?;
 
-        let mut c =
-            best_compilations(policy_cache, subs[0].1.as_ref(), lw * sat_prob, dissat_prob)?;
+        let c = best_compilations(policy_cache, subs[0].1.as_ref(), lw * sat_prob, dissat_prob)?;
 
-        compile_tern!(&mut a1, &mut b2, &mut c, [rw, lw]);
-        compile_tern!(&mut b1, &mut a2, &mut c, [rw, lw]);
+        insert_ternary(policy_cache, &a1, &b2, &c, rw, lw)?;
+        insert_ternary(policy_cache, &b1, &a2, &c, rw, lw)?;
     };
 
     let dissat_probs = |w: f64| -> Vec<Option<f64>> {
@@ -439,18 +460,13 @@ impl CompilerExtData {
         }
     }
 
-    fn and_or(a: Self, b: Self, c: Self) -> Self {
-        let aprob = a.branch_prob.expect("andor, a prob must be set");
-        let bprob = b.branch_prob.expect("andor, b prob must be set");
-        let cprob = c.branch_prob.unwrap_or(0.0);
-
+    fn and_or(a: Self, b: Self, c: Self, lprob: f64, rprob: f64) -> Self {
         let adis = a
             .dissat_cost
             .expect("BUG: and_or first arg(a) must be dissatisfiable");
-        debug_assert_eq!(aprob, bprob); //A and B must have same branch prob.
         Self {
             branch_prob: None,
-            sat_cost: aprob * (a.sat_cost + b.sat_cost) + cprob * (adis + c.sat_cost),
+            sat_cost: lprob * (a.sat_cost + b.sat_cost) + rprob * (adis + c.sat_cost),
             dissat_cost: c.dissat_cost.map(|cdis| adis + cdis),
         }
     }
@@ -497,18 +513,7 @@ impl CompilerExtData {
         Pk: MiniscriptKey,
         Ctx: ScriptContext,
     {
-        match *fragment {
-            Terminal::AndOr(ref a, ref b, ref c) => {
-                let atype = get_child(&a.node, 0);
-                let btype = get_child(&b.node, 1);
-                let ctype = get_child(&c.node, 2);
-                Self::and_or(atype, btype, ctype)
-            }
-            Terminal::Thresh(ref thresh) => {
-                Self::threshold(thresh.k(), thresh.n(), |n| get_child(&thresh.data()[n].node, n))
-            }
-            _ => unreachable!(),
-        }
+        unreachable!()
     }
 }
 
@@ -640,6 +645,29 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> AstElemExt<Pk, Ctx> {
                 Arc::new(Miniscript::FALSE),
             ))?,
             comp_ext_data: CompilerExtData::and_n(left.comp_ext_data, right.comp_ext_data),
+        })
+    }
+
+    fn and_or(
+        a: &Self,
+        b: &Self,
+        c: &Self,
+        l_weight: f64,
+        r_weight: f64,
+    ) -> Result<Self, types::Error> {
+        Ok(Self {
+            ms: Self::compose_typeck_only(Terminal::AndOr(
+                Arc::clone(&a.ms),
+                Arc::clone(&b.ms),
+                Arc::clone(&c.ms),
+            ))?,
+            comp_ext_data: CompilerExtData::and_or(
+                a.comp_ext_data,
+                b.comp_ext_data,
+                c.comp_ext_data,
+                l_weight,
+                r_weight,
+            ),
         })
     }
 
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -13,7 +13,7 @@ use sync::Arc;
 
 use crate::miniscript::context::SigType;
 use crate::miniscript::limits::{MAX_PUBKEYS_IN_CHECKSIGADD, MAX_PUBKEYS_PER_MULTISIG};
-use crate::miniscript::types::{self, ErrorKind, ExtData, Type};
+use crate::miniscript::types::{self, ErrorKind, Type};
 use crate::miniscript::ScriptContext;
 use crate::policy::Concrete;
 use crate::prelude::*;
@@ -280,6 +280,10 @@ impl CompilerExtData {
         Self { branch_prob: None, sat_cost: left.sat_cost + right.sat_cost, dissat_cost: None }
     }
 
+    fn and_n(left: Self, right: Self) -> Self {
+        Self { branch_prob: None, sat_cost: left.sat_cost + right.sat_cost, dissat_cost: None }
+    }
+
     fn or_b(l: Self, r: Self) -> Self {
         let lprob = l
             .branch_prob
@@ -353,7 +357,7 @@ impl CompilerExtData {
     fn and_or(a: Self, b: Self, c: Self) -> Self {
         let aprob = a.branch_prob.expect("andor, a prob must be set");
         let bprob = b.branch_prob.expect("andor, b prob must be set");
-        let cprob = c.branch_prob.expect("andor, c prob must be set");
+        let cprob = c.branch_prob.unwrap_or(0.0);
 
         let adis = a
             .dissat_cost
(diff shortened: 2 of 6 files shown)


src/primitives/positive_f64.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub const ONE: Self = Self(1.0);
+
+    /// Constant used in unit tsets
+    #[cfg(test)]
+    pub const ONE_QUARTER: Self = Self(0.25);
+
+    /// Attempts to create a [`PositiveF64`] from an ordinary `f64`.
+    pub fn new(f: f64) -> Option<Self> {
+        // Can likely make this const in Rust 1.83
+        if f > 0.0 {
```
Abeeujah: nit: Pending the MSRV upgrade for this to be made const, this can be rewritten to use the `then_some` method available on `bool`.
```rust
(f > 0.0).then_some(Self(f))
```

src/primitives/positive_f64.rs
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+
+    /// Given an [`Option<PositiveF64>`], if it is `Some` then add it to the value.
+    /// Otherwise return the unmodified value.
+    ///
+    /// Returns the sum (or original value). Does not modify in-place.
+    #[must_use]
+    pub fn conditional_add(self, other: Option<Self>) -> Self {
+        if let Some(other) = other {
```
Abeeujah: nit: This can also be written to make use of the `map_or` method available on `Option`
```rust
other.map_or(self, |i| i + self)
```

src/primitives/positive_f64.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// Internally clones the iterator and runs it twice, so best to only use
+    /// this with reference-based iterators obtained with e.g. `slice.iter()`
+    /// rather than "owning" iterators like you'd get from `vec.into_iter()`.
+    pub fn normalized_iter<I>(iter: I) -> NormalizedIterator<I>
+    where
+        I: Iterator<Item = Self> + Clone,
+    {
+        // Compute the sum of all the items in the iterator. Because all items in
+        // the iterator are positive, this will be 0 iff the iterator is empty.
+        let sum = iter.clone().map(|x| x.0).sum::<f64>();
```
Abeeujah, line 49: nit: iff -> if

src/primitives/positive_f64.rs
```diff
@@ -0,0 +1,179 @@
...
+        Self(t.k() as f64 / t.n() as f64) // casts okay, worst case wil lose precision
+    }
+
+    /// One minus the ratio `k` / `n` of a threshold, as a [`PositiveF64`]. Guaranteed
+    /// to be in the half-open range `[0, 1)`.
+    ///
+    /// Returns `None` if the return value would be 0, which is impermissible for the
+    /// [`PositiveF64`] type.
+    pub fn one_minus_k_over_n<const MAX: usize, T>(t: &Threshold<T, MAX>) -> Option<Self> {
+        if t.is_and() {
```
Abeeujah, line 70: nit: This could also use the `then_some` method, if adopted, the tradeoffs here would be conciseness over clarity
```rust
(!t.is_and()).then_some(Self(1.0 - (t.k() as f64 / t.n() as f64)))
```

src/policy/compiler.rs
```diff
@@ -104,6 +90,226 @@ impl fmt::Display for CompilerError {
...
+    subs: &[(NonZeroU32, Arc<Concrete<Pk>>)],
+    sat_prob: PositiveF64,
+    dissat_prob: Option<PositiveF64>,
+) -> Result<(), CompilerError> {
+    let total = PositiveF64::from(subs[0].0) + PositiveF64::from(subs[1].0);
+    let lw = PositiveF64::from(subs[0].0) / total;
+    let rw = PositiveF64::from(subs[1].0) / total;
+
+    //and-or
+    if let (Concrete::And(x), _) = (subs[0].1.as_ref(), subs[1].1.as_ref()) {
```
Abeeujah, line 143: I think this pattern here, could be simplified to just take the reference being used in this `if` block, and the one unused should not be evaluated at all, same for Line 158 below

src/policy/compiler.rs
```diff
@@ -104,6 +90,226 @@ impl fmt::Display for CompilerError {
...
+                            new_ext,
+                            sat_prob,
+                            dissat_prob,
+                        )?;
+                    }
+                }
+            }
+        }
+    };
+    if let (_, Concrete::And(x)) = (&subs[0].1.as_ref(), subs[1].1.as_ref()) {
```
Abeeujah, line 165: Same as stated above, the 0th value should not be evaluated since it is unused.

src/policy/compiler.rs
```diff
@@ -104,6 +90,226 @@ impl fmt::Display for CompilerError {
...
+    for l in l_comp[2].values() {
+        for r in r_comp[3].values() {
+            if let Ok(new_ext) = AstElemExt::or_i(r, l, rw, lw) {
+                insert_best_wrapped(policy_cache, policy, ret, new_ext, sat_prob, dissat_prob)?;
+            }
+        }
+    }
+
+    Ok(())
+}
```
Abeeujah, line 238: This function could benefit from Macros, it would greatly reduce the repetition.

Abeeujah (commented):
This is well executed, I identified a few nits which are optional or could be done in a follow up.

src/primitives/positive_f64.rs
apoelstra: It would be better to use `then` rather than `then_some`, I suppose, because strictly speaking what you wrote results in a temporarily-existing invalid `Self` in the case that `f <= 0`. But good idea.

src/primitives/positive_f64.rs
apoelstra: No, iff is correct.

src/primitives/positive_f64.rs
apoelstra: Hah, I think this is too noisy.

src/policy/compiler.rs
apoelstra: Sometimes we want to execute both sides of this. So there is no way to combine the two clauses into one.

src/policy/compiler.rs
apoelstra: The whole compiler needs to be rewritten to be non-recursive. So I don't want to make the code harder to modify ahead of doing that.

src/primitives/positive_f64.rs
Abeeujah: I believe, `then_some` returns `Some(T)` if the `bool` is true , or `None` otherwise; which looks very much like what the function is doing.

src/primitives/positive_f64.rs
Abeeujah: Yeah true, today's years old when I learned that.

src/primitives/positive_f64.rs
Abeeujah: Yeah, I thought as much

src/policy/compiler.rs
Abeeujah: Great planning ahead, hopefully we get to eliminate a couple of the double nested loops too in the planned rewrite.

Abeeujah (commented):
.

Abeeujah (approved):
ACK [424f23b](https://github.com/rust-bitcoin/rust-miniscript/pull/988/commits/424f23bfa60be83551260555f88e12c7ae391c74)

yancyribbens:
@Abeeujah can you stand on your head?

Abeeujah:
> @Abeeujah can you stand on your head?

I try, I try @yancyribbens

yancyribbens:
@Abeeujah

How many digits of pi can you calculate?

Abeeujah:
> @Abeeujah
> 
> How many digits of pi can you calculate?

I usually just end up with a massive headache after 3.14 🤣

yancyribbens:
> I usually just end up with a massive headache after 3.14 🤣

Ha! ok, you pass the Turing test :P

apoelstra:
@Abeeujah is not a LLM and has been contributing usefully to rust-bitcoin for quite some time.

yancyribbens:
> @Abeeujah is not a LLM and has been contributing usefully to rust-bitcoin for quite some time.

Yes, that is what I had meant by "passes Turing test", although maybe that was confusing.

My apologies @Abeeujah, we have an ongoing discussion in IRC to try to identify bot activity, and we thought your coding patterns might have been that of a bot/LLM.

apoelstra:
Rebased on master.

src/primitives/positive_f64.rs
apoelstra: Clippy prefers `then_some` to `then` which I don't like, but okay, I'll use `then_some`..

apoelstra:
Addressed @Abeeujah's comments except the "should use a macro" one which I want to play with a bit.

src/primitives/positive_f64.rs
Abeeujah: :sweat_smile: `then` is almost equivalent to just wriiting the if else block, whereas `then_some` is more intuitive for this usecase and fits everything nicely in one line.

Abeeujah:
> Addressed @Abeeujah's comments except the "should use a macro" one which I want to play with a bit.

You'd most likely have a lot of fun doing this, all the loop bodies are the same, only the provided arguments they evaluate differs.

apoelstra:
Okay, used closures in place of macros and significantly reduced the code repetition. Also added the missing ops impls for `PositiveF64`, renamed `and_or_0` to `and_n` which is the name used throughout the codebase, and cleaned up the history a bit.

I think this is ready to go now.

apoelstra:
On 347c334e4891dab1192f6adc79ddb822b2944cfa successfully ran local tests

Abeeujah (approved):
ACK 347c334e4891dab1192f6adc79ddb822b2944cfa Successfully ran local tests

apoelstra:
On c55dd43cc65fcf6b0d2dfa853120c7643d32779c successfully ran local tests

src/primitives/positive_f64.rs
```diff
@@ -0,0 +1,172 @@
...
+
+use crate::Threshold;
+
+/// Ordered f64 for comparison.
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub struct PositiveF64(f64);
+
+impl PositiveF64 {
+    /// The smallest representable value of a [`PositiveF64`].
+    pub const EPSILON: Self = Self(f64::EPSILON);
```
portlandhodl: Nit on PositiveF64::EPSILON:
- f64::EPSILON is the gap between 1.0 and the next representable float (≈ 2.2e-16)
- f64::MIN_POSITIVE is the smallest positive normal f64 (≈ 2.2e-308)


The doc says "the smallest representable value", which describes MIN_POSITIVE, not EPSILON. Given the constant's only use — a dummy initializer in enumerate_leaves, where it is always overwritten before being read MIN_POSITIVE seems like the better fit IMO

Also since the initial value is never read anyways ... a nit of a nit could be...

```rust
let mut found: Option<(Reverse<PositiveF64>, Arc<Self>)> = None;
// ...
found = Some((*p, Arc::clone(pol)));
// ...
let (prob, curr_policy) = found.expect("no_more_enum would have been set");
```

That would eliminates the dead Arc::new(Self::Unsatisfiable) dummy per loop iteration, and keeps a misleadingly-named constant out of the new public API.

src/primitives/positive_f64.rs
```diff
@@ -0,0 +1,172 @@
...
+    {
+        // Compute the sum of all the items in the iterator. Because all items in
+        // the iterator are positive, this will be 0 iff the iterator is empty.
+        let sum = iter.clone().map(|x| x.0).sum::<f64>();
+        NormalizedIterator { iter, sum }
+    }
+
+    /// The 'n' value of a threshold, as a [`PositiveF64`]
+    pub fn n<const MAX: usize, T>(t: &Threshold<T, MAX>) -> Self {
+        Self(t.n() as f64) // cast okay, worst case wil lose precision
```
portlandhodl: Nit: spelling "wil" -> "will"

Also on line 64

src/primitives/positive_f64.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /// The ratio `k`/`n` of a threshold, as a [`PositiveF64`]. Guaranteed to be
+    /// in the half-open range `(0, 1]`.
+    pub fn k_over_n<const MAX: usize, T>(t: &Threshold<T, MAX>) -> Self {
+        Self(t.k() as f64 / t.n() as f64) // casts okay, worst case wil lose precision
+    }
+
+    /// One minus the ratio `k` / `n` of a threshold, as a [`PositiveF64`]. Guaranteed
+    /// to be in the half-open range `[0, 1)`.
```
portlandhodl: Nit: should this be (0, 1)? The Some payload can never be 0; as stated below, that case returns None.

src/policy/compiler.rs
```diff
@@ -310,63 +389,41 @@ impl CompilerExtData {
...
 
-    fn or_b(l: Self, r: Self) -> Self {
-        let lprob = l
-            .branch_prob
-            .expect("BUG: left branch prob must be set for disjunctions");
-        let rprob = r
-            .branch_prob
-            .expect("BUG: right branch prob must be set for disjunctions");
+    fn and_n(left: Self, right: Self) -> Self {
+        Self { sat_cost: left.sat_cost + right.sat_cost, dissat_cost: None }
```
portlandhodl: and_n models dissat_cost as None (infinity), but and_n(a,b) == andor(a,b,0) is dissatisfiable via a. The 0 branch fails for free; so the cost is just a's dissat cost. 

Possible fix: https://github.com/portlandhodl/rust-miniscript/commit/6115bd2c23a3dd35fd5b893c35265f5cf03b936d

src/policy/compiler.rs
```diff
@@ -115,6 +101,142 @@ impl fmt::Display for CompilerError {
...
+        Ok(())
+    };
+
+    insert_binary(&l_comp[0], &r_comp[0], lw, rw, AstElemExt::or_b)?;
+    insert_binary(&r_comp[0], &l_comp[0], rw, lw, AstElemExt::or_b)?;
+    insert_binary(&l_comp[0], &r_comp[2], lw, rw, AstElemExt::or_d)?;
+    insert_binary(&r_comp[0], &l_comp[2], rw, lw, AstElemExt::or_d)?;
+    insert_binary(&l_comp[1], &r_comp[3], lw, rw, AstElemExt::or_c)?;
+    insert_binary(&r_comp[1], &l_comp[3], rw, lw, AstElemExt::or_c)?;
+    insert_binary(&l_comp[2], &r_comp[3], lw, rw, AstElemExt::or_i)?;
```
portlandhodl, line 232: L232 - L235 

The or_i candidate order changed vs. the pre-refactor code (positions 2 and 4 swapped: (r3,l2) now precedes (r2,l3)). This can flip the winner of a tie breaker. If byte stable replication is desired then this would need to be swapped.

a31f1be9fe2707de26012bd072af487d97f6644f stated  "Other than that I
tried to preserve the existing logic (and ran some local fuzztests quite
extensively to verify)."

portlandhodl (commented):
Some of my general thoughts about c55dd43cc65fcf6b0d2dfa853120c7643d32779c

src/primitives/positive_f64.rs
apoelstra: Oh, you're right. I think I wrote the comment in an earlier iteration when I was calling this `NonNegativeF64` and allowing 0 and attempting to use this type for weights as well as probabilities. (Then I realized this was not needed and would cause division to be fallible, so I changed my tune.)

src/primitives/positive_f64.rs
apoelstra: Great idea. I really didn't like this constant. I think when I introduced it I thought I was going to need it in more than one place.

Also great point that the `Arc::new(Unsatisfiable)` load probably has a cost that outweighs the additional `panic` branch (if the compiler doesn't eliminate it, which it likely can do by control-flow analysis).

apoelstra: And thanks for the trivia about `EPSILON` and `MIN_POSITIVE`! I did not know about `MIN_POSITIVE`. I wonder what 1.0 over this value is.

src/policy/compiler.rs
apoelstra: Nice catch! Yeah, I think I had confused the two branches of the `or` in `and_n` with the two satisfy/dissatisfy possibilities.

Only one of the `or` branches is accessible, which let to our old code calling `compile_tern!(&mut left, &mut q_zero_right, &mut zero_comp, [1.0, 0.0]);` (which my commit message should've highlighted as the offending line; this commit is kinda hard to read without that hint). But both satisfy/dissatisfy remain possible.

I'll ask Claude to find a unit test which demonstrates this.

apoelstra: Done. Claude wasted $3 of tokens before I gave up and switched to ChatGPT, which took 20 minutes to do it (which I'm a little surprised they let me do on the $20/mo tier..).

src/policy/compiler.rs
apoelstra: This is really weird and I wonder how I did this. Probably fat-fingering in vim.

apoelstra:
Addressed all of @portlandhodl's comments.

portlandhodl (approved):
ACK: 41b75a0508d5ff3f15676de28da486d35ba6eac0 - tests pass, comments / thoughts addressed concisely, diff looks great, no additional notes.

src/primitives/positive_f64.rs
```diff
@@ -0,0 +1,169 @@
...
+
+impl From<PositiveF64> for f64 {
+    fn from(value: PositiveF64) -> Self { value.0 }
+}
+
+impl From<NonZeroU32> for PositiveF64 {
+    fn from(value: NonZeroU32) -> Self { Self(f64::from(u32::from(value))) }
+}
+
+macro_rules! impl_op {
```
Rob1Ham, line 106: `PositiveF64`'s safe arithmetic does not preserve the invariant required by its `Eq` and `Ord` implementations. For example, `PositiveF64::new(f64::INFINITY).unwrap() / PositiveF64::new(f64::INFINITY).unwrap()` produces NaN, which is not equal to itself and panics in `Ord::cmp`. Similarly, dividing the smallest positive subnormal by two produces zero. This is also internally reachable through repeated weighted-policy multiplication: roughly 34 nested `or(1@...,4294967295@...)` branches underflow the low-probability path to zero. Once these values enter ordered maps, sets, heaps, or compiler cache keys, their ordering assumptions no longer hold. The arithmetic needs an explicit policy that preserves the domain, such as checked operations, documented saturation, or a representation that cannot underflow. If infinity remains accepted, `Div` also needs to account for `inf / inf`.

src/policy/compiler.rs
```diff
@@ -1182,20 +1135,18 @@ pub fn best_compilation<Pk: MiniscriptKey, Ctx: ScriptContext>(
...
 {
     best_compilations(policy_cache, policy, sat_prob, dissat_prob)?
         .into_iter()
-        .filter(|&(key, _)| {
-            key.ty.corr.base == types::Base::B && key.dissat_prob == dissat_prob.map(OrdF64)
-        })
+        .filter(|&(key, _)| key.ty.corr.base == types::Base::B && key.dissat_prob == dissat_prob)
         .map(|(_, val)| val)
-        .min_by_key(|ext| OrdF64(ext.cost_1d(sat_prob, dissat_prob)))
+        .min_by_key(|ext| PositiveF64::new(ext.cost_1d(sat_prob, dissat_prob)))
```
Rob1Ham, line 1149: Using `Option<PositiveF64>` directly as the `min_by_key` key gives invalid costs the best ordering. If `cost_1d` returns zero or NaN, `PositiveF64::new` returns `None`; Rust orders `None` before every `Some`, so an invalid candidate can be selected as the minimum rather than rejected or placed last. This can become reachable when an underflowed zero probability is multiplied by an infinite cost (`0.0 * INFINITY` is NaN). Could this explicitly reject non-finite/non-positive costs or map them after valid costs? The same issue appears in `best` below.

src/primitives/positive_f64.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Internally clones the iterator and runs it twice, so best to only use
+    /// this with reference-based iterators obtained with e.g. `slice.iter()`
+    /// rather than "owning" iterators like you'd get from `vec.into_iter()`.
+    pub fn normalized_iter<I>(iter: I) -> NormalizedIterator<I>
+    where
+        I: Iterator<Item = Self> + Clone,
+    {
+        // Compute the sum of all the items in the iterator. Because all items in
+        // the iterator are positive, this will be 0 iff the iterator is empty.
+        let sum = iter.clone().map(|x| x.0).sum::<f64>();
```
Rob1Ham, line 49: The direct sum can overflow even when every input is a finite valid `PositiveF64`. For `[f64::MAX, f64::MAX]`, `sum` becomes infinity and normalization returns `[0.0, 0.0]` rather than `[0.5, 0.5]`, contradicting both the normalization contract and the positivity invariant. An overflow-resistant approach could scale values by their maximum before summing. Tests for maximum finite values, disparate magnitudes, subnormals, and infinity would cover this boundary.

src/primitives/positive_f64.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    /// Takes an iterator over [`PositiveF64`] and produces a new iterator where
+    /// each item is divided so that they all total to 1.
+    ///
+    /// On an empty iterator, returns a new empty iterator.
+    ///
+    /// Internally clones the iterator and runs it twice, so best to only use
+    /// this with reference-based iterators obtained with e.g. `slice.iter()`
+    /// rather than "owning" iterators like you'd get from `vec.into_iter()`.
+    pub fn normalized_iter<I>(iter: I) -> NormalizedIterator<I>
```
Rob1Ham, line 43: This public method returns `NormalizedIterator<I>`, but downstream callers cannot name that type because it lives under the private `primitives` module and is not re-exported. Type inference permits immediate iteration, but downstream functions cannot store or return the concrete iterator type. Could `NormalizedIterator` be re-exported, this return `impl Iterator<Item = PositiveF64>`, or the method remain crate-private if it is only intended for compiler use?

src/expression/mod.rs
```diff
@@ -679,7 +680,7 @@ impl<'a> Tree<'a> {
 }
 
 /// Parse a string as a u32, forbidding zero.
-pub fn parse_num_nonzero(s: &str, context: &'static str) -> Result<u32, ParseNumError> {
+pub fn parse_num_nonzero(s: &str, context: &'static str) -> Result<NonZeroU32, ParseNumError> {
```
Rob1Ham, line 683: This changes the public return type from `u32` to `NonZeroU32`. The PR also changes the public `Policy::Or` and `TreeChildren::Or` payloads from `usize` to `NonZeroU32`, so existing downstream typed calls, construction, and pattern matching will stop compiling. If this targets the next major release, these changes should be called out in migration notes. Otherwise, retaining this signature and adding an internal nonzero helper would avoid an unrelated source break.

Rob1Ham (commented):
Some free daybreak (openai's cyber model) review!

Extended review after PortlandHODL's latest pass. The ordinary-policy test suite passes, and I did not find further regressions in disjunction propagation, threshold formulas, candidate ordering, Taproot enumeration, or Huffman construction. I did find several additional numeric-boundary and API issues, detailed inline. The main concern is that safe `PositiveF64` arithmetic can invalidate the assumptions required by its `Eq` and `Ord` implementations.

src/primitives/positive_f64.rs
portlandhodl: Verified ... though the bug is reachable only by downstream users calling the public method with adversarial floats.

src/expression/mod.rs
portlandhodl: Seems intentional as that is the point of this PR. Maybe a CHANGELOG entry for the next release stating that this breaks existing public APIs

src/primitives/positive_f64.rs
portlandhodl: Verified, in my opinion this would require an adversarial tap tree to hit. Famous last words ... "nobody would create a policy like this". This does not however make the finding untrue.

The result would simply be a sub-optimal spend (higher fees)

src/policy/compiler.rs
portlandhodl: No impact on this crate. Finding is true, would have to assume public api use.

apoelstra:
On 41b75a0508d5ff3f15676de28da486d35ba6eac0 successfully ran local tests

src/primitives/positive_f64.rs
apoelstra: Good catch. Will do in a followup but `NormalizedIterator` should be re-exported.

Insane that we needed such a heavy LLM to find something that the Rust compiler 100% could detect statically if it wanted to. (But it doesn't, because this sort of thing is a standard method of sealing traits.)

apoelstra:
Will file an issue for `PositiveF64` needing some sort of saturating logic to prevent you from reaching infinity or 0 (or we should just special-case all the arithmetic that would produce NaN; e.g. redefine inf/inf to be 1.0).

Will also file an issue, or maybe just stick a fix into my next PR, for `NormalizedIterator` not being exported.

apoelstra:
https://github.com/rust-bitcoin/rust-miniscript/issues/1033

PR #991, opened 2026-06-21
miniscript: replace `ExtParams`, `Ctx::check_*` methods, and more, with `ValidationParams`

apoelstra:
This series of commits starts a couple of parallel tracks:

* All of `ScriptContextError`, `LiftError` and `PolicyError` are merged into `ValidationError`, since they all represent essentially the same kinds of errors (except the previous error types were unequal and overlapping, and it was basically random which checks were done where)
* Their error variants in `Error` are removed and not replaced; many methods now return `ValidationError` or another more-specific type rather than `Error`
* We do validation of all rules in all `Descriptor` constructors; many previously-legal things (like `wsh(1)` which is an anyone-can-spend and should be forbidden by sanity rules) are no longer legal according to the library.
* Many error messages are improved.

### DIFF
--- a/src/descriptor/segwitv0.rs
+++ b/src/descriptor/segwitv0.rs
@@ -12,7 +12,7 @@ use bitcoin::{Address, Network, ScriptBuf, Weight};
 
 use crate::descriptor::write_descriptor;
 use crate::expression::{self, FromTree};
-use crate::miniscript::context::{ScriptContext, ScriptContextError};
+use crate::miniscript::context::ScriptContext;
 use crate::miniscript::limits::MAX_PUBKEYS_PER_MULTISIG;
 use crate::miniscript::satisfy::{Placeholder, Satisfaction, Witness};
 use crate::plan::AssetProvider;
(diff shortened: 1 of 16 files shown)


src/descriptor/segwitv0.rs
```diff
@@ -38,14 +38,17 @@ impl<Pk: MiniscriptKey> Wsh<Pk> {
...
         Ok(Self { ms })
     }
 
     /// Create a new sortedmulti wsh descriptor
-    pub fn new_sortedmulti(thresh: Threshold<Pk, MAX_PUBKEYS_PER_MULTISIG>) -> Result<Self, Error> {
+    pub fn new_sortedmulti(
+        thresh: Threshold<Pk, MAX_PUBKEYS_PER_MULTISIG>,
+    ) -> Result<Self, ValidationError> {
+        // The context checks will be carried out inside new function for
+        // sortedMultiVec
```
trevarj: c90c148e : 

Is this a TODO? I can't seem to find any new function for sortedmulti context validation. 

Also, is `sortedMultiVec` referring to the old type that I removed in ddc42c7b?

src/miniscript/mod.rs
```diff
@@ -366,6 +366,7 @@ mod private {
             // Malleability is only a top-level check since you can fix malleability
             // in some cases by adding wrappers.
             if !params.allow_malleability && !self.is_non_malleable() {
+                println!("{:?}", self);
```
trevarj: remove

src/descriptor/segwitv0.rs
apoelstra: Ah, yes, this is a rebase problem after #915. An old version of this PR had code to update `SortedMultiVec` but I think I dropped that commit. I'll track it down.

apoelstra:
After this PR we should add the `allow_bare` field suggested by a clanker in https://github.com/rust-bitcoin/rust-miniscript/pull/1021

At the same time we should re-add the `NonStandardBareScript` error variant, which this PR removes without replacing.

apoelstra:
PR on ice because I need to add a bunch of `Descriptor::new_*_with_params` methods to bypass the new checks (in our PSBT parser we use these to attempt to spend malleable/nonstandard scripts). Will revisit when I run out of commits on the #1034 branch and/or I find the energy/LLM cycles to add these new methods.

PR #1001, opened 2026-07-08
Drop obsolete unused-import workaround

satsfy:
The `#[cfg(not(test))]` guard on the secp256k1 imports worked around an `unused_imports` warning tracked in rust-lang/rust#121684 (2024). That bug is fixed, so the guard is no longer needed.

Built with `cargo +nightly-2026-06-26 build --all-targets`, a recent nightly version, and had no warning.

This is not a priority at all, but I caught it during a codebase read.


satsfy:
Force push rebases

apoelstra (approved):
ACK f6c561062eeb1eef164ca7f9aada7ef0bfe55dfc; successfully ran local tests

PR #1004, opened 2026-07-11
fix(key_map): do not fulfill BIP32 requests outside descriptor path

noahjoeris:
### Description

`GetKey` bypassed descriptor path validation on fingerprint match, allowing a master xprv for `…/0/*` to sign sibling paths like `…/9/7`.

Remove the early return and require `matches()` before key derivation. Adds a regression test.

### DIFF
--- a/src/descriptor/key_map.rs
+++ b/src/descriptor/key_map.rs
@@ -128,11 +128,7 @@ impl GetKey for DescriptorSecretKey {
                     Ok(None)
                 }
             }
-            (Self::XPrv(descriptor_xkey), ref key_request @ KeyRequest::Bip32(ref key_source)) => {
-                if let Some(key) = descriptor_xkey.xkey.get_key(key_request.clone(), secp)? {
-                    return Ok(Some(key));
-                }
-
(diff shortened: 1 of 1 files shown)


noahjoeris:
@apoelstra friendly ping. I think this is important to fix asap

apoelstra:
Yeah, I think you're right. I'm torn about whether it's reasonable for a user to rely on this -- it seems like if you only want to sign with a certain key, you should only make that key available to your signer.

But certainly, if a user *does* rely on this, we shouldn't be surprising them like that.

In general, Miniscript has a bias toward "if the user has asked to sign, do anything possible to make the signing happen" on the assumption that if the user isn't allowed to make that decision, presumably they're physically unable to make that decision.

apoelstra:
Sorry for the long delay. Should be ACKed/merged today.

If you could backport this to 12.x and 13.x (target branches `release-12.x` and `release-13.x`) I'd really appreciate it.

apoelstra (approved):
ACK 7043fa08f9c6adcdb6a5c8da64061b1ebb7dbd48; successfully ran local tests

PR #1005, opened 2026-07-14
Manual weekly update to cargo-rbmt (to 0.5.1)

satsfy:
Supersedes https://github.com/rust-bitcoin/rust-miniscript/pull/1003
Supersedes https://github.com/rust-bitcoin/rust-miniscript/pull/1008

The bump is straight to 0.5.1 to leverage a bug fix in the minor version.

The new [rbmt 0.5.0 changeset](https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/compare/cargo-rbmt-0.4.1...cargo-rbmt-0.5.0) modifies subcommands, breaking two CI jobs:
- `docsrs` became `docs`, which by default is `docs.rs` mode and accepts  `--no-docsrs`.
- `run` now injects `-p <name>` for every workspace member. `descriptor-fuzz` uses `old_miniscript` dep.


satsfy:
Drafted this pr to wait on [0.5.1 benches fix for rbmt](https://github.com/rust-bitcoin/rust-bitcoin/pull/6547#issuecomment-4973885594), which would render a commit in this PR obsolete.

satsfy:
Hey @nyonson, I also had a problem with rbmt 0.5.0 about disk space. So the CI failure on 03891c4 was fixed with b505b70e1990952cfc37216dcec15f71bc001c92, in which I reduced the log volume with [log discarding configs](https://doc.rust-lang.org/cargo/reference/config.html).

Is this correct? Or is there some objection to doing it?

nyonson:
I don't know much about those log settings, but after a brief reading it seems sensible to me. I can't think off the top of my head what changed in `0.5.0` to cause more artifact output. In the past the Github image has also changed out from under us and caused disk space issues.

nyonson:
I believe `CARGO_PROFILE_TEST_DEBUG` inherits from `CARGO_PROFILE_DEV_DEBUG`, so unless setting it to a different value than dev, I don't think it is necessary to set both?

satsfy:
Great point, I kept only the one essential flag. Realized that the problem actually is not related to rbmt 0.5.0, but a pre-existing issue. It induced flaky CI and just happened to affect my previous run. I kept it included on this PR because it is small and targeted.

satsfy:
Force push removed the docrs job as discussed in https://github.com/rust-bitcoin/rust-bitcoin/pull/6547. Now bumping to rbmt 0.5.1.

apoelstra (approved):
ACK de09b4fd4c5fbee31f0190418e8a5b587a968819; successfully ran local tests

PR #1007, opened 2026-07-22
Link ANTLR4 Miniscript grammar

Nuhiat-Arefin:
Adds a README link to the ANTLR4 grammar for Miniscript expressions in the `antlr/grammars-v4` repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,10 @@ We use a custom Rust compiler configuration conditional to guard the bench mark 
 benchmarks use: `RUSTFLAGS='--cfg=bench' cargo +nightly bench benchmarks`.
 
 
+## Other Resources
+
+An [ANTLR4 grammar](https://github.com/antlr/grammars-v4/tree/master/miniscript) for Miniscript expressions is available in the ANTLR `grammars-v4` repository.
+
 ## Release Notes
 
 See [CHANGELOG.md](CHANGELOG.md).


satsfy (commented):
~~Concept NACK~~
EDIT: cancelled the NACK given more recent comments                                             

From my point of view, this doesn't add value for rust-miniscript users. It reads as promotion for the [author's own grammar](https://github.com/antlr/grammars-v4/tree/master/miniscript). The grammar also only covers surface syntax with the type system out of scope, so it's of limited use for real tooling anyway. If anything, it should not be occupying space on the project's main README.md because this has not been a need of rust-miniscript users up until now.

ANTLR also does not support Rust targets. Any use-case I can think of is built on top of rust-miniscript, not within it.

Could you provide some stronger justification for including this antlr grammar?

trevarj:
Concept NACK from me as well. 

We aren't using ANTLR4 for parser generation (or any parser/lexer generator), so I don't really see the correlation.

apoelstra:
I'm gonna say concept ACK.

It's true that this has nothing to do with rust-miniscript, the Rust library. But this library is also the central hub for Miniscript research in general, since the other main implementation is built into Core and not designed to be so extensible.

apoelstra:
We may want to move this eventually to an "other resources" section. And I guess we ought to link Dmitry's Alloy implementation too, which actually found some bugs here.

apoelstra:
@Nuhiat-Arefin can you add an "Other Resources" section, above "Release Notes", and add this there instead of near the top of the README?

Nuhiat-Arefin:
> @Nuhiat-Arefin can you add an "Other Resources" section, above "Release Notes", and add this there instead of near the top of the README?

Done

Nuhiat-Arefin:
> Concept NACK
> 
> From my point of view, this doesn't add value for rust-miniscript users. It reads as promotion for the [author's own grammar](https://github.com/antlr/grammars-v4/tree/master/miniscript). The grammar also only covers surface syntax with the type system out of scope, so it's of limited use for real tooling anyway. If anything, it should not be occupying space on the project's main README.md because this has not been a need of rust-miniscript users up until now.
> 
> ANTLR also does not support Rust targets. Any use-case I can think of is built on top of rust-miniscript, not within it.
> 
> Could you provide some stronger justification for including this antlr grammar?

The main justification I had in mind is grammar-based and differential fuzzing. Random byte level mutations tend to produce malformed expressions that are rejected during shallow syntax checks. A grammar aware generator can preserve enough structure to produce valid or near valid Miniscript expressions, allowing fuzzing to exercise deeper paths in rust-miniscript. It can also help compare syntactic acceptance across independent implementations.

satsfy:
> grammar-based and differential fuzzing [...] grammar aware generator can preserve enough structure to produce valid or near valid Miniscript expressions, allowing fuzzing to exercise deeper paths in rust-miniscript.

More effective fuzzing would be a good outcome. I think the follow-up here would be to implement a PoC for grammar fuzzing. Which tools can do that? How to achieve it in rust-miniscript?

apoelstra (approved):
ACK 1ab21c13bbca3278390a51a08adf3aa84e330604; successfully ran local tests

Nuhiat-Arefin:
> > grammar-based and differential fuzzing [...] grammar aware generator can preserve enough structure to produce valid or near valid Miniscript expressions, allowing fuzzing to exercise deeper paths in rust-miniscript.
> 
> More effective fuzzing would be a good outcome. I think the follow-up here would be to implement a PoC for grammar fuzzing. Which tools can do that? How to achieve it in rust-miniscript?

I am not proposing integrating this grammar into `rust-miniscript`’s fuzzing infrastructure. Independent researchers and other implementations can use tools such as LibAFL Nautilus or Grammarinator to generate structured inputs for fuzzing and differential testing.

I have already conducted grammar-based differential fuzzing with LibAFL Nautilus and reported issues in nBitcoin, embit, and `bitcoinerlab/miniscript`.

The README addition is only intended to make the grammar discoverable as an independent Miniscript resource that the community can maintain.

satsfy (approved):
ACK 1ab21c13bbca3278390a51a08adf3aa84e330604

Moving link to "other resources" and, now knowing the goal of making rust-miniscript a central hub, it tips the scale for this PR so I approve.

If I may ask, feel free to ignore these if they take too much of your time, but the answers would be useful for anyone who wants to fuzz miniscript later:
- What did the fuzzer find? Links to the issues would help.
- How does the grammar get into the fuzzer? Is there a conversion step?
- How do you decide an input is a bug? The parser disagreeing with the grammar is not a bug by itself, since the grammar accepts whitespace and rust-miniscript does not.

PR #1010, opened 2026-07-27
13.x: do not fulfill BIP32 requests outside descriptor path

noahjoeris:
Backport of #1004.

Fixes bypassing descriptor path validation for `GetKey` on sibling paths.


apoelstra (approved):
ACK 651715a58d5f4dd728817a0941deb16de2245520; successfully ran local tests

PR #1013, opened 2026-08-03
Handle empty key derivation paths

### LINKED ISSUE #996: Planner/assets helpers panic on raw no-origin keys with matching fingerprint and non-empty asset path
## Summary

`rust-miniscript`'s public planning/assets helpers can panic on a raw no-origin key when it is checked against an asset entry whose fingerprint matches and whose derivation path is non-empty.

Verified locally on commit `6131095c8ee99ba9a91ee004b1d387dd8418766a`.

This affects the public planning surface around:

* `Assets::from_iter(DescriptorPublicKey)`
* `Descriptor::into_plan(...)`
* Planner key-availability helpers that call `is_key_direct_child_of(...)`

## Minimal reproduction

```rust
use std::str::FromStr;

use miniscript::descriptor::{DefiniteDescriptorKey, DescriptorPublicKey};
use miniscript::plan::Assets;
use miniscript::Descriptor;

fn main() {
    let have = DescriptorPublicKey::from_str(
        "[0d2a9f1a/1]02244fccfacfccf7a6de0b9e7d8a1376acf3b1d3c573b973eaa601fde25c50b956",
    )
    .unwrap();

    let assets = Assets::from_iter([have]);

    let desc = Descriptor::<DefiniteDescriptorKey>::from_str(
        "wsh(pk(02a71967cb7bcd569e6768f393e660d6773013a1099e1352c0487ffd7a4c78c87e))",
    )
    .unwrap();

    println!("{:?}", desc.into_plan(&assets));
}
```

## Observed behavior

Locally, the planning path panics instead of returning “key unavailable”.

In a debug build, the repro panics at `src/plan.rs:535` with:

```text
attempt to subtract with overflow
```

This is not debug-only.

In release builds, `definite_path_len - 1` wraps to `usize::MAX`, and the next slice expression still panics because it tries to evaluate:

```rust
&pk_derivation_path[..usize::MAX]
```

against an empty derivation path.

## Why this seems to happen

Current source has:

```rust
fn is_key_direct_child_of(
    pk: &DefiniteDescriptorKey,
    derivation_path: &bip32::DerivationPath,
) -> bool {
    for pk_derivation_path in pk.full_derivation_paths() {
        if &pk_derivation_path == derivation_path {
            return true;
        }

        let definite_path_len = pk_derivation_path.len();
        if derivation_path.as_ref() == &pk_derivation_path[..(definite_path_len - 1)] {
            return true;
        }
    }

    false
}
```

The reachability gate is real.

`is_key_direct_child_of(...)` is only reached after:

```rust
pk.master_fingerprint() == keysource.0
```

in `Assets::has_ecdsa_key(...)` at `src/plan.rs:544-549`, and in the analogous Taproot helper paths.

For raw no-origin `Single` keys, `DescriptorPublicKey::master_fingerprint()` does not return an all-zero sentinel. In the current `Single` no-origin branch, it computes the fingerprint from the key bytes via `XKeyIdentifier` and takes the first four bytes at `src/descriptor/key.rs:823-835`.

For the raw descriptor key in the repro:

```text
02a71967cb7bcd569e6768f393e660d6773013a1099e1352c0487ffd7a4c78c87e
```

I verified locally that:

```text
hash160(pubkey)[..4] = 0d2a9f1a
```

which exactly matches the asset fingerprint used in:

```text
[0d2a9f1a/1]02244fccfacfccf7a6de0b9e7d8a1376acf3b1d3c573b973eaa601fde25c50b956
```

So the fingerprint equality gate genuinely passes before the panic.

Once that gate passes, the underflow follows from the empty-path case.

For raw single keys without origin:

* `full_derivation_paths()` reports an empty path
* The asset-side public origin syntax can provide a matching fingerprint and a non-empty derivation path

That means the first equality check fails, because `[] != [1]`, and then the second branch evaluates:

```rust
definite_path_len - 1
```

with:

```rust
definite_path_len == 0
```

## Affected helper families

The same helper is used by:

* `has_ecdsa_key(...)`
* `has_taproot_internal_key(...)`
* `has_taproot_script_key(...)`

So the same empty-path panic shape should affect ECDSA and Taproot planning lookups, not just `wsh(pk(...))`.

febyeji:
Fixes rust-bitcoin/rust-miniscript#996.

- Prevent an underflow when the planner handles a raw key with no derivation path.
- Add a regression test based on the issue reproduction, validated with OpenAI Codex and reviewed by the author.

### DIFF
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -532,8 +532,10 @@ fn is_key_direct_child_of(
         }
 
         let definite_path_len = pk_derivation_path.len();
-        if derivation_path.as_ref() == &pk_derivation_path[..(definite_path_len - 1)] {
-            return true;
+        if let Some(parent_path_len) = definite_path_len.checked_sub(1) {
+            if derivation_path.as_ref() == &pk_derivation_path[..parent_path_len] {
+                return true;
+            }
         }
     }
 
@@ -1242,5 +1244,20 @@ mod test {
                 4
             )
         );
+    }
+
+    #[test]
+    fn raw_key_empty_path_does_not_panic() {
+        let asset_key = DescriptorPublicKey::from_str(
+            "[0d2a9f1a/1]02244fccfacfccf7a6de0b9e7d8a1376acf3b1d3c573b973eaa601fde25c50b956",
+        )
+        .unwrap();
+        let assets = Assets::from_iter([asset_key]);
+        let desc = Descriptor::<DefiniteDescriptorKey>::from_str(
+            "wsh(pk(02a71967cb7bcd569e6768f393e660d6773013a1099e1352c0487ffd7a4c78c87e))",
+        )
+        .unwrap();
+
+        assert!(desc.into_plan(&assets).is_err());
     }
 }


apoelstra (approved):
ACK de22d418cc93bcef93fd8fc4e93789bfbff5b8ff; successfully ran local tests; thanks!!

PR #1015, opened 2026-08-06
fix(typecheck): correct d: and hash-fragment witness counting

guggero:
Two of the values that #859 (f57cac0 "typeck: pull satisfaction/
dissatisfaction limits into struct") moved into `SatData` ended up in the
wrong field, so `ExtData` no longer agrees with the witness the satisfier
produces:

* `cast_dupif` used to add `(w + 2, s + 1)` to the (witness size,
  scriptsig size) pair and 1 to the element count. It now adds 1 to the
  witness size and 2 to the element count: the witness size took the
  scriptsig's increment and the count took the witness size's. A `d:`
  satisfaction pushes a single element holding `0x01`, which is two bytes
  in a witness (its length prefix and the byte itself) and one byte in a
  scriptsig (`OP_1`), one element either way. `Placeholder::PushOne`
  sizes that element as two, and `or_i`, whose `<1>` selector is the very
  same element, still adds `(2, 1)`.

* The four hash fragments used to have a `stack_elem_count_dissat` of 1
  and now have a `dissat_data.max_witness_stack_count` of 2, which is the
  value of the `exec_stack_elem_count_dissat` that sat next to it.
  Dissatisfying a hash fragment pushes a single 32-byte element, as the
  33-byte `max_witness_stack_size` beside it already says.

The witness size of `d:` is the consequential one, because it is an
under-estimate and `max_satisfaction_size` feeds `max_weight_to_satisfy`
for every descriptor type: a bare, sh, wsh or tr descriptor whose
miniscript contains a `d:` wrapper under-states the weight of its input by
one weight unit per wrapper, so a fee computed from it lands below the
intended feerate. The two element counts are over-estimates, which only
cost their users precision.

Both are in 13.0.0, 13.0.1 and 13.1.0. This area of the same refactor has
needed one follow-up correction already (f3ea32c, the `thresh` execution
stack count).

Found while differential-testing an independent port of this crate, which
disagreed on the witness size of the expressions whose satisfaction goes
through a `d:` wrapper (242 of ~16k) and on the element count of the hash
dissatisfactions.

### DIFF
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -2206,4 +2206,48 @@ mod tests {
             "The Miniscript corresponding Script cannot be larger than 10000 bytes, but got 10275 bytes."
         );
     }
+    #[test]
+    fn satisfaction_size_matches_template() {
+        // The satisfaction sizes stored in `ExtData` must agree with the
+        // witness template the satisfier produces, whose elements are sized by
+        // `Placeholder::size`.
+        //
+        // A `d:` satisfaction pushes a single `<1>` element, which takes two
+        // bytes on the witness stack (its length prefix plus the byte itself),
+        // and dissatisfying a hash fragment pushes a single 32-byte element.
+        struct HasSigFor(bitcoin::PublicKey);
+
+        impl crate::plan::AssetProvider<bitcoin::PublicKey> for HasSigFor {
+            fn provider_lookup_ecdsa_sig(&self, pk: &bitcoin::PublicKey) -> bool { *pk == self.0 }
+
+            fn check_older(&self, _: bitcoin::relative::LockTime) -> bool { true }
+        }
+
+        let keys = pubkeys(2);
+        let hash = sha256::Hash::hash(&[]);
+
+        for ms_str in [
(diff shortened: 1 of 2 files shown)


apoelstra:
Wow, thanks! This may take a bit of time to review.

apoelstra:
Can you run `cargo +nightly fmt` on this?

apoelstra:
Ok, actually it didn't take me so long. There were really only two fixes and I confirmed that both were regressions I introduced in #859. I did not verify that your test fails before and passes after your fix, but the fixes look correct (and match the previous code).

Thanks for finding this!

We will want to backport to the release-13.x branch.

utACK 0eb97dee559ea78daee2b76a261f9e69ed4f58b5 -- I have a local CI system that will check these but it will probably also fail because of formatting.

guggero:
> Can you run `cargo +nightly fmt` on this?

Done!

Thanks a lot for the review :tada:

apoelstra:
Thanks! Queued for ACK. I also took the time to verify that both your new test cases fail on master.

apoelstra (approved):
ACK 7fac44ca7934ab17f8f78aa1fc47c7e658439735; successfully ran local tests

PR #1016, opened 2026-08-07
Return Taproot compiler errors

### LINKED ISSUE #995: `compile_tr*` panics on recoverable Taproot-construction errors instead of returning normal errors
## Summary

`rust-miniscript`'s Taproot policy compiler currently converts at least two recoverable Taproot-construction failures into panics via internal `expect(...)` calls instead of surfacing normal errors.

Verified locally on commit `6131095c8ee99ba9a91ee004b1d387dd8418766a`.

Confirmed variants locally:

### 1. Uncompressed chosen/internal key

Affected entrypoints:

* `Policy::<bitcoin::PublicKey>::compile_tr(...)`
* `Policy::<bitcoin::PublicKey>::compile_tr_native(...)`
* `Policy::<bitcoin::PublicKey>::compile_to_descriptor(DescriptorCtx::Tr(..))`

These panic with:

```text
compiler produces sane output: ContextError(UncompressedKeysNotAllowed)
```

By contrast:

* `Policy::<bitcoin::PublicKey>::compile_tr_private_experimental(...)`

returns:

```text
Err(ContextError(UncompressedKeysNotAllowed))
```

### 2. Overdeep Huffman TapTree from a skewed root-level disjunction

Affected entrypoints:

* `Policy::compile_tr(...)`
* `Policy::compile_tr_native(...)`
* `Policy::compile_tr_private_experimental(...)`
* `Policy::compile_to_descriptor::<Tap>(DescriptorCtx::Tr(..))`

These panic with:

```text
huffman tree cannot produce depth > 128 given sane weights: TapTreeDepthError
```

## Representative repro 1: uncompressed chosen internal key

```rust
use std::panic::AssertUnwindSafe;
use std::str::FromStr;

use miniscript::bitcoin::PublicKey;
use miniscript::policy::concrete::Policy;

fn main() {
    let uncompressed = PublicKey::from_str(
        "0479be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8",
    )
    .unwrap();

    let policy = Policy::<PublicKey>::Trivial;

    let res = std::panic::catch_unwind(AssertUnwindSafe(|| {
        policy.compile_tr(Some(uncompressed))
    }));

    println!("{res:?}");
}
```

Observed locally:

* `Policy::<PublicKey>::Trivial.compile_tr(Some(uncompressed))` panics with:

```text
compiler produces sane output: ContextError(UncompressedKeysNotAllowed)
```

* `compile_tr_native(Some(uncompressed), ...)` panics with the same error.
* `compile_to_descriptor(DescriptorCtx::Tr(Some(uncompressed)))` panics with the same error.
* The same payload passed to `compile_tr_private_experimental(...)` returns:

```text
Err(ContextError(UncompressedKeysNotAllowed))
```

I also rechecked a useful control locally: a two-branch policy containing one uncompressed key and one compressed key compiled successfully through `compile_tr(...)` and `compile_tr_native(...)`.

That narrows this crash to cases where the chosen internal key itself is uncompressed, rather than any appearance of an uncompressed key anywhere in the policy.

## Representative repro 2: overdeep Huffman TapTree

This public manual policy shape is enough:

* Build a right-comb of `Policy::Or`
* Use weights `(1, 1)` at every `Or`
* Use unique `pk(Ki)` leaves

```rust
use std::sync::Arc;

use miniscript::policy::concrete::Policy;

fn main() {
    let mut leaves = Vec::new();
    for i in 0..131 {
        leaves.push(Policy::Key(format!("K{i}")));
    }

    let policy = build_weighted_or(&leaves);
    let _ = policy.compile_tr(None::<String>);
}

fn build_weighted_or(leaves: &[Policy<String>]) -> Policy<String> {
    if leaves.len() == 2 {
        return Policy::Or(vec![
            (1, Arc::new(leaves[0].clone())),
            (1, Arc::new(leaves[1].clone())),
        ]);
    }

    Policy::Or(vec![
        (1, Arc::new(leaves[0].clone())),
        (1, Arc::new(build_weighted_or(&leaves[1..]))),
    ])
}
```

Observed locally:

* 130 leaves: succeeds
* 131 leaves: panics with:

```text
huffman tree cannot produce depth > 128 given sane weights: TapTreeDepthError
```

I also rechecked this locally with tiny standalone repro binaries on both entry paths:

* Manual `Policy::Or` construction
* Parsed textual concrete policy

Both showed the same threshold, so this is reachable from parsed textual concrete policy as well as manual enum construction.

## Why this seems to happen

There are two separate `expect(...)` sites turning recoverable Taproot-construction failures into panics.

### 1. `Descriptor::new_tr(...)` error is unwrapped

Public `compile_tr(...)` / `compile_tr_native(...)` call `Descriptor::new_tr(...)` and unwrap it with:

```rust
expect("compiler produces sane output")
```

`Descriptor::new_tr(...)` itself returns `Result<_, Error>`.

When the chosen or caller-supplied internal key is uncompressed, the recoverable error:

```text
ContextError(UncompressedKeysNotAllowed)
```

is converted into a panic there.

By contrast, `compile_tr_private_experimental(...)` uses:

```rust
Descriptor::new_tr(...)?
```

which is why the same input returns a normal error on that entrypoint.

### 2. `TapTree::combine(...)` error is unwrapped

The shared `with_huffman_tree(...)` helper does:

```rust
TapTree::combine(...)
    .expect("huffman tree cannot produce depth > 128 given sane weights")
```

But `TapTree::combine(...)` itself is defined as:

```rust
pub fn combine(left: Self, right: Self) -> Result<Self, TapTreeDepthError>
```

So once a skewed Huffman construction would exceed depth 128, that recoverable `TapTreeDepthError` is also converted into a panic.

One source-level asymmetry appears to be that `CompilerError` currently has no variant carrying either `ContextError` or `TapTreeDepthError`, which helps explain why the public `compile_tr(...)` / `compile_tr_native(...)` path currently uses `expect(...)` instead of error propagation.

febyeji:
## Summary
Fix two cases where Taproot policy compilation could panic instead of returning an error. Resolves #995.

## Follow-up

While working on this fix, I noticed that the current Huffman construction can produce a tree deeper than Taproot allows, even though arranging the same leaves differently could produce a valid tree. (regarding `HuffmanTreeDepthExceeded`)

Would it be worth developing a fallback that finds another valid tree when the current Huffman construction exceeds the depth limit of Taptree? 

One option might be [a length-limited Huffman construction](https://doi.org/10.1145/79147.79150), which minimizes weighted path length subject to a maximum depth. (The current construction optimizes weighted path length without maximum-depth constraint.)

I would appreciate feedback on whether this would make sense as a follow-up.

### DIFF
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -57,6 +57,11 @@ pub enum CompilerError {
     /// In a Taproot compilation, no "unspendable key" was provided and no in-policy
     /// key could be used as an internal key.
     NoInternalKey,
+    /// The selected Taproot internal key is uncompressed.
+    UncompressedTaprootInternalKey,
+    /// A Huffman merge during Taproot compilation would place a leaf beyond
+    /// Taproot's maximum tree depth.
+    HuffmanTreeDepthExceeded,
     /// When compiling to Taproot, policy had too many Tapleaves
     TooManyTapleaves {
         /// Number of Tapleaves inferred from the policy.
@@ -88,6 +93,12 @@ impl fmt::Display for CompilerError {
                 "At least one spending path has exceeded the standardness or consensus limits",
             ),
             Self::NoInternalKey => f.write_str("Taproot compilation had no internal key available"),
+            Self::UncompressedTaprootInternalKey => {
+                f.write_str("Taproot compilation selected an uncompressed internal key")
+            }
+            Self::HuffmanTreeDepthExceeded => {
+                f.write_str("Taproot Huffman tree construction would exceed maximum depth 128")
+            }
             Self::TooManyTapleaves { n, max } => {
                 write!(f, "Policy had too many Tapleaves (found {}, maximum {})", n, max)
             }
@@ -116,6 +127,8 @@ impl error::Error for CompilerError {
             | ImpossibleNonMalleableCompilation
             | LimitsExceeded
             | NoInternalKey
+            | UncompressedTaprootInternalKey
+            | HuffmanTreeDepthExceeded
             | TooManyTapleaves { .. }
             | IfFragmentInNativeLeaf { .. } => None,
             PolicyError(e) => Some(e),
--- a/src/policy/concrete.rs
+++ b/src/policy/concrete.rs
@@ -1270,6 +1270,17 @@ mod compiler_tests {
     use super::*;
     use crate::policy::Concrete;
 
+    fn nested_equal_odds_policy(n_keys: usize) -> Policy<String> {
+        assert!(n_keys >= 2);
+
+        let mut source = format!("pk(K{})", n_keys - 1);
+        for i in (0..n_keys - 1).rev() {
+            source = format!("or(1@pk(K{i}),1@{source})");
+        }
+
+        Policy::from_str(&source).expect("generated policy must parse")
+    }
+
     #[test]
     fn test_gen_comb() {
         let policies: Vec<Arc<Concrete<String>>> = vec!["pk(A)", "pk(B)", "pk(C)", "pk(D)"]
@@ -1318,6 +1329,36 @@ mod compiler_tests {
         let desc = policy.compile_tr(None).unwrap();
         // pk(A) promoted to the internal key, leaving the script tree empty
         assert_eq!(desc.to_string(), "tr(A)#xyg3grex");
+    }
+
+    #[test]
+    #[should_panic(expected = "compiler produces sane output")] // will be fixed in next commit
+    fn uncompressed_key_error() {
+        let uncompressed = bitcoin::PublicKey::from_str(
+            "0479be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798\
+             483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8",
+        )
+        .unwrap();
+        let policy = Policy::<bitcoin::PublicKey>::Trivial;
+
+        assert!(policy.compile_tr(Some(uncompressed)).is_err());
+        assert!(policy.compile_tr_native(Some(uncompressed), 128).is_err());
+    }
+
+    #[test]
+    #[should_panic(expected = "huffman tree cannot produce depth > 128 given sane weights")] // will be fixed in next commit
+    fn huffman_depth_error() {
+        // Nesting produces geometrically decreasing Tapleaf probabilities without exceeding the
+        // parser's integer limit.
+        let policy = nested_equal_odds_policy(131);
+        // Selecting an internal key leaves 130 script-path leaves. The weighted Huffman tree
+        // reaches depth 129 and exceeds the TapTree depth limit.
+        assert!(policy.compile_tr(None).is_err());
+        assert!(policy.compile_tr_native(None, 1024).is_err());
+        assert!(matches!(
+            policy.compile_tr_private_experimental(None),
+            Err(Error::TapTreeDepthError(_))
+        ));
     }
 
     #[test]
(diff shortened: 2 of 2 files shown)


apoelstra:
>While working on this fix, I noticed that the current Huffman construction can produce a tree deeper than Taproot allows,

Can you describe an example of this? If your "right-skewed or" which has equal weight for every branch actually triggers this, that is a serious bug in our huffman encoding, since a 128-depth tree does not resemble the optimal huffman encoding at all.

In general, can you split your unit tests into separate commits from the fixes, so that it's easy to cherry-pick the tests onto the original branch (and various fix candidates)?

apoelstra:
It shouldn't even be possible to express a policy where the huffman tree exceeds depth 128 because you'd need to write two weights that differed by a factor 2^128 and we refuse to parse numbers outside of `[1, 2^32-1]`.

febyeji:
Thanks. I was imprecise about the example I had in mind. I was thinking of a nested policy like `or(1@pk(K0), 1@or(1@pk(K1), ...))`, which produces geometrically decreasing Tapleaf probabilities without exceeding the parser’s integer limit. 

```rust
use std::str::FromStr;
use miniscript::policy::Concrete;

fn main() {
    const N_KEYS: usize = 131;

    let mut source = format!("pk(K{})", N_KEYS - 1);
    for i in (0..N_KEYS - 1).rev() {
        source = format!("or(1@pk(K{i}),1@{source})");
    }
    
    let policy = Concrete::<String>::from_str(&source).unwrap();
    let error = policy.compile_tr(None).unwrap_err(); // reaches `expect` and panics in the original code
    
    println!("{error:?}");
}
```

I haven’t reviewed the Huffman implementation in detail, but I don’t suspect an issue with it. My point is that, for a nested policy like this, the resulting Huffman tree can exceed the TapTree depth limit even though the same set of Tapleaves could fit arranged in a more balanced shape.

For this PR, I’ll split the tests into a separate commit from the implementation.

apoelstra:
This PR now has a commit with tests, none of which fail, followed by behavioral changes, none of which are tested. It also contains no example of a policy which can trigger the maximum-depth check and your vague description of one does not make sense to me.

febyeji:
The previous split was intended to keep the tests separate and cherry-pickable, but I agree that it did not clearly show the failing tests and how the fix addressed them. I’ve rewritten the PR as three commits: a test reproducing the panic, the fix, and exact assertions for the new error variants. Please let me know if the change is still unclear.

apoelstra:
Neat! Yes, I see. Because `or(1@A,1@B)` gives weight 1/2 to both branches, you can easily get (1/2)^n for arbitrarily high n without actually typing large numbers anywhere. (Nor do we even represent large numbers in memory anywhere.)

I agree that just returning an error here is the most sensible thing, given that the only callsites are already fallible. It might make more sense to just cap the probabilities for branches below at 1/2^64, say, but it's pretty annoying and complicated to modify the Huffman algorithm to do this.

Very disappointed that our fuzztests never found this.

Can you tweak 3a22596d7192a90f3ab05fa2e98e17a6142bc0dd to add either `#[ignore] // will be fixed in next commit` or `#[should_panic] // will be fixed in next commit` to your failing unit tests? We want `cargo test` to succeed on every commit.

febyeji:
Thanks for the review! I've changed the first commit to mark both tests with `#[should_panic]`.

apoelstra (approved):
ACK 278572aeeb9285dd5a98ba12edf00ec05d6c1aa1; successfully ran local tests

PR #1019, opened 2026-08-11
Wallet Policy (BIP-0388) corrections

trevarj:
After using a powerful (Claude Fable 5) model to review the
wallet_policy module and check it for correctness and API suggestions,
I went back and forth with it over a few iterations to make a bunch of
fixes. These will be better to get in before a release has the
WalletPolicy stuff since there are some breaking API changes.

Commit by commit is the way to review. Also wouldn't mind more LLM
crosschecking, even though I had it review the changes multiple times
after making some tweaks myself.

### DIFF
--- a/src/descriptor/wallet_policy/mod.rs
+++ b/src/descriptor/wallet_policy/mod.rs
@@ -4,7 +4,6 @@ use core::fmt::{self, Display};
 use core::str::FromStr;
 
 use bitcoin::bip32;
-use bitcoin::hashes::{hash160, ripemd160, sha256};
 
 use super::key::{maybe_fmt_master_id, XKeyParseError};
 use super::{DerivPaths, DescriptorKeyParseError, DescriptorMultiXKey, DescriptorXKey, Wildcard};
@@ -182,30 +181,8 @@ impl Translator<KeyExpression> for WalletPolicyTranslator {
         }))
     }
 
-    // Hash terminals: KeyExpression stores hashes as hex `String` (for
-    // template round-tripping), DescriptorPublicKey uses the concrete
-    // `bitcoin::hashes::*::Hash` types. Parse the hex string into the
-    // binary form during materialization.
-
-    fn sha256(&mut self, s: &String) -> Result<sha256::Hash, Self::Error> {
-        s.parse::<sha256::Hash>()
-            .map_err(|_| WalletPolicyError::TranslatorInvalidHashHex("sha256", s.clone()))
-    }
-
-    fn hash256(&mut self, s: &String) -> Result<crate::hash256::Hash, Self::Error> {
-        s.parse::<crate::hash256::Hash>()
-            .map_err(|_| WalletPolicyError::TranslatorInvalidHashHex("hash256", s.clone()))
-    }
-
-    fn ripemd160(&mut self, s: &String) -> Result<ripemd160::Hash, Self::Error> {
-        s.parse::<ripemd160::Hash>()
-            .map_err(|_| WalletPolicyError::TranslatorInvalidHashHex("ripemd160", s.clone()))
-    }
-
-    fn hash160(&mut self, s: &String) -> Result<hash160::Hash, Self::Error> {
-        s.parse::<hash160::Hash>()
-            .map_err(|_| WalletPolicyError::TranslatorInvalidHashHex("hash160", s.clone()))
-    }
+    // Both key types use the concrete `Hash` types for hash terminals.
+    translate_hash_clone!(KeyExpression);
 }
 
 impl Translator<DescriptorPublicKey> for WalletPolicyTranslator {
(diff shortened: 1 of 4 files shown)


src/descriptor/wallet_policy/mod.rs
```diff
@@ -153,7 +231,16 @@ impl WalletPolicy {
     pub fn from_descriptor_unchecked(
         descriptor: &Descriptor<DescriptorPublicKey>,
     ) -> Result<Self, WalletPolicyError> {
-        let mut translator = WalletPolicyTranslator { key_info: descriptor.iter_pk().collect() };
+        // One entry per distinct key, numbered in textual order. Must use
+        // `iter_pk` here; `translate_pk` walks the descriptor right-to-left.
```
apoelstra, line 235: In 40a0272755d96fb0e5df86ea2b2863e362cfe5b1:

Lol. Opened #1022.

src/descriptor/wallet_policy/mod.rs
```diff
@@ -303,6 +314,7 @@ impl WalletPolicy {
         // since @i is only accepted once @0..@i-1 have appeared.
         let mut used: Vec<BTreeSet<_>> = vec![];
         for key in self.template.iter_pk() {
+            check_placeholder_deriv(&key)?;
```
apoelstra, line 317: In adc470e05df1878fb8655f39dd95d2003597cbcc:

This change is fine but this line is modifying a `validate` function which is defined as

```rust
/// Validates the wallet policy template and its key information items.
#[must_use = "Wallet policy won't be considered valid until this is called"]
fn validate(self) -> Result<Self, WalletPolicyError> {
```

This feels like somebody (`git blame` says you :P) was abusing the `must_use` annotation to say "you must call this method on every descriptor". But actually the annotation just means "if you call this, you must use the return value". At the time you described the `validate` function as a `HACK` though you've removed the HACK comment in this PR, in 2ebe7262dd4678f392b7aac364a49ab7609beb71.

This is a private method and it is called in all constructors, I believe, except in `from_descriptor_unchecked`, but in a followup we should refactor things so that it's impossible not to call it:

* remove `from_descriptor_unchecked`
* move `WalletPolicy` into its own private module with only one constructor, `from_descriptor` (which inlines all the `validation` logic), and read-only accessors for the fields
* re-export `WalletPolicy` publicly and implement the existing public API entirely in terms of those two methods

src/descriptor/wallet_policy/mod.rs
```diff
@@ -112,6 +112,19 @@ impl Display for KeyInfo {
     }
 }
 
+/// Checks that the key information items are pairwise distinct: BIP-388 requires
+/// the deserialized public keys to be so. Only the compressed public key is
+/// compared, since two xpubs sharing one are never legitimate.
+fn check_keys_distinct(key_info: &[KeyInfo]) -> Result<(), WalletPolicyError> {
```
apoelstra, line 118: In 699154f5cb7d81b3a2d9cc03f350ebd4cbee7edb:

Note to reviewers: in general this function is obviously insufficient to guarantee no distinct keys. However BIP-388 points this out, and makes it work by restricting the set of allowable key expressions to a small set of derivation paths, all of which have the same length and therefore prevent the sort of "key prefix ambiguity" issues that have plagued Miniscript's attempts to prevent duplicate keys.

I believe in `master` we don't actually enforce the small set of expressions, but two commits after this, in 88e913fdd4120b08fe47da66ac24f4f35ed5a9aa, the extra validation is added.

src/descriptor/wallet_policy/mod.rs
apoelstra: Oh lol, in the next commit 9477a1a830fa4d45cdca60be2664813053230af5 you do remove `from_descriptor_unchecked`. I believe now you can move `WalletPolicy` into a private module which only has `from_descriptor` as a constructor, and you should be able to modify `TryFrom<&str>` to call that instead of directly constructing a `WalletPolicy` then validating it.

src/descriptor/wallet_policy/mod.rs
```diff
@@ -212,22 +213,26 @@ impl Translator<DescriptorPublicKey> for WalletPolicyTranslator {
     type Error = WalletPolicyError;
 
     fn pk(&mut self, pk: &DescriptorPublicKey) -> Result<Self::TargetPk, Self::Error> {
+        // One extraction serves both the index lookup and the placeholder.
+        let (origin, xkey, derivation_paths, wildcard) = match pk {
+            DescriptorPublicKey::XPub(x) => (
+                &x.origin,
+                &x.xkey,
+                DerivPaths::new(vec![x.derivation_path.clone()]).expect("always one path"),
```
apoelstra, line 221: In 9477a1a830fa4d45cdca60be2664813053230af5:

We should add a `DerivePaths::single` constructor so we don't need to `expect` here. Can do in a followup.

apoelstra:
Done reviewing 05b5a983b6ea9b5780aa7f29bea552dd72828490. Queued for ACK. I only made two actionable comments and they're both nits/code cleanliness. If you fix them I'd prefer you do so in a followup or additional commits. This PR was pretty hard to review so I'd rather not re-review a rebase.

trevarj:
> Done reviewing [05b5a98](https://github.com/rust-bitcoin/rust-miniscript/commit/05b5a983b6ea9b5780aa7f29bea552dd72828490). Queued for ACK. I only made two actionable comments and they're both nits/code cleanliness. If you fix them I'd prefer you do so in a followup or additional commits. This PR was pretty hard to review so I'd rather not re-review a rebase.

Thanks! I'll get a follow up going and then also do #1022 since it will affect this area too.

apoelstra (approved):
ACK 05b5a983b6ea9b5780aa7f29bea552dd72828490; successfully ran local tests

portlandhodl:
Post-merge ACK, with a caveat.

This changes the observable invocation order of Translator::pk (and the hash callbacks) from right-to-left to left-to-right. For stateless translators output is bit-identical, but a translator whose result depends on visitation order — e.g. one that assigns keys from a list by counter — will now attach the same keys to different leaves, silently producing a different descriptor/scriptPubKey/address for identical inputs (verified: filling wsh(and_v(v:pk(A),pk(B))) from a fixed key list yields bcrt1qqx5nrr8… before, bcrt1qkr42lc9… after, no error anywhere). After a  search of various wallets including BDK  none would be affected.

Thoughts/Nits 

- Document the left-to-right traversal order on Translator/translate_pk, plus a CHANGELOG entry calling out the behavior change explicitly.

- Add a regression test (recording translator asserting LTR visitation) — currently nothing in the suite fails if this is reverted, since all existing tests are order-agnostic.

PR #1025, opened 2026-08-15
fix(iter): translate_pk to iterate in left-to-right order

### LINKED ISSUE #1022: We should fix `translate_pk` to iterate in left-to-right order
It's kinda surprising that `Miniscript::translate_pk` (and therefore `Descriptor::translate_pk`) iterate through in right-to-left order.

The reason seems to be so that I could write lines like

```rust
Terminal::AndB(..) => {
    Terminal::AndB(translated.pop().unwrap(), translated.pop().unwrap())
}
```

instead of

```rust
Terminal::AndB(..) => {
    let right = translated.pop.unwrap();
    let left = translated.pop.unwrap();
    Terminal::AndB(left, right)
}
```

which hardly seems worth confusing users.

Noticed in https://github.com/rust-bitcoin/rust-miniscript/pull/1019

trevarj:
First commit is the actual conversion.

Second commit with the Ext trait can be dropped, but I didn't want to
ruin the readability of Terminal mappings.

Closes #1022

### DIFF
--- a/src/iter/mod.rs
+++ b/src/iter/mod.rs
@@ -9,13 +9,72 @@
 
 mod tree;
 
+#[cfg(all(not(feature = "std"), not(test)))]
+use alloc::vec;
+#[cfg(any(feature = "std", test))]
+use std::vec;
+
 pub use tree::{
     PostOrderIter, PostOrderIterItem, PreOrderIter, PreOrderIterItem, Tree, TreeLike,
     VerbosePreOrderIter,
 };
 
+use crate::prelude::*;
 use crate::sync::Arc;
-use crate::{Miniscript, MiniscriptKey, ScriptContext, Terminal};
+use crate::{Miniscript, MiniscriptKey, ScriptContext, Terminal, Threshold};
+
+/// Extension trait for the stacks used by algorithms which reconstruct a tree
+/// from a [`TreeLike::post_order_iter`].
+///
+/// Since a post-order iterator yields children left-to-right, the children of
+/// a node appear in that same order at the top of the stack. These methods pop
+/// them off and hand them back in that order, rather than reversed as a bare
+/// sequence of [`Vec::pop`] calls would.
+pub(crate) trait StackExt<T> {
+    /// Pops the top two elements off of the stack and passes them, in
+    /// left-to-right order, to `f`.
+    fn pop2<R>(&mut self, f: impl FnOnce(T, T) -> R) -> R;
+
+    /// Pops the top three elements off of the stack and passes them, in
+    /// left-to-right order, to `f`.
+    fn pop3<R>(&mut self, f: impl FnOnce(T, T, T) -> R) -> R;
+
+    /// Pops the top `n` elements off of the stack, in left-to-right order.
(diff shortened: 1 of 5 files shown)


apoelstra:
In 66c6c9e71909287f19f8231537e846dbf11407e6:

Let's keep `rtl_post_order_iter`. It's part of the API and implementors of `TreeLike` get it for free. (And I use `TreeLike` in a bunch of downstream projects, though I don't know that I ever use the rtl iterator..)

trevarj:
> In [66c6c9e](https://github.com/rust-bitcoin/rust-miniscript/commit/66c6c9e71909287f19f8231537e846dbf11407e6):
> 
> Let's keep `rtl_post_order_iter`. It's part of the API and implementors of `TreeLike` get it for free. (And I use `TreeLike` in a bunch of downstream projects, though I don't know that I ever use the rtl iterator..)

restored it and amended into the first commit.

apoelstra:
In b413b9303b987d6e1dc8bc8929e674fdc71796a2:

I'd suggest squashing these two commits since the second largely overwrites the first. I like the idea of pulling this into a `StackExt` trait. Though all the methods should be `#[inline]` I think.

Then, I don't like the use of `split_off`, which allocates a whole new vector. You should be able to use `Vec::drain` to "directly" turn the end of the stack vector into an iterator.

trevarj:
> In [b413b93](https://github.com/rust-bitcoin/rust-miniscript/commit/b413b9303b987d6e1dc8bc8929e674fdc71796a2):
> 
> I'd suggest squashing these two commits since the second largely overwrites the first. I like the idea of pulling this into a `StackExt` trait. Though all the methods should be `#[inline]` I think.
> 
> Then, I don't like the use of `split_off`, which allocates a whole new vector. You should be able to use `Vec::drain` to "directly" turn the end of the stack vector into an iterator.

roger that. i had it separate in case i needed to drop it off

apoelstra (approved):
ACK ec491c41617dc3481b69d06778c915e87c62f512; successfully ran local tests

portlandhodl:
ACK ec491c41617dc3481b69d06778c915e87c62f512: tests pass. 

Notes: For stateless translators output is bit-identical, but a translator whose result depends on visitation order — e.g. one that assigns keys from a list by counter — will now attach the same keys to different leaves, silently producing a different descriptor/scriptPubKey/address for identical inputs (verified: filling wsh(and_v(v:pk(A),pk(B))) from a fixed key list yields bcrt1qqx5nrr8… before, bcrt1qkr42lc9… after, no error anywhere).  I am not aware of any project that implements a custom Translator, but it might be worth noting.

Nit: I think it would be worth adding a CHANGELOG entry calling out the behavior change explicitly.

portlandhodl (approved):

apoelstra:
Yeah, agreed on the changelog entry. Ok to wait until we actually cut a release. I'll notice this PR title as a "meaningful behavior change" worth calling out, and try to remember to bold it.

I think that if anybody *was* depending on the order, they'd probably have already complained about it. But even most kinds of stateful translators (e.g. ones that are maintaining a counter) are still order independent, I think.

PR #1031, opened 2026-08-22
refactor(wallet_policy): follow up on BIP-388 review

trevarj:
Follow up to #1019 with incorporated feedback.

- add an infallible `DerivPaths::single` constructor
- move `WalletPolicy` and `KeyInfo` into their own private modules
- number key information during left-to-right descriptor translation, now that #1025 is merged
- centralize template and key-information invariant checks in `check_policy`
- validate replacement key information before mutating a policy

### DIFF
--- a/src/descriptor/wallet_policy/policy.rs
+++ b/src/descriptor/wallet_policy/policy.rs
@@ -1,0 +1,275 @@
+// SPDX-License-Identifier: CC0-1.0
+
+use core::fmt::{self, Display};
+use core::str::FromStr;
+
+use bitcoin::bip32;
+
+use super::key_expression::{KeyExpression, KeyIndex};
+use super::{to_key_info, KeyInfo, WalletPolicyError};
+use crate::descriptor::{DerivPaths, DescriptorMultiXKey, Wildcard};
+use crate::{BTreeSet, Descriptor, DescriptorPublicKey, ToString, Translator, Vec};
+
+/// Checks that the key information items are pairwise distinct: BIP-388 requires
+/// the deserialized public keys to be so. Only the compressed public key is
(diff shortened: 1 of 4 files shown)


src/descriptor/wallet_policy/policy.rs
```diff
@@ -0,0 +1,275 @@
...
+        && paths.len() == 2
+        && matches!((step(&paths[0]), step(&paths[1])), (Some(a), Some(b)) if a != b)
+    {
+        Ok(())
+    } else {
+        Err(WalletPolicyError::TemplateValidationInvalidPlaceholderDeriv)
+    }
+}
+
+struct WalletPolicyTranslator {
```
apoelstra: In 4974394b1884d24343443ac81fedb8d6e6f1b1fb:

I would suggest going further than this:

* Make `WalletPolicy::from_template` `pub(super)`
* Similarly add `as_template` and `into_template` methods which are also `pub(super)`

Then you can move `WalletPolicyTranslator`, both your existing public constructors, `n_keys`, and the `Display`/`FromStr`/`TryFrom` impls, all out of policy.rs and up to mod.rs.

This will leave policy.rs very small and easy for a reviewer to determine that every single constructor obeys the invariants. It would have only the `pub(super)` methods, `key_info` and `set_key_info`, and the private `check_*` methods in it.

trevarj: Nice, thanks. done in 28da2fa

trevarj: Looks better when reviewing all commits at once now.

apoelstra (approved):
ACK 28da2fab98cce9e8032e1080652b2d16f1a90a42; successfully ran local tests

PR #1034, opened 2026-08-23
compiler: several optimizations and cleanups

apoelstra:
This is a followup to #988 and starts a project to optimize the compiler. It pulls apart the compiler module into multiple submodules, inlines the `CompilerExtData` struct (which since #988 is just a pair of floats which never change after construction) (previously it also had a `branch_prob` value that would be mutated during compilation).

This also cleans up the cast-computation logic, type checking within the compiler, and eliminates a few other places where we were passing miscellaneous closures around. It leaves the codebase in a much cleaner state.

Some of the commits have benchmark data showing a significant speedup, but I don't have a spare computer right now to run clean before and after benchmarks on the whole branch. Will update the PR with that data once I have it.

It also greatly speeds up the existing regression test between the compiler on master and the one released on 12.x. I ran 150 MM iterations just now, plus a whole bunch when I was developing the branch some months ago.


apoelstra:
On d4a97c6e3521acbd9c44c24e01ca78b0ef6479b8 successfully ran local tests

src/policy/compiler/ext_data.rs
```diff
@@ -0,0 +1,395 @@
...
+        Ok(Self {
+            ms: Self::compose_typeck_only(
+                Terminal::AndB(Arc::clone(&left.ms), Arc::clone(&right.ms)),
+                types::Type::and_b(left.ms.ty, right.ms.ty)?,
+            ),
+            sat_cost: left.sat_cost + right.sat_cost,
+            dissat_cost: match (left.dissat_cost, right.dissat_cost) {
+                (Some(l), Some(r)) => Some(l + r),
+                _ => None,
+            },
```
Abeeujah: This can leverage the `zip` and `map` methods on `Option<T>`
```suggestion
            dissat_cost: left.dissat_cost.zip(right.dissat_cost).map(|(l, r)| l + r),
```

src/policy/compiler/ext_data.rs
```diff
@@ -0,0 +1,395 @@
...
+                } else {
+                    Some(2.0 + ldis)
+                }
+            } else if let Some(ldis) = left.dissat_cost {
+                Some(2.0 + ldis)
+            } else if let Some(rdis) = right.dissat_cost {
+                Some(1.0 + rdis)
+            } else {
+                None
+            },
```
Abeeujah: > // Complex if/let is better as is.

Unsure what the argument is, but this whole if/let can be simplified using `match`
```suggestion
            dissat_cost: match (
                left.dissat_cost.map(|l| 2.0 + l),
                right.dissat_cost.map(|r| 1.0 + r),
            ) {
                (Some(l), Some(r)) => Some(l.min(r)),
                (left, right) => left.or(right),
            },
```

apoelstra: Nice! Clippy wanted me to turn it into a flat series of `} else if {`s.

apoelstra:
Addressed @Abeeujah's commenst.

apoelstra:
On 8fd590e0aabca52eeab0881afe15d83984cfbf73 successfully ran local tests

apoelstra:
8fd590e0aabca52eeab0881afe15d83984cfbf73 needs rebase

apoelstra:
On 482b0a13abc83c540b7ea1809b309cf8990ff600 successfully ran local tests

apoelstra:
482b0a13abc83c540b7ea1809b309cf8990ff600 needs rebase

PR #1036, opened 2026-08-24
fix(miniscript): compare thresh k and tree size in Terminal Eq/Ord

portlandhodl:
`Terminal::eq` never compared the `k` threshold or child count of `Thresh` fragments (they fell into the discriminant-only catch-all arm), and both `Terminal::eq` and `Terminal::cmp` zipped the two pre-order iterators without checking that they had the same length. As a result structurally distinct miniscripts could compare equal, and could order `Ordering::Equal`:

```
thresh(2,pk(A),s:pk(B)) == thresh(1,pk(A),s:pk(B))
thresh(2,pk(A),s:pk(B)) == thresh(2,pk(A),s:pk(B),s:pk(C))
```

while hashing differently (`Hash` includes `k` and the child count), violating the Eq/Hash and Eq/Ord contracts for `Miniscript`, `Descriptor` and `TapTree`, which all delegate to these impls, and corrupting sorted and hashed collections of descriptors in downstream users.

**Fix:** compare the number of pre-order nodes up front in both impls (a tree now orders before any tree it is a strict pre-order prefix of), and compare the `Thresh` threshold `k` and child count in `eq`.

**Tests:** regression tests cover `thresh` differing in `k`/child count, mutual Eq/Ord/Hash consistency (including antisymmetry and cross-instance equality) over a list of structurally distinct scripts, survival of distinct scripts in `BTreeSet`/`HashSet`, and descriptor-level equality/sort+dedup. All fail without the fix.

### DIFF
--- a/src/miniscript/decode.rs
+++ b/src/miniscript/decode.rs
@@ -232,6 +232,11 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> PartialEq for Terminal<Pk, Ctx> {
                 (Self::Hash256(h1), Self::Hash256(h2)) if h1 != h2 => return false,
                 (Self::Ripemd160(h1), Self::Ripemd160(h2)) if h1 != h2 => return false,
                 (Self::Hash160(h1), Self::Hash160(h2)) if h1 != h2 => return false,
+                (Self::Thresh(th1), Self::Thresh(th2))
+                    if th1.k() != th2.k() || th1.n() != th2.n() =>
+                {
+                    return false
+                }
                 (Self::Multi(th1), Self::Multi(th2)) if th1 != th2 => return false,
                 (Self::SortedMulti(th1), Self::SortedMulti(th2)) if th1 != th2 => return false,
                 (Self::MultiA(th1), Self::MultiA(th2)) if th1 != th2 => return false,
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -1264,6 +1264,25 @@ mod tests {
     }
 
     #[test]
+    fn descriptor_eq_distinguishes_thresh_policies() {
+        let two_of_two = Descriptor::<String>::from_str("wsh(thresh(2,pk(A),s:pk(B)))").unwrap();
+        let one_of_two = Descriptor::<String>::from_str("wsh(thresh(1,pk(A),s:pk(B)))").unwrap();
+        let two_of_three =
+            Descriptor::<String>::from_str("wsh(thresh(2,pk(A),s:pk(B),s:pk(C)))").unwrap();
+
+        assert_ne!(two_of_two, one_of_two);
+        assert_ne!(two_of_two, two_of_three);
+        assert_ne!(two_of_two.cmp(&one_of_two), core::cmp::Ordering::Equal);
+        assert_ne!(two_of_two.cmp(&two_of_three), core::cmp::Ordering::Equal);
+        assert_eq!(two_of_two, two_of_two.clone());
+
+        let mut sorted = vec![two_of_two, one_of_two, two_of_three];
+        sorted.sort();
+        sorted.dedup();
+        assert_eq!(sorted.len(), 3);
+    }
+
+    #[test]
     fn desc_rtt_tests() {
         roundtrip_descriptor("c:pk_k()");
         roundtrip_descriptor("wsh(pk())");
--- a/src/miniscript/display.rs
+++ b/src/miniscript/display.rs
@@ -331,7 +331,18 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Ord for Terminal<Pk, Ctx> {
             cmp::Ordering::Less => cmp::Ordering::Less,
             cmp::Ordering::Greater => cmp::Ordering::Greater,
             cmp::Ordering::Equal => {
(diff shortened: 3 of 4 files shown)


src/miniscript/decode.rs
```diff
@@ -220,7 +220,22 @@ impl<Pk: MiniscriptKey, Ctx: ScriptContext> Clone for Terminal<Pk, Ctx> {
...
+    /// `Thresh`. Its children follow it in the iteration and are compared
+    /// by the iterator itself, but its threshold `k` and child count `n`
+    /// must be compared explicitly when its node is visited.
+    ///
+    /// Because `zip` stops at the end of the shorter iterator, we first
+    /// check that both trees yield the same number of nodes. Otherwise a
+    /// tree would compare equal to any other tree that has it as a
+    /// pre-order prefix, e.g. `thresh(2,A,B)` and `thresh(2,A,B,C)`.
     fn eq(&self, other: &Self) -> bool {
+        if self.pre_order_iter().count() != other.pre_order_iter().count() {
```
apoelstra: In a187133c0646bca5194aa6e261485a3fbfaf9fad:

This is super inefficient. Both these iterators are kinda costly, allocate stacks and do a bunch of accounting etc. I also don't believe it's possible for this check to fail without one of the below checks failing, since we are checking `thresh.n()` and every other combinator has a fixed number of children.

So we should drop this check.

apoelstra:
In f3417ad7ed3b170ffd6263f88b5dd6532773c519:

Normally I'd complain about all this algorithmic detail in a doccomment, but in a docomment on `Eq` it sorta makes sense that anybody reading it is curious about how equality is implemented.

src/miniscript/decode.rs
portlandhodl: Ack, dropped; 100% the .count() comparison was redundant. Good catch.

portlandhodl:
> In [f3417ad](https://github.com/rust-bitcoin/rust-miniscript/commit/f3417ad7ed3b170ffd6263f88b5dd6532773c519):
> 
> Normally I'd complain about all this algorithmic detail in a doccomment, but in a docomment on `Eq` it sorta makes sense that anybody reading it is curious about how equality is implemented.

Thought: given LLMs could easily generate the same docs, want me to just remove to keep consistency?

apoelstra:
>Thought: given LLMs could easily generate the same docs, want me to just remove to keep consistency?

Yeah, I'd appreciate it. Good point that anybody curious can just ask a LLM how the comparison works (and ask whatever their "real" question is, e.g. how does it handle key equality checks or whether it uses recursion).

portlandhodl:
> > Thought: given LLMs could easily generate the same docs, want me to just remove to keep consistency?
> 
> Yeah, I'd appreciate it. Good point that anybody curious can just ask a LLM how the comparison works (and ask whatever their "real" question is, e.g. how does it handle key equality checks or whether it uses recursion).

a9e0479c310aed5f540d7ff2a619a6f096ff41b3: Complete

portlandhodl:
I did some performance analysis on my host regarding the performance characteristics of this pull. This test harness was created outside this PR but should give an idea if any performance regressions would be considered unacceptable. Under normal use nothing sticks out as unacceptable in my opinion.

### Performance impact

Micro-benchmark: `eq`/`cmp` on a 63-node balanced `thresh` miniscript (~2.5 KB script — near the wsh standardness limit), baseline `d2ec504` vs this PR, both compiled into the same release+LTO binary, best-of-9 × 2000 iterations. 

**Benchmark environment**

| | |
|---|---|
| CPU | AMD Ryzen 5 7600X (6 cores / 12 threads) |
| RAM | 32 GB |
| OS | Ubuntu 26.04 LTS, Linux 7.0.0-29-generic, x86_64 |
| Toolchain | rustc 1.95.0 (59807616e 2026-04-14) |

| Operation | Before | After | Change | Notes |
|---|---:|---:|---:|---|
| `eq` — identical trees | 2368 ns | 2482 ns | ~+4% | noise band; new `thresh` arm is two O(1) integer compares |
| `eq` — differ in `thresh` `k` | 2344 ns | 66 ns | **~35× faster** | now exits at the first differing node (old code walked the whole tree *and* returned the wrong answer) |
| `eq` — no-`thresh` control | 2115 ns | 2239 ns | ~noise | new match arm never hit |
| `cmp` — identical trees | 4380 ns | 7320 ns | +67% | two extra pre-order counting passes |
| `cmp` — differ at first leaf | 298 ns | 3209 ns | ~11× | O(distance) → O(n), but only **+2.9 µs** absolute |
| `cmp` — differ at root fragment name | 6 ns | 6 ns | none | fast path unchanged |
| `hash` control (unchanged code) | 1688 ns | 1683 ns | ~0% | harness stable |
| BTreeSet stress: 256 inserts of near-identical scripts | 2.80 ms | 4.81 ms | +72% | adversarial worst case, pure `cmp` cost |

apoelstra:
Lol at the benchmarks showing massive speedups in cases where we now correctly notice that we're comparing different trees.

src/miniscript/decode.rs
apoelstra: In 2476148d9c6b8ffb29a50ee4d3444104cbd2cd96:

There's another node count added in `Terminal::cmp` which is redundant.

PR #1037, opened 2026-08-26
types: make Malleability an enum

guggero:
This is an alternative approach to https://github.com/rust-bitcoin/rust-miniscript/pull/1032, following the suggestion here: https://github.com/rust-bitcoin/rust-miniscript/pull/1032#issuecomment-5380985927.

With this PR we change the `types::Malleability` type to erase the `f/e/s` properties if something is malleable, by turning it into an enum.

The [BIP will be updated](https://github.com/bitcoin/bips/pull/2267) to match this change. And, assuming there is agreement on the approach, #1032 would be closed.


src/miniscript/mod.rs
```diff
@@ -1345,8 +1345,12 @@ mod tests {
         match (ms, valid) {
             (Ok(ms), true) => {
                 assert_eq!(format!("{:x}", ms.encode()), expected_hex);
-                assert_eq!(ms.ty.mall.non_malleable, non_mal);
-                assert_eq!(ms.ty.mall.signed, need_sig);
+                assert_eq!(ms.is_non_malleable(), non_mal);
+                // The "s" property is only reported for a non-malleable
+                // expression, so the expected value only applies to those.
+                if non_mal {
+                    assert_eq!(ms.requires_sig(), need_sig);
```
apoelstra: In c7a495d36721090530c7745c413cdc36c468c9ef:

Rather than gating the check on `non_mal`, I think we should change the assertion to `non_mal && need_sig`. And we should promise that this method returns false for malleable script.

apoelstra: Ah, your third commit adds such a test. We should do the same here.

src/miniscript/analyzable.rs
```diff
@@ -9,11 +9,17 @@ use crate::prelude::*;
...
-    /// Whether all spend paths of miniscript require a signature
-    pub fn requires_sig(&self) -> bool { self.ty.mall.signed }
+    /// Whether all spend paths of miniscript require a signature.
+    ///
+    /// Returns `false` for a malleable miniscript. The "s" property only
+    /// describes the satisfactions of an expression that is non-malleable to
+    /// begin with, so nothing is known for a malleable one, and `false` is the
+    /// assumption that keeps a caller from relying on a signature being
+    /// required.
+    pub fn requires_sig(&self) -> bool { self.ty.mall.signed().unwrap_or(false) }
```
apoelstra: In 53993fc947323348859c90cb6de4c06704dc1b3f:

In addition to updating the docs we should also deprecate this and rename it to `non_malleable_and_requires_sig`.

We could also implement it as `matches!(self.ty.mall, Malleability:::NonMalleable { signed: true, .. })` I think.  I dunno if that's any clearer than what you've written. Just suggesting it in case you like it.

src/miniscript/mod.rs
guggero: Yeah, that makes sense, combined with your other comment. Though I moved this update of the assertion to the second commit (now a37a1605) where it makes a bit more sense.
I kept the additional test as its own commit, but I can fold it into a37a1605 as well if you prefer that.

src/miniscript/analyzable.rs
guggero: Agree, I've done that in a new refactor commit (4817382f).

guggero:
Thanks a lot for the review, I've addressed your comments!

src/miniscript/analyzable.rs
```diff
@@ -10,18 +10,24 @@ use crate::prelude::*;
...
     /// Returns `false` for a malleable miniscript. The "s" property only
     /// describes the satisfactions of an expression that is non-malleable to
     /// begin with, so nothing is known for a malleable one, and `false` is the
     /// assumption that keeps a caller from relying on a signature being
     /// required.
-    pub fn requires_sig(&self) -> bool {
+    #[deprecated(
+        since = "TBD",
+        note = "check the `signed` property of the malleability type instead"
+    )]
```
apoelstra, line 24: In 4817382fc202d44e0447a031ff6af6b7fead1d5b:

This deprecation is in the wrong place and its message makes no sense. We should deprecate the `requires_sig` method, telling users to use `non_malleable_and_requires_sig` instead. And we should not deprecate the new method.

Did Claude do this (introducing a new method and immediately deprecating it)?

guggero: Hmm, I might have misunderstood your [earlier comment](https://github.com/rust-bitcoin/rust-miniscript/pull/1037#discussion_r3889775276).

The `requires_sig` method did exist before this PR. And your comment stated: "In addition to updating the docs we should also deprecate this and rename it to `non_malleable_and_requires_sig`.".
Which is what I did (with Claude, yes).

But it sounds like you might have meant something like "introduce a new, fixed method called `non_malleable_and_requires_sig`"?

And yes, the deprecation comment definitely isn't my best work, happy to use a better wording if you have a suggestion.

apoelstra:
Overall this looks good (other than the deprecation thing). ChatGPT 5.6 Sol points out that this slightly messes up our satisfaction algorithm.

In particular, we pass `root_has_sig` as a top-level property into the satisfier, which is supposed to mean "all branches require a signature to be satisfier". We use this to change how failed timelocks work.

If there *is* a signature, then we return `Witness::Impossible` since the signature constrains the timelock values and if they're not high enough, they're not high enough. If there is *not* a signature, we return `Witness::Unavailable` meaning that we can't satisfy the branch, but some 3rd party maybe could, by messing up the timelock values in the transaction.

This was already a crude check that would sometimes report timelock-checks as malleable when they weren't, and now it's even cruder, but I guess it's worth noting. We should maybe file a bug to "fix" the satisfier to keep better track of whether it's included a not-3rd-party-removable signature anywhere, and use *that* rather than this `root_has_sig` boolean. Since #895 this should be much easier than it would've been....though it may still require backtracking. (See the `regression_895` test for a different example of timelock crap causing wrong behavior that would require backtracking to fix.) (And I really don't want to add backtracking since it will turn a linear algorithm into a worst-case exponential one.)

I would suggest that, at least, we add a regression test to this PR, and call it `regression_1037`, which demonstrates the "wrong" timelock-malleability behavior.

guggero:
I've pushed up commit fab8f9b6 that adds the requested `regression_1037` test. And I created https://github.com/rust-bitcoin/rust-miniscript/issues/1048 to track that as well. I also switched from Fable to Astra, hoping that produces more usable results and fewer "claudisms" in the generated texts...

About the deprecation note, I think there might be a misunderstanding. Can you please take a [look at my latest inline comment](https://github.com/rust-bitcoin/rust-miniscript/pull/1037#discussion_r3935280149)?

src/miniscript/analyzable.rs
apoelstra: Yes, "rename and deprecate" means "introduce a new method with the new name, make the old one pass through to it, and deprecate the old one".

And the message refers to the `signed` property of the `Malleability` type but this PR removes that property (or rather, makes the type an enum where only one variant has the property).

guggero: Okay, my bad, sorry. I've done that now.

PR #1038, opened 2026-08-27
positive_f64: saturate arithmetic at MIN_POSITIVE, define inf/inf as 1

### LINKED ISSUE #1033: `PositiveF64` needs better bounds-checking for extreme values
In #988 we introduced the `PostiveF64` type which guarantees that it represents a positive `f64`, i.e. it will not hold 0.0, any negative finite number, `-inf` or `NaN`. It implements `Eq` although its `PartialEq` implementation passes through to `f64`, which is permissible only if we avoid `NaN`s.

However:

* You can construct *positive* infinity directly with `PositiveF64::new`, which by itself is fine, but then you can divide this by itself to get `NaN`.
* Or you can divide 1.0 by this to get 0.0, I think.
* By using the sort of deeply-nested-`or` policy from #1016 you can get arbitrarily small `PositiveF64` values, and eventually you can underflow it to 0.0

These extreme values require you to use a crazy policy and/or directly call the API with extreme values. So I think it'd be fine if internally we:

* Limited the `f64` below by `f64::MIN_POSITIVE` and documented that if you get this value then attempt to keep dividing it by stuff, the division will be a no-op.
* Defined `inf / inf` to be 1.0
* Defined `<finite> / inf` to be `f64::MIN_POSITIVE`

Alternately, we could avoid the `inf` special cases by clamping the value *above*, though this seems annoying to do and hard to do consistently.

portlandhodl:
PositiveF64 guarantees that it holds a positive, non-NaN f64, but its
arithmetic could escape this invariant at the extremes: inf / inf
produced NaN (silently breaking the Eq contract), a finite value
divided by inf produced 0.0, and repeated division (e.g. through
deeply-nested-or policies) could underflow a probability to 0.0.

Harden the arithmetic as described in issue #1033:

- Multiplication and division results that would underflow to a
  subnormal number or to 0.0 now saturate at f64::MIN_POSITIVE, exposed
  as PositiveF64::MIN_POSITIVE. Once this value is reached, further
  division (or multiplication by values less than one) is a no-op.
- inf / inf is defined to be 1.0, making NaN unreachable and the Eq
  implementation sound.
- A finite value divided by inf underflows to 0.0 and therefore
  saturates at MIN_POSITIVE like any other underflow.
- NormalizedIterator uses the same hardened division, closing the
  compiler-side underflow path.

Document the invariant and saturation semantics on the type (replacing
the stale "Ordered f64 for comparison" doc line left over from OrdF64)
and add targeted unit tests for the extreme-value behavior, including
the reversed-iteration path the compiler uses.

Fixes #1033

### DIFF
--- a/src/primitives/positive_f64.rs
+++ b/src/primitives/positive_f64.rs
@@ -7,13 +7,35 @@ use core::{cmp, hash, ops};
 
 use crate::Threshold;
 
-/// Ordered f64 for comparison.
+/// A positive floating-point number.
+///
+/// This type guarantees that the contained value is positive: it will never
+/// hold 0.0, a negative number, `-inf` or `NaN`. (Positive infinity *is* a
+/// permissible value.) This guarantee makes it safe to implement [`Eq`], even
+/// though the underlying [`PartialEq`] implementation passes through to `f64`.
+///
+/// To uphold the guarantee, arithmetic on this type saturates below at
+/// [`PositiveF64::MIN_POSITIVE`]: any operation whose result would underflow
+/// to a subnormal number or to 0.0 instead yields [`PositiveF64::MIN_POSITIVE`].
+/// This means that once you obtain [`PositiveF64::MIN_POSITIVE`], dividing it
+/// further (or multiplying it by values less than one) is a no-op.
+///
+/// Division involving infinity is also defined so as to avoid `NaN` and 0.0:
+/// `inf / inf` is defined to be 1.0, and `<finite> / inf` underflows to 0.0
+/// and therefore yields [`PositiveF64::MIN_POSITIVE`].
 #[derive(Copy, Clone, PartialEq, Debug)]
 pub struct PositiveF64(f64);
 
 impl PositiveF64 {
     /// The constant one.
     pub const ONE: Self = Self(1.0);
+
+    /// The smallest value that arithmetic on [`PositiveF64`] can produce.
+    ///
+    /// This is the smallest positive normal `f64`. Operations whose results
+    /// would underflow below this value saturate here instead; see the type
+    /// documentation for more detail.
+    pub const MIN_POSITIVE: Self = Self(f64::MIN_POSITIVE);
 
     /// Constant used in unit tsets
     #[cfg(test)]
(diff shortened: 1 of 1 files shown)


apoelstra:
In 695c2f834b3366c3d2b4ac95d0aaebc43b7c6759:

I'm glad that you tested the behavior of `NormalizedIterator` when infinities are present. Currently its behavior is:

* if one infinitity is present, it becomes 1.0 and everything else becomes `MIN_POSITIVE` (correct)
* if multiple infinities are present, they *all* become 1.0 and everything else becomes `MIN_POSITIVE` (not correct)

I think we should just document that the result is wrong if you have multiple infinities. We never hit this case in our own usage, and I think the 'correct' way to implement it is to add a bunch of extra logic where we do an additional pass to count all the infinities, then yield `1 / <# infinities>` for each infinity, and it'll be slow and complicated to review for no benefit to this crate.

src/primitives/positive_f64.rs
```diff
@@ -103,6 +125,40 @@ impl From<NonZeroU32> for PositiveF64 {
...
+/// however, underflow to a subnormal number or to 0.0; in this case we
+/// saturate at `f64::MIN_POSITIVE`.
+fn clamp_positive(f: f64) -> f64 {
+    // Note that this comparison is false for NaN, so NaNs would pass through
+    // unchanged. As described above, NaN is unreachable here.
+    if f < f64::MIN_POSITIVE {
+        f64::MIN_POSITIVE
+    } else {
+        f
+    }
```
Abeeujah: This looks like an [f64::max](https://doc.rust-lang.org/std/primitive.f64.html#method.max) operation, and it upholds the invariants here
```suggestion
    f.max(f64::MIN_POSITIVE)
```

apoelstra: Agreed, though we need to keep the comment saying that `f` being `NaN` is impossible here. (Though if we *did* get a NaN, `f64::max` would turn it into `f64::MIN_POSITIVE` which I guess is an improvement.)

apoelstra (approved):
ACK d0beebff8db18641bbb7b73e3ef7d014f3f8aaaa; successfully ran local tests

PR #1039, opened 2026-08-31
13.x: descriptor: fix taproot tree formatting

apoelstra:
Commit 552e8440 ("tr: flatten TapTree type") changed TapTree from a recursive enum into a depth-preorder leaf list. The formatter only tracked depth changes between adjacent leaves, which loses binary subtree boundaries for some unbalanced trees and can render descriptors that do not parse back.

Format the flat representation with an explicit child-count stack so completed binary subtrees are closed at the right time.

Backport of #953. I literally just `jj duplicate`d the commits without looking at them. May warrant re-review.

### DIFF
--- a/src/descriptor/tr/taptree.rs
+++ b/src/descriptor/tr/taptree.rs
@@ -89,27 +89,34 @@ fn fmt_helper<Pk: MiniscriptKey>(
     f: &mut fmt::Formatter,
     mut fmt_ms: impl FnMut(&mut fmt::Formatter, &Miniscript<Pk, Tap>) -> fmt::Result,
 ) -> fmt::Result {
-    let mut last_depth = 0;
+    let mut child_counts = Vec::<u8>::with_capacity(TAPROOT_CONTROL_MAX_NODE_COUNT);
+
     for item in view.leaves() {
-        if last_depth > 0 {
+        let depth = usize::from(item.depth());
+
+        if !child_counts.is_empty() {
             f.write_str(",")?;
         }
 
-        while last_depth < item.depth() {
+        while child_counts.len() < depth {
             f.write_str("{")?;
-            last_depth += 1;
+            child_counts.push(0);
         }
+
         fmt_ms(f, item.miniscript())?;
-        while last_depth > item.depth() {
+
+        if let Some(child_count) = child_counts.last_mut() {
+            *child_count += 1;
+        }
+        while let Some(2) = child_counts.last() {
             f.write_str("}")?;
-            last_depth -= 1;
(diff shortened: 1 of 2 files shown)


apoelstra:
lol `jj duplicated` made separate duplicates of all three commits. One sec..

apoelstra:
Never mind, I just put the bookmark on the wrong commit because the `jj duplicate` order was the opposite of what I expected. Should be good now.

src/descriptor/tr/taptree.rs
```diff
@@ -282,5 +289,33 @@ impl<Pk: MiniscriptKey> TapTreeBuilder<Pk> {
     }
 
     #[inline]
-    pub(super) fn finalize(self) -> TapTree<Pk> { TapTree { depths_leaves: self.depths_leaves } }
+    pub(super) fn finalize(self) -> TapTree<Pk> {
+        assert!(!self.depths_leaves.is_empty(), "TapTreeBuilder must not finalize an empty tree");
```
portlandhodl, line 293: There is no possible way to hit this condition today as the 

TapTreeBuilder: only created after root.verify_toplevel("tr", 1..=2) guarantees a second child exists, so tap_tree is a real node then every path through the loop either pushes a leaf or aborts before finalize.

So why?

portlandhodl (approved):

src/descriptor/tr/taptree.rs
apoelstra: See https://github.com/rust-bitcoin/rust-miniscript/pull/953#issuecomment-4512267086

This came out of a **long** fight with some bot that opened the original PR.

apoelstra: Previously the code was doing all sorts of stupid contortions to handle the impossible case. I pointed out that it is impossible, but only because our private constructors are called correctly, and suggested adding an assertion that this is true.

portlandhodl: Heh, ACK

portlandhodl (approved):

apoelstra:
@portlandhodl FYI unless you type "ACK <commit ID>" where `<commit ID>` is (at least 8 characters of) the tip commit's ID, I can't do anything with your approvals

It doesn't matter whether or not you use the Github "approve" workflow or just post a comment.

portlandhodl:
ACK: bb5004f6fed9bdf419b6203fc3870eb3d623ff9e

apoelstra:
On bb5004f6fed9bdf419b6203fc3870eb3d623ff9e successfully ran local tests

PR #1040, opened 2026-09-01
fix(miniscript): count exactly k satisfactions in thresh max-size bounds

portlandhodl:
A `thresh(k, ...)` witness satisfies exactly k children: the script checks
the running sum against k with `OP_EQUAL`, so any witness with a different
number of satisfied children is invalid. The maximum-satisfaction
computation in `ExtData::threshold` nevertheless selected k+1 children as
satisfied (`i <= k` instead of `i < k`), e.g. computing the maximum witness
size of `thresh(1,pk(A),s:pk(B))` as 146 bytes (two signatures) instead of
74 (one signature plus one empty dissatisfaction).

This inflates `Miniscript::max_satisfaction_weight()` and everything
derived from it, making fee/weight estimates overestimate by up to one
largest-child satisfaction per thresh node.

Found by differential testing against Bitcoin Core's miniscript
implementation (`Node::GetWitnessSize`/`GetStackSize` agree with the fixed
values, and with the reference implementation's satisfaction table,
which permits exactly k satisfactions).

### DIFF
--- a/src/miniscript/mod.rs
+++ b/src/miniscript/mod.rs
@@ -2113,6 +2113,23 @@ mod tests {
     }
 
     #[test]
+    fn test_thresh_sat_data_exact_k_satisfactions() {
+        let ms =
+            Miniscript::<String, Segwitv0>::from_str_insane("thresh(1,pk(A),s:pk(B))").unwrap();
+        let sat = ms.ext.sat_data.expect("thresh(1,...) is satisfiable");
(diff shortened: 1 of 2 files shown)


apoelstra:
Nice! This is a pretty subtle difference in outputs.

apoelstra (approved):
ACK d7801d1405849e1b1eb062ef4e6f1334e253d8c1; successfully ran local tests

PR #1041, opened 2026-09-02
WIP: Demonstrate usage of the bitcoin-units compat layer

satsfy:
Built on top of [units 1.0 compat layer](https://github.com/rust-bitcoin/rust-bitcoin/pull/6259). Mirrors the [rust-psbt](https://git.rust-bitcoin.org/rust-bitcoin/rust-psbt/pulls/195) and [corepc](https://git.rust-bitcoin.org/rust-bitcoin/corepc/pulls/672) demos.

Done against version 13.x, so that I can test on BDK which is in 13.x. 

Changes:
- Expected MSRV bump to 1.74.0 (to be reduced later back to 1.63), forced by units 1.0. Cannot be feature gated bcs a public field type would break other crates in the graph.
- Patch bitcoin to the 0.32.101 compat layer, stable types come from `bitcoin::compat`.
- One commit per migrated type for Sequence, Weight, absolute and relative LockTime.
- Stable types in the public API. No conversation paths presented so that user is forces into stable types.
- bitcoind-tests and embedded need their own `[patch.crates-io]` or they resolve bitcoin 0.32.6 (todo: nothing in them was migrated here, awaiting primitives).
- A lint rule for each type prevents regression, forcing user to put lint exception.

I noticed that `TxIn`, `TxOut` and `Transaction` (from primitives) are responsible for all lint exceptions, because they are using the unstable types, awaiting primitives compat layer.


apoelstra:
FWIW from miniscript 14 onward we should just bump the MSRV.

satsfy:
@tcharding just pinging

src/psbt/finalizer.rs
```diff
@@ -485,7 +486,7 @@ mod tests {
     fn finalize_skips_already_finalized_input() {
         let tx = Transaction {
             version: transaction::Version::ONE,
-            lock_time: absolute::LockTime::ZERO,
+            lock_time: bitcoin::absolute::LockTime::ZERO,
```
tcharding, line 489: Why introduce `SequenceUnstable` but use `bitcoin::absolute::LockTime`?

This is @apoelstra's place but FWIW I'm currently leaning towards preference for the latter (ie using path to distinguish and not use an alias). Whichever we choose we should be uniform.

src/plan.rs
```diff
@@ -877,7 +876,7 @@ mod test {
             (
                 vec![],
                 vec![],
-                Some(Sequence(1000).to_relative_lock_time().unwrap()),
+                Some(relative::LockTime::from_consensus(1000).unwrap()),
```
tcharding, line 879: So this change is interesting. How you did it in the other place (changing to `bitcoin::Sequence` confused me, so I guess this is the clearer way. Not sure though.

src/psbt/finalizer.rs
apoelstra: `Sequence` is part of a Bitcoin transaction and there is only one kind.

vs `LockTime` where there are two kinds, absolute and relative, so I prefer using one layer of path.

The `bitcoin::` part I have no opinion about.

src/plan.rs
apoelstra: It was a little alarming to see the direct `Sequence(1000)` constructor which would bypass all invariants (though for `Sequence`, unlike the locktime types, I believe there are no invariants). So this is definitely clearer.

I'm not sure what "the other place" you're referring to is.

PR #1044, opened 2026-09-04
compiler: reject policy keys invalid in the target script context

portlandhodl:
Supersedes and closes #761.

`Concrete::pk(xonly).compile::<Segwitv0>()` used to panic (`Terminal creation must always succeed`, reported in #761). Since the compiler rewrite (ea221af, part of the recent compiler refactor series), it no longer panics — it **silently returns `Ok`** with a Miniscript that is invalid for the target context, e.g. `wsh(pkh(<32-byte x-only key>))`: a fundable address whose coins can never be spent (a 32-byte key is not a valid segwit-v0 ECDSA pubkey, so `CHECKSIG` can never succeed). Verified end-to-end against Bitcoin Core 26.0 consensus (libbitcoinconsensus): funding succeeds, and the owner's spend with the correct private key and a valid signature is rejected.

The same hole lets uncompressed keys compile under Segwitv0/Tap, and x-only keys under Legacy/Bare.

**Fix:** call `Ctx::check_pk` when compiling `Concrete::Key` fragments — the single point where policy keys become terminals, so `multi`/`multi_a` and all public entry points (`compile`, `compile_to_descriptor`, `compile_tr*`) are covered. Invalid keys now produce `CompilerError::ContextError`, reusing the informative `ScriptContextError` messages from #760. (`CompilerError` loses its `Copy`/`Hash` derives since `ScriptContextError` has neither; nothing in-tree relied on them.)

**Test:** `context_invalid_keys` covers x-only keys under Segwitv0/Legacy/Bare (Err) and Tap (Ok), the `multi` path, and uncompressed keys under Segwitv0/Tap (Err) and Legacy (Ok). Adapted from #761; @shesek is credited as co-author.

### DIFF
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -12,7 +12,7 @@ use std::error;
 
 use sync::Arc;
 
-use crate::miniscript::context::SigType;
+use crate::miniscript::context::{ScriptContextError, SigType};
 use crate::miniscript::limits::{MAX_PUBKEYS_IN_CHECKSIGADD, MAX_PUBKEYS_PER_MULTISIG};
 use crate::miniscript::types::{self, ErrorKind, Type};
 use crate::miniscript::ScriptContext;
@@ -25,7 +25,7 @@ type PolicyCache<Pk, Ctx> = BTreeMap<
     BTreeMap<CompilationKey, AstElemExt<Pk, Ctx>>,
 >;
 /// Detailed error type for compiler.
-#[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]
+#[derive(Clone, PartialEq, Eq, Debug)]
 pub enum CompilerError {
     /// `And` fragments only support two args.
     NonBinaryArgAnd,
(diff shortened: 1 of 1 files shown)


apoelstra:
Thanks for finding this! Yeah, this was definitely not an intentional regression.

Can you

1. Move the regression test to a separate commit
2. Also check that you can't compile uncompressed keys into segwit v0 policies?

src/policy/compiler.rs
```diff
@@ -263,6 +268,11 @@ impl From<policy::concrete::PolicyError> for CompilerError {
     fn from(e: policy::concrete::PolicyError) -> Self { Self::PolicyError(e) }
 }
 
+#[doc(hidden)]
+impl From<ScriptContextError> for CompilerError {
```
apoelstra: In d2b2290348a307c25ee1069f72ff6a5a132be4a8:

Please drop this `from` impl and add a `.map_err(CompilerError::ContextError` wherever it's used. These `From` impls hide error conversions behind `?`s, making the code harder to read and follow, plus they're an API commitment that `doc(hidden)` does little to absolve us of. (Users won't avoid this because it's missing from docs ... they'll just type `?` and it'll "just work" and they'll expect it to continue working.)

portlandhodl: Thanks for the solid reasoning behind this! Easy enough as a change to make. Will be more cognizant of this going forward.

portlandhodl:
> * Move the regression test to a separate commit
> * Also check that you can't compile uncompressed keys into segwit v0 policies?

1. 545c89a694d3207d2d7164e4287f9e61d965b13c
2. This was already covered in the previous tests but now made more explicit.

apoelstra:
Oh, https://github.com/rust-bitcoin/rust-miniscript/pull/991 fixes this bug. I'd kinda prefer to close this in favor of that, since this one introduces new usage of `ScriptChecks` (on its way out) and it specifically checks keys. Vs #991 which does all the "non-toplevel" checks which include a few other things and which can be configured by the user via `ValidationParams`.

I will cherry-pick your tests (or at least check that they're covered elsewhere).

portlandhodl:
More than fine.

PR #1046, opened 2026-09-04
plan: fix underreported witness and scriptSig sizes for wsh and sh() wrapping

portlandhodl:
Reimplements #1045, where these bugs were originally found and reported by @guggero.

Fixes two size-accounting bugs in `Plan` fee-estimation helpers that caused input sizes (and therefore fee estimates) to be underreported:

- **Count the witness script in `witness_size` for wsh and sh(wsh).** `Plan::satisfy()` appends the witness script as the final witness element for wsh and sh(wsh) descriptors (#897), but `witness_size()` only summed the template, underestimating the witness by the whole script (varint length prefix plus script bytes).

- **Fix scriptSig size for sh(wpkh) and sh(wsh).** The scriptSig of a nested-segwit input is a push of the witness program, and `scriptsig_size()` missed the outer push opcode byte: sh(wpkh) is 24 bytes on the wire, not 23, and sh(wsh) is 36 bytes, not 35. Fee estimates were 4 WU per input too low.

- **Regression tests:** hold `witness_size()` and `scriptsig_size()` to the exact wire size of a real satisfaction and `satisfaction_weight()` to `Descriptor::max_weight_to_satisfy()`, for both a sigless wsh/sh(wsh) path (no signature size variance) and — using 72-byte signatures that exactly fill the 73-byte-per-signature budget documented on `types::extra_props::SatData` — signature-bearing wpkh, sh(wpkh), wsh and sh(wsh) paths.

Assisted-by: Kimi kimi-latest

### DIFF
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -1283,4 +1283,69 @@ mod test {
 
         assert!(desc.into_plan(&assets).is_err());
     }
+
+    /// A [`Satisfier`] that knows hash preimages only.
+    ///
+    /// A plan sizes a signature at its 73-byte maximum while a real one varies
+    /// in length, so a spend path that carries no signature is the one whose
+    /// witness has a size a plan can be held to byte for byte.
+    struct Preimages(BTreeMap<sha256::Hash, [u8; 32]>);
+
+    impl Satisfier<DefiniteDescriptorKey> for Preimages {
+        fn lookup_sha256(&self, h: &sha256::Hash) -> Option<[u8; 32]> { self.0.get(h).copied() }
+    }
+
+    #[test]
+    fn sizes_match_satisfaction() {
+        use bitcoin::hashes::Hash as _;
+
+        let preimages = [[1u8; 32], [2u8; 32]];
+        let hashes = preimages.map(|p| sha256::Hash::hash(&p));
+        let satisfier = Preimages(hashes.iter().copied().zip(preimages).collect());
+        let assets = Assets::new().add(hashes[0]).add(hashes[1]);
+
+        // Sigless, so it has to go around the sanity checks of `from_str`.
+        let ms = Miniscript::<DefiniteDescriptorKey, Segwitv0>::from_str_insane(&format!(
+            "and_v(v:sha256({}),sha256({}))",
+            hashes[0], hashes[1]
+        ))
+        .unwrap();
+
+        for desc in [
+            Descriptor::new_wsh(ms.clone()).unwrap(),
+            Descriptor::new_sh_wsh(ms).unwrap(),
+        ] {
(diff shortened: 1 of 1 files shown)


apoelstra:
What needs to be done to undraft this?

portlandhodl:
Minimum: I need to squash commits 4 + 5.

Bigger picture I am investigating : while holding these sizes byte-for-byte against a real satisfaction, I found the problem doesn't stop at segwit. The legacy
`sh(<miniscript>)` path is worse than a size bug:

- `Plan::satisfy()` never appends the redeem script for `DescriptorType::Sh`.
  The scriptSig it produces fails the P2SH hash check, so the spend is
  consensus-invalid, not just mispriced.
- `scriptsig_size()` accordingly omits the redeem script push entirely.
- It also uses the witness item-count varint as the scriptSig length
  prefix. Those agree for small scripts and diverge once the scriptSig
  passes 252 bytes, which is routine for P2SH miniscript.
  
  
Then this: the old push_slice path encoded OP_1 (thresh/or branch selectors) as OP_PUSHBYTES_1 0x01, non-minimal, so even sub-252-byte sh() scriptSigs could differ byte-for-byte from get_satisfaction().

src/plan.rs
```diff
@@ -257,7 +257,20 @@ impl<Pk: MiniscriptKey + ToPublicKey> Plan<Pk> {
...
+                        .len();
+                    template_size + varint_len(script_len) + script_len
+                        - varint_len(self.template.len())
+                        + varint_len(self.template.len() + 1)
+                }
+                _ => template_size,
+            }
         } else {
             0
         }
```
Abeeujah, line 276: Nit, this can be done if you have to address any other changes, for scannability, I would implement the function like this:
```rust
pub fn witness_size(&self) -> usize {
    let desc_type = self.descriptor.desc_type();

    if desc_type.segwit_version().is_none() {
        return 0;
    }

    let template_size = witness_size(self.template.as_ref());
    match desc_type {
        DescriptorType::Wsh | DescriptorType::ShWsh => {
            let script_len = self
                .descriptor
                .explicit_script()
                .expect("wsh descriptors have explicit script")
                .len();
            let template_len = self.template.len();
            let varint_diff = varint_len(template_len + 1) - varint_len(template_len);
            template_size + varint_len(script_len) + script_len + varint_diff
        }
        _ => template_size,
    }
}
```

- This reduces the amount of calls to `desc_type` from two to one.
- Reduces Logic nesting by early returning `0` if it's not a witness program.
- Separates the `varint` arithmetic from the witness size arithmetic.

Abeeujah (commented):
ACK ab32aa1e4b209eefc81284f3044f2ba21c5de54a

PR #1049, opened 2026-09-10
 ci: consolidate rbmt version configuration and align local workflows

### LINKED ISSUE #945: CI: Further improve `cargo-rbmt` usage

ANAMASGARD:
Closes #945
  
 - Moved the cargo-rbmt version pin into Cargo.toml so rbmt can check that the installed version matches the repository’s configuration.  Update CI installation and weekly automation to read this pin, then remove rbmt-version.

 - Route local formatting, docs and benchmarks through rbmt to use the same pinned toolchain and defaults as CI. Keep the existing lockfile helper because rbmt’s lock command changes dependency-resolution behavior.

-  Validated formatting, docs, benchmark compilation, workflow syntax and updater behavior.

### DIFF
--- a/.github/actions/setup-rbmt/action.yml
+++ b/.github/actions/setup-rbmt/action.yml
@@ -8,10 +8,8 @@ runs:
       shell: bash
       run: |
         cargo install \
-          --git https://git.rust-bitcoin.org/rust-bitcoin/rust-bitcoin-maintainer-tools \
-          --rev "$(cat rbmt-version)" \
-          cargo-rbmt \
-          --locked
+          --locked \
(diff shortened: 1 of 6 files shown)


apoelstra:
Can we please copy/paste the approach used in https://github.com/rust-bitcoin/rust-bitcoin/pull/6547?

It's more compact (using grep rather than sed) but more importantly, I don't want a proliferation of different LLM-generated hacks to parse Cargo.toml. Let's stick with one.

PR #1050, opened 2026-09-17
Use `serialize()` instead of `to_vec()` for signatures

### LINKED ISSUE #687: Remove unnecessary allocations in `signatures.to_vec`
Ah, yep, this serialization is unnecessary (though it's unrelated to this PR I think and can be done separately).

_Originally posted by @apoelstra in https://github.com/rust-bitcoin/rust-miniscript/pull/679#discussion_r1601598150_
            
Use [`serialize()`](https://docs.rs/bitcoin/latest/bitcoin/ecdsa/struct.Signature.html#method.serialize) instead for ECDSA and Schnorr sigs.

vadim-anfv:
Replace `to_vec()` with `serialize()` on signatures where an owned `Vec<u8>` is not needed.

The remaining `to_vec()` calls on signatures are left as is: an owned `Vec<u8>` is required there by public API or by surrounding code.

Closes #687


apoelstra (approved):
ACK 80fe83b81cb4bc78115c203fcf97194276c6c9dd; successfully ran local tests

